PR #1, opened 2020-05-18
Add rmDir to System.Directory

andrevidela:
I noticed it was missing while I was trying to fix the problem with paths starting with `~`


edwinb:
Thanks! By the way, I'm not going to keep updating the scheme with every change - as long as it's good enough to build the current state it's fine.

PR #2, opened 2020-05-19
Windows fixes

melted:
There's some nonsense when trying to install it yet. But this as far as I got during lunch, the bootstrap works and the libraries build at least.


edwinb:
Brilliant, thank you! Should I wait until the install is successfull, or are you happy for me to merge now and do the rest later?

melted:
You can merge now, it's useful as is.

edwinb:
Done, thanks! I have a Windows machine lying around, though I don't really know how to use it for anything other than playing old games :). I should see if I can get it running myself now.

melted:
The trick is to get msys2 and work just like in unix :-) https://www.msys2.org

PR #5, opened 2020-05-19
Make network test build on windows

melted:
Add a little to gitignore


melted:
It runs through install now (I might have to document though). So now it just have to find the prelude, and how hard can that be?

edwinb:
Great, thanks

PR #7, opened 2020-05-19
Fix chez016 test

chrrasmussen:
One more test passing!


edwinb:
Great, thanks! I hoped it would be easy enough to fix.

PR #8, opened 2020-05-19
Refactor Utils.String.stripQuotes to avoid double reversion on Strings

fabianhjr:
Small REPL test:
Screenshot from 2020-05-19 15-47-14
Dual of edwinb/Idris2#388

### DIFF
--- a/src/Utils/String.idr
+++ b/src/Utils/String.idr
@@ -3,6 +3,8 @@ module Utils.String
 %default total
 
 export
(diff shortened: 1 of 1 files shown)


fabianhjr:
Small repl test:
Screenshot from 2020-05-19 15-43-01

edwinb:
Thanks, always good to avoid a `strTail` too since this allocates unnecessarily.

PR #10, opened 2020-05-20
More fixes and improvements

ska80:
* Ignore build artifacts in 'tests' directory
* Remove unused variables in makefiles
* Add 'bootstrap-clean' rule to delete build artifacts from 'bootstrap' directory
* Add 'distclean' rule to delete all build artifacts from the source tree


edwinb:
Thanks for the updates - actually one of those things explains the difficulty I was having with OSX yesterday! It's been a while since I used it seriously so I've forgotten things...

PR #82, opened 2020-05-20
Add System.Path

andylokandy:
Add a cross-platform path helper, which handles different formats depending on the platform. It can also be used to append paths, to update the extension and to extract the parent directory.

Hopefully, it can be used to self-host idris2 and, as a result, improves robustness on different platforms.

Some necessary list functions are added all together.

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -29,6 +29,15 @@ drop Z     xs      = xs
 drop (S n) []      = []
 drop (S n) (x::xs) = drop n xs
 
+||| Generate a list by repeatedly applying a partial function until exhausted.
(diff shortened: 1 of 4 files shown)


edwinb:
Thanks! I don't know what Travis isn't building this, but once it's all checked I'll merge.

andylokandy:
The trick works :-)

edwinb:
Aha, I was just digging through docs to see if I could force it somehow, but that does the job!

andylokandy:
LOL😆. By the way, is it a good time for me to submit the other functions of List in other PR?

edwinb:
I think it's fine to start adding other library functions now, yes.

PR #83, opened 2020-05-20
Fixing path: `idris2-boot.ss` instead of `idris2-boot.so`

buzden:
It seems that this is a typo in the `idris-boot` shell script.

### DIFF
--- a/bootstrap/idris2-boot
+++ b/bootstrap/idris2-boot
@@ -2,4 +2,4 @@
 
 DIR="`realpath $0`"
 export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:`dirname "$DIR"`/"idris2_app""
-${SCHEME} --script "`dirname $DIR`"/"idris2_app/idris2-boot.so" "$@"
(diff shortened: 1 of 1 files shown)


edwinb:
No, it's correct as is. It gets generated by `bootstrap.sh`. You can see it in action in the travis build.

buzden:
Oh, then I need to understand why it is not generating in my local build. Sorry.

edwinb:
I'm guessing that you must be getting an error message at the point where it tries to run `compile.ss`?

buzden:
> I'm guessing that you must be getting an error message at the point where it tries to run `compile.ss`?

Well, it seems that no. I realized that when Chez fails with exception, `make` continues as if everything is okay. I get
```
...
sh ./bootstrap.sh
compiling idris2_app/idris2-boot.ss with output to idris2_app/idris2-boot.so
Exception: attempt to reference unbound identifier make-thread-parameter at line 171, char 30 of idris2_app/idris2-boot.ss
...
```

somewhere at the top of lots of stuff that's printed when doing `make`. The strange thing is that Chez is compiled with threads support, I don't understand yet why `make-thread-parameter` is unbound.

PR #85, opened 2020-05-20
Add List folds without default value

andylokandy:


libs/base/Data/List.idr
```diff
@@ -400,6 +400,38 @@ mapMaybe f (x::xs) =
...
+||| Foldl without seeding the accumulator. If the list is empty, return `Nothing`.
+public export
+foldl1' : (a -> a -> a) -> List a -> Maybe a
+foldl1' f [] = Nothing
+foldl1' f (x::xs) = Just (foldl f x xs)
+
+||| Foldr without seeding the accumulator. If the list is empty, return `Nothing`.
+public export
+foldr1' : (a -> a -> a) -> List a -> Maybe a
+foldr1' f [] = Nothing
```
gallais, line 431: This one somehow escaped the refactoring via the non-primed version.

PR #86, opened 2020-05-20
Add a total version of the fromMaybe

andylokandy:
- fromJust : (v : Maybe a) -> IsJust v => a


edwinb:
I'm slightly surprised this isn't there in the Idris 1 Prelude actually, given how much else is... thanks! I can imagine uses for it (in fact, I have used it...)

PR #87, opened 2020-05-20
Remove stale src/IdrisPath.idr on each 'make clean' run

ska80:
${PREFIX} can now be properly changed on each rebuild


edwinb:
Oops, good idea, thanks.

PR #88, opened 2020-05-20
libs/network: Port FFI calls from deprecated interface to `%foreign` pragma

ohad:
Mostly tedious rebinding.

There's a slight wiggle in `Network.Socket.Data`, where this module also calls (at one point sneakily) foreign functions. I repeated the binding of the three functions it calls locally (without exporting them).


edwinb:
Great, thanks for doing this! It's a bit of a tedious job but needed doing...

At some point, I think it would also be good to move these into `libidris2_support` rather than having a separate library. You have made that much easier now at least :)

PR #89, opened 2020-05-20
Split Parser into Source(.idr) and Package (.ipkg) (Part 1)

fabianhjr:
Depends on #84.


edwinb:
This seems pretty uncontroversial and tidies things up nicely, thanks.

PR #90, opened 2020-05-20
Fix typo seperate -> separate in System/Path.idr

keram:
> Separate is often misspelled as seperate, a word that has no meaning and is simply a misspelling:

https://www.grammarly.com/blog/separate-seperate/

Noticed when looking at latest merges. Thank you!


gallais:
Thanks!

PR #92, opened 2020-05-21
Move System.Clock from 'contrib' to 'base'

ska80:


edwinb:
I wasn't sure whether to move this into base or not (which is why it accidentally ended up in the code twice) but I suppose the fact that it's used to build Idris 2 itself is an argument for it being in base.

andylokandy:
> I wasn't sure whether to move this into base or not (which is why it accidentally ended up in the code twice) but I suppose the fact that it's used to build Idris 2 itself is an argument for it being in base.

I'm not opposed to moving it into base, but I think the usage in the compiler is not a strong argument for moving things into base, e.g, rust's compiler can make use of all sorts of nightly features.

PR #93, opened 2020-05-21
Rename some file/dir manipulation functions for consistency

ska80:
Renamed:

fileRemove -> removeFile

dirOpen -> openDir
dirClose -> closeDir
rmDir -> removeDir

These changes follow the same naming conventions, like createDir, changeDir, readFile etc.

This change breaks building self-hosted Idris2-in-Idris2 with Idris2-in-Idris1 but bootstrapping with generated scheme code works.


edwinb:
Consistency is good, thanks - these names are left over from Idris 1, where I obviously didn't put much thought into them. I won't merge quite yet, though, because I think the change should also be made in https://github.com/edwinb/Idris2-boot so as not to break that bootstrap path. I'll do that when I get around to it...

edwinb:
Oh dear, I was just about to merge this, but I see there's a problem with the bootstrap files, and it's more annoying than I thought to resolve. Would you mind updating them please, then I'll merge and fix idris2-boot while I'm at it.

ska80:
@edwinb it is done

> Oh dear, I was just about to merge this, but I see there's a problem with the bootstrap files, and it's more annoying than I thought to resolve. Would you mind updating them please, then I'll merge and fix idris2-boot while I'm at it.

It is done

edwinb:
Thanks! I'll sort out Idris2-boot soon, unless someone beats me to it

PR #94, opened 2020-05-21
Windows support

melted:
213 of 218 tests passes, which is pretty good for now.

This needs a new bootstrap image to run to the end. But one can just pick the one from `build/exec` on the first run and copy it to the bootstrap folder. So I don't add that noise to this commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ MAJOR=0
 MINOR=2
 PATCH=0
 
+
 GIT_SHA1=
 ifeq ($(shell git status >/dev/null 2>&1; echo $$?), 0)
     # inside a git repo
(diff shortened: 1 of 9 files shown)


melted:
It's somewhat hacky, there's a new bootstrap-win.sh, when an if statement would have done.

melted:
Next step is to get some CI up and running. Maybe it's time to try out Github Actions?

gallais:
> Maybe it's time to try out Github Actions?

FYI travis also has windows support and that includes msys2 which seems like
a popular solution to build idris2 if I'm to believe the slack channel
https://docs.travis-ci.com/user/reference/windows/#how-do-i-use-msys2

ziman:
FWIW, we use Travis to build Windows binaries for another project and it works well.

edwinb:
Great, thank you! I wonder if this will work on Travis - I would like to keep the windows build working.

PR #96, opened 2020-05-21
Exclude `DN` from `Hashable Name`

### LINKED ISSUE #95: Implementations trigger unnecessary rebuild of dependent modules
# Steps to Reproduce

## Step 1

code in `B.idr`:
```idris
module B

export
moo : IO ()
moo = putStrLn "moo"

export
Show (() -> ()) where
  show f = "_"

export
boo : IO ()
boo = putStrLn "boo"
```
code in `Main.idr`:
```idris
module Main
import B
main : IO ()
main = B.boo
```

## Step 2

```bash
$ idris2 --check Main.idr
1/2: Building B (B.idr)
2/2: Building Main (Main.idr)
```

## Step 3

Add a blank line after the definition of `moo` and run `idris2 --check Main.idr`.

# Expected Behavior

```bash
$ idris2 --check Main.idr
1/2: Building B (B.idr)
```

# Observed Behavior

```bash
$ idris2 --check Main.idr
1/2: Building B (B.idr)
2/2: Building Main (Main.idr)
```

# Further observations

It seems that the hash of interface implementations somehow includes the `FC` of the whole implementation block because `Main` gets rebuilt whenever I make a modification that changes this FC, for example:
* add a blank line above the implementation so it gets displaced to different line numbers
* add a blank line between the end of the implementation and the first following token

Ordinary functions do not suffer from this. I can move the whole function definition to a different line number and `Main` won't get rebuilt. (When experimenting, beware that moving `boo` including its `export` modifier will change the end point of the `FC` of the preceding implementation. If you move just `boo` but leave its `export` where it was, `Main` will _not_ get rebuilt.)

Implementations also do not suffer from this as long as the modifications do not change their `FC`, it seems. For example, if you add a space after `show f`, `Main` will _not_ get rebuilt.

ziman:
File locations end up in module interface hashes, triggering unnecessary rebuilds. This patch ignores `DN` layers in hashes to prevent this behaviour.

Everything else, such as the `Show` instance for names, should be unaffected.

Fixes #95.

### DIFF
--- a/src/Core/Hash.idr
+++ b/src/Core/Hash.idr
@@ -64,6 +64,8 @@ Hashable String where
 export
 Hashable Name where
   hashWithSalt h (MN s _) = hashWithSalt h s
+  hashWithSalt h (DN _ n) = hashWithSalt h n
+  hashWithSalt h (NS ns n) = hashWithSalt (hashWithSalt h ns) n
   hashWithSalt h (Resolved i) = hashWithSalt h i
   hashWithSalt h n = hashWithSalt h (show n)
 


edwinb:
I wonder if having the number in the display is worth it? The purpose of a DN is to show a name that's useful to the programmer without leaking internal details. Maybe it should just say "Show implementation in Error2.idr" or possible better still "Show <type> implementation in <file>"?

Anyway, thanks for the fix! I've been wondering why this sometimes happens but hadn't tried to pin down the reason...

ziman:
I think it would be easier to read `--dumpcases` and the generated Scheme output with the number in.

ziman:
I see what you mean now; I'll change it.

edwinb:
Thanks!

PR #100, opened 2020-05-21
Use Control.Delayed from the 'contrib' package

ska80:


edwinb:
I'm not sure we should do this in general for modules in `contrib` because they might change, and I would like to keep the Idris2 source itself as self contained as possible, given that everything else somehow depends on it. In practice, that means only using things from `base` if we can (there is a dependency on `network` too).

There's a few things in the Idris2 source that have small variations from library code, or that I added because they didn't exist in base library code. Probably these should feed back to the base libraries eventually.

ska80:
Should I keep this PR in sync with the master for a while until it is decided how to proceed next or close it?

edwinb:
Hmm, this has been lingering at the bottom of the PR list for a while. By now we're using a couple of things from `contrib` elsewhere. Maybe we should consider whether they belong in `base` instead. I'll merge this, to save the repetition.

PR #102, opened 2020-05-21
Convert to full names before computing hashes

### LINKED ISSUE #101: Resolved names affect module hashes
# Steps to Reproduce

I haven't been able to quickly isolate a small MFE, sorry! I'd like to file the issue, anyway.

Removing [this line](https://github.com/ziman/itt-idris/blob/251cb2c55046d666c7cba4b3007cdfccb5f812a7/Core/TT.idr#L100) changes the hash of `Core.TT` and causes rechecking of dependent modules, even though the instance is only `export`.

# Expected Behavior

Modules that depend on `Core.TT` do not get rechecked.

# Observed Behavior

Modules that depend on `Core.TT` get rechecked.

# Hypothesised cause

I put a `printLn` into `Core.Context.addHash`, ran Idris on the module with and without the clause in question, and `diff`ed the outputs. That produced a plausible reason for the observed behaviour:
```idris
97c97
< ("ADD-HASH", {0 q : Type} -> {0 n : $resolved1030} -> ((1 name : String) -> ((1 qv : q[2]) -> ((1 type : ($resolved3697 q[3] n[2])) -> ($resolved3696 q[4] n[3])))))
---
> ("ADD-HASH", {0 q : Type} -> {0 n : $resolved1030} -> ((1 name : String) -> ((1 qv : q[2]) -> ((1 type : ($resolved3698 q[3] n[2])) -> ($resolved3697 q[4] n[3])))))
```

The types of names bound in this signature contain resolved names but the numbers are different. These numbers can indeed affect the hash:

```idris
Hashable Name where
  hashWithSalt h (MN s _) = hashWithSalt h s
  hashWithSalt h (DN _ n) = hashWithSalt h n
  hashWithSalt h (NS ns n) = hashWithSalt (hashWithSalt h ns) n
  hashWithSalt h (Resolved i) = hashWithSalt h i
  hashWithSalt h n = hashWithSalt h (show n)
```
I'm not sure how to best fix this since `hashWithSalt` does not have access to the context to retrieve the real name that is actually part of the interface of the module.

I have not verified that this is indeed the cause (it could probably be done by hashing `Resolved _` to a constant value to see if that stops the problem from manifesting).

ziman:
Fixes #101 but there's more left to address and it has no effect in practice yet. I'm marking this as a draft until it's done.

### DIFF
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -885,6 +885,9 @@ clearCtxt
     resetElab : Options -> Options
     resetElab = record { elabDirectives = defaultElab }
 
+-- Beware: if your hashable thing contains (potentially resolved) names,
+-- it'll be better to use addHashWithNames to make the hash independent
+-- of the internal numbering of names.
 export
 addHash : {auto c : Ref Ctxt Defs} ->
           Hashable a => a -> Core ()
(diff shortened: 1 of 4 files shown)


ziman:
It seems I made a mistake during testing; now it correctly avoids rebuilding in the cases it did not before.

Issue #104, opened 2020-05-21
--find-ipkg causes error when there is no .ipkg file in parent directories


ohad:
# Steps to Reproduce
Create a valid .idr file in a directory where the directory and its parents are without `.ipkg` files:
```
$ cat Temp.idr
h : Bool
h = True
$ ls *.ipkg
$
```

# Expected Behavior
```
$ idris2 -c Temp.idr 
1/1: Building Temp (Temp.idr)
$ idris2 -c Temp.idr  --find-ipkg
$
```
# Observed Behavior
```
$ idris2 -c Temp.idr 
1/1: Building Temp (Temp.idr)
$ idris2 -c Temp.idr  --find-ipkg
Uncaught error: Parse error (Parse error: Expected name (next tokens: [symbol =, identifier contrib, end of input]))
$
```

ohad:
Sorry, I can't reproduce this, because I accidentally *did* have a malformed `.ipkg` file in a directory/parent directory.

PR #105, opened 2020-05-22
Emacs mode compatibility

ohad:
First steps towards emacs idris-mode compatibility

To make sure the protocol doesn't jam, supply stub implementation to
all protocol commands.

Report to REPL when an unimplemented command is used.

Notes
-----
0.  Fix disambiguation issue (this also happens on master, so might cause a conflict if someone fixes it first).

1. We don't support add-proof-clause. I guess all proof-related stuff
   should be removed from the protocol

2. Some types are stub types too. That's because we need to `SExpable`
   more structured types like `PTerm`. I'll leave that to the future.

3. I've lifted `REPLResult` into `IDEResult` inside `IDEMode/REPL.idr`.
   That's because some results, like `who-calls` only make sense in
   the context of an IDE, not in the context of a user-facing REPL.

   The Editing commands should be moved from the REPL into the IDE and
   called from the REPL.

   I leave that to the future, once more of the protocol is
   implemented.

4. Export a few functions from the REPL so that the IDE can call them.

5. There's one outstanding issue with the emacs idris-mode: it
   currently calls the unsupported `:consolewidth` REPL command. This
   is harmless, and can wait until @edwinb decides whether we should
   support it in the future or not.

6. There was a bug as to how holes are returned to the user. The
   format isn't documented in the protocol, so we'll need to reproduce
   it, perhaps from the idris-mode elisp sources.


abailly:
re holes: Most of the action takes place in file [idris-hole-list.el](https://github.com/idris-hackers/idris-mode/blob/master/idris-hole-list.el#L118) where idris-mode takes a hole name, a list of premises and a conclusion and displays them as a tree. The result should look like:

```
- + subway.fromChars [P]
 `--                 b : Type
     ----------------------------------------
      subway.fromChars : List Char -> Subway

```

jfdm:
@ohad Thanks for this! Emacs mode compatibility has been on my radar but not enough time to get to it thanks! Incidentally, you might have noticed that I have performed some care-taker actions on idris-hackers/idris-mode If you need more changes to idris-mode itself then let me know. I will be happy to shepherd anything.

ohad:
@jfdm there's still plenty to do, but at least now I can use emacs with the existing functionality (exciting!).

We'll likely need some minor changes to idris-mode (e.g., #104 ), but I guess whatever change would need to be backwards compatible with `idris1`.

edwinb:
Oh that was clever, I failed to resolve the conflicts competently. Sorry! I have no idea how to undo this...

edwinb:
Oh, yes, I can fix it... hang on!

PR #107, opened 2020-05-22
Fix windows tests

melted:
This makes all the tests pass on Windows.

There is one real oddity in the test runner that I will note in a comment.

### DIFF
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -7,6 +7,7 @@ import Data.Strings
 import System
 import System.Directory
 import System.File
+import System.Info
 import System.Path
 
 %default covering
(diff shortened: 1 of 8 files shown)


tests/Main.idr
```diff
@@ -203,17 +216,17 @@ runTest opts testPath
                            else print FileNotFound
                          pure False
                      | Left err => do print err
-                                      pure False
-
-                 if (out == exp)
+                                      pure False 
+                 let result = normalize out == normalize exp
+                 if normalize out == normalize exp
```
melted, line 221: Here I should of course use `result`in the if-condition. But I can't! It bombs with a really weird error.

melted:
I have a CI setup on Github Actions that works. I didn't have a lot of luck with MSYS2 on Travis (there apparently has been some recent change in it that makes it crash hard on travis), and on Github Actions you can just import a step from somebody else.

tests/Main.idr
melted:
```
Main.idr:222:26--223:21:While processing right hand side of runTest at Main.idr:166:1--231:1:
While processing right hand side of runTest,runTest' at Main.idr:206:9--231:1:
        ()
and
        ?a [locals in scope: testPath, opts, {_:1513}, {_:1522}, out, {_cn:1}, exp]
```

melted: I get the same if I ditch normalize and just put in `out == exp` in result.

edwinb:
The `let` error seems like something I should investigate separately. That is odd.

PR #109, opened 2020-05-22
`with` expressions for name disambiguation

ziman:
In https://github.com/edwinb/Idris2-boot/pull/308, we talked about restricting disambiguation for better error messages. (See that pull request for a motivating example.)

While some aspects of the original proposal were controversial, it seems that most people were okay with notation like `with Prelude.(>>=) do ...`. This pull request is a port of the [original implementation](https://github.com/edwinb/Idris2-boot/pull/359).

This patch adds new syntax `with NS.name <term>` and `with [NS.name1, NS.name2, ...] <term>` to determine disambiguation for the specified names in `<term>` without search.

The names given need not be _fully_ qualified, so `with Vect.Nil` will work for `Data.Vect.Nil`, as long as that's the only possible disambiguation, and you can even say `with Nil []` if there's only one `Nil` in scope.

BTW, the syntax allows `with Prelude.pure the (Maybe _) (pure 4)` without parenthesising the RHS. Not sure if I should restrict it; it's syntactically unambiguous because the list of identifiers is unambiguous but it looks kind of strange to me. However, I definitely want to support `with ... do ...` without parenthesising the `do` expression so I decided to leave it at this for now. Suggestions welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ Language changes:
 * A %spec flag on functions which allows arguments to be marked for partial
   evaluation, following the rules from "Scrapping Your Inefficient Engine"
   (ICFP 2010, Brady & Hammond)
+* To improve error messages, one can use `with NS.name <term>`
+  or `with [NS.name1, NS.name2, ...] <term>` to disable disambiguation
+  for the given names in `<term>`. Example: `with MyNS.(>>=) do ...`.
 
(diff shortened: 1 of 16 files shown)


edwinb:
I was wanting this a couple of days ago too, thanks for fixing up for the new version. I think this is good to go following the discussion over on Idris2-boot.

PR #111, opened 2020-05-22
minus proofs on Nats from Idris1

err0r500:
It's just a copy / paste from Idris 1 source code adding the "minus" proofs on Nats


edwinb:
Thanks for porting!

PR #113, opened 2020-05-23
Split Source and Package Parser (Part 2)

fabianhjr:
This has two commits, the first one refactors SourceToken a bit and the second one implements an (unpolished) Package Lexer/Parser and refactors Idris/Package.idr to use this new Lexer/Parser.

Feedback is appreciated. First part is #89.

One pro is that package names can now have dashes in their name UwU

## Cleanup of SourceToken

There are many organizational and consistency changes, one of the more important ones being `Literal` being rename `StringLit` to correspond with `CharLit`, `IntegerLit`, `DoubleLit` and extracting more common code to Parser/Lexer/Common.

## Change to `ParseError tok` and `ParseError`

`ParseError tok` was renamed `ParsingError tok` and `ParseError` was refactored to `ParseError tok` to accept a token type.

### DIFF
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -15,6 +15,12 @@ import Data.Strings
 import Data.StringTrie
 import Data.These
 
+import Parser.Package
+import System
+import Text.Parser
+import Utils.Binary
+import Utils.String
+
 import Idris.CommandLine
 import Idris.ModTree
 import Idris.ProcessIdr
@@ -23,13 +29,6 @@ import Idris.REPLOpts
 import Idris.SetOptions
 import Idris.Syntax
 import Idris.Version
-import Parser.Lexer.Source
-import Parser.Source
-import Utils.Binary
-
-import System
-import Text.Parser
-
 import IdrisPaths
 
 %default covering
@@ -114,7 +113,7 @@ data DescField : Type where
   PPreclean    : FC -> String -> DescField
   PPostclean   : FC -> String -> DescField
 
-field : String -> SourceRule DescField
+field : String -> Rule DescField
 field fname
       = strField PVersion "version"
     <|> strField PAuthors "authors"
@@ -134,43 +133,41 @@ field fname
     <|> strField PPostinstall "postinstall"
     <|> strField PPreclean "preclean"
     <|> strField PPostclean "postclean"
-    <|> do exactIdent "depends"; symbol "="
-           ds <- sepBy1 (symbol ",") unqualifiedName
+    <|> do exactProperty "depends"
+           equals
+           ds <- sep packageName
            pure (PDepends ds)
-    <|> do exactIdent "modules"; symbol "="
-           ms <- sepBy1 (symbol ",")
-                      (do start <- location
-                          ns <- nsIdent
-                          end <- location
-                          Parser.Core.pure (MkFC fname start end, ns))
-           Parser.Core.pure (PModules ms)
-    <|> do exactIdent "main"; symbol "="
+    <|> do exactProperty "modules"
+           equals
+           ms <- sep (do start <- location
+                         m <- moduleIdent
+                         end <- location
+                         pure (MkFC fname start end, m))
+           pure (PModules ms)
+    <|> do exactProperty "main"
+           equals
            start <- location
-           m <- nsIdent
+           m <- namespacedIdent
            end <- location
-           Parser.Core.pure (PMainMod (MkFC fname start end) m)
-    <|> do exactIdent "executable"; symbol "="
-           e <- unqualifiedName
-           Parser.Core.pure (PExec e)
+           pure (PMainMod (MkFC fname start end) m)
+    <|> do exactProperty "executable"
+           equals
+           e <- (stringLit <|> packageName)
+           pure (PExec e)
   where
-    getStr : (FC -> String -> DescField) -> FC ->
-             String -> Constant -> SourceEmptyRule DescField
-    getStr p fc fld (Str s) = pure (p fc s)
-    getStr p fc fld _ = fail $ fld ++ " field must be a string"
-
-    strField : (FC -> String -> DescField) -> String -> SourceRule DescField
-    strField p f
+    strField : (FC -> String -> DescField) -> String -> Rule DescField
+    strField fieldConstructor fieldName
         = do start <- location
-             exactIdent f
-             symbol "="
-             c <- constant
+             exactProperty fieldName
+             equals
+             str <- stringLit
              end <- location
-             getStr p (MkFC fname start end) f c
+             pure $ fieldConstructor (MkFC fname start end) str
 
-parsePkgDesc : String -> SourceRule (String, List DescField)
+parsePkgDesc : String -> Rule (String, List DescField)
 parsePkgDesc fname
-    = do exactIdent "package"
-         name <- unqualifiedName
+    = do exactProperty "package"
+         name <- packageName
          fields <- many (field fname)
          pure (name, fields)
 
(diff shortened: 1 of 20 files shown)


fabianhjr:
This is the final part of splitting lexing/parsing of sources (.idr) and package (.ipkg) files.

It isn't as polished but works with minimal changes.

src/Parser/Lexer/Common.idr
```diff
@@ -45,3 +52,37 @@ ident : Flavour -> Lexer
...
+export
+identAllowDashes : Lexer
+identAllowDashes = ident AllowDashes
+
+namespaceIdent' : Bool -> Lexer
+namespaceIdent' b = ident Capitalised <+> many leavingLastIdent
+  where
+    leavingLastIdent : Lexer
+    leavingLastIdent = if b then (is '.' <+> ident Capitalised)
+                       else (is '.' <+> ident Capitalised <+> expect (is '.'))
```
ziman, line 73: What's the meaning of this `Bool` flag? Could you add a comment or maybe rename `b`?

src/Parser/Rule/Package.idr
```diff
@@ -0,0 +1,69 @@
...
+import public Parser.Lexer.Package
+import public Parser.Rule.Common
+
+import Data.List
+
+%default total
+
+public export
+PackageRule : Type -> Type
+PackageRule = Rule PackageToken
```
ziman, line 12: Given that we already are in a module called `Package`, wouldn't it be better to name this just `Rule` and then use `Package.Rule` whenever it matters and just `Rule` whenever unambiguous?

`SourceRule` would then be `Source.Rule` or just `Rule` most of the time, too.

I think that module qualification is better than baking information into identifiers, both semantically and for convenience, too.

src/Parser/Lexer/Common.idr
fabianhjr: Hi @ziman a comment has been added.

src/Parser/Rule/Package.idr
fabianhjr: Renamed {Package,Source}Token -> Token and {Package,Source}Rule -> Rule as per the suggestion, it does work quite nicely.

fabianhjr:
Hi @ziman, thanks for the feedback, those two points have been added to this merge request.

ziman:
Great, thanks! I've got some more questions, I'll put them as comments in the diff.

src/Parser/Lexer/Common.idr
fabianhjr: > How does your lexer disambiguate them?

It doesn't, it uses it in a different lexer for packages

fabianhjr:
Hey @ziman, I think this is way cleaner now, thanks for the feedback.

src/Idris/Package.idr
```diff
@@ -134,43 +133,40 @@ field fname
...
-           m <- nsIdent
+           m <- namespacedIdent
            end <- location
-           Parser.Core.pure (PMainMod (MkFC fname start end) m)
-    <|> do exactIdent "executable"; symbol "="
-           e <- unqualifiedName
-           Parser.Core.pure (PExec e)
+           pure (PMainMod (MkFC fname start end) m)
+    <|> do exactProperty "executable"
+           e <- getString
```
ziman: what about renaming this function to `stringLit` or something that's consistent with your other token names?

src/Idris/Package.idr
```diff
@@ -134,43 +133,40 @@ field fname
...
   where
-    getStr : (FC -> String -> DescField) -> FC ->
-             String -> Constant -> SourceEmptyRule DescField
-    getStr p fc fld (Str s) = pure (p fc s)
-    getStr p fc fld _ = fail $ fld ++ " field must be a string"
-
-    strField : (FC -> String -> DescField) -> String -> SourceRule DescField
-    strField p f
+    strField : (FC -> String -> DescField) -> String -> Rule DescField
+    strField fieldConstructor fieldName
```
ziman, line 159: I like the descriptive names!

src/Idris/Package.idr
```diff
@@ -425,6 +421,12 @@ clean pkg
              delete $ ttFile ++ ".ttc"
              delete $ ttFile ++ ".ttm"
 
+getParseErrorLoc : String -> ParseError Token -> FC
+getParseErrorLoc fname (ParseFail _ (Just pos) _) = MkFC fname pos pos
+getParseErrorLoc fname (LexFail (l, c, _)) = MkFC fname (l, c) (l, c)
+getParseErrorLoc fname (LitFail _) = MkFC fname (0, 0) (0, 0) -- TODO: Remove this unused case
```
ziman, line 428: A TODO here; do we still need to remove it?

src/Parser/Lexer/Common.idr
```diff
@@ -45,3 +43,26 @@ ident : Flavour -> Lexer
...
+export
+identAllowDashes : Lexer
+identAllowDashes = ident AllowDashes
+
+namespaceIdent : Lexer
+namespaceIdent = ident Capitalised <+> many (is '.' <+> ident Capitalised <+> expect (is '.'))
+
+export
+namespacedIdent : Lexer
+namespacedIdent = namespaceIdent <+> opt (is '.' <+> identNormal)
```
ziman, line 64: Oh, nice!

I think it even keeps the accuracy that you originally wanted (my suggestion wouldn't).

ziman: Perhaps `namespaceIdent` could be called just `namespace`?

src/Parser/Lexer/Package.idr
```diff
@@ -0,0 +1,68 @@
...
+
+%default total
+
+public export
+data Token
+  = Assignment
+  | Comment String
+  | EndOfInput
+  | DotSepIdent (List String)
+  | Ident String
```
ziman: What about removing `Ident x` in favour of `DotSepIdent [x]`?

src/Parser/Lexer/Package.idr
```diff
@@ -0,0 +1,68 @@
...
+  show (Comment str) = "Comment: " ++ str
+  show EndOfInput = "EndOfInput"
+  show (DotSepIdent dsid) = "DotSepIdentifier: " ++ dotSep dsid
+  show (Ident i) = "Identifier: " ++ i
+  show Separator = "Separator"
+  show Space = "Space"
+  show (StringLit s) = "StringLit: " ++ s
+
+assignment : Lexer
+assignment = is '='
```
ziman, line 35: This should probably be called `equals` or something like that; not all uses of an equals sign are assignments (and they are not in package descriptions).

Also, an assignment (statement) would normally consist of a LHS, an equals sign, and a RHS, such as `x = 3`; but here, we're just talking about the symbol itself.

src/Parser/Lexer/Source.idr
```diff
@@ -12,46 +12,50 @@ import Utils.String
...
-  | CharLit String
+data Token
+  -- Literals
+  = CharLit String
   | DoubleLit Double
+  | IntegerLit Integer
+  | StringLit String
+  -- Identifiers
+  | HoleIdent String
+  | Ident String
```
ziman, line 23: Perhaps here, `Ident x` should go in favour of `DotSepIdent [x]`, too.

src/Parser/Lexer/Source.idr
```diff
@@ -205,7 +193,9 @@ fromOctLit str
                   Nothing => 0 -- can't happen if the literal lexed correctly
                   Just n => cast n
 
-rawTokens : TokenMap SourceToken
+%hide Prelude.reverse
```
ziman, line 196: Is this still necessary? Wouldn't it be better to qualify the references to `reverse` with their namespace, such as `List.reverse`?

src/Parser/Rule/Package.idr
```diff
@@ -0,0 +1,79 @@
...
+public export
+PackageEmptyRule : Type -> Type
+PackageEmptyRule = EmptyRule Token
+
+export
+assignment : Rule ()
+assignment = terminal "Expected assignment"
+                      (\x => case tok x of
+                                  Assignment => Just ()
+                                  _ => Nothing)
```
ziman, line 23: I'd call this "equals" as well.

src/Idris/Package.idr
fabianhjr: I added it, yes since this parser doesn't use Unlit and so can never have a LitFail.

src/Parser/Lexer/Common.idr
fabianhjr: Impossible since namespace is a keyword in Idris2

PR #115, opened 2020-05-23
Windows CI

melted:


edwinb:
Thanks, now we have no excuse not to keep Windows support working :)

PR #118, opened 2020-05-23
Fix comparison in gen_expected.sh

melted:


melted:
Have to check the travis log to see if the noise disappears before merging.

PR #120, opened 2020-05-23
Fix Windows build

melted:


melted:
This fixes breakage, there's just more breakage beyond it. So commit.

PR #121, opened 2020-05-23
Add popen and pclose from Idris 1

timsueberkrueb:
Mostly copied from Idris 1. I have only tested it on Ubuntu though.
Is this the right place for the test case?

### DIFF
--- a/libs/base/System/File.idr
+++ b/libs/base/System/File.idr
@@ -38,6 +38,10 @@ prim__writeLine : FilePtr -> String -> PrimIO Int
 prim__eof : FilePtr -> PrimIO Int
 %foreign "C:fflush,libc 6"
 prim__flush : FilePtr -> PrimIO Int
+%foreign support "idris2_popen"
+prim__popen : String -> String -> PrimIO FilePtr
+%foreign "C:pclose,libc 6"
+prim__pclose : FilePtr -> PrimIO ()
 
 %foreign support "idris2_removeFile"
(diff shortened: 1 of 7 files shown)


edwinb:
Yes, that's right for the test case, thanks for porting, and sorry to be slow to comment. In looks like the CI needs a poke though.

melted:
It's using the old CI setup, it would be best if this was rebased on master before merging.

timsueberkrueb:
The test case fails on Windows but unfortunately without any error output after "opened".

melted:
It says "No entry for pclose" quite a bit higher up in the output, guess that went to stderr. I can try to run it later after work.

melted:
It's _pclose on windows, probably best to just put in an idris2_pclose: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/pclose?view=vs-2019

But in the long run it would be nice if we could put such differences in the %foreign in some way.

melted:
I tried to put an if statement in one of those %foreign functions, but the compiler didn't buy it :-)

melted:
Another problem, I'm looking at it now.

timsueberkrueb:
Thank you @melted. Looks like Windows is happy now.

timsueberkrueb:
There's probably a better way to write this test case, though.

melted:
Guess you could have put an if in `run` instead and used cygpath to get a windows path, but this is fine.

PR #122, opened 2020-05-23
Add additional information on requirements in INSTALL.md

keram:
and improve formatting in terminal.

Why:
This may help people prevent or resolve following issues
when running `make bootstrap SCHEME=chez`

- missing clang
```
make: clang: Command not found
make[1]: Entering directory 'support/c'
make[1]: clang: Command not found
```
- missing Chez thread support
```
Compiling idris2_app/idris2-boot.ss with output to idris2_app/idris2-boot.so
Exception: attempt to reference unbound identifier
make-thread-parameter at line 171, char 30 of idris2_app/idris2-boot.ss
```

`$ less INSTALL.md` before:
install-less-before

`$ less INSTALL.md` after:
install-less-after

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,18 +1,20 @@
 Installing
 ==========
 
-The easiest way to install is via the existing generated Scheme code. The
-requirements are:
+The easiest way to install is via the existing generated Scheme code.
+The requirements are:
 
-* A Scheme compiler; either Chez Scheme (default), or Racket
-* `bash`, with `realpath`. On Linux, you probably already have this. On
-  a Mac, you can install this with `brew install coreutils`.
+* A Scheme compiler; either Chez Scheme (default), or Racket.
+* `bash`, with `realpath`. On Linux, you probably already have this.
+  On a Mac, you can install this with `brew install coreutils`.
 
 On Windows, it has been reported that installing via `MSYS2` works
 (https://www.msys2.org/). On Raspberry Pi, you can bootstrap via Racket.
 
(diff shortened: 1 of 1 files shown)


melted:
We can remove `CC := clang`, the problem that was fixed by that is gone.

keram:
> We can remove `CC := clang`, the problem that was fixed by that is gone.

You mean remove that from `config.mk` right?
You also mean that compiling using gcc is now as fast or faster than when using clang?
There used to be comment saying that clang was much faster so it make sense to use it 
https://github.com/idris-lang/Idris2/blob/1212e569037c832cd0ecc89abd4106b8ccc7138a/config.mk#L5
I have noticed you removed that comment, Are there any benchmarks?

ziman:
There's barely any C left in Idris 2 so the compiler doesn't really matter anymore.

melted:
It used to be that Idris 2 was compiled by Idris 1 to a huge C file that clang handled much better. Now there's just a small support lib that is compiled in seconds by any compiler.

keram:
Removed the mention of clang and also created the followup PR to remove it from config.mk as suggested. https://github.com/idris-lang/Idris2/pull/183 
Here I also replaced my original note about Chez 
```
If you build your Chez Scheme from source please ensure it is build with 
[native thread support](https://github.com/cisco/ChezScheme/blob/master/BUILDING#L89-L91).
```
with similar note from `tutorial/starting.rst` about which I did not know when creating this PR.
@ziman Matus feeling brave to help get rid of those warnings ? :wink:

PR #123, opened 2020-05-23
Make System.Path work with String

andylokandy:
Lift String to be the primitive of System.Path, that is, use pure String to represent paths, and the original Path ADT has become the internal util used to manipulate path.

I made this change because I've attempted to replace System.Path in the compiler itself, and I've found that the ADT path is kind of useless in practice, because it forces every function uses Path argument to be wrapped by a do-block and to parse every argument in advance, including the hard-code one, and to `show path` in every call to file functions.

The previous module interface is adopted from Rust [std](https://doc.rust-lang.org/std/path/struct.Path.html) and [FilePath](https://github.com/haskell/filepath) of Haskell, both of whom manipulate Path in String, so I switched to stick closer to them.

One thing weird is that this PR works fine in my testing project, nevertheless, it will cause **`Exception: invalid memory reference. Some debugging context lost`** when I run the compiler that is modified to use System.Path, I suspect it may indicate some bugs in libidris2_support.

### DIFF
--- a/libs/contrib/System/Path.idr
+++ b/libs/contrib/System/Path.idr
@@ -25,43 +25,44 @@ export
 pathSeparator : Char
 pathSeparator = if isWindows then ';' else ':'
 
-||| Windows' path prefixes.
-|||
-||| @ UNC Windows' Uniform Naming Convention, e.g., a network sharing
-|||   directory: `\\host\c$\Windows\System32`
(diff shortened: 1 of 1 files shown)


andylokandy:
I'm going to investigate more how well the api integrates with the compiler in action.

andylokandy:
Closed due to https://github.com/idris-lang/Idris2/pull/169

PR #127, opened 2020-05-23
Add GitHub CI action that bootstraps on Ubuntu

wchresta:
The latest commits broke bootstrapping on Linux, even though we have CI using Travis already that should detect this problem.

This CR adds a bootstrap build for Ubuntu using GitHub actions, similarly to the ones we have for windows. This hopes to achieve more stable and faster feedback on builds and their problems.

The proposed action shows the current build failure, as shown here: https://github.com/wchresta/Idris2/runs/702631251
The current Travis build for the same commit is currently stuck: https://travis-ci.org/github/idris-lang/Idris2/builds/690433878

This will not fix the build failure for Linux but rather make the failure obvious through (hopefully) better CI. We can leave both builds in, Travis and GitHub, and observe which works better for us. Eventually, we'll want to get rid of one of them.

### DIFF
--- a/.github/workflows/ci-ubuntu.yml
+++ b/.github/workflows/ci-ubuntu.yml
@@ -1,0 +1,24 @@
+name: Ubuntu CI
+on:
+  push:
+    branches:
+      - '*'
+    tags:
+      - '*'
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


wchresta:
I'll probably add commands for both, bootstrapping via scheme and via racket.

edwinb:
Let's give it a go! Yes, I'd like to make sure the racket process works too, especially now I have the tests running successfully I don't want to break them...

edwinb:
Given that this clearly works just as well as travis, maybe we should remove travis now and add a racket bootstrap too in its place?

PR #128, opened 2020-05-23
Allow cli given option overriding for ipkg cmds

jfdm:
As it was in Idris1 being able to override *some*, or introduce new, options to an Idris IPKG is beneficial. For example, generate code for multiple codegens from a single source.

This PR is the beginning of adding this ability to Idris2. I have fixed some other minor issues with how Idris handle's IPKG files as well.

I will get round to this as and when I can. This PR is there to let people know something is being done.

### DIFF
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -10,6 +10,7 @@ import Core.Options
 import Core.Unify
 
 import Data.List
+import Data.Maybe
 import Data.StringMap
 import Data.Strings
 import Data.StringTrie
@@ -263,11 +264,14 @@ compileMain mainn mmod exec
 build : {auto c : Ref Ctxt Defs} ->
         {auto s : Ref Syn SyntaxInfo} ->
         {auto o : Ref ROpts REPLOpts} ->
-        PkgDesc -> Core (List Error)
-build pkg
+        PkgDesc ->
+        List CLOpt ->
+        Core (List Error)
+build pkg opts
     = do defs <- get Ctxt
          addDeps pkg
          processOptions (options pkg)
+         preOptions opts
          runScript (prebuild pkg)
          let toBuild = maybe (map snd (modules pkg))
                              (\m => snd m :: map snd (modules pkg))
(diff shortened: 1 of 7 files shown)


tests/idris2/pkg003/expected
```diff
@@ -0,0 +1,15 @@
+1/1: Building Main (Main.idr)
+Not all command line options can be used to override package options.
+
+Overridable options are:
+        --quiet
+        --verbose
+        --timing
+        ---dumpcases <file>
```
gallais, line 8: typo: 3 dashes

tests/idris2/pkg003/run
```diff
@@ -0,0 +1,9 @@
+$1 --build testpkg.ipkg
+rm -f  build/
+$1 --build testpkg.ipkg --quiet
+$1 --build testpkg.ipkg --verbose
+$1 --build testpkg.ipkg --timing
+$1 --build testpkg.ipkg --codegen gambit
+$1 --build testpkg.ipkg --ide-mode
+$1 --build malformed-package-name
+$1 --build non-existent-package.ipkg
```
gallais, line 9: Shouldn't you `rm -rf build` at the end too to cleanup after yourself?

tests/idris2/pkg003/expected
jfdm: ah yes! thanks

tests/idris2/pkg003/run
jfdm: that would be good.

jfdm:
This *should* be good to go!

src/Idris/Package.idr
```diff
@@ -432,51 +440,101 @@ getParseErrorLoc fname _ = replFC
...
+         else do Right (pname, fs) <- coreLift $ parseFile file
+                                          (do desc <- parsePkgDesc file
+                                              eoi
+                                              pure desc)
+                     | Left (FileFail err) => throw (FileErr file err)
+                     | Left err => throw (ParseFail (getParseErrorLoc file err) err)
+                 pkg <- addFields fs (initPkgDesc pname)
+                 case cmd of
+                      Build => do [] <- build pkg opts
+                                     | errs => coreLift (exitWith (ExitFailure 1))
```
gallais, line 472: I was surprised to see we are not doing anything with the errors but
it turns out that `buildMod` has already emitted all the warnings &
errors. So answering my own question: these are bubbling out but we
do not need to do anything with them.

jfdm:
@gallais I am sure we can do the filtering with respect to a predefined list of overridable options (might not require `List CLOpt` but something slightly different) that is also also used to generate the error message. To be fair [`options` in `CommandLine.idr`](https://github.com/idris-lang/Idris2/blob/master/src/Idris/CommandLine.idr#L106) is not covering wrt to [`CLOpt`](https://github.com/idris-lang/Idris2/blob/master/src/Idris/CommandLine.idr#L39). But that is a more interesting problem to solve.

PR #130, opened 2020-05-24
add Data.List.Quantifiers

clayrat:
Ported from https://github.com/clayrat/Idris-dev/blob/master/libs/base/Data/List/Quantifiers.idr with some refactoring


libs/base/Data/List/Quantifiers.idr
```diff
@@ -0,0 +1,84 @@
...
+
+export
+Uninhabited (Any p Nil) where
+  uninhabited (Here _) impossible
+  uninhabited (There _) impossible
+
+||| Eliminator for `Any`
+export
+anyElim : (Any p xs -> b) -> (p x -> b) -> Any p (x :: xs) -> b
+anyElim _ f (Here p)  = f p
```
gallais, line 24: That looks more like a `case` than an `elim` as we are not calling
ourselves recursively in the `There` case (and the target type is
not dependent).

clayrat: Yeah, this was just copied from Idris1, and I'm not even sure what's the use case for it, maybe let's just remove it?

edwinb: Me neither... if it turns out anyone needs it I'm sure we'll welcome the PR!

PR #133, opened 2020-05-24
some proofs and views around lists

rgrover:
Added a few views and and proofs. In some cases, the code was borrowed from Idris1.

### DIFF
--- a/libs/contrib/Data/List/Equalities.idr
+++ b/libs/contrib/Data/List/Equalities.idr
@@ -18,7 +18,7 @@ consCong2 Refl Refl = Refl
 ||| Equal non-empty lists should result in equal componens after destructuring 'snoc'.
 export
 snocCong2 : {x : a} -> {xs : List a} -> {y : b} -> {ys : List b} ->
-            (xs ++ [x]) = (ys ++ [y]) -> (xs = ys, x = y)
+            (snoc xs x) = (snoc ys y) -> (xs = ys, x = y)
 
 snocCong2 {xs = []} {ys = []} Refl = (Refl, Refl)
 snocCong2 {xs = []} {ys = z :: ys} prf = snocCong2 prf
@@ -54,6 +54,6 @@ lengthDistributive [] ys = Refl
 lengthDistributive (x :: xs) ys = cong S $ lengthDistributive xs ys
 
 export
-lengthSnoc : (x : _) -> (xs : List a) -> length (xs ++ [x]) = S (length xs)
+lengthSnoc : (x : _) -> (xs : List a) -> length (snoc xs x) = S (length xs)
 lengthSnoc x [] = Refl
 lengthSnoc x (_ :: xs) = cong S (lengthSnoc x xs)
--- a/libs/contrib/Data/List/Equalities.idr
+++ b/libs/contrib/Data/List/Equalities.idr
@@ -17,14 +17,16 @@ consCong2 Refl Refl = Refl
 
 ||| Equal non-empty lists should result in equal componens after destructuring 'snoc'.
 export
-snocCong2 : (x, y : a) -> (xs, ys : List a) -> (xs ++ [x]) = (ys ++ [y]) -> (xs = ys, x = y)
-snocCong2 x x [] [] Refl = (Refl, Refl)
-snocCong2 x y [] (z :: ys) prf = snocCong2 x y [] (z :: ys) prf
-snocCong2 x y (w :: xs) [] prf = snocCong2 x y (w :: xs) [] prf
-snocCong2 x y (w :: xs) (z :: ys) prf =
+snocCong2 : {x : a} -> {xs : List a} -> {y : b} -> {ys : List b} ->
+            (xs ++ [x]) = (ys ++ [y]) -> (xs = ys, x = y)
+
+snocCong2 {xs = []} {ys = []} Refl = (Refl, Refl)
+snocCong2 {xs = []} {ys = z :: ys} prf = snocCong2 prf
+snocCong2 {xs = w :: xs} {ys = []} prf = snocCong2 prf
+snocCong2 {xs = w :: xs} {ys = z :: ys} prf =
   let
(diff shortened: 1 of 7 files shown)


libs/contrib/Data/List/Equalities.idr
```diff
@@ -0,0 +1,60 @@
...
+appendCong2 : {x1 : List a} -> {x2 : List a} ->
+              {y1 : List b} -> {y2 : List b} ->
+              x1 = y1 -> x2 = y2 -> x1 ++ x2 = y1 ++ y2
+appendCong2 {x1=[]} {y1=(_ :: _)} Refl _ impossible
+appendCong2 {x1=(_ :: _)} {y1=[]} Refl _ impossible
+appendCong2 {x1=[]} {y1=[]} _ eq2 = eq2
+appendCong2 {x1=(_ :: _)} {y1=(_ :: _)} eq1 eq2 =
+  consCong2
+    (fst $ consInjective eq1)
+    (appendCong2 (snd $ consInjective eq1) eq2)
```
gallais, line 42: Instead of calling `fst` and then `snd` on `consInjective eq1`,
why not use `let (hd, tl) = consInjective eq1 in`?

libs/contrib/Data/List/Equalities.idr
```diff
@@ -0,0 +1,60 @@
...
+consInjective Refl = (Refl, Refl)
+
+||| Two lists are equal, if their heads are equal and their tails are equal.
+consCong2 : {x : a} -> {xs : List a} -> {y : b} -> {ys : List b} ->
+            x = y -> xs = ys -> x :: xs = y :: ys
+consCong2 Refl Refl = Refl
+
+||| Equal non-empty lists should result in equal componens after destructuring 'snoc'.
+export
+snocCong2 : {x : a} -> {xs : List a} -> {y : b} -> {ys : List b} ->
```
gallais, line 20: I would argue this is a good opportunity to add `snoc` as a
definition (in `Data.List`, I guess?).

libs/contrib/Data/List/Equalities.idr
rgrover: Updated PR. Thanks. This was code borrowed from Idris1.

libs/contrib/Data/List/Equalities.idr
rgrover: Do you mean to suggest something like:

```idris
snoc : List a -> a -> List a
snoc [] x = [x]
snoc (y :: xs) x = y :: snoc xs x
```

gallais: I would be happy with it just being an alias for `xs ++ [x]`.

src/Idris/Main.idr
```diff
@@ -71,6 +71,7 @@ updateEnv
          defs <- get Ctxt
          addPkgDir "prelude"
          addPkgDir "base"
+         addPkgDir "contrib"
```
clayrat, line 74: I think `contrib` was deliberately excluded here - like in Idris1 you're supposed to load it manually.

rgrover: Thanks. I'll remove this.

libs/contrib/Data/List/Equalities.idr
gallais: Can we use the `snoc` definition in these theorems
now that it has been added?

libs/contrib/Data/List/Reverse.idr
```diff
@@ -0,0 +1,93 @@
...
+reverseAppend [] ys = sym (appendNilRightNeutral (reverse ys))
+reverseAppend (x :: xs) ys =
+  rewrite reverseCons x (xs ++ ys) in
+    rewrite reverseAppend xs ys in
+      rewrite reverseCons x xs in
+        sym $ appendAssociative (reverse ys) (reverse xs) [x]
+
+||| A recursive definition of reverse.
+export
+reverseRec : List a -> List a
```
gallais, line 45: I wonder whether it's worth calling this one `slowReverse`?
Or to declare it as a `0 reverse : ...` to make sure it never
shows up in people's runtime code?

ziman: I definitely like the idea of using `0` to mark specification functions.

On the other hand, maybe you sometimes just want to avoid unnecessary proofs when a simple/slow reverse would do.

ziman: Oh by the way, if this function is a spec, shouldn't it be `public export`?

rgrover: Thanks. Updated PR.

libs/contrib/Data/List/Equalities.idr
rgrover: yes. updated PR.

libs/contrib/Data/List/Equalities.idr
```diff
@@ -0,0 +1,61 @@
...
+mapAppendDistributive _ [] _ = Refl
+mapAppendDistributive f (x :: xs) ys =
+  cong (f x ::) $ mapAppendDistributive f xs ys
+
+||| List.length is distributive over appending.
+export
+lengthDistributive : (xs, ys : List a) -> length (xs ++ ys) = length xs + length ys
+lengthDistributive [] ys = Refl
+lengthDistributive (x :: xs) ys = cong S $ lengthDistributive xs ys
+
```
gallais, line 57: One lemma is called `mapAppendDistributive` but the other is called `lengthDistributive`
(no mention of `Append`). We should settle on one naming style and enforce it throughout
the library.

rgrover: Fixed. And also made `snocCong2` total.

edwinb:
Do we believe this is ready to merge? As a general rule, I'm happy for things in `contrib` to be entirely experimental so I don't want to be too fussy...

gallais:
Looks great to me!

PR #134, opened 2020-05-24
Minor sed fix to bootstrapping - for GNU sed?

kristleifur:
This makes the `sed` calls for bootstrapping work with GNU sed (tested against 4.8) by changing `-i ''` to `-i=''`

And as a bonus?: Also swap the `|` delimiter for `!` to avoid collisions with a symbol that the shell imbues with power. That's opinionated of me and I'm unsure if it's polite to do so but here goes.

(fwiw, the "meta-sed" command to do this in case it's useful:)
```fish
sed -i='' -r s/"([ \t]sed)(.*)(-i) ('')/\1\2\3=\4"/g\;s/"(sed .*s)\|(.+?)\|(.+?)\|(g)"/"\1!\2!\3!\4"/g Makefile
```

(I was play-hacking on idris2 last night – it was most thrilling to see bootstrapping stuff fly into place as I was exploring! at least that's how it appeared! There's much magic afoot these days ...)


edwinb:
This is the sort of thing where I'm not bothered how it looks as long as it works :). My only worry at the moment is that it hasn't run against the new macOS CI, and this is the sort of thing that tends to break on differene OSes. I wonder, is there a way to force this to rerun on the macOS CI? Maybe a force push?

kristleifur:
I’ll look into it from that perspective! The framing/context is useful to know where to look :) 🙏

gallais:
The merge (don't know if it's standard github or [refined github](https://github.com/sindresorhus/refined-github) but
I have a big button "Update Branch" that does this automatically)
was enough to start a new build with the macOS target included.

kristleifur:
Ah 💡 I'm going to close this PR; I misread the Makefile in my hurry to hack things together. I also understand the context everything is in a lot better. I'll close this and propose a slightly different mechanism, and leave the inconsequential `!` vs `|` stuff out.

Issue #63, opened 2020-05-20
one type annotation in let means all other definitions must have type annotation


edwinb:
<a href="https://github.com/shmish111"><img src="https://avatars2.githubusercontent.com/u/934267?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [shmish111](https://github.com/shmish111)**
_Sunday Apr 26, 2020 at 13:15 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/326_

----

If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!

# Steps to Reproduce

Check
```
f : a -> b
f a' =
    let x : a
        x = a'
        y = 1
    in ?xyz

g : a -> b
g a' =
    let x : a
        x = a'
        y : Int
        y = 1
    in ?xyz
```

# Expected Behavior

Type checks fine

# Observed Behavior

```
Errors (1)
Test.idr:217:9
While processing right hand side of f at Test.idr:214:1--220:1:
No type declaration for Test.y
```

It can be 'fixed' by having a separate let block for things with type annotations and things without. If this is expected behaviour then:

1. I couldn't find any documentation about it
2. I found it very unintuitive. Indeed it took me a while to work out that I could have an additional block with the unannotated items

edwinb:
<a href="https://github.com/gallais"><img src="https://avatars1.githubusercontent.com/u/1094905?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [gallais](https://github.com/gallais)**
_Sunday Apr 26, 2020 at 17:48 GMT_

----

This bit of the parser is responsible: once we have seen `let` we commit to
either a block of let-binders or a block of local definitions. We could instead
allow interleavings of the two.
 https://github.com/edwinb/Idris2/blob/19426ecd141f93c34268a53ed09b5e48e64ba73b/src/Idris/Parser.idr#L553-L570

edwinb:
<a href="https://github.com/edwinb"><img src="https://avatars1.githubusercontent.com/u/255926?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [edwinb](https://github.com/edwinb)**
_Monday Apr 27, 2020 at 12:50 GMT_

----

I think it's reasonable to allow interleavings of the two different `let` forms.

edwinb:
[...]

PR #136, opened 2020-05-24
[ re #63 ] interleaving of let-binder and local definitions

gallais:
I still need to add test cases. This may not quite fix the issue
yet because I was forced to grab a whole `nonEmptyBlock` of
`topDecl` after seeing one.

Otherwise,

```
let f : Ty
    f = g
```

gets parsed
as  the topDecl   `f : Ty`
and the letBinder `f = g`

Ideally we should post-process the `topDecl` blocks to group definitions
by name & turn the ones without a corresponding type declaration back
into mere let-binders.


gallais:
Brought up to speed. I still have a couple of failing test cases because I am not grouping
things properly e.g. I seem to group

```idris
let f : List A -> ()
    f [] = ()
    f (x :: xs) = ()
in ...
```

as

```idris
let f : List A -> ()
    f [] = ()
in
let f (x :: xs) = ()
in ...
```

which is clearly pretty bad. :sweat_smile:

gallais:
These merge conflicts are really horrible. I think I'll start from scratch.

PR #142, opened 2020-05-24
[prelude] Use `foldl` for sum and product

### LINKED ISSUE #132: Why are `sum` and `product` not tail recursive?
Today I noticed that `sum` and `product` are defined with `foldr` rather than `foldl`: https://github.com/idris-lang/Idris2/blob/master/libs/prelude/Prelude.idr#L626-L634

It makes sense with Haskell because it's lazy. But Idris is an eager language and would it make more sense to make them tail recursive so that they don't stack overflow when given a large list of numbers? Or am I missing something?

jinwoo:
To make them tail recursive. Fixes #132.


jinwoo:
As discussed in https://github.com/idris-lang/Idris2/issues/132, now provides `sum'` and `product'` separate from `sum` and `product`. `sum'` and `product'` are only `export` rather than `public export`.

jinwoo:
I'm not sure of the CI failures

```
Uncaught error: idris2.ipkg:60:5--60:15:IdrisPaths not found
Makefile:91: recipe for target 'clean' failed
```

Doesn't seem related to this PR.

wchresta:
Seems to me the build failures might be related to new tests added in https://github.com/idris-lang/Idris2/commit/1ea54e764be3ab3abcb161e6a1e7322a9a51cb70
For example:
```
Expected:
"1/1: Building wcov (wcov.idr)[..] Main.wbar is not covering [..] in 1204\n[...]"
Given:
"1/1: Building wcov (wcov.idr)[..] Main.wbar is not covering [..] in 1206\n[...]"
```

This might indicate some lines numbers might have changed due to your commit and you might have to run `make test` interactively and create new golden test cases.

edwinb:
Indeed, the tests are fine, they just need updating. We should fix the display so it doesn't leak the internal names in any case, some day...

jinwoo:
Okay. I rebased and updated the golden files.

jinwoo:
But the change in `tests/typedd-book/chapter14/expected` seems wrong to me:
```
Hangman.idr:66:1--67:1:hangman is not covering:
	Calls non covering function prim__strIndex
```

jinwoo:
Updated the golden files for real. I was struggling with that before realizing that I have to `make install` first before `make test` :)

Some golden file changes don't seem related to this PR. Some coverage tests are already failing HEAD. Hope my change fixes all of them.

buzden:
Just to save couple of minutes of your time, I realized that you did the same test change in `linear009` as I tried but it is not actual since #160 is merged. You'll need to revert `linear009` after rebasing.

jinwoo:
Thank you, @buzden . Done.

Issue #141, opened 2020-05-24
Problem running Idris2 on Windows shell


zenntenn:
When trying to run idris2 from the command line I get the error:

`Exception: (while loading libidris2_support.dll) The specified module could not be found`

Manually copying libwinpthread-1.dll from msys64\mingw64\bin to $PREFIX\bin\idris2_app\ fixes the issue as a workaround.

melted:
Fixed in #143

PR #143, opened 2020-05-24
Remove dependency on pthreads on Windows

melted:
This eliminates a dependency on libwinpthread-l that has been documented to cause confusion when it's absent. An alternative would be to always copy it around when we build. But as we only use nanosleep from there it is much better to eliminate that function on windows.

It's possible to do a bit better with a lot more work, but people will just have to make do with millisecond resolution.


melted:
Fixes #141

PR #149, opened 2020-05-25
Split bootstrap into separate build and test stage

### LINKED ISSUE #125: Separate testing from bootstrapping
Currently, `make bootstrap` will both build and test Idris2. The [test command is part of `bootstrap.sh`](https://github.com/idris-lang/Idris2/blob/master/bootstrap.sh#L49) which is also used to build.

To be able to separate building from testing for automatic builds (like the [nix expression we are working on](https://github.com/NixOS/nixpkgs/pull/88645)), it would be very useful to be able to expose bootstrapping/building and testing separately.

This request is to split `make bootstrap` into two targets: `bootstrap-build` and `bootstrap-test` such that they can be called as separate stages in automatic builds.

wchresta:
* Refactor bootstrap and bootstrap-rkt scripts
* Move the execution of the test phase after bootstrapping from bootstrap
  scripts into the Makefile. This allows separate execution of build
  and test seperately.
* Fixes #125

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,13 +1,18 @@
 #!/bin/sh
 
-if [ -z "$SCHEME" ]
+set -e  # Exit on any error
+
+echo "bootstrapping SCHEME=$SCHEME IDRIS2_VERSION=$IDRIS2_VERSION"
+if [ -z "$SCHEME" ] || [ -z "$IDRIS2_VERSION" ]
 then
-    echo "SCHEME not set. Invoke with SCHEME=[name of chez executable]"
+    echo "Required ENV not set."
     exit 1
 fi
 
 # Compile the bootstrap scheme
+# TODO: Move boot-build to Makefile in bootstrap/Makefile
 cd bootstrap
+echo "Building idris2-boot from idris2-boot.ss"
 ${SCHEME} --script compile.ss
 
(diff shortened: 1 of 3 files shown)


bootstrap.sh
```diff
@@ -1,13 +1,18 @@
 #!/bin/sh
 
-if [ -z "$SCHEME" ]
+set -e  # Exit on any error
+
+echo "bootstrapping SCHEME=$SCHEME IDRIS2_VERSION=$IDRIS2_VERSION"
+if [ -z "$SCHEME" ] || [ -z "$IDRIS2_VERSION" ]
 then
-    echo "SCHEME not set. Invoke with SCHEME=[name of chez executable]"
+    echo "Required ENV not set."
```
edwinb, line 8: I expect forgetting to set `SCHEME` will be a common error (I've done it myself many times) so could the error message still report specifically if `SCHEME` is missing? I guess it'll get `IDRIS2_VERSION` from the Makefile.

edwinb: Actually I'm just going to merge then update this I think.

wchresta: Thanks. I think in the future we should actually have a configure stage that searches for a suitable scheme without the user having to provide it manually.

PR #154, opened 2020-05-25
Read Idris version from a single file

ska80:
Instead of hardcoding it in all over build scripts and makefiles


melted:
As the bootstrap scripts are run from the makefile, it could export IDRIS_VERSION to them.

ska80:
> As the bootstrap scripts are run from the makefile, it could export IDRIS_VERSION to them.

Right.  I Will update shortly.

ska80:
I will close this PR as #149 already solved this problem

PR #157, opened 2020-05-25
Fix typos in documentation and comments

keram:


docs/source/tutorial/multiplicities.rst
```diff
@@ -112,7 +112,7 @@ If we use ``x`` for one part of the pair...
 The same happens if we try defining ``duplicate x = (?help, x)`` (try it!).
 
 In order to avoid parsing ambiguities, if you give an explicit multiplicity
-for a variable as with he argument to ``duplicate``, you need to give the
+for a variable as with the argument to ``duplicate``, you need to give the
```
keram, line 115: I have bit of trouble understand this sentence even though I think i get what it tries to say:
What about this @edwinb ?: 
```
In order to avoid parsing ambiguities, if you give an explicit multiplicity
for a variable as with the argument to ``duplicate``, you need to give it
a name too.
```

edwinb: That sounds better, yes

edwinb: Should I wait for this update before merging?

keram: will do in a sec 👍

edwinb: Thanks! I'll take a chance on CI still being okay with that :)

PR #161, opened 2020-05-25
Add missing totality annotations

fabianhjr:
In some cases the totality checker hanged and `%default covering` was used. Some cases the totality checker didn't detect totality when reducing recursivly on an argument and `assert_total`s were added.

### DIFF
--- a/src/Algebra.idr
+++ b/src/Algebra.idr
@@ -4,6 +4,8 @@ import public Algebra.ZeroOneOmega
 import public Algebra.Semiring
 import public Algebra.Preorder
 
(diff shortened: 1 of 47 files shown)


edwinb:
If there's places where it seems to be hanging it'd be useful to know so we can investigate... if you get time would you mind picking a representative example and putting it in an issue report? Ta.

fabianhjr:
Will do @edwinb, though I have to be sure it is hanging and not just taking too much time. (So will leave running on the simplest one for a bit and see if it typechecks)

PR #162, opened 2020-05-25
`idris2/linear009` test fix

buzden:


edwinb:
Hi, thanks for this - I'm doing a slightly different fix in #160 though (it should reduce the types).

buzden:
So am I right that it's means to inline the types when showing the hole's type as late as possible to make types look as close to what was written by hand as possible?

PR #167, opened 2020-05-26
Sync idris2api.ipkg with idris2.ipkg

ska80:


melted:
Looks good! Thanks!

gallais:
What's the difference between the packages? It won't build the executable if
we call `idris2 --build` on the API one?

melted:
yes, the API is for using it as a library

PR #168, opened 2020-05-26
Possible copy-paste mistake in typesfuns.rst

asdfjasdf:
In the intial code listing we currently have "x = 42" but in following listings "x == 9 * 9 + 13" and "x == 8 * 8 + 30" return true.


gallais:
Thanks!

Issue #140, opened 2020-05-24
Exception: attempt to apply non-procedure "Unrecognised input"


andylokandy:
# Steps to Reproduce

1. In an empty directory, add a file `Main.idr`:

```idris
module Main

-- comment out this line to disable the bug
import Text.Parser
import Language.JSON

main : IO ()
main = putStrLn $ show $ parse "{\"a\":2}"
```

2. Copy directory `src/Text` of Idris2 repo to the previous directory. 

3. Run `idris2 --exec main -p contrib Main.idr`.
# Expected Behavior

Output `Just {"a":2}`

# Observed Behavior

```
Exception: attempt to apply non-procedure #(9 0 0 #(1 "Unrecognised input" #<procedure ext-1 at _tmpchez.ss:83008>) #(1 "Unrecognised input" #<procedure ext-1 at _tmpchez.ss:83008>))
```

andylokandy:
Does it seem that the function `parse` of `libs/contrib/Text` is falsely unified with the one in `./Text`?

edwinb:
I can't reproduce this behaviour, sorry! Do you have everything up to date? Although I can't think of any recent changes that would have fixed it...

andylokandy:
The problem still exists in the latest commit. In addition, with gambit backend, it outputs `Segment fault`. I'll pack the demo up, and upload it somewhere ...

```
$ idris2 -v
Idris 2, version 0.2.0-1ea54e764
$ idris2 --exec main -p contrib Main.idr
Exception: attempt to apply non-procedure #(9 0 0 #(1 "Unrecognised input" #<procedure ext-1 at _tmpchez.ss:83468>) #(1 "Unrecognised input" #<procedure ext-1 at _tmpchez.ss:83468>))
$ idris2 --exec main -p contrib Main.idr --cg gambit
Segmentation fault
```

ziman:
I can reproduce it, too. Looking into it.

edwinb:
I'm still not getting the error!

    edwin@yaffle:~/Research/Idris/Issues/140$ cat Main.idr 
    module Main

    -- comment out this line to disable the bug
    import Text.Parser
    import Language.JSON

    main : IO ()
    main = putStrLn $ show $ parse "{\"a\":2}"
    edwin@yaffle:~/Research/Idris/Issues/140$ idris2 --exec main Main.idr -p contrib Main.idr
    Just {"a":2}

[...]

PR #169, opened 2020-05-26
Use Path in the compiler

andylokandy:
1. Move the changes in https://github.com/idris-lang/Idris2/pull/123 from `libs/contrib/System/Path.idr` to `src/Utils/Path.idr` due to the namespace resolution issue (https://github.com/idris-lang/Idris2/issues/140).

2. Refactor the compiler code to manipulate paths using `Utils.Path` instead of String appending.

I think the `Utils.Path` could eventually be moved into `base`, just like other utils modules in the compiler.


src/Utils/Path.idr
```diff
@@ -0,0 +1,437 @@
...
+dirSeparator = if isWindows then '\\' else '/'
+
+||| The character that separates multiple paths.
+export
+pathSeparator : Char
+pathSeparator = if isWindows then ';' else ':'
+
+||| Windows' path prefixes of path component.
+public export
+data Volumn
```
melted, line 32: Should that be Volume?

src/Utils/Path.idr
```diff
@@ -0,0 +1,437 @@
...
+import Text.Lexer
+import Text.Parser
+import Text.Quantity
+
+infixr 5 </>
+infixr 7 <.>
+
+
+isWindows : Bool
+isWindows = os `elem` ["windows", "mingw32", "cygwin32"]
```
melted, line 18: We should probably move this to System.Info, it exists in Core.Options too.

andylokandy: Yes! I'm still working in the process. Maybe I should change the title a bit...

melted:
You can press "Convert to draft" in the sidebar if you don't want anyone to merge it accidentally.

andylokandy:
> You can press "Convert to draft" in the sidebar if you don't want anyone to merge it accidentally.

But I can't run the checks in a draft.

melted:
No need of a PR for the checks, they should run in your repo too, check the Actions tab.

andylokandy:
@melted Thanks!

andylokandy:
@melted PTAL

melted:
It looks good! I'll wait with merging for a bit to see if there are any other comments.

PR #170, opened 2020-05-26
Add edits to Chapter 1 section

gdevanla:
Using Idrs 1.3.3, there is a disambigation error between Prelude.Strings.(++) and Prelude.List.(++).

```
Hello.idr:6:15-62:
  |
6 | showAve str = "The average length is " ++ (show (average str))
  |               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
When checking right hand side of showAve with expected type
        String

Can't disambiguate since no name has a suitable type:
        Prelude.List.++, Prelude.Strings.++
```
Using explicit (++) resolves this issue.

### DIFF
--- a/docs/source/typedd/typedd.rst
+++ b/docs/source/typedd/typedd.rst
@@ -39,6 +39,14 @@ In ``AveMain.idr`` and ``Reverse.idr`` add:
 .. code-block:: idris
 
     import System.REPL -- for 'repl'
+ 
+To resolve disambiguation, use ``Prelude.Strings.(++)`` while concatenating strings in ``showAverage``. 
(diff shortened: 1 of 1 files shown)


gdevanla:
Even though I propose this change, I am guessing there is a better way to do this. Open to suggestions.

edwinb:
The page isn't about changes needed in Idris 1.3.3, though, it's for changes needed in Idris 2.

gdevanla:
Great, thanks. I am not sure if this is applicable to Idris 2, since I have not tried it. Will close it.

PR #172, opened 2020-05-26
Remove unused imports from Idris.Main

fabianhjr:


src/Idris/Main.idr
```diff
@@ -1,35 +1,27 @@
 module Main
 
-import Core.Binary
-import Core.Context
+import Data.Strings
+import System
```
keram, line 4: Is there a convention on ordering imports in Idris ? I feel like the version before this change followed non-strict alphabetical order so i just wonder if it is convention or personal taste ?

fabianhjr: I wait, didn't mean to commit this, will ammend.

keram: haha, didn't realised this was mistake. But good work in general 👍

edwinb:
This one probably needs the CI to restart since the import fixing patch might have caught some problems...

fabianhjr:
Rebased all my open PRs and pushed.

PR #183, opened 2020-05-27
Remove clang from config.mk

keram:
Why:
As sugested in review
https://github.com/idris-lang/Idris2/pull/122#issuecomment-633284853
It is not necessary anymore and by removing it, it makes one step
easier for people who don't have clang installed to build the idris2.

Not sure if this is the best way to test but I ran `time make bootstrap SCHEME=scheme` successfuly 
before and after with these results:
Before:
```
real    4m9.884s
user    3m40.771s
sys     0m28.138s
```
after:
```
real    4m17.707s
user    3m47.847s
sys     0m28.751s
```
Which confirms there is no significant difference (reddit may disagree :))


keram:
Unrelated but by running few times `make bootstrap.. `make all..` make me wonder 
If we could(how to) fix these warnings clean way:
```
cc -Wall  -fPIC -O2   -c -o idris_directory.o idris_directory.c
idris_directory.c: In function ‘idris2_currentDirectory’:
idris_directory.c:11:4: warning: ignoring return value of ‘getcwd’, declared with attribute warn_unused_result [-Wunused-result]
   11 |    getcwd(cwd, 1024);
      |    ^~~~~~~~~~~~~~~~~
cc -Wall  -fPIC -O2   -c -o idris_buffer.o idris_buffer.c
idris_buffer.c: In function ‘idris2_readBufferFromFile’:
idris_buffer.c:145:5: warning: ignoring return value of ‘fread’, declared with attribute warn_unused_result [-Wunused-result]
  145 |     fread((buf->data), sizeof(uint8_t), (size_t)len, f);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ar rc libidris2_support.a idris_file.o idris_support.o idris_net.o getline.o idris_directory.o idris_buffer.o
ranlib libidris2_support.a
```
My C knowledge is pretty limited so I find this as good learning oportunity to learn from more experienced. 
Thank you.

edwinb:
The `clang` default was just because there was a significant difference in build times when we used Idris1's C runtime. I doubt it'll make any difference at all now.

We should tidy up the warnings though... I missed them because `clang` didn't complain.

PR #199, opened 2020-05-30
added more functions to Control.App.Console

karroffel:
Now all common console IO functions available from the
prelude are available through the `Control.App.Console`
interface.

Added:
- putChar
- getChar
- getCharLn
- print
- printLn

Renamed:
- getStr to getLine


edwinb:
Thanks! Good to see these being extended, I'd like to experiment more with the App interfaces.

PR #200, opened 2020-05-30
Allow absolute paths in --find-ipkg

edwinb:
Otherwise they need to be relative to know where the root of the module
tree is, so report an error with absolute paths in other cases.

### DIFF
--- a/src/Core/Core.idr
+++ b/src/Core/Core.idr
@@ -126,6 +126,7 @@ data Error : Type where
      CyclicImports : List (List String) -> Error
      ForceNeeded : Error
      InternalError : String -> Error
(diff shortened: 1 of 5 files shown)


andylokandy:
I think it's too strict to force to use a relative path. If the absolute path doesn't have a common part with source dir, I'd like to throw an `UserError` in `pathToNs`.

edwinb:
If you invoke Idris at the command line, the current location is expected to be the root of the module tree, and so far relative paths have always been assumed. So this merely checks that the existing assumption holds, and fixes the problem with idris2-vim. Further improvements welcome!

PR #207, opened 2020-05-31
Make assert_smaller and assert_total linear

mb64:
This is necessary in order to `assert_total` a linear function.


mb64:
I added an explanation for the multiplicity 0 argument in assert_smaller, but for the most part, I think these functions should Just Work™.

If you're not writing a linear function, you won't care about the multiplicities – they don't affect anything (as evidenced by the fact that all the existing code using these functions was oblivious to the change in multiplicity).

If you are writing a linear function, the multiplicity 0 argument might be a little weird, so it's worth mentioning in the docs, but other than that, the functions are very unsurprising: they don't affect how many times things are consumed.

I'm not sure why the CI is failing – I don't know what changing the docs could have done to alter the compiler output.

clayrat:
Regarding the CI it looks like removing the implicit declarations has shifted the internal numbering scheme for other implicits in Builtin/Prelude, and the ideMode tests rely on this internal numbering. It's probably ok to just replace the expected values for 3 ideMode test with the new output.

gallais:
> If you are writing a linear function, the multiplicity 0 argument might be a little weird, so it's worth mentioning in the docs, but other than that, the functions are very unsurprising: they don't affect how many times things are consumed.

My motivation is that this may be the first contact with quantities a newcomer
will have: they write a Haskell-like function, get told off by the termination
checker and learn about `assert_*` but are puzzled by the types.

I agree that the quantities are unsurprising *once you have a good understanding
of their meaning*.

mb64:
That's a very good point.  I rebased, updated tests, and did my best to make the documentation as clear as I could.  It's a hard thing to explain, though, so if you have suggestions for improvements I would be happy to change it.

PR #217, opened 2020-06-01
Factors of natural numbers

Sventimir:
This introduces to the `contrib` package the notion of a factor of a natural number (in `Data.Nat.Factor`) and of g.c.d. of two natural numbers. Also several important properties are proven. The purpose of these definitions is to build a decent implementation of rational numbers, similar to Haskell's `Data.Ratio` module, but ensuring that only one representation of each ratio is valid and – of course – prove all algebraic properties of rationals.

Module `Data.Fin.Extra` is ported from Idris1, but one theorem there fails to type check for some reason I couldn't work out. I decided to leave it commented out. If anyone can explain, why it does not work, I'll be happy to fix and uncomment it.

I'm also not convinced to the name of `Data.Nat.Equational` module. If anyone can suggest a better name, I'll gladly change it to something more appropriate. Maybe these theorems should just be moved to `Data.Nat` in `base`?


melted:
Looks good, let's merge.

Issue #76, opened 2020-05-20
Can't pass argument of type %World to foreign function


edwinb:
<a href="https://github.com/mokshasoft"><img src="https://avatars1.githubusercontent.com/u/19352700?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mokshasoft](https://github.com/mokshasoft)**
_Tuesday May 12, 2020 at 06:21 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/368_

----

Error when compiling an FFI function with a callback function of type PrimIO (), using Gambit but not Chez. A callback function of String -> PrimIO () seems to work better.

https://gist.github.com/mokshasoft/e0d3b39b2be5bbd766d93605a4d7e48d

# Steps to Reproduce
1. Run the Makefile in the gist.
```shell
idris2 -c --cg chez NoBug1.idr
idris2 -c --cg gambit NoBug1.idr
idris2 -c --cg chez Bug1.idr
idris2 -c --cg gambit Bug1.idr
idris2 -c --cg chez NoBug1.idr -o c_nobug1
compiling <dir>/c_nobug1.ss with output to <dir>/c_nobug1.so
idris2 -c --cg gambit NoBug1.idr -o g_nobug1
sh: -exe: command not found
idris2 --cg chez Bug1.idr -o c_bug1
compiling <dir>/c_bug1.ss with output to <dir>/c_bug1.so
idris2 --cg gambit Bug1.idr -o g_bug1
Uncaught error: Bug1.idr:1:1--4:1:Can't pass argument of type %World to foreign function
make: *** [Makefile:10: bug1] Error 1
```

# Expected Behavior
If passing functions of type String -> PrimIO () to an FFI works it seems like passing functions of type PrimIO () should also work.

# Observed Behavior
Both files pass the --check compile step. The Gambit CG cannot handle FFI callbacks of PrimIO () and prints error "Uncaught error: Bug1.idr:1:1--4:1:Can't pass argument of type %World to foreign function".

edwinb:
<a href="https://github.com/edwinb"><img src="https://avatars1.githubusercontent.com/u/255926?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [edwinb](https://github.com/edwinb)**
_Wednesday May 13, 2020 at 11:48 GMT_

----

[...]

PR #219, opened 2020-06-01
Fix "Can't pass argument of type %World to foreign function" issue in gambit ffi

normalcoder:
This resolves https://github.com/idris-lang/Idris2/issues/76


edwinb:
Thanks!

PR #223, opened 2020-06-02
Try to load libidris2_support on older Windows

dkeehl:
`GetSystemTimePreciseAsFileTime ` was introduced in Windows 8. On earlier Windows systems, it might compile (with a newer SDK), but the result dll cannot be loaded. For these systems, use `GetSystemTimeAsFileTime ` instead.


melted:
Thanks!

PR #227, opened 2020-06-02
add Data.Fuel

clayrat:


edwinb:
I guess this is a useful and common enough idiom that it's handy to have in the library. Thanks!

PR #232, opened 2020-06-03
Remove redundant thin functions and replace with insertNames

mb64:
There are three functions; `Core.TT.thin`, `Core.CaseTree.thinTree`, and `Core.CompileExpr.thin`; which duplicate much of the code of `Core.TT.insertNames`, `Core.CaseTree.insertCaseNames`, and `Core.CompileExpr.insertNames`.

This PR removes them, and replaces their (very infrequent) use by equivalent uses of `insertNames`.

I'm not sure how much you try to keep a stable module API, but this would be a breaking change.  If that's something that's important, I can add short wrapper functions in place of the originals:

```idris
%inline
export
thin : {outer, inner : _} ->
       (n : Name) -> Term (outer ++ inner) -> Term (outer ++ n :: inner)
thin n t = insertNames [n] t
```

### DIFF
--- a/src/Compiler/CompileExpr.idr
+++ b/src/Compiler/CompileExpr.idr
@@ -370,9 +370,9 @@ mutual
 --                              let env : SubstCEnv args (MN "eff" 0 :: vars)
 --                                      = mkSubst 0 (CLocal fc First) pos args in
 --                              do sc' <- toCExpTree n sc
---                                 let scope = thin {outer=args}
---                                                  {inner=vars}
---                                                  (MN "eff" 0) sc'
+--                                 let scope = insertNames {outer=args}
+--                                                         {inner=vars}
+--                                                         [MN "eff" 0] sc'
 --                                 pure $ Just (CLet fc (MN "eff" 0) False scr
 --                                                   (substs env scope))
                 _ => pure Nothing -- there's a normal match to do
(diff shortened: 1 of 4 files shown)


edwinb:
This looks fine to me, I think it is good to simplify these things. One question I will always ask just to confirm, though - does this affect the performance of "make all"? I expect it will make no difference at all, but it can be easy to miss things so I've got a bit cautious with any changes to the core!

mb64:
I rebased and compiled three times.  There doesn't seem to be any noticeable difference.

```shell
$ time idris2 --build idris2.ipkg # master building master

real	2m45.264s
user	2m28.902s
sys	0m15.947s

$ git checkout no-thin
$ make clean support src/IdrisPaths.idr
$ time idris2 --build idris2.ipkg # master building no-thin

real	2m46.542s
user	2m31.053s
sys	0m15.064s

$ # install idris from no-thin ...
$ make clean support src/IdrisPaths.idr
$ time idris2 --build idris2.ipkg # no-thin building no-thin

real	2m44.911s
user	2m29.589s
sys	0m14.879s
```

edwinb:
Great, thanks! I didn't expect anything measurable, but I do want to encourage the habit since I've been surprised in the past... I'll merge now.

PR #238, opened 2020-06-05
Fix test reporting when in interactive mode.

jfdm:
The current approach has the wrong boolean test.
When the command executed cleanly it will return 0.
A non-zero answer implies that `git diff` errored.

### DIFF
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -198,8 +198,8 @@ runTest opts testPath
               ]
             Just exp => do
               putStrLn "Golden value differs from actual value."
-              code <- system "git diff expected output"
-              when (code /= 0) $ printExpectedVsOutput exp out
+              code <- system "git diff --exit-code expected output"
+              when (code < 0) $ printExpectedVsOutput exp out
               putStrLn "Accept actual value as new golden value? [yn]"
           b <- getAnswer
           when b $ do Right _ <- writeFile "expected" out


gallais:
Thank you! This was driving me mad.

I wonder whether `--word-diff=color` would be a better default than git's
brutal line-based diffs. When I was looking at the diffs introduced by refolding
natural numbers it was often hard to spot the `Z` that turned into a `0`.

jfdm:
Actually, my original intuition was a little off.

The above was wrong!

When runing in interactive mode, the diff between expected and output is not being show,
Supposedly, Git-diff returns 0 on successful run and a non-zero for error.
However, this is not being picked up by the test suite.

We change git-diff's error reporting to be more diff like in which:

+ +ve indicates that there is a diff.
+ 0 indicates the files are the same.
+ -ve program error.

This should make diff reporting more robust.

jfdm:
I second the word-diff, but tbh the default git diffs are not showing when I was running the test suite. I do not know what I happening here.

gallais:
I am confused by your last message: is this PR ready? If so I'd be happy
to merge it ASAP (& I can then add the `--word-diff=color` in another
commit if you think this one shouldn't bring that change in too).

jfdm:
Sorry, for being late on this: twas a long weekend.

Please do merge and extend.

Of note, my local version still wasn't showing the results of the git diff. This way ensures that the fall back manual presentation is working as expected.

gallais:
I am still not getting any diff. Strange.

PR #239, opened 2020-06-05
add parens for some TTImp Shows

MarcelineVQ:
helps with readability since these, especially named-IPi, come up a lot
didn't change everything that could need it like PiInfo or BindMode
PiInfo rarely has DefImplicit (so far) and BindMode hasn't come up a lot (so far)
reduced indentation for TTImp Show implementation

change example:
from `%pi Rig0 Implicit Just {arg:7394} _ (%pi RigW AutoImplicit Nothing ...`
to `%pi Rig0 Implicit (Just {arg:7394}) _ (%pi RigW AutoImplicit Nothing ...`


MarcelineVQ:
oop, I shoulda ran the tests after that, I'll fix that

src/TTImp/TTImp.idr
```diff
@@ -120,62 +120,67 @@ mutual
        Unique : AltType
        UniqueDefault : RawImp -> AltType
 
+  private
+  wrapMaybe : Show a => Maybe a -> String
+  wrapMaybe x@(Just _) = "(" ++ show x ++ ")"
+  wrapMaybe x@Nothing  = show x
+
```
fabianhjr, line 127: I think this would be better on src/Utils/String.idr in case someone might want to reuse it on other Show instances.

MarcelineVQ: Yeah I was no the fence about that. Wasn't sure if it was too specific a use to put in a Util module. I don't mind adding it though.
Can you think of a better name than wrapMaybe? wrapShowMaybe is feeling a little verbose though it's direct

fabianhjr: Oh now that I think about naming it might be a good use case for named implementations

`[wrapJust] Show (Maybe a)` and then when used it would be `show @{wrapJust}`

MarcelineVQ: Neat application of instances but don't think I'd want to have an extra Show instance hanging around in the codebase just for that, in the end I'm just wrapping two uses of Just :>
(modular) wrapping functions have more potential for re-use than a specific Show instance and are slightly more discoverable since people are likely to look in Utils like you did.

MarcelineVQ:
I just saw that we have showPrec! So this patch gets quite a bit more direct.
@edwinb Is it correct etiquette to squish these commits down to one or do you not care either way?

fabianhjr:
It is great etiqutte to rebase and squash as appropriate :3

edwinb:
`showPrec` is definitely the way to go here. I think I was in some kind of hacking frenzy when I bashed that instance out. Deriving will be nice :). Anyway, this looks much neater, thanks.

I'm not especially fussy about squashing commits. Others may be. I'm happy to leave it to your Skill and Judgement in general.

PR #242, opened 2020-06-05
Add Algebra interfaces and laws

nickdrozd:
This PR ports over some Algebra stuff from Idris 1. It is mostly similar: the interfaces are split into syntactic versions and verified versions. The latter are named like `VMonoid` etc. That can be changed, but I think it's the most ergonomic name.

Many of the theorems are stated but not proved. They can all be proved in Idris 1 [1], but the proofs don't work here. I would expect them all to work. There seem to be a few different bugs involved, but I don't understand the details.

[1] See https://github.com/idris-lang/Idris-dev/blob/master/libs/contrib/Control/Algebra/Laws.idr and https://github.com/idris-lang/Idris-dev/pull/4848

### DIFF
--- a/libs/contrib/Control/Algebra/Laws.idr
+++ b/libs/contrib/Control/Algebra/Laws.idr
@@ -1,0 +1,331 @@
+module Control.Algebra.Laws
+
+import Prelude
+import Control.Algebra
+
+%default total
+
+-- Monoids
+
+||| Inverses are unique.
+public export
+uniqueInverse : MonoidV ty => (x, y, z : ty) ->
+  y <+> x = neutral {ty} -> x <+> z = neutral {ty} -> y = z
+uniqueInverse x y z p q =
+  rewrite sym $ monoidNeutralIsNeutralL y in
+    rewrite sym q in
+      rewrite semigroupOpIsAssociative y x z in
+  rewrite p in
+    rewrite monoidNeutralIsNeutralR z in
+      Refl
+
+-- Groups
+
+||| Only identity is self-squaring.
+public export
+selfSquareId : Group ty => (x : ty) ->
+  x <+> x = x -> x = neutral {ty}
+selfSquareId x p =
+  rewrite sym $ monoidNeutralIsNeutralR x in
+    rewrite sym $ groupInverseIsInverseR x in
+  rewrite sym $ semigroupOpIsAssociative (inverse x) x x in
+    rewrite p in
+      Refl
+
+||| Inverse elements commute.
+public export
+inverseCommute : Group ty => (x, y : ty) ->
+  y <+> x = neutral {ty} -> x <+> y = neutral {ty}
+inverseCommute x y p = selfSquareId (x <+> y) prop where
+  prop : (x <+> y) <+> (x <+> y) = x <+> y
+  prop =
+    rewrite sym $ semigroupOpIsAssociative x y (x <+> y) in
+      rewrite semigroupOpIsAssociative y x y in
+    rewrite p in
+      rewrite monoidNeutralIsNeutralR y in
+        Refl
+
+||| Every element has a right inverse.
+public export
+groupInverseIsInverseL : Group ty => (x : ty) ->
+  x <+> inverse x = neutral {ty}
+groupInverseIsInverseL x =
+  inverseCommute x (inverse x) (groupInverseIsInverseR x)
+
+||| -(-x) = x
+public export
+inverseSquaredIsIdentity : Group ty => (x : ty) ->
+  inverse (inverse x) = x
+-- inverseSquaredIsIdentity x =
+--   let x' = inverse x in
+--     uniqueInverse
+--       x'
+--       (inverse x')
+--       x
+--       (groupInverseIsInverseR x')
+--       (groupInverseIsInverseR x)
+
+||| If every square in a group is identity, the group is commutative.
+public export
+squareIdCommutative : Group ty => (x, y : ty) ->
+  ((a : ty) -> a <+> a = neutral {ty}) ->
+  x <+> y = y <+> x
+-- squareIdCommutative x y p =
+--   let
+--     xy = x <+> y
+--     yx = y <+> x
+--       in
+--   uniqueInverse xy xy yx (p xy) prop where
+--     prop : (x <+> y) <+> (y <+> x) = neutral {ty}
+--     prop =
+--       rewrite sym $ semigroupOpIsAssociative x y (y <+> x) in
+--         rewrite semigroupOpIsAssociative y y x in
+--       rewrite p y in
+--         rewrite monoidNeutralIsNeutralR x in
+--           p x
+
+||| -0 = 0
+public export
+inverseNeutralIsNeutral : Group ty =>
+  inverse (neutral {ty}) = neutral {ty}
+-- inverseNeutralIsNeutral {ty} =
+--   let e = neutral {ty} in
+--     rewrite sym $ cong inverse (groupInverseIsInverseL e) in
+--       rewrite monoidNeutralIsNeutralR $ inverse e in
+--         inverseSquaredIsIdentity e
+
+||| -(x + y) = -y + -x
+public export
+inverseOfSum : Group ty => (l, r : ty) ->
+  inverse (l <+> r) = inverse r <+> inverse l
+-- inverseOfSum {ty} l r =
+--   let
+--     e = neutral {ty}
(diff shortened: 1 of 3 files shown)


nickdrozd:
I've included commented-out the proofs that I would expect to work, but don't. In some cases, the mirror image of the proof does work. For instance, this works:

```idris
cancelLeft : VGroup ty => (x, y, z : ty) ->
  x <+> y = x <+> z -> y = z
cancelLeft x y z p =
  rewrite sym $ monoidNeutralIsNeutralR y in
    rewrite sym $ groupInverseIsInverseR x in
      rewrite sym $ semigroupOpIsAssociative (inverse x) x y in
        rewrite p in
      rewrite semigroupOpIsAssociative (inverse x) x z in
    rewrite groupInverseIsInverseR x in
  monoidNeutralIsNeutralR z
```

But this doesn't:

```idris
cancelRight : VGroup ty => (x, y, z : ty) ->
  y <+> x = z <+> x -> y = z
cancelRight x y z p =
  rewrite sym $ monoidNeutralIsNeutralL y in
    rewrite sym $ groupInverseIsInverseL x in
      rewrite semigroupOpIsAssociative y x (inverse x) in
        rewrite p in
      rewrite sym $ semigroupOpIsAssociative z x (inverse x) in
    rewrite groupInverseIsInverseL x in
  monoidNeutralIsNeutralL z
```

libs/contrib/Control/Algebra/Laws.idr
```diff
@@ -0,0 +1,332 @@
...
+--     rewrite sym $ groupInverseIsInverseR x in
+--   rewrite sym $ semigroupOpIsAssociative (inverse x) x x in
+--     rewrite p in
+--       Refl
+
+||| Inverse elements commute.
+public export
+inverseCommute : VGroup ty => (x, y : ty) ->
+  y <+> x = neutral {ty} -> x <+> y = neutral {ty}
+-- inverseCommute x y p = selfSquareId (x <+> y) prop where
```
rgrover, line 39: same as above. this can be uncommented.

libs/contrib/Control/Algebra/Laws.idr
```diff
@@ -0,0 +1,332 @@
...
+    rewrite monoidNeutralIsNeutralR z in
+      Refl
+
+-- Groups
+
+||| Only identity is self-squaring.
+public export
+selfSquareId : VGroup ty => (x : ty) ->
+  x <+> x = x -> x = neutral {ty}
+-- selfSquareId x p =
```
rgrover, line 28: Using Idris2 from 4aa1c1f44, this definition, when uncommented, compiles for me.

libs/contrib/Control/Algebra/Laws.idr
```diff
@@ -0,0 +1,332 @@
...
+--       rewrite semigroupOpIsAssociative y x y in
+--     rewrite p in
+--       rewrite monoidNeutralIsNeutralR y in
+--         Refl
+
+||| Every element has a right inverse.
+public export
+groupInverseIsInverseL : VGroup ty => (x : ty) ->
+  x <+> inverse x = neutral {ty}
+-- groupInverseIsInverseL x =
```
rgrover, line 52: can be uncommented (see above)

libs/contrib/Control/Algebra/Laws.idr
```diff
@@ -0,0 +1,332 @@
...
+--   inverseCommute x (inverse x) (groupInverseIsInverseR x)
+
+||| -(-x) = x
+public export
+inverseSquaredIsIdentity : VGroup ty => (x : ty) ->
+  inverse (inverse x) = x
+-- inverseSquaredIsIdentity x =
+--   let x' = inverse x in
```
rgrover, line 60: Idris2 isn't able to use `x' = inverse x` in its normalization. The following also reproduces the issue (the copiler complains about `prf`):
```idris
let x' = inverse x
    prf = the (x' = inverse x) Refl
 in ...
```

edwinb:
Thanks for these. I'd like to take a look at the ones which fail before merging this, just to avoid having things with types but no definitions in the library.

Also some bikeshedding - I'm not sure about having the `V` as a prefix, my preference would be for a suffix for tab completion reasons. I expect there's as many opinions on naming as there are people, however...

edwinb:
Actually it looks as if @rgrover has worked out all the problems, thanks for that! The `let` one is decribed here: https://idris2.readthedocs.io/en/latest/updates/updates.html#let-bindings

It's an annoyance, but better for overall consistency, with the core language as it stands.

libs/contrib/Control/Algebra/Laws.idr
rgrover: Based on what @edwinb suggested, we could aid Idris2 in the equality between `x'` and `inverse x` using a local function like:

```idris
||| -(-x) = x
public export
inverseSquaredIsIdentity : VGroup ty => (x : ty) ->
  inverse (inverse x) = x
inverseSquaredIsIdentity x =
  let
    lemma : (x' : ty) -> (x' = inverse x) -> inverse (inverse x) = x
    lemma x' prf =
      uniqueInverse
        x'
        (inverse (inverse x))
        x
        (rewrite sym prf in groupInverseIsInverseR x')
        (rewrite prf in groupInverseIsInverseR x)
   in lemma (inverse x) Refl
```

nickdrozd:
@edwinb I changed `VMonoid` to `MonoidV`, etc.

nickdrozd:
@rgrover I wasn't able to get any of the proofs to go through locally. I uncommented the proofs and pushed to see how the CI would react. It has failed, but I can't see the output.

libs/contrib/Control/Algebra/Laws.idr
nickdrozd:
```
Control/Algebra/Laws.idr:45:7--48:1:While processing right hand side of inverseCommute at Control/Algebra/Laws.idr:39:1--48:1:
While processing right hand side of inverseCommute,prop at Control/Algebra/Laws.idr:41:3--48:1:
Rewriting by neutral <+> ?ty x y p (\{rwarg:1434} => (rwarg = x <+> y)) (sym (semigroupOpIsAssociative x y (x <+> y))) (\{rwarg:1451} => (x <+> rwarg = x <+> y)) (semigroupOpIsAssociative y x y) (\{rwarg:1459} => (x <+> (rwarg <+> y) = x <+> y)) p = ?ty x y p (\{rwarg:1434} => (rwarg = x <+> y)) (sym (semigroupOpIsAssociative x y (x <+> y))) (\{rwarg:1451} => (x <+> rwarg = x <+> y)) (semigroupOpIsAssociative y x y) (\{rwarg:1459} => (x <+> (rwarg <+> y) = x <+> y)) p did not change type x <+> (neutral <+> y) = x <+> y
```

libs/contrib/Control/Algebra/Laws.idr
nickdrozd: Using 6898965a:
```
Control/Algebra/Laws.idr:33:7--35:1:While processing right hand side of selfSquareId at Control/Algebra/Laws.idr:28:1--35:1:
	neutral
and
	inverse x <+> x
```

libs/contrib/Control/Algebra/Laws.idr
nickdrozd:
```
Control/Algebra/Laws.idr:53:33--53:57:While processing right hand side of groupInverseIsInverseL at Control/Algebra/Laws.idr:52:1--55:1:
	Constraint (Monoid ty) (Constraint (MonoidV ty) conArg)
and
	Constraint (Monoid ty) (Constraint (Group ty) conArg)
```

libs/contrib/Control/Algebra/Laws.idr
nickdrozd: I suggest we avoid leaning on workarounds like this. As a user, I don't want to have to work around quirks in the type system. Ideally I would be able to express what I want to express without feeling like I have to walk on eggshells to satisfy the typechecker.

Especially in this particular case, the existing proof is clean and elegant, and this change obscures that.

nickdrozd:
I got rid of all the `let` bindings and uncommented the proofs (except for `inverseOfSum`, which I couldn't get to go through). There are no holes left, so unless anyone has any objections this should be good to go.

edwinb:
Sorry, this has been lingering for a while - everything looks fine to me though. Thanks for the contribution!

PR #244, opened 2020-06-06
Update docs to reflect changes to ipkg format

xoltar:
Just some of the differences I stumbled upon while trying to compile lightyear. There might be others!


edwinb:
Thanks for the update!

PR #245, opened 2020-06-06
Added fix to recvAll that doesn't use signal (Left 0)

FrederikVigen:
We have tried to use the recvAll functionality in the Network.Socket library from the contrib package.
It seems that when it runs recursively into the recv on an empty buffer, it is blocking and not receiving the clean exit (Left 0) signal, thus running infinitely and never returning.
We propose a solution to this issue that instead of waiting for a Left 0 signal, the function runs recursively until then buffer size is smaller than the reading size that it is trying to achieve.


edwinb:
Thanks for the fix. Can someone who knows more about networking than me give an opinion on whether this is the right way to solve the problem? Or, at least, better than what we currently have?

melted:
Looks OK to me!

PR #248, opened 2020-06-07
":let" command for REPL

vilunov:


edwinb:
Looks good, nice to see more of the REPL features from Idris 1 coming in. Thanks!

PR #252, opened 2020-06-08
Case block improvements

edwinb:
* Productivity checker wasn't looking under case, now it does
* Free standing case blocks assumed to return (), as in Idris 1

### DIFF
--- a/src/Core/Termination.idr
+++ b/src/Core/Termination.idr
@@ -157,6 +157,9 @@ mutual
       = findSC defs env g pats tm
   findSC defs env g pats tm
       = do let (fn, args) = getFnArgs tm
+           -- if it's a 'case' or 'if' just go straight into the arguments
+           Nothing <- handleCase fn args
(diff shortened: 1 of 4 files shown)


src/Core/Termination.idr
```diff
@@ -186,6 +189,14 @@ mutual
                  do scs <- traverse (findSC defs env Unguarded pats) args
                     pure (concat scs)
       where
+        handleCase : Term vars -> List (Term vars) -> Core (Maybe (List SCCall))
+        handleCase (Ref fc nt n) args
+            = do n' <- toFullNames n
+                 if caseFn n'
```
andrevidela, line 195:
```suggestion
            = if caseFn !(toFullNames n)
```
What about this?

edwinb: Possibly, but it elaborates to the same, and any day now there'll be another check on n' to see if it's the "if..then..else" function.

PR #257, opened 2020-06-08
Add finalisers to Chez and Racket back ends

edwinb:
This involves new primitives GCPtr and GCAnyPtr which are pointer types that have finalisers attached. The finalisers are run when the associated pointer goes out of scope, or at the end (so there's now an explicit call to the GC at the end).

In the test, I am assuming that the GC will only be called once, right at the end. Otherwise, the output isn't guaranteed to be deterministic! Let's see how this assumption holds...

I expect it'll be easy enough to add to the Gambit back end, but I'll wait for a patch for that.


melted:
I believe there should be a collect-request-handler in there for this to work as expected.

Issue #253, opened 2020-06-08
A view argument is not visible


jinwoo:
The `groupByN` function from exercise 1 in Chapter 10 of Type-driven Development with Idris doesn't work with Idris2.

Is this a valid language change in Idris2, or is it a bug in the compiler implementation?

# Steps to Reproduce

```idris
data TakeN : List a -> Type where
  Fewer : TakeN xs
  Exact : (n_xs : List a) -> TakeN (n_xs ++ rest)

takeN : (n : Nat) -> (xs : List a) -> TakeN xs
takeN 0 _ = Exact []
takeN (S _) [] = Fewer
takeN (S k) (x :: xs) = case takeN k xs of
                             Fewer => Fewer
                             Exact n_xs => Exact (x :: n_xs)

groupByN : (n : Nat) -> (xs : List a) -> List (List a)
groupByN n xs with (takeN n xs)
  groupByN n xs | Fewer = [xs]
  groupByN n (n_xs ++ rest) | Exact n_xs = n_xs :: groupByN n rest
```

# Expected Behavior

The code must be built without errors.

# Observed Behavior

The compiler says that `rest` is not accessible.

```
1/1: Building ex_10_1 (ex_10_1.idr)
ex_10_1.idr:15:63--16:1:While processing right hand side of with block in groupByN at ex_10_1.idr:15:3--16:1:
rest is not accessible in this context
Error(s) building file ex_10_1.idr
```

I have to modify the code like below to make it work. That is, change `rest` to an explicit argument to `Exact`.

```idris
data TakeN : List a -> Type where
  Fewer : TakeN xs
  Exact : (n_xs : List a) -> (rest : List a) -> TakeN (n_xs ++ rest)

takeN : (n : Nat) -> (xs : List a) -> TakeN xs
takeN 0 rest = Exact [] rest
takeN (S _) [] = Fewer
takeN (S k) (x :: xs) = case takeN k xs of
                             Fewer => Fewer
                             Exact n_xs rest => Exact (x :: n_xs) rest

groupByN : (n : Nat) -> (xs : List a) -> List (List a)
groupByN n xs with (takeN n xs)
  groupByN n xs | Fewer = [xs]
  groupByN n (n_xs ++ rest) | Exact n_xs rest = n_xs :: groupByN n rest
```

jinwoo:
[...]

PR #260, opened 2020-06-09
Add some comments about `TakeN` in chapter 10.

jinwoo:
The `rest` argument must be made explicit. Related to
https://github.com/idris-lang/Idris2/issues/253.


jinwoo:
I also notice that `Data.List.Views.SnocList` also has `x` and `xs` as explicit arguments: https://github.com/idris-lang/Idris2/blob/master/libs/base/Data/List/Views.idr#L80.

Would it be better to make them explicit implicit?

PR #262, opened 2020-06-09
Integer division

asandroq:
Add more primitives for number-theoretic integer division.


src/Core/Primitives.idr
```diff
@@ -65,33 +65,25 @@ castInt _ = Nothing
 castBits8 : Vect 1 (NF vars) -> Maybe (NF vars)
 castBits8 [NPrimVal fc (BI i)]
     = let max = prim__shl_Int 1 8 in
-          if i > 0 -- oops, we don't have `rem` yet!
-             then Just (NPrimVal fc (B8 (fromInteger i `mod` max)))
-             else Just (NPrimVal fc (B8 (max + fromInteger i `mod` max)))
+          Just (NPrimVal fc (B8 (fromInteger i `rem` max)))
```
chrrasmussen, line 68: I think the comment was meant to indicate that `mod` was implemented as `rem`, thus the workaround. Can you change the code from `rem` to `mod`, now that you have fixed the implementation of `mod`? (Also for `castBits16`, `castBits32`, `castBits64`.)

tests/chez/chez021/expected
```diff
@@ -12,7 +12,7 @@ False
 4294967295
 18446744073709551615
 1/1: Building Bits (Bits.idr)
-Main> ["257", "200", "248", "1", "255", "200", "254"]
+Main> ["257", "200", "-8", "1", "-1", "200", "254"]
 Main> ["True", "True"]
-Main> ["255", "65535", "4294967295", "18446744073709551615"]
+Main> ["-1", "-1", "-1", "-1"]
```
chrrasmussen, line 17: I think the expected values were correct before, as `Bits8` (through `Bits64`) are supposed to be unsigned integers. If you change the `castBitsNN` functions, the test output should match the previously expected values.

edwinb: Yes, these are here to test that the results of operations on the Bits primitives are still valid Bits values.

edwinb:
Thanks for these. And, sorry to be awkward, but they'll also need adding in Idris2-boot in order to preserve the bootstrapping path from Idris 1.

edwinb:
Or, maybe a better alternative in general, is to resist using any just added features or taking advantage of recent fixes in the Idris 2 code base itself. I know it's hard to resist, but it will be smoother overall, and at some future release we'll define a new baseline to work from. Right now, that baseline is Idris2-boot.

asandroq:
Thanks @chrrasmussen and @edwinb for taking a look. I am actually feeling silly due to the mistake with the Bits types. I'm going to abandon this PR and send a better one.

asandroq:
> Or, maybe a better alternative in general, is to resist using any just added features or taking advantage of recent fixes in the Idris 2 code base itself. I know it's hard to resist, but it will be smoother overall, and at some future release we'll define a new baseline to work from. Right now, that baseline is Idris2-boot.

@edwinb What does this mean exactly, that no code inside the compiler should use these primitives?

PR #263, opened 2020-06-09
Integer division

asandroq:


libs/prelude/Prelude.idr
```diff
@@ -443,6 +452,12 @@ Abs Integer where
 
 public export
 Integral Integer where
+  quot x y
+      = case y == 0 of
+             False => prim__quot_Integer x y
```
gallais, line 457: Rather than relying on the default error message generated by
an uncaught pattern, we could perhaps use `idris_crash` and
return a nice error?

asandroq: Fixed in the new PR

PR #265, opened 2020-06-09
In IDE mode, look for a ipkg file before loading

edwinb:
This takes the responsibility of finding the ipkg away from IDE mode, which seems sensible given that we can do it ourselves. If there isn't one, it'll load from the local directory as always.

(I won't merge this immediately - please shout if it's likely to break anything. I think it's a better way to go, though, in the end.)

### DIFF
--- a/src/Idris/IDEMode/REPL.idr
+++ b/src/Idris/IDEMode/REPL.idr
@@ -20,6 +20,7 @@ import Data.So
 import Idris.Desugar
 import Idris.Error
 import Idris.ModTree
+import Idris.Package
 import Idris.Parser
 import Idris.Resugar
 import Idris.REPL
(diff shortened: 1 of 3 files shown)


jfdm:
This is nice.

Just to be clear, and reading the source, the behaviour is not to find the first ipkg file one encounters regardless of name, but try to find the named one.

There is an issue with Idris1 emacs mode in which it gets confused over which ipkg file to use if there are multiple ones present: iirc it finds the first ipkg file regardless of name. Given that we can now explicitly say which one we want to load this should address that issue.

gallais:
https://github.com/agda/agda/issues/4526 has taught me that we have to be careful about exploring
the directory tree structure looking for a configuration file.

Hopefully won't be a problem here as people will not be interactively editing
files in a subdirectory of `/nix/store/`.

edwinb:
@jfdm No, it looks for any ipkg in the parent directories. The filename you give `findIpkg` is the thing you want to load. It returns the path of that file relative to the first `.ipkg` it finds.

@gallais One problem I have encountered (with `--find-ipkg`, which vim uses) is finding the wrong `.ipkg` when working in a subdirectory of the main project. This is why I had to add dummy `.ipkg` files in the IDE mode tests. A refinement that would probably address this would be to check whether the file you're trying to load is in the appropriate source directory.

PR #269, opened 2020-06-10
[ fix ] make some exports public

gallais:
I have managed to track down the reason why not enough computation was
happening on the idris2 branch of tparsec to get the STLC example running.
We use these two functions at the type level but they are not publically exported.

It turns out that by simply asking for the type of `test` in the repl, I would
not get the top-level type I had declared but rather the one Idris inferred: a 
big expression clearly stuck on `mapMaybe`, and then on `head'`.

Having to make some functions `public` in the main repo on a per-project
basis does not seem like a tractable solution.

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -332,7 +332,7 @@ init [x] = []
 init (x::y::ys) = x :: init (y::ys)
 
 ||| Attempt to get the head of a list. If the list is empty, return `Nothing`.
-export
+public export
 head' : List a -> Maybe a
 head' []      = Nothing
 head' (x::xs) = Just x
@@ -402,7 +402,7 @@ intercalate sep xss = concat $ intersperse sep xss
 
 ||| Apply a partial function to the elements of a list, keeping the ones at which
 ||| it is defined.
-export
+public export
 mapMaybe : (a -> Maybe b) -> List a -> List b
 mapMaybe f []      = []
 mapMaybe f (x::xs) =


edwinb:
The Idris 1 libraries were generally happen to export things publicly. I wonder if, given that a lot of these things are somehow primitive, we should be happier to do that. This has come up a lot...

jfdm:
I think this is symptomatic of a wider problem in dependently typed programming.

As we all know there is a trade off here between: how we as *library authors* want to expose our functions; and we as *developers* want to use these functions. Sometimes you want to use a function at the type-level only to find that it was exported from a module but not public export. This has bitten me on far too many occasions.

I wonder: Do we need to provide a 'feature' that allows us to override the library author's decision for type-level computations, or should we state that user's are stuck with the decision's made...

gallais:
One thing we could already relax: we could allow non-publically exported
functions to reduce when evaluating *closed terms* because the result will
be stable as long as the function is kept extensionally the same, no matter
whether the actual implementation changes.

ziman:
I don't think there's any conflict here.

Some functions are "specifications" besides just being computation, such as `head'` and `mapMaybe` or `concat`; they are a short name for a more complicated operation. They will pretty much never change; there's "only one way to do it". These should be `public export`.

Other functions,  such as `fastAppend`, are just opaque and their specification is nothing more than their name. These are `export`. They do their magic in a way that you should not care about. They may be changed as you come up with new optimisations. If you need to know _what_ they do, you need to add a spec function and prove (or postulate) they're equivalent to it. (You could theoretically make them public export and thus define their meaning by their implementation. But then you can't change the implementation.)

We had a similar case in https://github.com/idris-lang/Idris2/issues/132: there was the inefficient but convenient for reasoning sum-by-foldr, and the efficient but inconvenient sum-by-foldl.

If you need a function to reduce in a type signature, and it doesn't, there are two possible reasons:
1. The function is genuinely a spec and it should be marked as `public export` but is marked as `export`. This is a library bug and needs to be fixed.
2. The function is not meant to be transparent, its implementation is not part of the API, and you should not depend on its internals. You need to change your approach; maybe you need a formal spec for this function first.

The fact that you can't depend on the internals of an `export` function does not mean you can't mention it in type signatures. You are free to talk _about_ it as much as you like.

gallais:
Telling people that they may only use inefficient functions at the top level
is a great way to shoot one's foot off though.

ziman:
Hmm, what do you mean by "top level"? Can you give an example?

jfdm:
@ziman Maybe we should write this up and add to the docs as a guiding principle on module API design. Not all these decisions will be intuitive to those coming from non-dependently typed languages.

gallais:
Sorry, I meant type level. If you have any sort of serious computation at the
type level you are going to want to use efficient data structures and functions.

ziman:
I think I get what you mean now. Does your closed-term suggestion address it fully or are there other things that you have in mind?

Maybe there are two kinds of type-level computation; one is "to get the result" (which uses closed terms and feels like staging), and the other is "to talk about things" (which doesn't).

gallais:
I don't know if it fully addresses it but it would at least solve the problem
with the unit tests in https://github.com/gallais/idris-tparsec.

PR #272, opened 2020-06-11
Use 2^64 constant instead of calculation

### LINKED ISSUE #268: Building idris2 with the bootstrap version gives RTS error following attempt of allocating a huge heap
Hi there,

I wanted to update my Idris2 (currently on version 0.1.0-45d02a11b, from before the self-hosting of Idris2).
I'm running trouble for compiling the latest Idris2, both with my old Idris2 and with the official bootstrapping version (sometimes called Idris2-in-Idris1, from here https://github.com/edwinb/Idris2-boot ).

I'll only detail the trouble with the bootstrap version, as I assume this is more annoying that it does not build with this version of Idris2 that is specifically designed for it. (and my current idris2 could be a bit broken, which should not be the case of the bootstrap version).

I did manage to install Idris2boot with Idris1 (I'm using the Version 1.3.2-git:0417c53fb).
I also added the folder $HOME/idris2boot into my path (in my .profile file), as I already did for the $HOME/idris2 folder.
(by the way, we should update [the installation file of the bootstrap version](https://github.com/edwinb/Idris2-boot/blob/master/INSTALL.md) which tells that "The default installation directory is $HOME/.idris2" as the idris2boot executable is in fact installed in ./idris2boot).

Thus, If I do :
$ idris2boot

It runs the REPL of idris2boot ( Version 0.1.1-e9611b4ee ) and I effectively get the message that this is a bootstrap version, no longer developed. Everything is ok here.

Now, I want to install the latest Idris2 using this bootstrap version.

I've just cloned it.
Then I do :
$ make distclean

to make sure that everything is clean.

And following the [installation procedure of Idris2](https://github.com/idris-lang/Idris2/blob/master/INSTALL.md), I did :

$ make all IDRIS2_BOOT=idris2boot

Which goes well at the start :

> cc -Wall  -fPIC -O2   -c -o idris_file.o idris_file.c
> cc -Wall  -fPIC -O2   -c -o idris_support.o idris_support.c
> cc -Wall  -fPIC -O2   -c -o idris_net.o idris_net.c
> cc -Wall  -fPIC -O2   -c -o getline.o getline.c
> cc -Wall  -fPIC -O2   -c -o idris_directory.o idris_directory.c
> idris_directory.c: In function ‘idris2_currentDirectory’:
> idris_directory.c:11:4: warning: ignoring return value of ‘getcwd’, declared with attribute warn_unused_result [-Wunused-result]
>     getcwd(cwd, 1024);
>     ^~~~~~~~~~~~~~~~~
> cc -Wall  -fPIC -O2   -c -o idris_buffer.o idris_buffer.c
> idris_buffer.c: In function ‘idris2_readBufferFromFile’:
> idris_buffer.c:145:5: warning: ignoring return value of ‘fread’, declared with attribute warn_unused_result [-Wunused-result]
> [...]
> echo 'module IdrisPaths' > src/IdrisPaths.idr
> echo 'export idrisVersion : ((Nat,Nat,Nat), String); idrisVersion = ((0,2,0), "d71bfef73")' >> src/IdrisPaths.idr
> echo 'export yprefix : String; yprefix="/home/franck/.idris2"' >> src/IdrisPaths.idr
> idris2boot --build idris2.ipkg
> 1/144: Building Core.FC (src/Core/FC.idr)
> 2/144: Building Core.Name (src/Core/Name.idr)

And it's only when it reaches the step 40/144 that it fails with the following error message :

40/144: Building Core.Primitives (src/Core/Primitives.idr)
RTS ERROR: Unable to allocate heap. Requested 8751446623104 bytes.
Makefile:57: recipe for target 'build/exec/idris2' failed
make: *** [build/exec/idris2] Error 1

If I try to run it again :

40/144: Building Core.Primitives (src/Core/Primitives.idr)
RTS ERROR: Unable to allocate heap. Requested 8735164991968 bytes.
Makefile:57: recipe for target 'build/exec/idris2' failed
make: *** [build/exec/idris2] Error 1

So it looks like it tries to allocate a *huge* amount of memory (8 terabytes, right?), and this is not always the exact same amount.

Any idea on what could cause that, and how I could try to fix it?
I did looked briefly at Primitives.idr, but I must admit this is a bit too cryptic for me.

Any idea would be most welcome! Many thanks.

Franck

melted:
This is a workaround to fix the bootstrap
Fixes #268


src/Core/Primitives.idr
```diff
@@ -147,13 +150,14 @@ strSubstr [NPrimVal fc (I start), NPrimVal _ (I len), NPrimVal _ (Str str)]
     = Just (NPrimVal fc (Str (prim__strSubstr start len str)))
 strSubstr _ = Nothing
 
+
 add : Constant -> Constant -> Maybe Constant
 add (BI x) (BI y) = pure $ BI (x + y)
 add (I x) (I y) = pure $ I (x + y)
 add (B8 x) (B8 y) = pure $ B8 $ (x + y)
 add (B16 x) (B16 y) = pure $ B16 $ (x + y) `mod` (prim__shl_Int 1 16)
 add (B32 x) (B32 y) = pure $ B32 $ (x + y) `mod` (prim__shl_Int 1 32)
```
gallais, line 159: Does it make sense to perform the same refactoring for all the other
`prim__shl__Int 1 N` used in this file?

melted: Yes, it would also look nicer to use hexadecimal literals.

melted:
Weird `(the Bits64 (fromInteger (-1)))` works in the main but not in testsMax

melted:
Hah, it's because it can't handle the hex literal for 2^64

src/Core/Primitives.idr
```diff
@@ -147,13 +155,14 @@ strSubstr [NPrimVal fc (I start), NPrimVal _ (I len), NPrimVal _ (Str str)]
     = Just (NPrimVal fc (Str (prim__strSubstr start len str)))
 strSubstr _ = Nothing
 
+
 add : Constant -> Constant -> Maybe Constant
 add (BI x) (BI y) = pure $ BI (x + y)
 add (I x) (I y) = pure $ I (x + y)
 add (B8 x) (B8 y) = pure $ B8 $ (x + y)
```
melted, line 162: This is wrong, it should of course do a mod there too.

gallais: Why is this bug only showing up after the refactoring? O_O

melted: It didn't, the expected values in the test were wrong,  differed between the REPL tests and the :exec main test.

PR #274, opened 2020-06-11
Make Buffer more primitive

edwinb:
Meaning that the FFI is aware of it, so you can send arbitrary byte data
to foreign calls.

This means that we no longer need the hacky way of reading and writing
binary data via scheme, so can have a more general interface for reading
and writing buffer data in files.

It will also enable more interesting high level interfaces to binary
data, with C calls being used where necessary.

Note that the Buffer primitive are unsafe! They always have been, of
course... so perhaps (later) they should have 'unsafe' as part of their
name and better high level safe interfaces on top.

### DIFF
--- a/libs/base/Data/Buffer.idr
+++ b/libs/base/Data/Buffer.idr
@@ -3,6 +3,13 @@ module Data.Buffer
 import System.Directory
 import System.File
 
+-- Reading and writing binary buffers. Note that this primitives are unsafe,
+-- in that they don't check that buffer locations are within bounds.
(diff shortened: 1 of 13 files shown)


edwinb:
Oh good, it needs more support in Idris2-boot doesn't it... I could fix that by updating the bootstrap scheme but that wouldn't help with Idris2-boot.

edwinb:
On second thoughts, Idris2-boot does this differently so I think we're fine. I'll take a chance on updating the scheme and if I'm wrong fix it in Idris2-boot.

Issue #270, opened 2020-06-11
use of auto-implicits leads to cryptic error message


rgrover:
# Steps to Reproduce

Compile the following:

```idris
import Data.So

-- extending `a` with bounds
public export
data BoundExtended : (a : Type) -> Type where
  т   : BoundExtended a -- top (maximum value)
  ⊥   : BoundExtended a -- bottom (minimum value)
  Val : a -> BoundExtended a

-- lift (<=) over bounded-extended values
public export
le : Ord a => BoundExtended a -> BoundExtended a -> Bool
le ⊥ _ = True
le (Val x) (Val y) = x <= y
le _ т = True
le _ _ = False

-- A binary search tree with bounded-extended element type
public export
data BST : BoundExtended a -> BoundExtended a -> Type where
  Leaf : BST l u
  Node : Ord a => {pivot : a} -> BST l (Val pivot) -> BST (Val pivot) u
       -> So (l `le` (Val pivot)) => So ((Val pivot) `le` u) => BST l u

total
insert : Ord a => {l, u: BoundExtended a} -> (newElem : a) -> BST l u
       -> So (l `le` (Val newElem)) => So ((Val newElem) `le` u) => BST l u
insert newElem Leaf = Node {pivot=newElem} Leaf Leaf
insert newElem tree@(Node {pivot} lt rt) with (newElem <= pivot)
  insert newElem tree@(Node lt rt) | result = ?rhs
```

# Expected Behavior

There should be no compilation error.

# Observed Behavior

```
BST.idr:31:42--31:52:While processing right hand side of insert at BST.idr:31:1--33:1:
Multiple solutions found in search. Possible correct results:
	conArg
	conArg
Error(s) building file BST.idr
```

gallais:
> There should be no compilation error.

When you pattern match on `Node`, you expose a proof of `Ord a`. You now have
two possible solutions for `Ord a`: the one provided by `insert` and the one
provided by `Node` and there is no reason for Idris to believe that they are
equal. So it cannot make a decision. Arguably the error message is quite bad*
(we were just discussing that yesterday on slack).

Minimised test case:

```idris
f : Num a => a
f = g where

  g : Num a => a
  g = 0
```

rgrover:
@gallais: Thanks for providing a smaller case that produces the same kind of error message. 

One possible workaround for your example is:
```idris
f : Num a => a
f = g where
  g : a
  g = 0
```

[...]

PR #275, opened 2020-06-11
[ re #270 ] Mention the target type of the auto search

gallais:


edwinb:
That's definitely an improvement, thanks!

PR #280, opened 2020-06-11
Add badge for the bootstrap from Idris 1

melted:


melted:
Looks good! No need to wait for the CI for this.

PR #281, opened 2020-06-11
Make fastAppend a foreign call

ziman:
`Data.Strings.fastAppend` was a `schemeCall`; this patch makes it a foreign call, so it can be implemented on non-Scheme backends.

### DIFF
--- a/libs/base/Data/Strings.idr
+++ b/libs/base/Data/Strings.idr
@@ -20,7 +20,7 @@ foldl1 f (x::xs) = foldl f x xs
 -- enough room in advance so there's only one allocation, rather than lots!
 %foreign
     "scheme:string-concat"
-    "node:lambda:__prim_idris2js_stringConcat"
+    "javascript:lambda:__prim_idris2js_stringConcat"
 export
 fastConcat : List String -> String
 
--- a/libs/base/Data/Strings.idr
+++ b/libs/base/Data/Strings.idr
@@ -20,7 +20,7 @@ foldl1 f (x::xs) = foldl f x xs
 -- enough room in advance so there's only one allocation, rather than lots!
 %foreign
     "scheme:string-concat"
-    "javascript:lambda:__prim_idris2js_stringConcat"
+    "javascript:lambda:(xs)=>''.concat(...__prim_idris2js_array(xs))"
 export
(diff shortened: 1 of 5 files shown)


ziman:
The tests are failing on renumbered `MN`s; I'll fix it.

libs/base/Data/Strings.idr
```diff
@@ -15,15 +15,12 @@ partial
 foldl1 : (a -> a -> a) -> List a -> a
 foldl1 f (x::xs) = foldl f x xs
 
--- This works quickly because when string-append builds the result, it allocates
+-- This works quickly because when string-concat builds the result, it allocates
 -- enough room in advance so there's only one allocation, rather than lots!
+%foreign
+    "scheme:string-concat"
 export
 fastAppend : List String -> String
```
gallais, line 23: If we are going to fix the FFI name & the comment,
should we also fix the exported name? This is a
breaking change but it also means we won't have to
keep living with this confusing name.

ziman: This is used in the compiler, too, so it would require a rebootstrap, i guess?

Maybe we could add `fastAppend` as an alias to `fastConcat` and then mark it as `%deprecated` once someone adds `%deprecated`.

Not sure what to do here.

gallais: Sounds good.

libs/base/Data/Strings.idr
```diff
@@ -16,15 +16,19 @@ partial
 foldl1 : (a -> a -> a) -> List a -> a
 foldl1 f (x::xs) = foldl f x xs
 
--- This works quickly because when string-append builds the result, it allocates
+-- This works quickly because when string-concat builds the result, it allocates
 -- enough room in advance so there's only one allocation, rather than lots!
+%foreign
+    "scheme:string-concat"
+    "node:lambda:__prim_idris2js_stringConcat"
```
ska80, line 23: AFAIK if we want this to work on node and the browser `node` should be replaced with `javascript` according to https://github.com/idris-lang/Idris2/blob/master/docs/source/backends/javascript.rst#javascript-ffi-specifiers

ziman: Thank you!

PR #282, opened 2020-06-11
Fix MkRecord signature

ether42:
I'm probably doing something wrong in the test (I must admit I don't have much experience and it's been a long time since I've played with Idris), feel free to request changes (if I define `axes` as a subroutine of `mkPoint` by using `where`, the totality checker will complain but maybe it's unrelated?).

If I `declare [ mkPoint 4 ]`, I'll get `Main.(.z) is already defined` which seems about right but I couldn't figure out if there is a way to properly namespace fields (adding `NS [type, "Main"]` doesn't help, any idea?).

### DIFF
--- a/libs/base/Language/Reflection/TTImp.idr
+++ b/libs/base/Language/Reflection/TTImp.idr
@@ -132,7 +132,7 @@ mutual
   data Record : Type where
        MkRecord : FC -> (n : Name) ->
                   (params : List (Name, TTImp)) ->
-                  (conName : Maybe Name) ->
+                  (conName : Name) ->
                   (fields : List IField) ->
                   Record
 
--- a/tests/idris2/reflection004/refdecl.idr
+++ b/tests/idris2/reflection004/refdecl.idr
@@ -1,6 +1,22 @@
+import Data.Nat
+import Data.Vect
 import Language.Reflection
 
 %language ElabReflection
+
+axes : (n : Nat) -> {auto gt : GT n 0} -> {auto lte : LTE n 4} -> Vect n String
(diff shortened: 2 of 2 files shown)


melted:
@edwinb is this a fix valid?

edwinb:
Sorry I'm well behind on PRs, but this does look valid to me - it fixes an inconsistency between the TTImp representation and the reflected type. Not sure I understand what you mean by properly namespacing fields though, do you mean when building a term via reflection?

ether42:
Really sorry for the late reply.

> Not sure I understand what you mean by properly namespacing fields though, do you mean when building a term via reflection?

I wasn't able to place definitions in a 'sub' namespace via the reflection.
The idea was to create the following structure (from the added test example):
```
Main
    Point3D
            MkPoint
            .x
            .y
            .z
            ...
    Point4D
            MkPoint
            .x
            .y
            .z
            .w
            ...
```
Basically, have each record in its own namespace to avoid conflicting names (e.g.: `.z` to access the z dimension of the point), because this:
```idris
declare [ mkPoint 3 ]
declare [ mkPoint 4 ]
```
will fail with `Main.(.z) is already defined` (which isn't surprising).

I remember trying `inCurrentNS`, `NS [type, "Main"]` (to create a namespace under `Main`), and a bunch of other things but IIRC the helpers to access the records'  fields were all created under `Main` (instead of `Main.Point3D` and `Main.Point4D`).

gallais:
> the helpers to access the records' fields were all created under `Main`

Are they not created under `Main.Point3D` and `Main.Point4D` respectively?

PR #284, opened 2020-06-11
Various stdlib updates

clayrat:
Add helpers and proofs for Bool, Nat, Fin, So and SortedMap, accumulated from various Idris projects. I've also moved boolean properties from `contrib` to `base` and renamed them to simply `Data.Bool`, to align with Nat and List properties.


melted:
Let's merge!

PR #287, opened 2020-06-12
Safe indexing of lists

jfdm:


jfdm:
This is annoying, you update the reflection tests locally and you discover that the test runner in the CI provides a new different number.

melted:
What does that number even signify?

melted:
It's this code 

```
tryGenSym : Elab a
tryGenSym
   = do n <- genSym "plus"
        ns <- inCurrentNS n
        fail $ "failed after generating " ++ show ns
```
Maybe genSym is too brittle to be relied on for predictable output?

gallais:
The solution is to add a bit of `sed` magic to filter out the numbers from
the output.

melted:
It's a sequence number that very much depends on the exact versions of stuff. It's a bad thing to have in a test. We should censor that output (filter all digits to "X"?) or just not show the variable.

melted:
> The solution is to add a bit of `sed` magic to filter out the numbers from
> the output.

We could mangle the output in the idris file already.

melted:
I have implemented a simple fix, I'll make a PR soon.

jfdm:
@melted Thanks for updating the reflection test. I've rebased on master so this should pass. Fingers crossed...

PR #288, opened 2020-06-12
Remove hardcoded default CG from command line options

ska80:


src/Core/TTC.idr
```diff
@@ -731,14 +731,14 @@ TTC CDef where
...
+  toBuf b Gambit = tag 2
 
   fromBuf b
       = case !getTag of
              0 => pure Chez
-             2 => pure Racket
-             3 => pure Gambit
+             1 => pure Racket
+             2 => pure Gambit
              _ => corrupt "CG"
```
gallais, line 743: You should probably change the TTC version number if you are going to change
the meaning of some tags.

src/Core/TTC.idr
```diff
@@ -731,14 +731,14 @@ TTC CDef where
 export
 TTC CG where
   toBuf b Chez = tag 0
-  toBuf b Racket = tag 2
-  toBuf b Gambit = tag 3
+  toBuf b Racket = tag 1
+  toBuf b Gambit = tag 2
```
melted, line 736: These changes just break the TTC file format for no gain.

ska80:
> TTC changes must be reverted.(OUTDATED: I'm not a fan of this commit overall, I don't think this should be merged. Chez is the only code generator that works reasonably well at this point (Racket is slow, Gambit can't bootstrap).)
> 
> EDIT: I see that it still leaves the default so that is fine.

I didn't mean to change the default GC at all :)

PR #289, opened 2020-06-12
Remove unsolved holes from `base`.

jfdm:
Having unsolved holes in a 'core' library unneccessarily pollutes the list of holes shown to the user.
Thus, having unfilled holes in a 'core' library is not right.
These constructs can be re-added once the holes have been filled in.


melted:
Looks like Reflection003 is spooking again.

jfdm:
Rebased on master with fixed reflection test. Should now pass *fingers crossed*...

PR #299, opened 2020-06-13
Debug printing for buffers

melted:
I'm working on making network use buffers, while doing that I added this little lib to dump buffers to strings.


melted:
It looks pretty good:
Skärmklipp 2020-06-13 22 11 50

PR #302, opened 2020-06-14
Integer division 3

asandroq:
This is a new attempt at merging integer division. I am creating separate pull requests because the boot Scheme files change too much and making new PRs keeps history cleaner.

This time I didn't use the new primitives inside the compiler itself. Nevertheless, the definition of `mod` changed and I had to adjust the casting primitives inside the compiler. I don't know if that could have been avoided.

I am also now using `idris_crash` instead of relying on partial pattern matching for the implementation in the prelude.


edwinb:
Hi, sorry this has been sitting here so long. I'm working oldest to newest in an attempt to resolve our PR backlog - I do think we need these, but it's been so long that it's hard to see how to merge it easily. Therefore, I think I'll have to close this but if you think it can easily be revived for the current code base, please do reopen and I'll merge it soon!

PR #303, opened 2020-06-14
Implement support for other codegens via the Idris2 API

melted:
The idea is to make everything accessible via the API, so
codegen implementors will be able to reuse all infrastructure by calling
the appropriate function in Idris.Driver supplying their codegen.


melted:
Still lacking tests and dokumentation. It works very well, though.

melted:
An example codegen:
```idris
module Main

import Core.Context
import Compiler.Common
import Idris.Driver

compile : Ref Ctxt Defs -> (execDir : String) ->
          ClosedTerm -> (outfile : String) -> Core (Maybe String)
compile defs dir term file = do coreLift $ putStrLn "I'd rather not."
                                pure $ Nothing

execute : Ref Ctxt Defs -> (execDir : String) -> ClosedTerm -> Core ()
execute defs dir term = do coreLift $ putStrLn "Maybe in an hour."

lazyCodegen : Codegen
lazyCodegen = MkCG compile execute

main : IO ()
main = mainWithCodegens [("lazy", lazyCodegen)]
```
compile with `idris2 -p idris2 -p contrib -p network Lazy.idr -o lazy-idris2`

ziman:
Long awaited Lazy Idris is here!

PR #310, opened 2020-06-15
Replace FFI implementation of 'hex' function with a plain Idris version

chrrasmussen:
The motivation for this change is to remove the usage of FFI in the `Idris.IDEMode.Commands.hex` function so that code generators have one less thing to worry about.

This change is covered by the `ideModeNNN` tests.

Notes about the changes in this pull request:
- Changed `Utils.Hex.asHex` to be a total function. Before one could supply a negative integer and it would loop forever. There was also an edge case where calling `asHex 0` would result in an empty string (instead of `"0"`).
- Moved `Compiler.Scheme.Racket.showRacketChar.pad` to `Utils.Hex` because it is used in two places.
- Renamed `sendLine` to `sendStr` because it doesn't send a newline.

### DIFF
--- a/src/Compiler/Scheme/Racket.idr
+++ b/src/Compiler/Scheme/Racket.idr
@@ -56,14 +56,8 @@ showRacketChar : Char -> String -> String
 showRacketChar '\\' = ("\\\\" ++)
 showRacketChar c
    = if c < chr 32 || c > chr 126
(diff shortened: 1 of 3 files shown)


melted:
Looks a lot less hacky than my implementation in Debug.Buffer.

PR #312, opened 2020-06-16
Update Chez support for *BSD operating systems

ska80:


melted:
Does it work otherwise? Because I wondered about those hardcoded `make` calls in bootstrap.sh, they look like something that would be BSD-unfriendly.

ska80:
> Does it work otherwise?

I have successfully managed to bootstrap Idris2 with Chez Scheme on OpenBSD 6.7 and it does work (at least passes all tests :))

> Because I wondered about those hardcoded `make` calls in bootstrap.sh, they look like something that would be BSD-unfriendly.

Yes, calls to `make` and `realpath` have to be replaced with `gmake` and `grealpath` (from `coreutils` package) correspondingly.  For now, I just symlinked them into my local `PATH`.  I will send another PR that adds support for OpenBSD.

Issue #305, opened 2020-06-14
unhelpful error message while unifying identical types


rgrover:
I get a cryptic error message while lifting a `Preorder` relationship to `Maybe`. The type-checker is unable to unify identical types.

# Steps to Reproduce

Using the definition of `Preorder` from Idris1:

```idris
interface Preorder a (rel : a -> a -> Type) where
  transitive : (x : a) -> (y : a) -> (z : a) -> rel x y -> rel y z -> rel x z
  reflexive : (x : a) -> rel x x

-- Proof of Preorder relationship over Maybe values.
export
data MaybePreorder : Maybe a -> Maybe a -> Type where
  AnythingIsBetter : MaybePreorder Nothing x
  Lift : Preorder a rel => rel x y -> MaybePreorder (Just x) (Just y)

-- If `a` has a Preorder relationship `rel`, then `Maybe a` has a lifted
-- Preorder relationship.
Preorder a rel => Preorder (Maybe a) MaybePreorder where
  reflexive Nothing = AnythingIsBetter
  reflexive (Just x) = Lift $ reflexive {rel=rel} x

  transitive Nothing y z xy yz = AnythingIsBetter
  transitive (Just x) (Just y) (Just z) (Lift v) (Lift w) =
    Lift $ transitive x y z v w
```

# Expected Behavior

The code above should type check.

# Observed Behavior

```
Test.idr:21:39--21:41:While processing right hand side of transitive at Test.idr:20:3--148:1:
Can't solve constraint between:
	rel x y
and
	rel x y
```

If I remove `{rel=rel}` from the definition of `transitive`:
```idris
  transitive (Just x) (Just y) (Just z) (Lift v) (Lift w) =
    Lift $ transitive x y z v w
```
I get a variant:
```
Test.idr:21:31--148:1:While processing right hand side of transitive at Test.idr:20:3--148:1:
Can't solve constraint between:
	rel y z
and
	rel y z
Error(s) building file Test.idr
```

I also get a different error message when I leave out `{rel=rel}` from the defintion of `reflexive`, but let's leave that for now.

gallais:
This is #35 

In the meantime, if you're wondering how to get out of this predicament:
parametrise `MaybePreorder` over `rel`. At the moment there is no connection
whatsoever between the (implicitly-bound) `rel` used in `Lift` and the one
brought into scope by the `Preorder a rel =>` constraint.

rgrover:
@gallais:
[...]

PR #314, opened 2020-06-16
Proposal: Relation interfaces

nickdrozd:
Currently some proofs have names like `lteRefl`, `lteTransitive`, `factorReflexive`, etc. These names say the same thing, namely that some relation is reflexive or transitive, but they do so in an ad-hoc and disorganized manner.

This PR adds some interfaces that can act as a universal language for these kinds of relations. Specifically, it adds the interfaces `Reflexive` and `Transitive`. There are more that could be added, like `Symmetric` and `Antisymmetric`, but this is a start. `Reflexive` and `Transitive` are implemented for `LTE` and `Factor`. Again, there are more of these to do, but this is a start.

Two questions to discuss are whether this is a good at all, and if so, whether this is the right way to do it. Obviously I think this is a good idea, but I'm not wed to the particular design. Initially I tried an approach that used explicit arguments all around, then I switched to implicit arguments, which match the existing relations better. Compare the two commits here to see the difference.

Idris 1 contained a `contrib` module called `Order`. It defined things like partial order directly in terms of these properties. This approach is more flexible, allowing for properties to be composed in different ways (along the lines of [this suggestion](https://groups.google.com/d/msg/idris-lang/VZVpi-QUyUc/reZxLJSOAQAJ)). This PR doesn't deal with orders, but they could be defined easily enough.

### DIFF
--- a/libs/base/Control/Relation.idr
+++ b/libs/base/Control/Relation.idr
@@ -1,0 +1,20 @@
+||| Properties of binary relations
+module Control.Relation
+
+%default total
+
+public export
+interface Reflexive (ty : Type) (rel : ty -> ty -> Type) where
+  reflexive : (x : ty) -> rel x x
+
+public export
+interface Transitive (ty : Type) (rel : ty -> ty -> Type) where
+  transitive : (x, y, z : ty) -> rel x y -> rel y z -> rel x z
+
+-- Equality implementations ------------
+
+Reflexive ty Equal where
+  reflexive _ = Refl
+
+Transitive ty Equal where
+  transitive _ _ _ xy yz = trans xy yz
--- a/libs/base/Control/Relation.idr
+++ b/libs/base/Control/Relation.idr
@@ -5,16 +5,16 @@ module Control.Relation
 
 public export
 interface Reflexive (ty : Type) (rel : ty -> ty -> Type) where
-  reflexive : (x : ty) -> rel x x
+  reflexive : {x : ty} -> rel x x
 
 public export
 interface Transitive (ty : Type) (rel : ty -> ty -> Type) where
-  transitive : (x, y, z : ty) -> rel x y -> rel y z -> rel x z
+  transitive : {x, y, z : ty} -> rel x y -> rel y z -> rel x z
 
 -- Equality implementations ------------
 
 Reflexive ty Equal where
-  reflexive _ = Refl
+  reflexive = Refl
 
 Transitive ty Equal where
-  transitive _ _ _ xy yz = trans xy yz
+  transitive xy yz = trans xy yz
--- a/libs/contrib/Data/Nat/Factor.idr
+++ b/libs/contrib/Data/Nat/Factor.idr
@@ -86,17 +86,17 @@ oneSoleFactorOfOne (S (S k)) (CofactorExists (S j) prf) =
 ||| Every natural number is factor of itself.
 export
 Reflexive Nat Factor where
-  reflexive n = CofactorExists 1 $ rewrite multOneRightNeutral n in Refl
+  reflexive {x} = CofactorExists 1 $ rewrite multOneRightNeutral x in Refl
 
 ||| Factor relation is transitive. If b is factor of a and c is b factor of c
 ||| is also a factor of a.
 export
 Transitive Nat Factor where
-  transitive a b c (CofactorExists qb prfAB) (CofactorExists qc prfBC) =
+  transitive {x} (CofactorExists qb prfAB) (CofactorExists qc prfBC) =
     CofactorExists (qb * qc) $
         rewrite prfBC in
         rewrite prfAB in
-        rewrite multAssociative a qb qc in
+        rewrite multAssociative x qb qc in
         Refl
 
 multOneSoleNeutral : (a, b : Nat) -> S a = S a * b -> b = 1
@@ -264,7 +264,7 @@ minusFactor {a} {b} (CofactorExists qab prfAB) (CofactorExists qa prfA) =
 ||| A decision procedure for whether of not p is a factor of n.
 export
 decFactor : (n, d : Nat) -> DecFactor d n
-decFactor Z Z = ItIsFactor $ reflexive Z
+decFactor Z Z = ItIsFactor $ reflexive {x = Z}
 decFactor (S k) Z = ItIsNotFactor $ ZeroNotFactorS k
 decFactor n (S d) =
         let Fraction n (S d) q r prf = Data.Fin.Extra.divMod n (S d) in
@@ -334,7 +334,7 @@ oneCommonFactor a b = CommonFactorExists 1
 export
 selfIsCommonFactor : (a : Nat) -> {auto ok : LTE 1 a} -> CommonFactor a a a
 selfIsCommonFactor Z {ok} = absurd $ succNotLTEzero ok
-selfIsCommonFactor (S k) = CommonFactorExists (S k) (reflexive $ S k) (reflexive $ S k)
+selfIsCommonFactor (S k) = CommonFactorExists (S k) (reflexive {x = S k}) (reflexive {x = S k})
 
 
 -- Some helpers for the gcd function.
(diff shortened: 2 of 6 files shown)


libs/base/Control/Relation.idr
```diff
@@ -0,0 +1,12 @@
...
+
+%default total
+
+public export
+interface Reflexive (ty : Type) (rel : ty -> ty -> Type) where
+  reflexive : {x : ty} -> rel x x
+
+public export
+interface Transitive (ty : Type) (rel : ty -> ty -> Type) where
+  transitive : {x, y, z : ty} -> rel x y -> rel y z -> rel x z
```
Sventimir, line 12: From what I see, you have to supply these implicit arguments explicitly anyway. I believe practicality should be a guideline as to which arguments should be implicit and which explicit. If the compiler can't infer these by itself, then in my opinion they should be explicit. These curly brackets are noisy and I'd prefer to avoid writing them. If you can't do that most of the time, probably the argument shouldn't be implicit after all.

libs/base/Control/Relation.idr
```diff
@@ -0,0 +1,12 @@
+||| Properties of binary relations
+module Control.Relation
+
+%default total
+
+public export
+interface Reflexive (ty : Type) (rel : ty -> ty -> Type) where
```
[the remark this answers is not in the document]
nickdrozd: What you're describing is possible in Idris 1, but it is not currently possible in Idris 2. As inference improves, I would be happy to get rid of redundancies like this. That said, it isn't the worst thing in the world to have a little extra explicitness. `Reflexive Nat LTE` can be read as: "`LTE` is `Reflexive` over `Nat`".

libs/base/Control/Relation.idr
nickdrozd: These "explicit implicit" arguments are required because implementations may need to case-split on them. For example,

```idris
Reflexive Nat LTE where
  reflexive {x = Z} = LTEZero
  reflexive {x = S k} = LTESucc $ reflexive {x = k}
```

That's not a `Transitive` instance, but you get the idea.

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -358,7 +358,7 @@ notLteAndGt (S k) (S j) aLteB aGtB = notLteAndGt k j (fromLteSucc aLteB) (fromLt
 gcd_step : (x : Search) ->
     (rec : (y : Search) -> Smaller y x ->  (f : Nat ** GCD f (left y) (right y))) ->
     (f : Nat ** GCD f (left x) (right x))
-gcd_step (SearchArgs Z _ bLteA {bNonZero}) _ = absurd . succNotLTEzero $ lteTransitive bNonZero bLteA
+gcd_step (SearchArgs Z _ bLteA {bNonZero}) _ = absurd . succNotLTEzero $ transitive {ty = Nat} bNonZero bLteA
```
nickdrozd, line 361: @Sventimir Here's a case where an "implicit" argument must be specified explicitly. Strangely, _some_ argument must be specified, but it doesn't matter _which_ argument is specified.

This does not work:

```idris
transitive bNonZero bLteA
```

But the following all do work:

```idris
transitive {ty = Nat} bNonZero bLteA
```

```idris
transitive {rel = LTE} bNonZero bLteA
```

```idris
transitive {x = S Z} bNonZero bLteA
```

```idris
transitive {y = a} bNonZero bLteA
```

```idris
transitive {z = Z} bNonZero bLteA
```

So nothing in particular needs to be specified, but one thing or another does need to be specified. I don't know anything about the internals, but as a user it feels like the typechecker needs a little nudge in the right direction to get started. Hopefully inference will improve and these nudges won't be required.

libs/base/Control/Relation.idr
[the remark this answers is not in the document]
Sventimir: Maybe you're right. But if it's impossible, it's all academic anyway.

libs/base/Control/Relation.idr
Sventimir: I'm not complaining about their existence. They're obviously required. What bothers me is the necessity of writing these curly braces when calling a function. Pattern matching on them is not that bad, because you do it only once and besides pattern matching when implementing a function is often unavoidable. My point is about calling the function. If the compiler can't infer the argument most of the time, then it should probably be explicit. You made a good point below, though.

libs/contrib/Data/Nat/Factor.idr
Sventimir: That's interesting. I'm guessing that's because of packing all the arguments to the `gcd_step` in this `Search` type. The compiler only sees this `Search` type and does not dig deep enough to understand that it's all about `Nat`s. Notice that from any of these arguments it can be inferred that we want `transitive` for `Nat`/`LTE`. So you're probably right in that the compiler needs just a little nudge.

libs/base/Control/Relation.idr
Sventimir: On the other hand in my experience, having all arguments implicit damages the readability. Suppose you indeed don't need to specify any of them, because the compiler can work all of them out on its own. However, when human reads that code, trying to understand it, he or she will often have a hard time figuring out, reflexivity of which relation and over which value is important here. So specifying the value is usually a nice hint for a human reader, event if it's not exactly relevant to the compiler. Moreover, even if the compiler can figure it out having the complete code is one thing, but these arguments also help it make sense of things in the presence of holes.

libs/base/Control/Relation.idr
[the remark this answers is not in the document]
gallais: I think it is indeed possible, see the test case in #305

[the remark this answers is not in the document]
nickdrozd: Do you have any concrete suggestions for how to do that? I think what we would like to see is:

```idris
interface Reflexive (rel : ty -> ty -> Type) where
  reflexive : {x : ty} -> rel x x

Reflexive LTE where
  reflexive {x = Z} = LTEZero
  reflexive {x = S k} = LTESucc $ reflexive {x = k}
```

But that turns up this error:

```
Control/Relation.idr:12:31--12:33:While processing type of reflexive at Data/Nat.idr:77:1--81:1:
When unifying ?ty [no locals in scope] and Nat
Pattern variable ty unifies with:
	Nat
```

libs/contrib/Data/Nat/Factor.idr
nickdrozd: It's even more general than that. Every example of required implicit arguments in this PR acts the same way, including the simple call to `reflexive` in `Data.List.Views`.

libs/base/Control/Relation.idr
nickdrozd: I consider the braced arguments as temporary workarounds. At some point in the future it may be possible to go through and delete all of them in one fell swoop.

libs/base/Control/Relation.idr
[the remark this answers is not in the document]
gallais: That looks like a bug.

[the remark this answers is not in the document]
Sventimir: I don't think so. I thought about putting `ty : Nat` in `{}` braces, but that is indeed a syntax error. However, omitting the argument altogether renders you unable to instantiate the interface with a concrete `ty` type, because Idris then demands it to be abstract (which seems reasonable). It's also consistent with behaviour for regular functions, where omitting a type argument means you can't refer to it in the definition.  I don't know how this was resolved in Idris1, but here I would like to be able to put `ty : Type` in `{}`, which again would be consistent with behaviour for functions. But that is probably a big change…

nickdrozd:
@edwinb @gallais Is there any interest in this? If not, I can trash it. It's a change from Idris 1, but I think it's more flexible.

https://github.com/idris-lang/Idris2/pull/543 ports over the old version of these relations, so these two PRs are in conflict.

abailly:
FWIW, I am fine thrashing #543 in favor of a more suitable interface. I don't think I have the knowledge to choose which version is better :)

Sventimir:
I haven't seen any serious argument against it, so I wouldn't trash it. It's a resonable proposal in my opinion. Of course it should be reviewed by someone competent enough to accept or reject it though…

PR #316, opened 2020-06-16
Fancy preorder reasoning

ohad:
Change the poor-man's preorder reasoning in `contrib` to use slightly faster data structures, following [Agda](https://github.com/agda/agda-stdlib/issues/622).

Since only the `TailRec` ursive version of `Data.List` seems to use these proofs, hopefully it'll be ok to change this now.

### DIFF
--- a/libs/contrib/Data/List/TailRec.idr
+++ b/libs/contrib/Data/List/TailRec.idr
@@ -19,9 +19,9 @@ module Data.List.TailRec
 import Syntax.PreorderReasoning
 import Syntax.WithProof
 
-import Data.List 
-import Data.Vect 
-import Data.Nat  
+import Data.List
+import Data.Vect
+import Data.Nat
 
 length_aux : List a -> Nat -> Nat
 length_aux [] len = len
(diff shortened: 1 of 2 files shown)


libs/contrib/Data/List/TailRec.idr
```diff
@@ -19,9 +19,9 @@ module Data.List.TailRec
 import Syntax.PreorderReasoning
 import Syntax.WithProof
 
-import Data.List
-import Data.Vect
-import Data.Nat
+import Data.List 
+import Data.Vect 
+import Data.Nat
```
gallais, line 25: Trailing whitespace

libs/contrib/Data/List/TailRec.idr
```diff
@@ -33,36 +33,25 @@ length xs = length_aux xs Z
...
-  ==| Refl,
-  Data.List.TailRec.length xs
-  ==| QED
-  ]
+length_ext xs = Calc $
+  |~ Data.List.length xs                
+  ~~ Data.List.length xs + 0             ...( sym $ plusZeroRightNeutral $ Data.List.length xs ) 
+  ~~ Data.List.TailRec.length_aux xs 0   ...( lemma 0 xs )                                       
+  ~~ Data.List.TailRec.length xs  ...( Refl )                                             
   where
```
gallais, line 41: Should we have a notation for qed? (Also: alignment)

ohad:
@gallais : both syntactic comments are because I reverted `import Data.List as DL`, so once we implemented `import .. as`, these will go away (just kidding, will push a fix).

Re. `QED`: it's not needed, we have a `DEQ` in the beginning of the proof.

PR #318, opened 2020-06-16
Add Homebrew installation instructions

### LINKED ISSUE #317: Mention Homebrew formula as alternative installation method
Usually, I would just do a PR adding those few lines of markdown but I'm unsure where do you want to place them 🤔 
If anyone can point me in the right direction I'll gladly update docs myself or someone else can do that if they like. It is distributed under `idris2` moniker: https://formulae.brew.sh/formula/idris2#default 🎉

vladimyr:
This adds Homebrew installation instructions to `INSTALL.md` and Getting Started guide.
Closes #317

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,11 +1,15 @@
-Installing
-==========
+# Installing
+
+- [Installing from source](#installing-from-source)
+- [Installing from a package mananager](#installing-from-a-package-manager)
+
+## Installing from source
 
 The easiest way to install is via the existing generated Scheme code.
 The requirements are:
 
-* A Scheme compiler; either Chez Scheme (default), or Racket.
-* `bash`, with `realpath`. On Linux, you probably already have this.
+- A Scheme compiler; either Chez Scheme (default), or Racket.
+- `bash`, with `realpath`. On Linux, you probably already have this.
   On a Mac, you can install this with `brew install coreutils`.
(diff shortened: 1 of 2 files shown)


vladimyr:
Note that Homebrew distribution does not require `bash` nor `realpath` and instead links directly to compiled chez script with the following hashbang: `#!/usr/local/opt/chezscheme/bin/chez --script` ⚠️

melted:
Thanks!

Nice as installing from brew is, I don't think we should lead off with an installation method that not everybody can use and doesn't give them a fresh install from git. I propose that you split that first section into "installing from source" and "installing from a package manager" and move the brew stuff to the later one. Then hopefully we can add more to that later.

melted:
You can also make links to the sections and put on top so people can just jump to the one they're interested in.

vladimyr:
@melted I mostly did it because looking through my own eyes I just want to find quickest possible way to install and be done with it so I can focus on more fun stuff. But that is pretty much addressed with your other comment suggesting to implement clickable goto-s. That surely sounds like a better option, so I'll update PR accordingly.

> Nice as installing from brew is, I don't think we should lead off with an installation method that not everybody can use and doesn't give them a fresh install from git.

FWIW you can get the latest version directly from git using brew and compile it on your own machine:
```
brew install --HEAD idris2
```

vladimyr:
@melted I updated PR as you suggested. Does it look ok now?

melted:
Sure!

PR #319, opened 2020-06-17
Add support for OpenBSD

ska80:
and probably for other *BSD operating systems with minor tweaks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,6 +146,7 @@ else
 endif
 	SCHEME=${SCHEME} \
 	IDRIS2_VERSION=${IDRIS2_VERSION} \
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -146,6 +146,7 @@ else
 endif
 	SCHEME=${SCHEME} \
 	IDRIS2_VERSION=${IDRIS2_VERSION} \
+	MAKE=${MAKE} \
```
melted, line 149: you could do an `export MAKE` in config.mk instead.

ska80: Thanks.  Also removed `${SCHEME}` and `${IDRIS2_VERSION}` as they are already exported too.

PR #321, opened 2020-06-17
Add comment about TTC files in TTC error

prozacchiwawa:
In using local builds of idris2 I often forget that there's a build dir containing TTC files and lose time to thinking something other than "you last compiled this with a previous version of idris2" is wrong.  This just adds to the error message.


gallais:
Thanks & sorry it took so long. I am not sure whether we are trying
to layout error messages to avoid long lines or not. But this is an
improvement over the current state of affairs anyways.

prozacchiwawa:
NP, and don't worry.  Thanks for having a look :-)

PR #322, opened 2020-06-17
Function mapping `Not (x=True)` to `x=False` was added for `Bool`s

buzden:
This function may be useful when using `decEq` for `Bool`s with `with` where previously `with ... proof` was used, e.g.:

```idris
f : X -> Y
f x with (decEq whatever True)
  | Yes p = rewrite p in ...
  | No n = rewrite notTrueIsFalse n in ...
```
(where previously we were used to write
```idris
f : X -> Y
f x with (whatever) proof p
  | True = rewrite p in ...
  | False = rewrite p in ...
```
)


buzden:
I don't know what is the meaning of the number in the changed test, thus I have no idea why they changed because of library change, but still...

gallais:
Thanks!

PR #323, opened 2020-06-18
Add missing modules in base.ipkg

berewt:
`Data.Fuel` was missing.
A few other in `Data.App` were missing as well, but they do not compile, I don't know if we must fix or remove them.


gallais:
Thanks! As for the other modules, I guess they can stay as
they are for now (we have quite a few commented out proofs
in various files because they are currently broken for
instance).

PR #324, opened 2020-06-18
Mention custom codegens in the change log

melted:


melted:
Heh, 7/8 CI runs are good enough for a changelog update

PR #309, opened 2020-06-15
Add 'optional' command line options

ska80:
Also combine `--ide-mode-socket` and `--ide-mode-socket-with` options into one single `--ide-mode-socket` option.

PR #327, opened 2020-06-18
Allow overriding build and output directory

chrrasmussen:
This pull request contains more changes than I originally intended:
1. Adds `--build-dir` and `--output-dir` and corresponding fields in the package description (See more detailed description below)
2. Adds `--source-dir` to match the existing `sourcedir` field in the package description (See overview at the bottom)
3. Renamed `dir_prefix` to `prefix_dir` to match the other names in the record
4. Fixes a bug regarding the `main` field in the package description (Previously it always used `Main` as the namespace)
5. Silences the compilation message when compiling Chez programs: `compiling <path>/print_hello.ss with output to <path>/print_hello.so`
   - I should probably emphasize that by silencing this message, there might be cases where you will miss seeing this message, for example when building a new Idris 2 executable. However, I believe silencing this message is for the better, because then we can control when/if we want to print such a message.

Point 4 and 5 showed up when adding the test case for point 1.

I recommend looking at the individual commits to understand the changes in isolation. If you want to I can split this pull request into multiple smaller pull requests.

## Description of `--build-dir` and `--output-dir`

This pull request adds the ability to change the location of the build directory (defaults to `build`) and the location of the output directory (defaults to `build/exec`). I changed the name of "executable directory" to "output directory" because the output isn't necessarily an executable. The output may consist of source files, or even a compiled library, depending on the code generator.

The code generators are now provided with the following path arguments:
- `compileExpr`: `tmpDir` and `outputDir`
- `executeExpr`: `tmpDir`

Since the `outputDir` can be chosen by the user, the code generator can decide to put temporary files in `tmpDir` and put the desired files in `outputDir`. By default, both `tmpDir` and `outputDir` will point to the same directory (`build/exec`). The code generator can assume that both paths exist. In the case of the Chez and Racket code generators, all their files will go to `outputDir`.

Concretely, I have added the following options to the `idris2` CLI:
```
  --source-dir <dir>                  Set source directory
  --build-dir <dir>                   Set build directory
  --output-dir <dir>                  Set output directory
```

And similar fields to the package description (overridable by CLI):
```
builddir = "build"
outputdir = "build/exec"
```

I have added a test (`idris2/pkg005`) that checks that the `builddir` and `outputdir` fields in the package description works. I have manually tested that all 3 code generators works when specifying the `--build-dir` and `--output-dir`.

~~This pull request will most certainly have conflicts with #309: I will rebase this PR after #309 is merged.~~

---

An overview of the available directory paths and how to override them:

| Directory   | CLI                    | REPL             | Package             | ENV |
|-------------|------------------------|------------------|---------------------|-----|
| working_dir |                        | `:cd`            |                     |     |
| source_dir  | `--source-dir` **New** |                  | `sourcedir`         |     |
| build_dir   | `--build-dir` **New**  |                  | `builddir` **New**  |     |
| output_dir  | `--output-dir` **New** |                  | `outputdir` **New** |     |
| prefix_dir  |                        |                  |                     | IDRIS2_PREFIX |
| extra_dirs  | `-p`                   |                  |                     | IDRIS2_PATH   |
| lib_dirs    |                        |                  |                     | IDRIS2_LIBS   |
| data_dirs   |                        |                  |                     | IDRIS2_DATA   |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ Command-line options changes:
 
 * Removed `--ide-mode-socket-with` option.  `--ide-mode-socket` now accepts an
   optional `host:port` argument.
+* Added options to override source directory, build directory and output
+  directory: `--source-dir`, `--build-dir`, `--output-dir`.
+  + These options are also available as fields in the package description:
+    `sourcedir`, `builddir`, `outputdir`.
 
 Compiler changes:
 
--- a/docs/source/backends/custom.rst
+++ b/docs/source/backends/custom.rst
@@ -27,13 +27,13 @@ Now create a file containing
     import Compiler.Common
     import Idris.Driver
 
-    compile : Ref Ctxt Defs -> (execDir : String) ->
+    compile : Ref Ctxt Defs -> (tmpDir : String) -> (outputDir : String) ->
             ClosedTerm -> (outfile : String) -> Core (Maybe String)
-    compile defs dir term file = do coreLift $ putStrLn "I'd rather not."
+    compile defs tmpDir outputDir term file = do coreLift $ putStrLn "I'd rather not."
                                     pure $ Nothing
 
-    execute : Ref Ctxt Defs -> (execDir : String) -> ClosedTerm -> Core ()
-    execute defs dir term = do coreLift $ putStrLn "Maybe in an hour."
+    execute : Ref Ctxt Defs -> (tmpDir : String) -> ClosedTerm -> Core ()
+    execute defs tmpDir term = do coreLift $ putStrLn "Maybe in an hour."
 
     lazyCodegen : Codegen
     lazyCodegen = MkCG compile execute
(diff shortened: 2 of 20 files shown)


chrrasmussen:
Thanks for merging #309! Will look into the failing check soon.

chrrasmussen:
After reproducing the error, I realised that the problem is that the package is generating and building a Chez program on a machine that only has Racket installed. Because of that I moved the test to the `chez` directory, as test `chez026`.

The pull request is ready to be reviewed/merged.

melted:
LGTM!

jfdm:
It will be good to have the documentation in docs/ updated where applicable as well. There is a lot of information in the github message at the start of this thread. That would be a good source of documentation to develop from.

chrrasmussen:
That's a good point! I found a few pages where it make sense to expand with this new information:
- https://idris2.readthedocs.io/en/latest/reference/packages.html
- https://idris2.readthedocs.io/en/latest/backends/index.html
- https://idris2.readthedocs.io/en/latest/backends/custom.html

I will look into it by the end of this week. Thanks for the nudge!

PR #332, opened 2020-06-19
VMCODE: Carry over required type names for named constructors without a tag

cypheon:
For constructors which don't have a tag (so case matching needs to be done based on the `Name`), the `Name` information is correctly carried over into the `CASE` instructions (in form of an `Either Int Name`), but is lost when converting the ANF `ACon` to a VMCode `MKCON`.

For reference, the current conversion from ANF to VMCode `CASE` looks like this:
```idris
    toVMConAlt (MkAConAlt n (Just tag) args code)
        = (Left tag, projectArgs 0 args ++ toVM t res code)
    toVMConAlt (MkAConAlt n Nothing args code)
        = (Right n, projectArgs 0 args ++ toVM t res code)
```

This commit slightly changes the constructor for the `MKCON` instruction to match the type of the `CASE` alternatives and changes the conversion from ANF to keep the original constructor name. As before, backends still need to do the `Name` -> unique Int mapping, if they want to perform the case match on "named constructors" efficiently.

### DIFF
--- a/src/Compiler/VMCode.idr
+++ b/src/Compiler/VMCode.idr
@@ -30,7 +30,7 @@ data VMInst : Type where
      DECLARE : Reg -> VMInst
      START : VMInst -- start of the main body of the function
      ASSIGN : Reg -> Reg -> VMInst
-     MKCON : Reg -> (tag : Maybe Int) -> (args : List Reg) -> VMInst
+     MKCON : Reg -> (tag : Either Int Name) -> (args : List Reg) -> VMInst
      MKCLOSURE : Reg -> Name -> (missing : Nat) -> (args : List Reg) -> VMInst
      MKCONSTANT : Reg -> Constant -> VMInst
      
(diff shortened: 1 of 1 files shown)


gallais:
Can we had a test case where Idris generates such a thing?
(Should we not try to fix that to always have the more efficient
dispatching strategy?)

melted:
I wondered about this before, so I looked it up. These come from TyCons, and there are none in an idris 2 build. Guess it's better safe than sorry anyway.

melted:
Another oddity with VMCode: VMDef drops the foreign functions, so you have to pick them up from an earlier stage.

melted:
I'll merge this. I think we can do without a test case, as we have none currently about any of Lifted, ANF or VMCode and this edge case is hardly the first thing we should test.

cypheon:
That was quick, thanks for merging this!

There actually is code in a test case, which triggers this behaviour (that's how I discovered it):
`tests/chez/chez006/TypeCase.idr` (edited for brevity):
```idris
foo : (x : Type) -> String
foo Nat = "Nat"
[…]
main = do printLn (foo Nat)
```

Generated VMCode for `main` (linebreaks added):
```
Main.main = [0]: [DECLARE v1, DECLARE v2, DECLARE v4, DECLARE v3, DECLARE v5, START,
v1 := MKCLOSURE Main.{main:0} 1 (v0),
v2 := MKCLOSURE Main.{main:1} 2 (v0),
v4 := MKCON Left 0 (v1, v2),
v3 := MKCON Right Prelude.Nat (),
v5 := CALL Main.foo(v3),
[…]
```
Before, this would have looked like `v4 := MKCON Just 0 (v1, v2),` and `v3 := MKCON Nothing (),`, which left backends no chance to generate any meaning full code (for the `MKCON Nothing ()` part)

PR #333, opened 2020-06-19
Make Data.Vect linear

mb64:
This PR changes `Data.Vect.::` to be a linear constructor, like its `List` counterpart.

It further modifies many of the linear functions in `Data.Vect` to be annotated as such.

However, it falls short of making *everthing* that could be linear linear: higher-order functions remain nonlinear, to retain their use with nonlinear arguments.

I believe that although it adds some complexity, this change makes it overall conceptually simpler and much less surprising: one would expect that a function `{0 a : Type} -> (1 xs : Vect 4 a) -> Vect 3 a` would be impossible to write, but without this change, it's as easy as `tail`.  The goal of this change is to make `Vect` behave more "normally" wrt linearity.

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -182,7 +182,7 @@ merge = mergeBy compare
 public export
 reverse : (1 xs : Vect len elem) -> Vect len elem
 reverse xs = go [] xs
-  where go : Vect n elem -> Vect m elem -> Vect (n+m) elem
+  where go : (1 _ : Vect n elem) -> (1 _ : Vect m elem) -> Vect (n+m) elem
         go {n}         acc []        = rewrite plusZeroRightNeutral n in acc
         go {n} {m=S m} acc (x :: xs) = rewrite sym $ plusSuccRightSucc n m
                                        in go (x::acc) xs
@@ -284,8 +284,8 @@ zip3 (x::xs) (y::ys) (z::zs) = (x, y, z) :: zip3 xs ys zs
 public export
 unzip : (1 xs : Vect n (a, b)) -> (Vect n a, Vect n b)
 unzip []           = ([], [])
-unzip ((l, r)::xs) with (unzip xs)
-  unzip ((l, r)::xs) | (lefts, rights) = (l::lefts, r::rights)
+unzip ((l, r)::xs) = let (lefts, rights) = unzip xs
+                     in (l::lefts, r::rights)
 
 ||| Convert a vector of three-tuples to a triplet of vectors
 |||
@@ -295,9 +295,8 @@ unzip ((l, r)::xs) with (unzip xs)
 public export
 unzip3 : (1 xs : Vect n (a, b, c)) -> (Vect n a, Vect n b, Vect n c)
 unzip3 []            = ([], [], [])
-unzip3 ((l,c,r)::xs) with (unzip3 xs)
-  unzip3 ((l,c,r)::xs) | (lefts, centers, rights)
-      = (l::lefts, c::centers, r::rights)
+unzip3 ((l,c,r)::xs) = let (lefts, centers, rights) = unzip3 xs
+                       in (l::lefts, c::centers, r::rights)
 
 --------------------------------------------------------------------------------
 -- Equality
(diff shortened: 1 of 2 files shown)


ziman:
Nice!

libs/base/Data/Vect.idr
```diff
@@ -181,9 +180,9 @@ merge = mergeBy compare
 ||| reverse [1,2,3,4]
 ||| ```
 public export
-reverse : Vect len elem -> Vect len elem
+reverse : (1 xs : Vect len elem) -> Vect len elem
 reverse xs = go [] xs
```
gallais, line 184: Worth noting that we can have a [proof-free implementation of `reverse`
using `foldl`](http://agda.github.io/agda-stdlib/Data.Vec.Base.html#7735).

libs/base/Data/Vect.idr
```diff
@@ -224,7 +223,7 @@ fromList' {len} ys (x::xs) =
 ||| fromList [1,2,3,4]
 ||| ```
 public export
-fromList : (l : List elem) -> Vect (length l) elem
+fromList : (1 l : List elem) -> Vect (length l) elem
```
gallais, line 226: Why is this function so complicated?

```idris
fromList [] = []
fromList (x :: xs) = x :: fromList xs
```

ought to work! Consequently, I'm not sure what the use case for `fromList'` is.

libs/base/Data/Vect.idr
```diff
@@ -780,9 +780,15 @@ range {len=S _} = FZ :: map FS range
...
-transpose (x :: xs) = zipWith (::) x (transpose xs) -- = [| x :: xs |]
+transpose : {n : _} -> (1 array : Vect m (Vect n elem)) -> Vect n (Vect m elem)
+transpose []        = replicate _ []          -- = [| [] |]
+transpose (x :: xs) = helper x (transpose xs) -- = [| x :: xs |]
+  where -- Can't use zipWith since it doesn't preserve linearity
+    helper : (1 _ : Vect a elem)
+          -> (1 _ : Vect a (Vect b elem))
+          -> Vect a (Vect (S b) elem)
+    helper [] [] = []
+    helper (x::xs) (tl::tls) = (x::tl) :: helper xs tls
```
gallais, line 791: We cannot use `zipWith` but we could use applicative combinators.
Using Haskell notations, this is `(::) <$> xs <*> tls`.

gallais: Alternatively we could come up with a naming scheme & provide
linear variants of the higher order functions. e.g. `zipWith1`.

mb64: Applicative combinators don't work either: `map` only maps unrestricted arrows to unrestricted arrows, and similary for `<*>`.  (I'm also not yet convinced that it *should* work in the general case – that is a question for someone with more knowledge of category theory than me.)

As for having separate linear variants for the higher-order functions, I like the idea, at least until Idris gets multiplicity polymorphism (which I'm hoping it does).  I don't like the `1` suffix though, due to potential confusion with `fold[lr]1`.  However, I think that adding those should be the job of a different PR than this one.

libs/base/Data/Vect.idr
mb64: While `foldr (flip cons)` would be great, it's again the issue that `foldr` and its higher-order friends only talk about unrestricted functions, so we have to implement the fold manually.

libs/base/Data/Vect.idr
mb64: It makes it tail recursive, which I would guess is more important in a strict language like Idris so that you won't bash your stack trying to convert an especially long list to a `Vect`.  Not certain tho – perhaps @edwinb can comment here.

ziman: We've had this conversation about tail-recursive `public export` functions a few times recently ([#132](https://github.com/idris-lang/Idris2/issues/132#issuecomment-633285153), [#269](https://github.com/idris-lang/Idris2/pull/269#issuecomment-643168588)), so I'll just link what I'd repeat here. :)

There are issues with what I suggest, though, as @gallais pointed out, so take it just as an illustration why making everything tail recursive is not automatically a good thing.

mb64:
I rebased and updated the test output.  As for `fromList`, I'm choosing to leave it as is in this PR; I don't feel like I have the authority or expertise to decide its eventual fate.

gallais:
Fair enough. Thanks for the PR!

PR #334, opened 2020-06-19
Make reference and source agree about literate markdown

mb64:
In the CommonMark section of the reference for literate Idris, it says that both `.md` files and `.markdown` files are treated as Org-Mode style literate Idris.  I believe this should say CommonMark style.

When looking at the source to back this up, however, I saw that it only interprets `.md` files as CommonMark-style literate Idris files, and doesn't allow `.markdown` files.

This PR fixes both of these.


gallais:
Thank you!

PR #335, opened 2020-06-19
Vect updates

clayrat:
Added injectivity and DecEq instance, refactored `Vect.Elem` and moved it into a separate submodule to match `Data.List`.


libs/base/Data/Vect/Elem.idr
```diff
@@ -0,0 +1,73 @@
...
+  Here : Elem x (x::xs)
+  There : (later : Elem x xs) -> Elem x (y::xs)
+
+export
+Uninhabited (Elem x []) where
+  uninhabited Here impossible
+
+||| An item not in the head and not in the tail is not in the Vect at all
+export
+neitherHereNorThere : {x, y : a} -> {xs : Vect n a} -> Not (x = y) -> Not (Elem x xs) -> Not (Elem x (y :: xs))
```
gallais, line 22: Would `{0 x, y : a}` not do?

libs/base/Data/Vect/Elem.idr
```diff
@@ -0,0 +1,73 @@
...
+  replaceElem (x::xs) (There xinxs) y | (ys ** yinys) = (x :: ys ** There yinys)
+
+public export
+replaceByElem : (xs : Vect k t) -> Elem x xs -> t -> Vect k t
+replaceByElem (x::xs)  Here         y = y :: xs
+replaceByElem (x::xs) (There xinxs) y = x :: replaceByElem xs xinxs y
+
+public export
+mapElem : {0 xs : Vect k t} -> {0 f : t -> u} ->
+          Elem x xs -> Elem (f x) (map f xs)
```
gallais, line 49: I expect we can make it `(1 _ : Elem x xs)`.
Not sure whether we should always be as precise as possible in the lib?

clayrat: Let's try, I guess :)

melted:
Maybe we should do some sed magic for coverage010, it's irritating.

PR #337, opened 2020-06-20
change how unelabBinder shows names

MarcelineVQ:
unelabBinder was losing information for the purpose of displaying names nicely. This is typically wanted but when working with elabortation it's useful to have all the information you can get.  
Things like record field names would be lost when querying with GetType but are now retained.


edwinb:
This looks like a useful thing, and come to think of it is part of the part of the NoSugar flag. I don't know what's up with the CI. Maybe a glitch? Can we force it to restart somehow?

PR #343, opened 2020-06-20
Postfix projections

ziman:
This is a port of the [original PR](https://github.com/edwinb/Idris2-boot/pull/342).

Included changes:
* `.proj` is now allowed for any identifier `proj`, not only record projections.
* `.(complex expression)`, which works just like `.proj` but you can have an arbitrary expression on the RHS. This is feature-gated by `%language PostfixProjections`.
* The special `RF` constructor for names of the form `.recordField` is gone; it does not make much sense anymore with this more general syntax.

Even without `%language PostfixProjections` enabled, this patch expands the set of accepted programs by allowing non-record-projections in `.proj`. I'm not sure how to feature-gate this easily but maybe it's okay as it is.


ziman:
I'm not sure about the name `PostfixProjections`; maybe `ComplexPostfixProjections` would be more accurate? But I find that a bit too long.

ziman:
I added a patch that changes the following:
* Bare `.proj` without a preceding operand is no longer valid syntax, so `f $ .proj` no longer parses. You have to write `f $ (.proj)` or `f (.proj)` or `f proj`.
* We now support projection sections: `(.proj1.proj2 y z)` is equivalent to `(\x => x.proj1.proj2 y z)`. This is feature-gated by `%language PostfixProjections`.

docs/source/reference/records.rst
```diff
@@ -21,7 +20,7 @@ Lexical structure
 
 * ``.foo.bar.baz`` is three lexemes: ``.foo``, ``.bar``, ``.baz``
 
-* If you want ``Constructor.field``, you have to write ``(Constructor).field``.
+* If you want ``Constructor.proj``, you have to write ``(Constructor).proj``.
```
gallais, line 23: Is this `Constructor` a `TyCon` (i.e. you're using the record's namespace)?
Could we have a small example with a concrete record?

ziman: At lexing/syntax level, we do not (cannot) distinguish between namespaces, data constructors and type constructors. All the lexer sees is a "capitalised identifier, dot, uncapitalised identifier", and lexes it as a namespaced name. If you want something like `MkFoo.bar`, where `MkFoo` is a (syntactically) nullary data constructor, you need to write `(MkFoo).bar` because a capitalised identifier _not_ followed by a dot is lexed as an (unqualified) identifier, not as a module name.

`(Constructor).field` is literally sugar for `field Constructor`; the desugarer does not look at whether `field` is record field, and does not qualify it using the record namespace. It leaves that to the disambiguator.

For example:
```idris
record Foo where
  constructor MkFoo
  {auto field : ()}

MkFoo.field  -- fails, parsed as NS ["MkFoo"] (UN "field")
(MkFoo).field  -- succeeds, parsed as (field MkFoo)
Foo.field  -- succeeds, parsed as NS ["Foo"] (UN "field")

MkFoo.Foo  -- fails, parsed as NS ["MkFoo"] (UN "Foo")
MkFoo.(Foo.field)  -- succeeds, parsed as (Foo.field MkFoo)
MkFoo.Foo.field  -- fails, parsed as NS ["Foo", "MkFoo"] (UN "field")
(MkFoo.Foo).field  -- fails, parsed as (field MkFoo.Foo)
```
Should I put this in the doc?

gallais: > Should I put this in the doc?

I think so: a "nullary" constructor with (invisible) fields is exotic
enough that it would be easier to understand the original sentence
with an example.

ziman: Perhaps `Z.printLn` would be a better example?

ziman: I'll update it, thanks!

ziman: (Or `True.not`.)

gallais: Yes to all because these also demonstrate non-field examples! :D

ziman: I'll overhaul the doc, actually; this should no longer be called "record projections", either.

gallais: Beautiful!

gallais:
Are we happy with this PR? I am happy with it.
Just need a quick rebase & then we can merge it.

ziman:
I rebased it and fixed the tests. Let's hope the CI agrees, too.

Russoul:
@ziman expression `x .f y .g z` parses like `(x .f) (y .g) z`.
May it be `(x .f y) .g z` ?
## Motivation:
ability to use 'standard', 'scala-like' function application syntax:
suppose:
`map : m a -> (a -> b) -> m b`
`reduce : m a -> (a -> a -> a) -> a`
then this would be possible:
`list.map(\x => 2 * x).reduce(+)` == `reduce (map list (\x => 2 * x)) (+)`
or
`students.target("Peter").update(.expelled)` == `update (target students "Peter") (\s => expelled s)`

but currently it is:
`(list.map) ((\x => 2 * x).reduce(+))` == `map list (reduce (\x => 2 * x) (+))`
or
`(students.target) (("Peter").update(.expelled))`

ziman:
I can see where you're coming from but I think that the suggestion goes against the purpose of postfix projections, which is binding tighter than application. (Since you cannot define so tightly binding operators yourself, a change to the syntax was necessary.)

The suggestion would make `f x.y p.q` interpreted as `q (y p (f x))` rather than `f (y x) (q p)`, if I understand correctly, you'd lose the ability to access record fields in applications, which was the motivation for postfix projections.

I think that what you suggest would be better achieved by borrowing `infixl (&) : a -> (a -> b) -> b` from Haskell. Then you could write `list & map (2 *) & reduce (+)`. I think this should behave as you suggest, except that you'd write `&` instead of `.`.

Alternatively, you can currently write `list.(map (2 *)).(reduce (+) 0)`.

Russoul:
@ziman yeah, gotcha

PR #344, opened 2020-06-20
Updated documentation for export

TomasPuverle:
Based on discussion on #irc

### DIFF
--- a/docs/source/tutorial/modules.rst
+++ b/docs/source/tutorial/modules.rst
@@ -117,7 +117,7 @@ Meaning for Functions
   about doing this.
 
 .. note::
-
+    
    Type synonyms in Idris are created by writing a function. When
    setting the visibility for a module, it is usually a good idea to
    ``public export`` all type synonyms if they are to be used outside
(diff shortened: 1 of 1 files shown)


docs/source/tutorial/modules.rst
```diff
@@ -129,6 +129,19 @@ this effectively makes the function definition part of the module's API.
...
 
+.. note::
+    *For beginners*: 
+    If the function needs to be accessed only at runtime, use ``export``.
+    However, if it's also meant to be used at *compile* time (e.g. to prove 
+    a theorem), use ``public export``.  
+    For example, consider the function ``plus: Nat -> Nat -> Nat`` discussed
+    previously, and the following theorem: ``thm: plus Z m = m``.  
+    In order to to prove it, the type checker needs to reduce ``plus Z m`` to ``m`` 
+    (and hence obtain ``thm: m = m``).
```
karroffel, line 140: Just a nitpick but I think it's common to have spaces around the `:` in type signatures.

```suggestion
    For example, consider the function ``plus : Nat -> Nat -> Nat`` discussed
    previously, and the following theorem: ``thm : plus Z m = m``.  
    In order to to prove it, the type checker needs to reduce ``plus Z m`` to ``m`` 
    (and hence obtain ``thm : m = m``).
```

PR #345, opened 2020-06-21
HasIO interface for IO actions

edwinb:
Also updates the Prelude and some base libraries to use HasIO rather than using IO directly. This is a draft, since it hasn't covered all the libraries yet (and some need a bit more thought, especially the concurrency primitives) and the documentation needs updating, especially FFI.

My main motivation is that whenever I experiment with some new abstraction for effects or linearity, I need to lift all the primitive IO operations somehow, and I wouldn't have to if they were in an appropriate interface. `HasIO` is basically like `MonadIO` from Haskell. My hope is that if everything lives in `HasIO` then we'll all get in the habit of adding primitives this way rather than putting them directly in `IO`!

* Is there a better name than `HasIO`? I'm not especially attached to it (but I don't like the name `MonadIO`)
* Are there other things that should get this treatment?
* Is there some potential problem I haven't thought of?

In case of worries about performance, I rebuilt Idris 2 using these libraries, then rebuilt again using the result, and the build times I'm getting for that result are within the usual range even though Idris2 itself does quite a lot of IO operations. So, at least, I'm not worried about that.

### DIFF
--- a/docs/source/ffi/ffi.rst
+++ b/docs/source/ffi/ffi.rst
@@ -89,8 +89,8 @@ returns a primitive IO action:
 Internally, ``PrimIO Int`` is a function which takes the current (linear)
 state of the world, and returns an ``Int`` with an updated state of the world.
 In general, ``IO`` operations in an Idris program are defined as instances
-of the ``HasIO`` interface (or ``MonadIO``). We can convert a primitive
-operation to one usable in ``HasIO`` using ``primIO``:
+of the ``HasIO`` interface. We can convert a primitive operation to one usable
+in ``HasIO`` using ``primIO``:
 
 .. code-block:: idris
 
--- a/docs/source/ffi/ffi.rst
+++ b/docs/source/ffi/ffi.rst
@@ -94,7 +94,7 @@ in ``HasIO`` using ``primIO``:
 
 .. code-block:: idris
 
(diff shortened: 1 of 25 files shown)


edwinb:
Following some discussion on the slack, I've split this into `HasIO` and `MonadIO`. While I still don't especially like the name `MonadIO`, if it really does require `Monad` then I suppose it's fine...

docs/source/ffi/ffi.rst
```diff
@@ -88,17 +88,19 @@ returns a primitive IO action:
...
 state of the world, and returns an ``Int`` with an updated state of the world.
-We can convert this into an ``IO`` action using ``primIO``:
+In general, ``IO`` operations in an Idris program are defined as instances
+of the ``HasIO`` interface (or ``MonadIO``). We can convert a primitive
+operation to one usable in ``HasIO`` using ``primIO``:
 
 .. code-block:: idris
 
-    primIO : PrimIO a -> IO a
+    primIO : HasIO a => PrimIO a -> io a
```
MarcelineVQ, line 97: Should say `HasIO io` iiuc

edwinb: oh my. How does this even work?

edwinb: Oh right, it's the documentation :). I'll fix it.

PR #348, opened 2020-06-21
change namespace parser to have minimum indentation

MarcelineVQ:
The namespace parser was not requiring a minimum indentation and instead
based its indentation on the following line, which meant that a line like:
```
namespace Foo
foodef : Int
```
placed `foodef` into namespace `Foo` instead of the module's top level.
And so made it unclear when a namespace ends.


gallais:
I am not too familiar with namespaces. Can we have a test case
showing that `foodef` is not in scope in namespace `Foo`?

MarcelineVQ:
for sure, hopefully this test covers the cases people might have with changes in the future too

PR #351, opened 2020-06-21
Restore Bool operators precedence

petithug:
(&&) traditionally has higher precedence than (||).

Note that this commit requires to bootstrap again.


petithug:
Should I rebase that patch and submit it too to Idris2-boot?

melted:
You can do that, it's best to keep the base libs reasonably in sync.

petithug:
I make some test and patching idris2-boot seems unnecessary, probably because the code in that repo does not require the modification.

melted:
Let's merge this!

PR #366, opened 2020-06-25
Mutate linear values within a limited scope

andrevidela:
This is not a pull request intended for merging, rather I need help with a project I'm working on and linking to parts of the code is much easier here.

I'm trying to optimise the runtime for this specific situation

```idris
let 1 v : Constructed = ConstructValue 3
    1 w = update v in w
```

In this case `ConstructValue` create a linear value, which is then immediately passed to `update : Constructed -> Constructed`.

the `update` function currently creates a new copy of the value, but since we _know_ that the value is _linear_ and created _in scope_ we can deduce that it is _unique_ and therefore subject to mutation.

This patch has therefore 2 changes to make:
1. detect when a value is bound linearly by a constructor and declare it as "safe to mutate"
2. change the codepath for values that are "safe to mutate" to go through a "mutating" version of themselves

`1.` is achieved during [let elaboration](https://github.com/idris-lang/Idris2/compare/master...andrevidela:mutate-case#diff-f5b3bc2e95ce5a568780c8b68464ee1fR182), the right hand side is checked for reference to a constructor and if that's the case, tags the constructor with the intended linearity

`2.` is achieved  by adding "mutating cases" to each pattern match in [CExp](https://github.com/idris-lang/Idris2/compare/master...andrevidela:mutate-case#diff-c10a8933fe146f8734cffba72477feb4R51-R61)
There is an additional `CMut` constructor for CExp that indicates to the backend that the value refered needs to be mutated instead of constructed.

The problem I'm facing now is that when inspecting `DataCon` in [toCExp](https://github.com/idris-lang/Idris2/compare/master...andrevidela:mutate-case#diff-aa65bc544c7551c680d3ba5db189aaeaR235) in order to update the tag to the mutating version of it. Every value I inspect has `RigW`, as if my pass through `1.` didn't go through. However my logs indicate that the rig count has indeed been correctly replaced. Why? What am I missing? Is updating the rig count in `checkLet` not the right way to go?

Please do not waste your time reviewing any of it since all of it is a huge WIP with half baked bits and pieces all over the place.

### DIFF
--- a/src/Compiler/ANF.idr
+++ b/src/Compiler/ANF.idr
@@ -30,6 +30,7 @@ mutual
     AApp : FC -> (closure : AVar) -> (arg : AVar) -> ANF
     ALet : FC -> (var : Int) -> ANF -> ANF -> ANF
     ACon : FC -> Name -> (tag : Maybe Int) -> List AVar -> ANF
+    AMut : FC -> Name -> List AVar -> ANF
     AOp : FC -> PrimFn arity -> Vect arity AVar -> ANF
     AExtPrim : FC -> Name -> List AVar -> ANF
     AConCase : FC -> AVar -> List AConAlt -> Maybe ANF -> ANF
@@ -67,7 +68,7 @@ mutual
     show (AAppName fc n args)
         = show n ++ "(" ++ showSep ", " (map show args) ++ ")"
     show (AUnderApp fc n m args)
-        = "<" ++ show n ++ " underapp " ++ show m ++ ">(" ++ 
+        = "<" ++ show n ++ " underapp " ++ show m ++ ">(" ++
           showSep ", " (map show args) ++ ")"
     show (AApp fc c arg)
         = show c ++ " @ (" ++ show arg ++ ")"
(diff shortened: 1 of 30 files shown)


edwinb:
I'm still trying to work out all the details of what's needed in here - maybe I'm oversimplifying this, but I would have thought an easier and more general way to go would be to record the count at the point the constructor is checked (https://github.com/idris-lang/Idris2/blob/master/src/TTImp/Elab/App.idr#L59-L64) rather than post-processing in the `let`.

Even so, if you say it's being recorded as a `Rig1` properly, that might not be any help...

andrevidela:
So I've looked at this but I don't understand how I'm supposed to get the linearity of the binder from there. Additionally I tried simply setting the linearity at `linear` but the logs still show that the DataCon has linearity `RigW` (when converting to CExp) so even if that was the right way to go, it doesn't actually fix the problem

andrevidela:
I figured it out. It was converted to NDCon and back and the linearity was lost there.

PR #369, opened 2020-06-26
Make Prelude.unpack tail-recursive

cypheon:
Being tail-recursive doesn't matter that much for Chez Scheme with its (seemingly?) unbounded call stack. But when targeting a runtime with a smaller stack / bounded call stack size, it's one less thing to worry about.


edwinb:
Looks a good idea, thanks. My guess is that Chez's use of continuations internally will make it looks like it has a huge call stack, but it still won't be free, and other back ends will benefit.

PR #386, opened 2020-06-29
Add `%default total` to a few modules, implement Uninhabited for Bool and Maybe, and some minor simplifications

nickdrozd:


libs/base/Data/Fin.idr
```diff
@@ -15,47 +17,46 @@ data Fin : (n : Nat) -> Type where
...
-    (==) _ _ = False
+Uninhabited (FS k = FZ) where
+  uninhabited Refl impossible
 
-||| There are no elements of `Fin Z`
 export
-FinZAbsurd : Fin Z -> Void
-FinZAbsurd FZ impossible
+FSInjective : FS m = FS n -> m = n
+FSInjective Refl = Refl
```
gallais, line 35: Given the naming convention we seem to be adopting for `(succ/left/right/just)Injective`,
we should probably call this one `fsInjective`.

libs/base/Decidable/Equality.idr
```diff
@@ -123,16 +107,16 @@ implementation (DecEq a, DecEq b) => DecEq (a, b) where
 lemma_val_not_nil : (the (List _) (x :: xs) = Prelude.Nil {a = t} -> Void)
```
gallais, line 107: All of these names are pretty terrible and the lemmas not that
useful given they're direct corollaries of the fact that `(::)`
is injective.

They're not exported so we can maybe get rid of all of this
boilerplate & replace them with `p . (fst / snd) . consInjective`
in the `decEq` proof for list.

nickdrozd: Thanks, I couldn't figure out how to get that to work before.

gallais:
Thanks!

PR #389, opened 2020-06-30
List's `snoc` injectivity property was renamed approporiately.

buzden:
Existing in the `contrib` library `snoc`'s injectivity property was named as a congruence for some reason, this PR fixes it.


gallais:
Thanks!

PR #390, opened 2020-06-30
Describe what is meant by "use" in the context of linear types

redfish64:
I hit a roadblock trying to understand linear types, because I couldn't figure out what is meant to "use" a variable. I eventually think I figured it out, by reading "Linear types can change the world!", but it was not easy to do so. I think a better description of what "use" means would be very helpful for those unfamiliar with linear types, so I took a stab at it. Hope it helps.


gallais:
If you had issues with it, others probably will too.
Thanks for the doc!

PR #400, opened 2020-07-01
Fixed visibility of some list functions.

jfdm:


gallais:
Related discussion: #269

jfdm:
Just to clarify, I originally contributed the ported code and marked the functions as export, in a personal library I use these functions at the type-level but only realised the error when I looked at porting the library the other day.

gallais:
Let's just declare that things in `base` are stable enough to
do this kind of `public` export and we'll think again about
that problem if we start hitting it in more unstable parts like
`contrib`?

jfdm:
I agree, I think with `prelude` and `base` we should not assume how the code is to be used by authors. Which is something I took away from #269. 

So maybe we should ask that contributions to prelude and base *should* be `public export` unless there is a *very* good reason for them to be `export` or `private`.

PR #404, opened 2020-07-03
A port of `Subset` and `Exists` from Idris1.

jfdm:
The ports are rather straight forward and I have purposefully written
the documentation to be beginner friendly.

Note, I have diverged from Idris1 over the naming of the projection
functions to make them consistent with `Pair` and `DPair`.


jfdm:
My main rational is that I wasn't sure if I could capture the same usage constraints in record form. That is all.

jfdm:
A local check tells me one can, but a subsequent PR can be added that turns the data structures into records. Although one will now have to make Exists have an explicit parameter for the type of the witness..

gallais:
Right, you'd have to write:

```idris
  record Exists {0 type : Type} (this : type -> Type) where
```

Sounds like a bug to me though.

jfdm:
thanks!

polendri:
@jfdm Could you provide a usage example for `Subset`? I'm getting linearity errors when I try to pattern-match on `Element` in the manner that I see the Idris 1 `Subset` is used in `idris-containers`. Example:

```idris
import Data.DPair

data SBool : Bool -> Type where
     SFalse : SBool False
     STrue  : SBool True

sNot : (Subset.Subset Bool SBool) -> Bool
sNot (Element False SFalse) = True  -- Fails on v2.1.0 with "Can't match on SFalse (Erased argument)"
sNot (Element True STrue) = False

sNot' : (x : Bool) -> (0 s : SBool x) -> Bool
sNot' False SFalse = True   -- Without using Subset, matching on the erased argument succeeds.
sNot' True  STrue  = False  -- But should this not be equivalent to using Subset?
```

ziman:
This is a shortcoming of Idris, as far as I know.

Matching on forced erased constructors in Idris 2 is currently more conservative than it could be and it cannot see that the `SFalse`/`STrue` are forced constructors there. I think it currently does not look more than 1 layer deep or so, just to make cases like matching on `Refl` go through, but `Element` requires looking deeper.

polendri:
@ziman Ah, OK, it's helpful to know it's an issue of depth in the matching. Thanks!

It seems there's a workaround then of introducing an intermediate definition on the matched values inside `Element` (compiles successfully with v2.1.0):

```idris
import Data.DPair

data SBool : Bool -> Type where
     SFalse : SBool False
     STrue  : SBool True

sNot : (Subset.Subset Bool SBool) -> Bool
sNot (Element x y) = sNot' x y where
  sNot' : (b : Bool) -> (0 s : SBool b) -> Bool
  sNot' False SFalse = True
  sNot' True STrue = False
```

gallais:
Strange because [this works](https://github.com/idris-lang/Idris2/blob/master/src/Core/TT.idr#L524) and it's doing pattern-matching on a
forced runtime irrelevant argument [paired up inside of a record](https://github.com/idris-lang/Idris2/blob/master/src/Core/TT.idr#L476).

Is it because `SizedOf` is an ad-hoc definition rather than reusing
a generic data structure like `Subset`?

gallais:
@polendri Could you open a separate issue? It's very easy to lose
this discussion because it is in a merged PR rather than as an open
bug report.

polendri:
Separate issue filed as #661

PR #406, opened 2020-07-03
Add support for --log flag

andrevidela:
The flag takes a natural number as argument
It would be ideal to generalise the argument parser to be able
to parse any argument of any type, but right now we only add
`RequiredNat` as a new flag option.


melted:
Looks useful!

PR #408, opened 2020-07-04
Add concatBuffers and splitBuffer to Data.Buffer

melted:
And expand the buffer test.


libs/base/Data/Buffer.idr
```diff
@@ -274,3 +276,33 @@ resizeBuffer old newsize
...
+export
+concatBuffers : HasIO io => List Buffer -> io (Maybe Buffer)
+concatBuffers xs
+    = do let sizes = map prim__bufferSize xs
+         let (totalSize, revCumulative) = foldl scanSize (0,[]) sizes
+         let cumulative = reverse revCumulative
+         Just buf <- newBuffer totalSize
+              | Nothing => pure Nothing
+         traverse (\(b, size, watermark) => copyData b 0 size buf watermark) (zip3 xs sizes cumulative)
+         pure (Just buf)
```
gallais, line 291: I wonder whether this wouldn't be easier to read as something
written like (don't have a typechecker handy at the moment,
Haskell syntax):

```idris
let (totalSize, buffs) = flip runState 0 $ forM xs $ \ x -> do
                            offset <- get
                            let size = prim__bufferSize x
                            put (offset + size)
                            pure (x, size, offset)
(...)
traverse ... buffs
```

melted: I thought about it, it looks nicer, but I feel bad enough about dragging in `Data.List` as an import in a module that should be foundational. I don't want to import the state monad for a convenience. Besides, I quite like the functional style.

gallais: I would argue it is not just convenience as it's only one traversal instead of 4:

* map
* foldl
* reverse
* zip3

PR #411, opened 2020-07-04
Allow to override log level with package options

0nkery:
Seems like useful addition to ease packaging debugging.


jfdm:
Nice catch! I thought I had added it in when doing the original PR.

Thanks!

PR #412, opened 2020-07-05
Add command line option to typecheck package without codegen

0nkery:
```
$ time idris2 --build idris2.ipkg
Executed in   46,90 secs   fish           external 
   usr time   44,26 secs  118,00 micros   44,26 secs 
   sys time    2,46 secs  498,00 micros    2,46 secs 

$ time idris2 --checkpkg idris2.pkg
Executed in    1,68 secs   fish           external 
   usr time  1482,77 millis  132,00 micros  1482,64 millis 
   sys time  147,89 millis  591,00 micros  147,30 millis 
```

I find this command very helpful for some workflows.

### DIFF
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -508,7 +508,7 @@ processPackage cmd file opts
                       Install => do [] <- build pkg opts
                                        | errs => coreLift (exitWith (ExitFailure 1))
                                     install pkg opts
-                      Check => do
+                      Typecheck => do
                         [] <- check pkg opts
                           | errs => coreLift (exitWith (ExitFailure 1))
                         pure ()
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -276,6 +276,27 @@ compileMain mainn mmod exec
          compileExp (PRef replFC mainn) exec
          pure ()
 
+prepareCompilation : {auto c : Ref Ctxt Defs} ->
+                     {auto s : Ref Syn SyntaxInfo} ->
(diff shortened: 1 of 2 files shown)


melted:
I think it would be better named `--typecheck`. None of the other package options has a Pkg suffix either.

0nkery:
> I think it would be better named `--typecheck`. None of the other package options has a Pkg suffix either.

Good point, thanks. I've renamed option name in CLI and in code.

src/Idris/Package.idr
```diff
@@ -360,6 +360,23 @@ install pkg opts -- not used but might be in the future
...
+        Core (List Error)
+check pkg opts =
+  do
+    defs <- get Ctxt
+    addDeps pkg
+    processOptions (options pkg)
+    let toBuild = maybe (map snd (modules pkg))
+                        (\m => snd m :: map snd (modules pkg))
+                        (mainmod pkg)
+    buildAll toBuild
```
gallais, line 378: Can we refactor `build` to reuse this shared prefix? Also, looking at `build`,
it seems you are missing a call to [`preOptions`](https://github.com/idris-lang/Idris2/blob/71db21b19d42acff8ce53b8ee036287dcc67afc1/src/Idris/SetOptions.idr#L36):

> Options to be processed before type checking. Return whether to continue.

0nkery: Should we run also `preBuild`/`postBuild` scripts during `check`?

gallais:
After double-checking the point of the pre/post build phases, this looks good to me!
Thanks for the PR!

PR #414, opened 2020-07-05
Add a simple parser combinator library for strings

melted:
I plan to use for making parsers for URIs and HTTP headers and such things.

It doesn't have all that many combinators yet. I'll fill it out when I get around to making the parsers, if not sooner.

### DIFF
--- a/libs/contrib/Data/String/Parser.idr
+++ b/libs/contrib/Data/String/Parser.idr
@@ -14,6 +14,13 @@ record State where
     pos : Int
     maxPos : Int
 
+Show State where
+    show s = "(" ++ show s.input ++", " ++ show s.pos ++ ", " ++ show s.maxPos ++")"
+
+dumpState : State -> State
+dumpState s = unsafePerformIO $ do printLn s
+                                   pure s
+
 data Result a = Fail Int String | OK a State
 
 
@@ -86,7 +93,7 @@ satisfy f = P $ \s => do if s.pos < s.maxPos
 public export
 string : Monad m => String -> ParseT m ()
 string str = P $ \s => do let len = strLength str
-                          if s.pos+len < s.maxPos
+                          if s.pos+len <= s.maxPos
                               then do let head = strSubstr s.pos len s.input
                                       if head == str
                                         then pure $ OK () (S s.input (s.pos + len) s.maxPos)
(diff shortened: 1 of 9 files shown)


martinbaker:
How does this relate to the parser for generating Idris2 code:
Idris2/src/Parser/
Is there a way the full parser can be made available to the library at runtime? (as it is in Idris 1).
I realise this PR is a simple parser for stings but is there any way the the same code could be used to scale up to the full Idris compiler without duplication and the users having to learn several versions?

melted:
It's not related. That parser is already present in contrib as Text.Parser (and Text.Lexer). It's unsuitable for my purposes (I don't need tokenization, I can operate on the input string and not unpack it to a list).

It's a good idea to harmonize the names of the functions if they do the same thing.

libs/contrib/Data/String/Parser.idr
```diff
@@ -0,0 +1,151 @@
...
+record State where
+    constructor S
+    input : String
+    pos : Int
+    maxPos : Int
+
+Show State where
+    show s = "(" ++ show s.input ++", " ++ show s.pos ++ ", " ++ show s.maxPos ++")"
+
+||| A debug helper to print out the input state during parsing.
```
chrrasmussen, line 22: Was this function left here intentionally? I only ask because this function breaks referential transparency and, in my opinion, such functions should be marked in a way that makes that obvious. For example by naming it `unsafeDumpState` or putting it in a module that include `Debug`, `Unsafe` (or similar) in its name.

There exist a function that could be used for this use-case, in `Debug.Trace`:
```
trace : (msg : String) -> (result : a) -> a
trace x val = unsafePerformIO (do putStrLn x; pure val)
```

There could probably also be added a function like the following (name is copied from Haskell's `Debug.Trace` module):
```
traceShowId : Show a => (result : a) -> a
traceShowId val = trace (show val) val
```

Depending on where you'd want to use `dumpState`, maybe it makes sense to have a function like:
```
dumpState : HasIO io => ParseT io ()
dumpState = P $ \s => do
  printLn s
  pure (OK () s)
```

melted: I'll remove that, it's not exported anyway.

melted:
Added the `<?>` combinator too.

melted:
Huh, an unrelated test fails because of a dropped line on macOS only? Let's assume it's a stray cosmic ray and rerun it.

PR #415, opened 2020-07-05
Javascript

rbarreiro:
javascript backend


melted:
Can you make this a draft? This requires discussions, and @edwinb need to decide if we want to put in another backend.

melted:
Related, backends don't have to live in the idris2 source tree: https://idris2.readthedocs.io/en/latest/backends/custom.html

melted:
Hey, I could make it a draft!

edwinb:
On the "should we have another backend" question, I think while we should be careful not to have to maintain too many, JS is too useful to not be included by default. (And, as a tiebreaker, Idris 1 has a JS back end by default.)

Beyond that, I'm not really well qualified to comment, so I'll see how the discussion goes!

rbarreiro:
> Related, backends don't have to live in the idris2 source tree: https://idris2.readthedocs.io/en/latest/backends/custom.html

Yes I know, I don't mind making it a separate backend. When I started that possibility was not available so I forked. And since there is a request for javascript in CONTRIBUTING.md I decided to make a merge request, but I am perfectly fine to make it separate.

edwinb:
Oh, one thing that is worth checking is what adding a new back end does to build times - code generation from the generated Chez already takes a bit too long. I don't expect it'll be too disastrous, but let's at least be a bit careful at least until we have a way of generating code faster.

libs/prelude/Prelude.idr
```diff
@@ -1509,9 +1510,32 @@ prim__putChar : Char -> (1 x : %World) -> IORes ()
 %foreign "C:getchar,libc 6"
 %extern prim__getChar : (1 x : %World) -> IORes Char
 
+read_line_js : String
```
melted, line 1513: move this to a separate module and import it. 

why are there two read_line_js?

libs/base/System/File.idr
```diff
@@ -17,26 +17,71 @@ support fn = "C:" ++ fn ++ ", libidris2_support"
 libc : String -> String
 libc fn = "C:" ++ fn ++ ", libc 6"
 
+js_try_catch_lasterr_Int : String -> String
+js_try_catch_lasterr_Int x = "{try{" ++ x ++ ";return 0n}catch(e){process.__lasterr = e; return 1n}}"
```
melted, line 21: move all these functions to a separate module.

use camelcase names

libs/base/System/File.idr
```diff
@@ -17,26 +17,71 @@ support fn = "C:" ++ fn ++ ", libidris2_support"
...
 prim_error : FilePtr -> PrimIO Int
+
 %foreign support "idris2_fileErrno"
+         "node:lambda:()=>-BigInt(process.__lasterr.errno)"
 prim_fileErrno : PrimIO Int
 
+
+read_line_js : String
+read_line_js =
+   "(file_ptr =>{
```
melted, line 48: move to a separate module

camelcase names

melted: An alternative is to put functions such as these in a file in `support/js`.

edwinb: It'd probably be neater to put them in `support/js`, especially for consistency with the Scheme back ends

edwinb:
I've tried this out - build time doesn't increase significantly, so that's no problem, and the tests pass. I just have a couple of small requests:

* I had to update my node because the version I had on my system (Ubuntu LTS) was too old. Please could you add a note to the README saying what the minimum requirement is?
* For consistency, please could the generated executable be put under `build/exec` - at first I couldn't find them because that's where I looked!

Thanks for the contribution, anyway! I think this is something which will interested quite a few people (me included).

chrrasmussen:
Regarding the output location that @edwinb mentioned: Changing `let out = outfile` to `let out = outputDir </> outfile` should fix this (in `src/Compiler/ES/Javascript.idr` and `src/Compiler/ES/Node.idr`).

By default this will output the generated file(s) to `build/exec`, but by specifying `--output-dir` it is possible to change the output directory: `idris2 --cg node --output-dir . -o foo.js Main.idr` will generate `foo.js` in the current directory.

libs/base/System/File.idr
rbarreiro: > An alternative is to put functions such as these in a file in `support/js`.

I can go that way, I didn't went that way to avoid unnecessary definitions on the generated js. Should I do it?

melted: Yes, it will scale better than leaving javascript code scattered around the libs as string literals. If you are concerned about code bloat, you could put the functions in separate files.

rbarreiro:
I think I have addressed all issues, you may check again.

melted:
I wondered if the Imperative was more generally useful and could be moved out. But that can easily be done later.

src/Compiler/ES/Javascript.idr
```diff
@@ -0,0 +1,62 @@
...
+           , " </body>\n"
+           , "</html>"
+           ]
+
+||| Javascript implementation of the `compileExpr` interface.
+compileExpr : Ref Ctxt Defs -> (tmpDir : String) -> (outputDir : String) ->
+              ClosedTerm -> (outfile : String) -> Core (Maybe String)
+compileExpr c tmpDir outputDir tm outfile
+    = do es <- compileToJS c tm
+         let res = if toLower (takeLast 5 outfile) == ".html" then htmlHeader ++ es ++ htmlFooter else es
```
andylokandy, line 48:
```suggestion
         let res = if toLower (extension outfile) == "html" then htmlHeader ++ es ++ htmlFooter else es
```

edwinb:
If nobody else has any comments, I think this is good to go now (and, it certainly doesn't do any harm!).

Ideally we'll also need documentation (like the other backends: https://idris2.readthedocs.io/en/latest/backends/index.html) but this can be added in later PRs.

Thanks again for the contribution!

PR #417, opened 2020-07-05
Merge sort implementation for Vect

birchmd:
It seemed strange to me that [`Vect` includes a `mergeBy` function](https://github.com/idris-lang/Idris2/blob/master/libs/base/Data/Vect.idr#L163), but does not use it in a merge sort implementation. This implementation is inspired by [the corresponding one for `List`](https://github.com/idris-lang/Idris2/blob/master/libs/base/Data/List.idr#L516).


libs/contrib/Data/Vect/Sort.idr
```diff
@@ -0,0 +1,39 @@
...
+                mergeBy cmp
+                    (sortBy cmp (assert_smaller xs left))
+                    (sortBy cmp (assert_smaller xs right))
+        where
+            pqRewrite : (p + q = n) -> Vect n a -> Vect (p + q) a
+            pqRewrite equalityProof xs = rewrite equalityProof in xs
+
+            -- Divides n by 2 (rounding down), returning `p = n / 2` and
+            -- `q = n - (n / 2)`, along with a proof that `p + q = n`.
+            div2 : (n : Nat) -> (p ** (q ** (p + q = n)))
```
gallais, line 24: This is worth lifting to the toplevel.

birchmd: Sure, I can certainly do that. Do you think it should go in this module, or one related to `Nat`?

gallais: I would put it in the `Nat` one.

nickdrozd: The comment there sounds like it is making guarantees about the numbers returned. Could those guarantees be incorporated into the signature? Otherwise this claim admits of a trivial proof:

```idris
div2 : (n : Nat) -> (p ** (q ** (p + q = n)))
div2 n = (Z ** (n ** Refl))
```

birchmd: Agreed @nickdrozd I want to include something along the lines of `(p = q) or (p = S q)` (depending on if `n` is even or odd of course), in addition to the `p + q = n` constraint. I wasn't sure how to cleanly encode it into the signature though. I was wondering if I should include an additional lemma, aside from the main definition which proves this fact.

gallais: @birchmd Sounds like you want the [half](https://github.com/idris-lang/Idris2/blob/master/libs/base/Data/Nat/Views.idr#L19) function.

birchmd: @gallais Yes, that is exactly what I wanted! Thanks for the tip, I didn't know that was in the base library. Please see the updated implementation using `half`. https://github.com/idris-lang/Idris2/pull/417/commits/d5eb34489286e99657a1fbafe4e79fa2b5e32f97

gallais:
Thanks!

PR #418, opened 2020-07-06
Tiny doc fix about combination of `import public` and `import .. as`

buzden:
It seems that `public` is forgotten between `import` and `as` ;-) I checked!


gallais:
Thanks!

PR #419, opened 2020-07-06
Fix forward declaration of idris_closeDir

markuspf:
More of a hotfix. this was clearly not used anywhere before, so one'd either have to think about tests to make sure this kind of thing doesn't happen again, or if it's not needed remove it.


melted:
Thanks!

There are lots of unused stuff in the support lib, so it's not strange. I would prefer cleaning it up to keeping it around in case it becomes useful again.

markuspf:
I noticed it while working on a backend; I will certainly go through the support lib and keep chipping at it for some time!

PR #424, opened 2020-07-07
Remove a superflousy declared argument to idris2_openFile

markuspf:
This didn't cause a problem before as it was likely just ignored by the C
function. According to Edwin the extra argument is a leftover from when this
was a pure scheme call.


melted:
Thanks for the PR!

PR #425, opened 2020-07-07
[ fix #423 ] --repl should load the main file 

### LINKED ISSUE #423: --repl a pre-built package doesn't load the package
# Steps to Reproduce
```idris
$ cat foo.ipkg 
package foo

modules = Foo

main = Foo

$ cat Foo.idr 
module Foo

main : IO ()
main = putStrLn "Foo"

$ idris2 --clean foo.ipkg 
Removed: build/ttc/Foo.ttc
Removed: build/ttc/Foo.ttm
Removed: build/ttc
$ idris2 --repl foo.ipkg
1/1: Building Foo (Foo.idr)
Foo> :exec main
Foo
Foo> :quit
Bye for now!
$ idris2 --repl foo.ipkg
Main> :exec main
(interactive):1:1--1:1:Undefined name unsafePerformIO at:
1	:exec main
	

Main>
```
# Expected Behavior
The two calls to `idris2 --repl foo.ipkg` should behave the same, with the second one not building the package again.

# Observed Behavior
The second call, which indeed doesn't build the package, doesn't load the module.

gallais:


gallais:
Is the racket backend not implementing `putStrLn`?
The `Hi!` message simply does not show up in the output!

I'll investigate tomorrow unless somehow solves this in the meantime.

PR #427, opened 2020-07-08
Minor simplifications

nickdrozd:
No changes to behavior here, just a reduction in code.

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -6,6 +6,8 @@ import public Data.Fin
 
 import Decidable.Equality
 
+%default total
+
 public export
 data Vect : (len : Nat) -> (elem : Type) -> Type where
   ||| Empty vector
@@ -20,13 +22,13 @@ data Vect : (len : Nat) -> (elem : Type) -> Type where
 public export
 length : (xs : Vect len elem) -> Nat
 length [] = 0
-length (x::xs) = 1 + length xs
+length (_::xs) = 1 + length xs
 
 ||| Show that the length function on vectors in fact calculates the length
 export
-lengthCorrect : (0 len : Nat) -> (xs : Vect len elem) -> length xs = len
-lengthCorrect Z     []        = Refl
-lengthCorrect (S n) (x :: xs) = rewrite lengthCorrect n xs in Refl
+lengthCorrect : (xs : Vect len elem) -> length xs = len
+lengthCorrect []        = Refl
+lengthCorrect (_ :: xs) = rewrite lengthCorrect xs in Refl
 
 ||| If two vectors are equal, their heads and tails are equal
 export
@@ -44,7 +46,7 @@ vectInjective Refl = (Refl, Refl)
 ||| ```
 public export
 tail : Vect (S len) elem -> Vect len elem
-tail (x::xs) = xs
+tail (_::xs) = xs
 
 ||| Only the first element of the vector
 |||
@@ -53,7 +55,7 @@ tail (x::xs) = xs
 ||| ```
 public export
 head : Vect (S len) elem -> elem
-head (x::xs) = x
+head (x::_) = x
 
 ||| The last element of the vector
 |||
@@ -62,8 +64,8 @@ head (x::xs) = x
 ||| ```
 public export
 last : Vect (S len) elem -> elem
-last (x::[])    = x
-last (x::y::ys) = last $ y::ys
+last [x]        = x
+last (_::y::ys) = last $ y::ys
 
 ||| All but the last element of the vector
 |||
@@ -72,7 +74,7 @@ last (x::y::ys) = last $ y::ys
 ||| ```
 public export
 init : Vect (S len) elem -> Vect len elem
-init (x::[])    = []
+init [_]        = []
 init (x::y::ys) = x :: init (y::ys)
 
 ||| Extract the first `n` elements of a Vect.
@@ -90,8 +92,8 @@ take (S k) (x :: xs) = x :: take k xs
 ||| ```
 public export
 index : Fin len -> Vect len elem -> elem
-index FZ     (x::xs) = x
-index (FS k) (x::xs) = index k xs
+index FZ     (x::_)  = x
+index (FS k) (_::xs) = index k xs
 
 ||| Insert an element at a particular index
 |||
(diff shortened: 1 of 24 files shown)


src/TTImp/Elab/Ambiguity.idr
```diff
@@ -74,12 +74,8 @@ expandAmbigName mode nest env orig args (IVar fc x) exp
...
         = do let NS ns x = fullname def
                  | _ => pure True
-             if !(isVisible ns)
-                then if visibleInAny (!getNS :: !getNestedNS) (NS ns x)
-                                     (visibility def)
-                        then pure True
-                        else pure False
-                else pure False
+             pure $ !(isVisible ns) &&
+                    visibleInAny (!getNS :: !getNestedNS) (NS ns x) (visibility def)
```
gallais, line 78: I think that this one does not actually have the same meaning
as the `!`-prefixed expressions are effectful computations.
This gets desugared to:

```idris
a <- isVisible ns
b <- getNS
c <- getNestedNS
pure (a && visibleAny (b :: c) (NS n x) (visibility def))
```

instead of

```idris
a <- isVisible ns
if a then do
  b <- getNS
  c <- getNestedNS
  pure (visibleAny (b :: c) (NS n x) (visibility def))
else pure False
```

nickdrozd: Here is my reasoning for the change. To start off, we have

```idris
             if !(isVisible ns)
                then if visibleInAny (!getNS :: !getNestedNS) (NS ns x) (visibility def)
                        then pure True
                        else pure False
                else pure False
```

There are three exit points from this `if` expression, all of which are `pure`. Hoisting that out:

```idris
             pure $ if !(isVisible ns)
                       then if visibleInAny (!getNS :: !getNestedNS) (NS ns x) (visibility def)
                            then True
                            else False
                       else False
```

`visibleInAny` just returns `Bool`, so that whole interior `if` expression can be reduced:

```idris
             pure $ if !(isVisible ns)
                       then visibleInAny (!getNS :: !getNestedNS) (NS ns x) (visibility def)
                       else False
```

At this point the `if` expression should desugar to:

```idris
a <- isVisible ns
if a
	then visibleInAny (!getNS :: !getNestedNS) (NS ns x) (visibility def)
	else False
```

This is just to say that `a` and the `visibleInAny` call must both be true, so join them as a conjunction:

```idris
a <- isVisible ns
a && visibleInAny (!getNS :: !getNestedNS) (NS ns x) (visibility def)
```

Which in turn desugars to:

```idris
a <- isVisible ns
a && do b <- getNS
        c <- getNestedNS
		visibleInAny (b :: c) (NS ns x) (visibility def)
```

`&&` is lazy, so those `b` and `c` assignments won't get executed unless `a` has come back true.

Or does `!` have some kind of precedence that overrules `&&`? To me that would be strange and unexpected behavior. But then, I don't understand Idris side effects very well to begin with. There is probably something important that I'm missing.

gallais: `&&` is pure so its second argument cannot be a do block.
That's why it's all hoisted before the `pure`.

libs/base/Data/Vect.idr
```diff
@@ -18,15 +20,14 @@ data Vect : (len : Nat) -> (elem : Type) -> Type where
 %name Vect xs,ys,zs,ws
 
 public export
-length : (xs : Vect len elem) -> Nat
-length [] = 0
-length (x::xs) = 1 + length xs
+length : {1 len : Nat} -> (0 xs : Vect len elem) -> Nat
```
clayrat, line 23: This looks like a breaking change since `len` is made non-erasable. Perhaps we should have two separate `length` functions.

src/TTImp/Elab/Ambiguity.idr
edwinb: The ! gets lifted no higher than the beginning of the block (which would be the start of the `then` and `else` branches here). iirc the reason for the `if` rather than `&&` is because the ! will get executed even if it's not needed because it's lifted outside the `&&`. If you see something of the form `if x then True else False` that's typically the most likely explanation.

Although, in this case, the operations are just lookups, so might even be cheaper than the `Delay`. So clarity is probably more important.

libs/base/Data/Vect.idr
edwinb: Yes, `length` shouldn't be written like this because it will leak the length everywhere. I don't see a need for a second `length` function where you already have the length available. Perhaps if someone comes up with one, we'll add it.

src/TTImp/Elab/Ambiguity.idr
nickdrozd: For consistency with other similar uses, I'll change just the interior `if` block. As a user this behavior is surprising and confusing to me. I would have expected at least for the compiler to say something about it.

libs/base/Data/Vect.idr
nickdrozd: What's the use case for a `length` function for `Vect`? A `Vect` is already wearing a name tag that says "Hello, my length is ___", so why would its length need to be recursively calculated?

nickdrozd: I changed it back, in any case.

gallais: > so why would its length need to be recursively calculated?

The length is known at *compile* time but it is not stored at *run* time.
If you need to access the length at run time then you need to compute it
(or pass around a pair `(n ** Vec n a)`.

melted:
There's lot here, and we've had a lot of bootstrap breakage recently. I'm running the Idris 1->Idris2-boot->Idris2 bootstrap on this. Please don't merge until it's finished.

src/Idris/Package.idr
```diff
@@ -558,11 +558,10 @@ processPackageOpts opts
     = do (MkPFR (Just (cmd, f)) opts' err) <- filterPackageOpts (MkPFR Nothing Nil False) opts
              | (MkPFR Nothing opts _) => pure False
 
-         if err
```
melted, line 561: This gives an error when bootstrapping from Idris2-boot.

nickdrozd: I reverted the change.

nickdrozd:
I added one more thing, a change to the `length` proof in `TailRec`. @melted if there are no further install issues, this should be ready to rock.

melted:
It's all green. So let's merge!

PR #431, opened 2020-07-08
Make sure names derived from case and with blocks get unique names during LambdaLifting

cypheon:
PR #416 removed the unique index from `show` for `CaseBlock` and `WithBlock`, which is great for tests. But as a result, LambdaLift generates non-unique names, as it uses `show Name` by default to generate names for lifted lambdas. This commit explicitly includes the Case/With block index in the generated name to guarantee that it's unique.


melted:
Thanks!

Issue #116, opened 2020-05-23
Test runner fails to compile when using a let instead of the expression


melted:
Using a let instead of the expression itself makes the test runner `tests/Main.idr` fail in type checking.

# Steps to Reproduce

Insert `result` instead of `normalize out == normalize exp` on  https://github.com/idris-lang/Idris2/blob/61ec7757ed485d55058a3991d8072f4f1db8a310/tests/Main.idr#L221 in the test runner.

# Expected Behavior

No change in behavior

# Observed Behavior

It fails to compile with 
```
Main.idr:222:26--223:21:While processing right hand side of runTest at Main.idr:166:1--231:1:
While processing right hand side of runTest,runTest' at Main.idr:206:9--231:1:
Can't solve constraint between:
        ()
and
        ?a [locals in scope: testPath, opts, {_:1513}, {_:1522}, out, {_cn:1}, exp]
```

edwinb:
Here's a small program that has the same problem:

    f : Nat -> IO ()
    f x = do case x of
                Z => putStrLn "zero"
                S k => putStrLn "nonzero"
             putStrLn "Done"

(The rest of this is mostly rubberducking, but also by way of explanation...)

The error here comes from trying to infer a type for the `case` block. Since it doesn't know the type in advance, it invents a metavariable in the scope of the local context, so that's `?a x`. Checking the case branches leads to the following unification problems:

    ?a Z =?= ()
    ?a (S k) =?= ()

So, because `case` is dependent, the `x` gets substituted into the unknown type, and neither of the above problems can be solved by pattern unification, because we can only solve metavariables applied to a spine of unique variables.

The only plausible solution I can think of right now, which is really a hack, is to say that a `case` will be considered non-dependent if the type can't be inferred. I don't like this. In general, `case` does have to be dependent to do a lot of interesting things.

But, this sort of program really ought to type check! I will dig deeper into the unification literature...

gallais:
We could have `case_return_of_` for the dependent one and `case_of_` for the
non dependent one. Users can write `case e return _ of alts` if they believe
Idris should be able to figure out the predicate corresponding to the return
type.

edwinb:
Hmm, yes, I think the most common case is going to be non-dependent - at least for me, most uses of `case` are destructuring lets or binds, and if I have a dependent case I really mean it. So that seems like it could be a neat solution if nothing else is plausible.

edwinb:
[...]

PR #437, opened 2020-07-08
Fix bootstrap

melted:
This was broken in #425, because Idris2-boot hasn't implemented the
fix for #116


melted:
It bootstrapped with the fix in my repo.

PR #439, opened 2020-07-09
Add `Ord` implementation for `Either`

matthewess:
Noticed this seemed to be missing. If there's a reason this wasn't included, feel free to ignore.


matthewess:
Oh my, I seem to have neglected some relevant tests?

melted:
Seems like some internal value changed and it shows in a reflection test. Just update the `expected` file in `reflection003`.

gallais:
Thanks!

PR #444, opened 2020-07-10
Allow `default` to be identifier

andylokandy:
Make `default` to be a contextual keyword, which allows `default` to be an identifier when it's not in the position of default implicit argument and default record field.

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -536,7 +536,7 @@ mutual
   defaultImplicitPi fname indents
       = do start <- location
            symbol "{"
-           keyword "default"
+           exactIdent "default"
            commit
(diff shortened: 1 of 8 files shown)


gallais:
I am not a huge fan of this change.

Why `default` but not `public` or `export` or any of the other keywords? The fact that
this words' meaning is now context-sensitive is also not super exciting. :confused:

andylokandy:
Well, it was one of the help wanted in `CONTRIBUTING.md`, I don't know why it's gone. Is our preference changed?

And I think it is a breaking change not to support it: https://github.com/idris-lang/Idris-dev/blob/master/libs/effects/Effect/Default.idr

gallais:
I'm going to stick with keeping keywords reserved.

PR #448, opened 2020-07-10
Make the idris2 executable depend on the idris2 lib

melted:
Let's not keep two identical lists.


idris2.ipkg
```diff
@@ -1,182 +1,6 @@
-package idris2
+package idris2app
```
gallais, line 1: I would probably keep the package name defined by `idris2.ipkg` to be `idris2`
and make `idris2api.ipkg` define `idris2api` instead.

melted: This package name will never be used as a dependency or really anywhere, the lib will, so it gets the `idris2` name.

buzden: Maybe, then, rename `idris2api.ipkg` to `idris2lib.ipkg` and `idris.ipkg` to `idris2app.ipkg`?

melted: That's probably better, but I want to keep these changes unobtrusive.

PR #449, opened 2020-07-10
Include a (non-linear) definition for transpose that uses zipWith

ohad:
The non-linear version might be easier to reason about 'in practice', since client code may already have proved
stuff about zipWith that they want to reuse

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -251,6 +251,26 @@ zipWith : (f : a -> b -> c) -> (xs : Vect n a) -> (ys : Vect n b) -> Vect n c
 zipWith f []      []      = []
 zipWith f (x::xs) (y::ys) = f x y :: zipWith f xs ys
 
+||| Linear version
+public export
+lzipWith : (f : (1 _ : a) -> (1 _ : b) -> c)
+      -> (1 _ : Vect n a)
+      -> (1 _ : Vect n b)
+      -> Vect n c
+lzipWith f [] [] = []
+lzipWith f (x::xs) (y::ys) = f x y :: lzipWith f xs ys
+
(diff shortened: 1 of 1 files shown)


libs/base/Data/Vect.idr
```diff
@@ -782,6 +782,25 @@ range : {len : Nat} -> Vect len (Fin len)
 range {len=Z}   = []
 range {len=S _} = FZ :: map FS range
 
+--------------------------------------------------------------------------------
+-- Matrix transposition
+--------------------------------------------------------------------------------
+||| Specialises zipWith (::) to preserve linearity
+public export
+zipWithConsLinear : (1 _ : Vect a elem)
```
gallais, line 790: Why not `lzipWith : (1 a -> 1 b -> c) -> 1 Vec n a -> 1 Vec n b -> Vec n c`?

ohad: Since the `transpose` is `public export` I tried to keep what was there unchanged. Happy to go with the non-specialised version.

The more important question is whether `transpose` should really be `public export` if `transpose'` is `public export` and `transposeSpec` is `export`.

gallais: I think we should opt for the more principled solution if we are going to
export the function. It may break things right now but we won't have to live
with the ad-hoc functions for the rest of our lives.

ohad: What is the more principled solution?

gallais: The generic `lzipWith` rather than a specialised version for `(::)`.

ohad: Ah, I see. Done.

ohad: Thanks!

PR #450, opened 2020-07-10
[ fix #346 ] Pull List.length into prelude

### LINKED ISSUE #346: Pull List length into prelude

gallais:


gallais:
Ah! It looks like I need to learn how to update the bootstrap process.

PR #451, opened 2020-07-10
Add some implementation notes to the docs

edwinb:
These were lying around in Idris2-boot, but it's better to put them somewhere people might find them


edwinb:
By the way - if anyone finds anything interesting or that they think would be useful to explain, please feel free to add a new section here (probably as a new file, rather than part of the overview) and I will happily review and edit. I really need more people to know more about aspects of the core!

PR #452, opened 2020-07-10
Some things to add.

jfdm:
Some minor things to add that I have seen so far this past week.


libs/base/Data/Vect.idr
```diff
@@ -75,6 +75,14 @@ init : Vect (S len) elem -> Vect len elem
 init (x::[])    = []
 init (x::y::ys) = x :: init (y::ys)
 
+||| Extract the first `n` elements of a Vect.
+public export
+take : (n : Nat)
```
gallais, line 80: Note that you can be linear in `n`.

jfdm:
FIxed.

PR #463, opened 2020-07-14
Remove dependency to Vector length in deleteAt

berewt:
`deleteAt` required access to the vector length, which can be problematic when you want to ensure that the length is erased at compile time.


gallais:
Thanks!

PR #465, opened 2020-07-14
Wrap Javascript case clauses in brackets to prevent conflicting declarations

memoryruins:
Currently if one attempts to use (for example) `Language.JSON.parse`  from `contrib` while using the Javascript backend, an error is emitted in Firefox:

```
Uncaught SyntaxError: redeclaration of const p$27
note: Previously declared at line 2719, column 9
```

This is due to different case clauses sharing the same block scope by default and multiple let statements are being interpreted as duplicate declarations of the same variable name. See "Block-scope variables within switch statements" in <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch>

The solution is to wrap each case clause in brackets, giving each clause its own block scope.


gallais:
Thanks for adding the test! :heart:

PR #466, opened 2020-07-14
Move all CG code out of the main source tree

ska80:


melted:
I don't really see the motivation for this. Could you expand on it?

I guess @edwinb gets to decide if this is a suitable organization. The major problem that I see is that pulling out the codegens means that they can't be used by apps built on the API. Like, an enhanced REPL that uses the API can't do codegen.

ska80:
> I don't really see the motivation for this. Could you expand on it?

The motivation was to decouple CG code from the core.

> For example, an enhanced REPL that uses the API can't do codegen as there are no codegens in it.

Unfortunately I didn’t consider this use case.

I could move two essential Scheme cgs (Chez and Racket) back into the main source tree.  Could this solve this problem?  If this approach is still completely wrong I can close this PR :)

edwinb:
I think at least Chez, Racket and JS should be there. I'm not so sure about Gambit, but only because I don't know if anyone is willing to maintain it (and it's not currently complete enough to build Idris itself). But in general, I don't know what problem this aims to solve, especially since to some extent the code generators and the core are closely related.

PR #478, opened 2020-07-18
Install the api by default

melted:
It will simplify testing and should be expected behavior anyway.


melted:
Looks like it needs bit of actual thinking.

PR #481, opened 2020-07-21
Use the same naming convention for foreign primitives

ska80:


gallais:
Should these changes in base be replicated in https://github.com/edwinb/Idris2-boot?

melted:
No need, there are no calls directly to them from the compiler code (if there was there would have been changes in `src/` in this PR).

ska80:
> Should these changes in base be replicated in https://github.com/edwinb/Idris2-boot?

All tests have passed when I pushed to my repo including bootstrapping with Idris2-boot

PR #485, opened 2020-07-23
Add `--directive` cli option

melted:
It sends a codegen directive to the current codegen. It's useful
for sending options you don't want to hard code in the source. Examples
include enabling profiling or setting a target.


buzden:
Maybe it is bikeshedding, but isn't `--directive` too general name for this? I think, it's hard to tell that this option regards to codegens. Maybe, something like `--cg-directive` is better?

PR #488, opened 2020-07-24
Inconsistent std IO

Russoul:
`getChar, putChar` reads/writes exactly one byte from/to stdio
`putChar` writes exactly one byte
`putCharLn` writes the whole codepoint + \n

I think that should be pointed out more vigorously as their behaviour currently is confusing

### DIFF
--- a/libs/prelude/Prelude/IO.idr
+++ b/libs/prelude/Prelude/IO.idr
@@ -88,17 +88,17 @@ export
 getLine : HasIO io => io String
 getLine = primIO prim__getStr
 
-||| Write a single character to stdout.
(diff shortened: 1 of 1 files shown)


Russoul:
image

Russoul:
related issue #489

gallais:
I'd be happy to merge the documentation changes if they were modified so that
the text is grammatically correct. E.g. instead of inserting "one byte!" in the
middle of the description, add an extra line:

```idris
||| /!\ We assume the character is one byte only!
```

Russoul:
Sure, hope that's better

PR #491, opened 2020-07-24
Add list lemmas

clayrat:
Some generic list lemmas that were useful in formalizing some explicit substitution & modal calcull.

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -624,10 +624,10 @@ revAppend (v :: vs) ns
               Refl
 
 export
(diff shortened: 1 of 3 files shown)


libs/base/Data/List.idr
```diff
@@ -620,3 +622,12 @@ revAppend (v :: vs) ns
           rewrite sym (revAppend vs ns) in
             rewrite appendAssociative (reverse ns) (reverse vs) [v] in
               Refl
+
+export
+dropSum : (n, m : Nat) -> (l : List t) -> drop n (drop m l) = drop (n+m) l
```
gallais, line 627: I would call this `dropFusion` just like we can have `mapFusion`
(I looked for such a lemma but could not find it, strangely).

PR #492, opened 2020-07-25
Added explicit case to headEq to allow matching on Pi Types.

donovancrichton:
Add a missing explicit case to Core/CaseBuilder.idr: `headEq` to allow for matching on pi-types as per Edwin.
Code that fails before the fix:
https://gist.github.com/donovancrichton/2838abd65c12dbf0dc01806ac8eafb0f
Has been tested to work after the fix.
See: https://functionalprogramming.slack.com/archives/C043A0KTY/p1595633177089900


melted:
Thanks!

PR #501, opened 2020-07-27
Take into account syntactic sub-singletons when dotting in patterns

### LINKED ISSUE #499: Pattern-matching on irrelevant, single-constructor datatypes
# Steps to Reproduce
```idris
import Data.List

infix 1 \\

record DIPair (a : Type) (b : (x : a) -> Type) where
  constructor (\\) 
  fst : a
  0 snd : b fst

MyList : Type
MyList = DIPair (List Bool) (\bs => NonEmpty bs)

bar : MyList
bar = [True, False] \\ IsNonEmpty


{- doesn't type-check -}
foo1 : MyList -> Bool
foo1 (b :: bs \\ IsNonEmpty) = b

```
# Expected Behavior
Type-check

# Observed Behavior
```
 |   While processing left hand side of foo1 at Foo.idr:19:1--21:1:
 |   Can't match on IsNonEmpty (Erased argument) at
 |   19	foo1 (b :: bs \\ IsNonEmpty) = b
 |   	                 ^^^^^^^^^^
```

I'd expect this to type-check because this type-checks fine:
```idris
foo2 : (bs : List Bool) -> (0 _ : NonEmpty bs) -> Bool
foo2 (b :: bs) IsNonEmpty = b

aux : MyList -> Bool
aux ml = foo2 (fst ml) (snd ml)
```

ohad:
Fixes #499 

When an erased pattern argument in a non-erased context is of a
datatype that has 0 or 1 constructors, there is no need to dot it
(i.e., require that the canonical term is inferred by other
dependencies): it has to be the unique constructor that appears in the
source.

### DIFF
--- a/src/TTImp/Elab/App.idr
+++ b/src/TTImp/Elab/App.idr
@@ -319,10 +319,11 @@ mutual
               Maybe Name -> Nat -> ElabMode -> RigCount -> RawImp -> Core RawImp
   dotErased argty mn argpos (InLHS lrig ) rig tm
       = if not (isErased lrig) && isErased rig
-           then
-              -- if the argument type aty has a single constructor, there's no need 
-              -- to dot it
-              if length (!(possibleConstructors argty)) <= 1
+          then do
+            -- if the argument type aty has a single constructor, there's no need 
+            -- to dot it
+            mconsCount <- countConstructors argty
+            if mconsCount == Just 1 || mconsCount == Just 0
               then pure tm
               else 
                 -- if argpos is an erased position of 'n', leave it, otherwise dot if
@@ -333,21 +334,20 @@ mutual
                    if argpos `elem` safeErase gdef
                       then pure tm
                       else pure $ dotTerm tm
-           else pure tm
+          else pure tm
     where
-      
-      ||| Find the constructors of a normal form that is a fully
-      ||| applied type constructor
-      possibleConstructors : NF vars -> Core (List Name)
-      possibleConstructors (NApp _ (NRef (TyCon _ n) tycName) args) = 
+      ||| Count the constructors of a fully applied concrete datatype
+      countConstructors : NF vars -> Core (Maybe Nat)
+      countConstructors (NTCon _ tycName _ n args) = 
         if length args == n
         then do defs <- get Ctxt
                 Just gdef <- lookupCtxtExact tycName (gamma defs)
-                | Nothing => pure []
-                let (TCon _ _ _ _ _ _ datacons _) = gdef.definition                                                   | _ => pure []
-                pure datacons
-        else pure []
-      possibleConstructors _ = pure []
+                | Nothing => pure Nothing
+                let (TCon _ _ _ _ _ _ datacons _) = gdef.definition
+                | _ => pure Nothing
+                pure (Just (length datacons))
+        else pure Nothing
+      countConstructors _ = pure Nothing
     
       dotTerm : RawImp -> RawImp
       dotTerm tm
--- a/src/TTImp/Elab/App.idr
+++ b/src/TTImp/Elab/App.idr
@@ -315,11 +315,16 @@ mutual
               else pure ()
   checkPatTyValid fc defs env _ _ _ = pure ()
 
-  dotErased : {auto c : Ref Ctxt Defs} ->
+  dotErased : {auto c : Ref Ctxt Defs} -> (argty : NF vars) ->
               Maybe Name -> Nat -> ElabMode -> RigCount -> RawImp -> Core RawImp
-  dotErased mn argpos (InLHS lrig ) rig tm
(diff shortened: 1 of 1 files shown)


src/TTImp/Elab/App.idr
```diff
@@ -330,6 +335,20 @@ mutual
                       else pure $ dotTerm tm
            else pure tm
     where
+      
+      ||| Find the constructors of a normal form that is a fully
+      ||| applied type constructor
+      possibleConstructors : NF vars -> Core (List Name)
+      possibleConstructors (NApp _ (NRef (TyCon _ n) tycName) args) = 
+        if length args == n
```
edwinb, line 343: This is one reason the tests are failing - `NApp` is for stuck function applications, whereas type constructors are already in normal form, so not stuck. Instead, you need `NTCon`

src/TTImp/Elab/App.idr
```diff
@@ -330,6 +335,20 @@ mutual
...
+        then do defs <- get Ctxt
+                Just gdef <- lookupCtxtExact tycName (gamma defs)
+                | Nothing => pure []
+                let (TCon _ _ _ _ _ _ datacons _) = gdef.definition                                                   | _ => pure []
+                pure datacons
+        else pure []
+      possibleConstructors _ = pure []
```
edwinb, line 350: ...and this is the other reason (possibly also the line above). If you have a type which is anything other than a `NTCon`, you can't assume there's only one constructor, because it might be matching on a `Type` or some other primitive, e.g. `Int`.

I think otherwise this is a good way to resolve this issue, just that it needs to be careful to only allow things through which are concretely in `NTCon` form.

src/TTImp/Elab/App.idr
```diff
@@ -330,6 +335,20 @@ mutual
...
+      
+      ||| Find the constructors of a normal form that is a fully
+      ||| applied type constructor
+      possibleConstructors : NF vars -> Core (List Name)
+      possibleConstructors (NApp _ (NRef (TyCon _ n) tycName) args) = 
+        if length args == n
+        then do defs <- get Ctxt
+                Just gdef <- lookupCtxtExact tycName (gamma defs)
+                | Nothing => pure []
+                let (TCon _ _ _ _ _ _ datacons _) = gdef.definition                                                   | _ => pure []
```
edwinb, line 347: (Not related to the test failure and a bit nitpicky, but I was initially confused here by only having one case for `TCon _ ...` since the alternative was way off to the right. I usually prefer to put the alternatives on the next line, slightly indented.)

src/TTImp/Elab/App.idr
ohad: Thanks, sorry for misunderstanding, and it makes sense.

src/TTImp/Elab/App.idr
ohad: OK, I'll fix. There's another bug: using the empty list as a failure is exactly the wrong result to return.
I'll use a `Maybe` instead.

src/TTImp/Elab/App.idr
ohad: Ooops! I'll fix.

ohad:
I should also fix `with` similarly (or maybe the fix should be elsewhere?). 
I expect this to type-check (modifying the example from #499 ):
```idris
import Data.List

infix 1 \\

record DIPair (a : Type) (b : (x : a) -> Type) where
  constructor (\\) 
  fst : a
  0 snd : b fst

MyList : Type
MyList = DIPair (List Bool) (\bs => NonEmpty bs)

bar : MyList
bar = [True, False] \\ IsNonEmpty

{- doesn't type-check -}
foo1 : Bool
foo1 with (bar)
 foo1 | b :: bs \\ IsNonEmpty = b
```
but I get:
```
     Foo.idr line 20 col 19:
     While processing left hand side of with block in foo1 at Foo.idr:20:2--22:1:
     Can't match on IsNonEmpty (Erased argument) at
     20	 foo1 | b :: bs \\ IsNonEmpty = b
     	                   ^^^^^^^^^^
```

edwinb:
It might be trickier for `with` since I don't think it's at all clever about quantities yet. But that's a thing that should be done properly at some point, certainly. Thanks for the fix anyway! This looks fine now.

ohad:
Ooops, I didn't test the original bug after the second 'fix'.
In
```idris
infix 1 \\

record DIPair (a : Type) (b : (x : a) -> Type) where
  constructor (\\) 
  fst : a
  0 snd : b fst

MyList : Type
MyList = DIPair (List Bool) (\bs => NonEmpty bs)

foo : MyList -> Bool
foo (b :: bs \\ IsNonEmpty) = b
```
After sticking some logging commands, it looks like `countConstructors (NTCon _ tycName _ n args)` is never matched when type-checking `foo`

PR #502, opened 2020-07-29
Add another recent video talk on Idris 2

ska80:
Also, rearrange videos in reverse chronological order to show the most recent videos first


gallais:
Seems like a good idea. Thanks!

PR #513, opened 2020-08-02
export a few useful functions

shmish111:
I want to parse ipkg files so I needed to expose these functions


shmish111:
Closing to open from a different branch

PR #515, opened 2020-08-02
Add runtime-performance flag

andrevidela:
This flag tells the compiler to measure the performance of the
output program by wrapping the main function in a timing function.

It currently only supports the Chez backend.


andylokandy:
Maybe we can make it a standalone benchmark instead of a builtin functionality.

andrevidela:
I would love to know how to do that while getting around the startup time of the backend

melted:
There is a node package that gets the GC stats from V8. https://www.npmjs.com/package/gc-stats

andrevidela:
@melted Thanks for your comment, however I am not sure how this helps. Are you suggeting:

- To not implement this and rely on custom backend modification to expose runtime performance like @andylokandy suggested?
- To rely on external packages and to fail at runtime if they arent installed
- To force the installation of this package with the node backend
- To re-implement this package's functionality in the Node backend?

melted:
I don't see it as an insurmountable obstacle to require an npm package, especially for dev-time functionality, there's certainly a lot of useful stuff that are only available in that form. The code generation step could run npm down in build/exec if this functionality is turned on rather than having the developer install it manually.

andrevidela:
@melted If I understand correctly you're suggesting the third option "To force the installation of this package with the node backend". It would be nice if you could say it explicitly next time.

This indeed is not an insurmountable, but it has a lot of implication that you seem to have glossed over. I'm sure they are obvious to you, but they should also be documented. So I'm just going to write them down:

1. Requiring the NPM package means the generated bytecode will fail at runtime if the package is absent
2. Having the compiler run npm will fail when
    - NPM isn't installed
    - internet is down
    - NPM is down
3. relying on an external NPM package exposes you to the failures of this package, typically security issues or bug fixes. Therefore, there should be a policy about versioning that enforces usage of the latest and safest version of the package.

I'm fine with 1., we can just say that advanced compiler debugging feature require some work on the part of the user and we can write a simple tutorial that say "run `npm install gc-stats`".

I do think however that 2. is more trouble than necessary, as a rule of thumb, having the compiler make network requests is a dangerous road to go through.

I think we can give up on 3. with the same argument as 1. Advanced compiler features don't have to be as user-friendly and catch all possible failure modes possible. It's something we can easily fix by documenting the behavior "if you use `--runtime-performance` with `--gc node` make sure you always use the lastest version" without having the compiler make a network request to check what's the latest version.

Does that sound good?

PR #516, opened 2020-08-03
Make showParens public

mb64:
I'm currently duplicating it in my own project, and it would be nicer to just use the standard library version.


gallais:
Seems reasonable.

PR #520, opened 2020-08-04
[ refactor ] introduce List1 to remove impossible case

gallais:
This will need an Idris2-boot update.

I chose a cut-off point after the parser converting `List1` into `List`
so that this is not too big of a PR. This could however be propagated
further to guarantee that when we claim a to have a namespace qualifier,
we have a non-empty list of them...


nickdrozd:
`List1` is not a very descriptive name. How about `NonEmptyList` or something like that?

gallais:
We already have `IsNonEmpty` in `Data.List` which is a predicate
on lists. I don't find `List1` weird but then again [I'm used to it](https://github.com/agda/agda/blob/master/src/full/Agda/Utils/List1.hs).

gallais:
I've merged it in because the update to Idris2-boot to accommodate
this PR broke the build for other branches. We can still debate
changing the module name.

PR #523, opened 2020-08-04
Implement HasIO for StateT

mb64:
It requires a slightly more convoluted implementation, because of `liftIO`'s linear type.

It might make sense to export a linear `io_map : (a -> b) -> (1 _ : IO a) -> IO b` from `PrimIO`, similarly to `io_bind` -- it would make things like this a little nicer to write.


melted:
Thanks!

PR #525, opened 2020-08-06
add prim__getNullAnyPtr and prim__castPtr

thomasdziedzic:
This will enable us to get null pointers in Idris2 so that we can pass
them on to other c methods. This is useful when creating FFI bindings
for other libraries that expect optional null pointers to be passed into
the arguments.

For example, the samples/FFI-readline code uses
nullString, a similar method to get a null pointer in Idris2.

https://github.com/idris-lang/Idris2/blob/3d53c2874b4dcfe8ca4b6687b384be261faf74d4/samples/FFI-readline/readline_glue/idris_readline.c#L32-L34

Note: I tried to implement a generic getNullPtr via

```idris
%foreign "C:idris2_getNull,libidris2_support"
export
prim__getNullPtr : Ptr t
```

But I kept getting an error saying:

`Uncaught error: PrimIO.idr:95:1--97:25:Can't pass argument of type Type  to foreign function`

So I just implemented one specific for `String`s.

### DIFF
--- a/libs/prelude/PrimIO.idr
+++ b/libs/prelude/PrimIO.idr
@@ -89,6 +89,14 @@ schemeCall ret fn args = fromPrim (prim__schemeCall ret fn args)
 export
 prim__nullAnyPtr : AnyPtr -> Int
 
+%foreign "C:idris2_getNull,libidris2_support"
+export
+prim__getNullAnyPtr : AnyPtr
+
+%foreign "C:idris2_getNull,libidris2_support"
+export
+prim__getNullString : Ptr String
+
 prim__forgetPtr : Ptr t -> AnyPtr
 prim__forgetPtr = believe_me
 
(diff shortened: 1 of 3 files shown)


ohad:
I think the error on the more general `prim__getNullPtr` is because it has the type:
```idris
prim__getNullPtr : {0 t : Type} -> Ptr t
```

Maybe in the future we could relax the restriction on the FFI so that only relevant arguments need to have FFI types. Then we could define `prim__getNullPtr : Ptr t`.

melted:
You could put in the opposite of `prim__forgetPtr` to create a Ptr t from AnyPtr and just have the AnyPtr version.

ohad:
@melted I don't see what you mean, can you add type-signatures to what you propose please?

melted:
I mean just having an `prim__getNullAnyPtr` and add `prim__castPtr : AnyPtr -> Ptr t` with implementation `prim__castPtr p = believe_me`

ohad:
Right, the point is that `prim_castPtr` is not actually a foreign function, but an Idris-definable function. The `prim__` prefix is just to seem more consistent with the other functions.

PR #526, opened 2020-08-06
Add BuildExpressionParser to contrib

alexhumphreys:
Idris port of haskell's [`BuildExpressionParser`](https://hackage.haskell.org/package/parsec-3.1.13.0/docs/Text-Parsec-Expr.html). I figured with the parser combinators in contrib this might be a useful addition, but I'm not sure what the rules are for what ends up in contrib so feel free to close this PR if it's not relevant.

Also this is my first attempt at idris2, so I may have gotten a lot of stuff wrong, and I may have gone overboard with the `public export`s. The test seems reasonable but could probably be improved upon, plus I had to add a few combinators just to test (`natural`, `token`...) which maybe don't belong in the test file. Open to any feedback on how to improve this anyway.

### DIFF
--- a/libs/contrib/Data/String/BuildExpressionParser.idr
+++ b/libs/contrib/Data/String/BuildExpressionParser.idr
@@ -20,12 +20,8 @@ data Operator a = Infix (Parser (a -> a -> a)) Assoc
                 | Postfix (Parser (a -> a))
 
 public export
-LO : Type -> Type
-LO a = List (Operator a)
-
-public export
 OperatorTable : Type -> Type
-OperatorTable a = List (LO a)
+OperatorTable a = List (List (Operator a))
 
 public export
 BinaryOperator : Type -> Type
@@ -97,7 +93,7 @@ buildExpressionParser a operators simpleExpr =
     splitOp x (Prefix op) (rassoc, lassoc, nassoc, prefixx, postfix) = (rassoc, lassoc, nassoc, op :: prefixx, postfix)
     splitOp x (Postfix op) (rassoc, lassoc, nassoc, prefixx, postfix) = (rassoc, lassoc, nassoc, prefixx, op :: postfix)
 
-    makeParser : (a : Type) -> Parser a -> LO a -> Parser a
+    makeParser : (a : Type) -> Parser a -> List (Operator a) -> Parser a
     makeParser a term ops =
       let (rassoc,lassoc,nassoc
                ,prefixx,postfix) = foldr (splitOp a) ([],[],[],[],[]) ops
--- a/libs/contrib/Data/String/BuildExpressionParser.idr
+++ b/libs/contrib/Data/String/BuildExpressionParser.idr
@@ -1,0 +1,129 @@
+module Data.String.BuildExpressionParser
+
+import Control.Monad.Identity
+import Data.String.Parser
+
+fail : Monad m => String -> ParseT m a
+fail x = P $ \s => do pure $ Fail s.pos x
+
+public export
+data Assoc = AssocNone
+           | AssocLeft
+           | AssocRight
+
+public export
+data Operator a = Infix (Parser (a -> a -> a)) Assoc
+                | Prefix (Parser (a -> a))
+                | Postfix (Parser (a -> a))
+
+public export
+LO : Type -> Type
+LO a = List (Operator a)
+
+public export
+OperatorTable : Type -> Type
+OperatorTable a = List (LO a)
+
+public export
+BinaryOperator : Type -> Type
+BinaryOperator a = List (Parser (a -> a -> a))
+
+public export
+UnaryOperator : Type -> Type
+UnaryOperator a = List (Parser (a -> a))
+
+public export
+data Ops a = BinOp (BinaryOperator a) | UnOp (UnaryOperator a)
+
+public export
+ReturnType : Type -> Type
+ReturnType a = (BinaryOperator a, BinaryOperator a, BinaryOperator a, UnaryOperator a, UnaryOperator a)
+
+toParserBin : BinaryOperator a -> Parser (a -> a -> a)
+toParserBin [] = fail "couldn't create binary parser"
+toParserBin (x :: xs) = x <|> toParserBin xs
+
+toParserUn : UnaryOperator a -> Parser (a -> a)
+toParserUn [] = fail "couldn't create unary parser"
+toParserUn (x :: xs) = x <|> toParserUn xs
+
+ambigious : (assoc : String) -> (op : Parser (a -> a -> a)) -> Parser a
+ambigious assoc op = do op
+                        fail ("ambiguous use of a " ++ assoc ++ " associative operator")
+
+mutual
+  mkRassocP : (amLeft : Parser a) -> (amNon : Parser a) -> (rassocOp : Parser (a -> a -> a)) -> (termP : Parser a) -> (x : a) -> Parser a
+  mkRassocP amLeft amNon rassocOp termP x =
+   (do f <- rassocOp
+       y <- do z <- termP ; mkRassocP1 amLeft amNon rassocOp termP z
+       pure (f x y))
+   <|> (amLeft)
+   <|> (amNon)
+
+  mkRassocP1 : (amLeft : Parser a) -> (amNon : Parser a) -> (rassocOp : Parser (a -> a -> a)) -> (termP : Parser a) -> (x : a) -> Parser a
+  mkRassocP1 amLeft amNon rassocOp termP x = (mkRassocP amLeft amNon rassocOp termP x) <|> pure x
+
+mutual
+  mkLassocP : (amRight : Parser a) -> (amNon : Parser a) -> (lassocOp : Parser (a -> a -> a)) -> (termP : Parser a) -> (x : a) -> Parser a
+  mkLassocP amRight amNon lassocOp termP x =
+    (do f <- lassocOp
+        y <- termP
+        mkLassocP1 amRight amNon lassocOp termP (f x y))
+    <|> amRight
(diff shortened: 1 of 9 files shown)


libs/contrib/Data/String/BuildExpressionParser.idr
```diff
@@ -0,0 +1,132 @@
+-- Port of https://hackage.haskell.org/package/parsec-3.1.13.0/docs/Text-Parsec-Expr.html
+-- Original License	BSD-style: https://hackage.haskell.org/package/parsec-3.1.13.0/src/LICENSE
+
+module Data.String.BuildExpressionParser
```
melted, line 4: Maybe it could be called `Data.String.Parser.Expression`?

libs/contrib/Data/String/BuildExpressionParser.idr
```diff
@@ -0,0 +1,132 @@
...
+           | AssocRight
+
+public export
+data Operator a = Infix (Parser (a -> a -> a)) Assoc
+                | Prefix (Parser (a -> a))
+                | Postfix (Parser (a -> a))
+
+public export
+LO : Type -> Type
+LO a = List (Operator a)
```
melted, line 24: That's a cryptic name, it would be clearer to just use `List (Operator a)`.

tests/chez/chez028/ExpressionParser.idr
```diff
@@ -0,0 +1,79 @@
...
+natural : Parser Nat
+natural = do x <- some digit
+             pure (fromDigits x)
+
+skip : Parser a -> Parser ()
+skip = map (const ())
+
+space : Parser Char
+space = satisfy isSpace
+
```
melted, line 47: Some of the commonly useful utility functions in the test could be moved to Data.String.Parser (or maybe in a new separate module).

libs/contrib/Data/String/BuildExpressionParser.idr
alexhumphreys: Can do! Would I need to nest the file in a `parser/` parent dir too, or can I just rename it to `Expression.idr`? Still getting the hang of modules...

tests/chez/chez028/ExpressionParser.idr
alexhumphreys: Aye, was thinking that. Any preference on adding them to Data.String.Parser, or maybe adding an additional Data.String.Parser.Combinators or something?

libs/contrib/Data/String/BuildExpressionParser.idr
melted: The module path is the directory path, so you'd put it in a `Parser` dir.

tests/chez/chez028/ExpressionParser.idr
melted: Let's add them to Parser for now.

libs/contrib/Data/String/BuildExpressionParser.idr
alexhumphreys: Updated.

tests/chez/chez028/ExpressionParser.idr
alexhumphreys: Updated. I left `digit` and `natural` here as they're a little involved, but can also move those if you like.

libs/contrib/Data/String/BuildExpressionParser.idr
alexhumphreys: Updated.

libs/contrib/Data/String/Parser.idr
```diff
@@ -153,3 +153,33 @@ option def p = p <|> pure def
 export
 optional : Monad m => ParseT m a -> ParseT m (Maybe a)
 optional p = (p >>= \res => pure $ Just res) <|> pure Nothing
+
+||| Discards the result of a parser
+export
+skip : Parser a -> Parser ()
+skip = map (const ())
```
gallais, line 160: This is known as `ignore` and is available in the prelude:

```shell
Main> :t ignore
Prelude.ignore : Functor f => f a -> f ()
```

libs/contrib/Data/String/Parser/Expression.idr
```diff
@@ -0,0 +1,125 @@
...
+OperatorTable : Type -> Type
+OperatorTable a = List (List (Operator a))
+
+public export
+BinaryOperator : Type -> Type
+BinaryOperator a = List (Parser (a -> a -> a))
+
+public export
+UnaryOperator : Type -> Type
+UnaryOperator a = List (Parser (a -> a))
```
gallais, line 29: Should these lists be non-empty? Cf. `Data.List1` in `base`.

Once you use `List1` the `toParserX` functions are essentially `choice . List1.toList`.

libs/contrib/Data/String/Parser/Expression.idr
```diff
@@ -0,0 +1,125 @@
...
+  mkLassocP : (amRight : Parser a) -> (amNon : Parser a) -> (lassocOp : Parser (a -> a -> a)) -> (termP : Parser a) -> (x : a) -> Parser a
+  mkLassocP amRight amNon lassocOp termP x =
+    (do f <- lassocOp
+        y <- termP
+        mkLassocP1 amRight amNon lassocOp termP (f x y))
+    <|> amRight
+    <|> amNon
+
+  mkLassocP1 : (amRight : Parser a) -> (amNon : Parser a) -> (lassocOp : Parser (a -> a -> a)) -> (termP : Parser a) -> (x : a) -> Parser a
+  mkLassocP1 amRight amNon lassocOp termP x = mkLassocP amRight amNon lassocOp termP x <|> pure x
```
gallais, line 72: These have useful generalisations as combinators that could be moved to `Data.String.Parser`:

```
hchainl : Parser init -> Parser (init -> arg -> init) -> Parser arg -> Parser init
hchainr : Parser arg -> Parser (arg -> end -> end) -> Parser end -> Parser end
```

The `h` stands for "heterogeneous" as opposed to the usual homogeneous chain combinators.

tests/chez/chez028/ExpressionParser.idr
gallais: I think these ought to be moved too as they are so useful others will inevitably
need them. You can make `digit` a bit simpler by using a list of pairs of char +
meaning and relying on:

```idris
Data.List.lookup : Eq a => a -> List (a, b) -> Maybe b
```

alexhumphreys: Aye, updated there. Switched `digit` to be a `Parser (Fin 10)` like in Lightyear, makes a bit more sense than for it to be a `Parser Nat`.

libs/contrib/Data/String/Parser.idr
alexhumphreys: Ah cool, didn't know that. Updated. I left the `skip` parser, but maybe I should delete it and just use `ignore` in the other parsers.

libs/contrib/Data/String/Parser/Expression.idr
alexhumphreys: That's a good idea, tried it out but I ran into some trouble getting it to play nice with the `foldr splitOp` stuff on line 96. Could require a more invasive change...

libs/contrib/Data/String/Parser/Expression.idr
alexhumphreys: That definitely sounds cool, but I'm sure I have the skills to pull of a refactor like that 😅

I seem to have extra params (`Parser a`, and `a : Type`) than the type signatures you provided. Not sure I can square that difference...

Good to know that's what the `h` stands for!

libs/contrib/Data/String/Parser.idr
```diff
@@ -183,3 +185,33 @@ token s = lexeme (skip (string s)) <?> "token " ++ show s
...
+
+export
+digit : Parser (Fin 10)
+digit = do x <- satisfy isDigit
+           case lookup x digits of
+                Nothing => P $ \s => do pure $ Fail s.pos "not a digit"
+                Just y => P $ \s => Id (OK y s)
+  where
+    digits : List (Char, Fin 10)
+    digits = [ ('0', FZ)
```
gallais, line 197: We have overloading of literals for `Fin` (cf. `Data.Fin.fromInteger`) so
you should even be able to write `[ ('0', 0), ('1', 1), ... ]`

libs/contrib/Data/String/Parser/Expression.idr
alexhumphreys: Updated to use `List1`, and renamed some stuff. Not sure are the names much better but `ReturnType` was not very descriptive. 

The `toParserX` functions are now the same except for the signatures. Is there some type level magic I can use reduce that duplication? Tried a few things but no luck.

libs/contrib/Data/String/Parser.idr
alexhumphreys: Sweet! Updated.

alexhumphreys:
Well the user does provide and parsers/precedence using the type `OperatorTable a`, as you can see in the test file ExpressionParser.idr:14. I guess that could be an empty list, but then all you get a `termP` parser with a whole lot of `fail` parsers nearby.

But if the code's less readable, I'm happy to revert if that commit. And of course there could be some reasons to pass empty lists that I didn't think of. Either way I got to play about with `List1` which was cool.

alexhumphreys:
Updated to revert that commit. Also I added a test to check when there's more than 1 parser in those lists, which meant adding an `integer` parser. Let me know if I need to make further changes, or revert more stuff. 

I should probably clean up the commit history before this is merged too 😅

melted:
I'll just squash it.

PR #530, opened 2020-08-07
Fix implicit variable bindings around quoting/unquoting

### LINKED ISSUE #517: Error in quoting: Linearity checking failed (not a function type)
Here's a minimized example that gives me the error.  I'm using the current HEAD.

```idris
$ cat minimized.idr 
import Language.Reflection

pos : FC
pos = MkFC "somewhere" (0,0) (0,0)

thing : TTImp -> TTImp -> ?
thing f arg = `[ bad : ~(IApp pos f arg) ]
$ idris2 --no-banner minimized.idr 
1/1: Building minimized (minimized.idr)
minimized.idr:7:31--7:34:While processing right hand side of thing at minimized.idr:7:1--8:1:
Linearity checking failed on ?{P:pos:1969}_[] (pat arg : $resolved1492 => pat f : $resolved1492 => $resolved1611 not a function type) at:
7	thing f arg = `[ bad : ~(IApp pos f arg) ]
	                              ^^^

```

I did some snooping in the compiler, and it seems like this is the uncaught case:
<details>
<summary>git diff</summary>
<p>

```diff
diff --git a/src/Core/LinearCheck.idr b/src/Core/LinearCheck.idr
index 29e05c5..a0cef23 100644
--- a/src/Core/LinearCheck.idr
+++ b/src/Core/LinearCheck.idr
@@ -338,6 +338,7 @@ mutual
                       pure (App fc f' aerased,
                             glueBack defs env sc',
                             fused ++ aused)
+                NBind _ _ (PVar _ _ _) _ => throw $ GenericMsg fc "←— here is the error!"
                 NApp _ (NRef _ n) _ =>
                       do Just _ <- lookupCtxtExact n (gamma defs)
                               | _ => throw (UndefinedName fc n)
```

</p>
</details>  

I'm not sure what it's trying to tell me -- what's gone wrong here?

mb64:
Fixes #517.

There are some parts I'm not sure about -- I left comments for them.  I'd appreciate some guidance for those cases.

One of the things I wasn't sure about is what the story on recursive quoting is/should be.  For example, should this:
```idris
f :: TTImp -> Bool
f `(g `(List ~(x))) = ?f_rhs
f _ = False
```

bind `x` as a pattern variable for `f`, or as a literal `IUnquote _ (IVar _ (UN "x"))` pattern?


mb64:
Hmm, it seems that while this works:
```idris
getListType : TTImp -> Maybe TTImp
getListType `(List ~(ty)) = Just ty
getListType _ = Nothing
```
And this works:
```idris
myListDecl : TTImp -> List Decl
myListDecl ty = `[myThing : List ~(ty)]
```
This *doesn't* work:
```idris
getListTypeDecl : List Decl -> Maybe TTImp
getListTypeDecl `[myThing : List ~(ty)] = Just ty
getListTypeDecl _ = Nothing
```

I'm not sure why.  The ways in which it doesn't work are different with and without this patch.

edwinb:
Sorry to leave this lingering so long, it's been a busy semester.

I haven't thought about recursive quoting. My instinct is that names should only be considered pattern variables at the very top level (so they have to be under as many escapes as they are quotes) but I haven't thought that through very deeply.

mb64:
It looks like Scheme agrees with you there:
```scheme
> (let ((x 12)) `(,x `(,x) `(,,x)))
(12 `(,x) `(,12))
```

I'll see if I can rebase and implement this soon.

edwinb:
As far as I can tell there's still some unresolved details here, but the patch as it stands certainly makes progress. So, I've resolved the existing conflict and I'll merge it as long as the tests still pass, but we will need to come back to the details for recursive quoting.

edwinb:
oh dear, looks like it needs a bit more attention. I'll see if I can sort it out now and open a new PR with the updates.

PR #531, opened 2020-08-07
Makes timing output more intuative to read.

jfdm:
The timings flag (`--timing`) displays various timing information during the entire compilation process.
This is useful when performing adhoc profiling of the compiler.
However, the information is a mixture of timestamps and relative progress.
The information is inherently nested, and such nesting is not immediatly clear in the output.
This can be confusing.

This minor changes nests the timing information on output so that when aligned one can see the timestamps at the highest level, and progress during the elaboration stage.


andrevidela:
I love this!

melted:
Looks good!

PR #533, opened 2020-08-07
Refactor grammar for dependent pairs

### LINKED ISSUE #532: Parsing Parentheses is not performant.
# Steps to Reproduce

Take the following [straightforward implementation of the Intrinsically Typed STLC in Idris and an ill-typed program](https://gist.github.com/jfdm/1233cba2c3ab9473fe1b0be233aad3ec) and compile with Idris2.

# Expected Behavior

Idris2 will type check the file in a reasonable amount of time.

# Observed Behavior

Idris2 takes an unreasonable amount of time.

Specifically if we observe the timings (acquired using `--timing`) we see that there is nothing inherently wrong with the code and its elaboration.
Idris2 **just** takes a *long* time to parse.

```
1/1: Building Lambda (Lambda.idr)
TIMING Parsing Lambda.idr: 110.247s
TIMING Reading imports: 0.032s
TIMING Processing decls: 0.004s
TIMING Compile defs: 0.000s
TIMING Elaborating Lambda.idr: 110.285s
...
TIMING Build deps: 110.291s
TIMING Loading main file: 110.292s
```

This is potentially due to the amount of Parentheses on the RHS of the equation for `term`. If we comment out this term then compilation takes a reasonable amount of time.

edwinb:
As it was, there was significant backtracking for big expressions,
getting to the end, not finding a **, so having to try again for
application expressions. Fixes #532

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -230,23 +230,28 @@ mutual
                  pure (POp (MkFC fname start end) op l r))
                <|> pure l
 
-  dpair : FileName -> FilePos -> IndentInfo -> Rule PTerm
-  dpair fname start indents
+  dpairType : FileName -> FilePos -> IndentInfo -> Rule PTerm
+  dpairType fname start indents
       = do x <- unqualifiedName
            symbol ":"
            ty <- expr pdef fname indents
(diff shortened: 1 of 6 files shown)


jfdm:
Thanks!

PR #535, opened 2020-08-08
add `atan2` as a new builtin primitive

karroffel:
The `atan2` trigonometry function was missing.

This adds the function to the `prelude` and implements the builtin primitive in the scheme, JS and RefC backend.


melted:
Is `fastAtan2` faster in Idris? I don't think we should include such a function without benchmarking first.

edwinb:
I think prelude is the right place, but I agree with @melted that we should only include `fastAtan2` if it really is faster. We should certainly have the primitive (not sure why I didn't add it in the first place).

gallais:
No movement, I'm closing this but feel free to ask for it to be reopened if you
want to either drop the `fast` version or benchmark it to justify its inclusion.

karroffel:
Sorry for the long inactivity!

I can remove the `fastAtan2` for now since I also don't know if it's any faster! :sweat_smile: 

I realised that both Racket and Chez have either `atan2` or support a two parameter `atan`, libc also has `atan2` for doubles and floats and JavaScript has `Math.atan2`. Would it maybe make sense to make this function a builtin primitive? That might also have better performance than an approximation function done in Idris.

gallais:
Adding it as a builtin primitive would make sense given that's how most other
basic float functions are implemented.

karroffel:
I'll redo this with a new builtin primitive then, I'll probably get around to it this weekend! :+1:

gallais:
No progress so I'm closing this. Feel free to reopen another one.

PR #540, opened 2020-08-09
fix arity for blodwen-set-thread-data scheme proc

lodi:
I think the declaration for `System.Concurrency.Raw.setThreadData`...

```idris
%foreign "scheme:blodwen-set-thread-data"
prim__setThreadData : {a : Type} -> a -> PrimIO ()

setThreadData : {a : Type} -> a -> IO ()
setThreadData val = primIO (prim__setThreadData val)
```

...is passing an extra implicit parameter that the scheme code isn't expecting to receive.  I added an extra (unused) parameter... 

```diff
-(define (blodwen-set-thread-data a)
+(define (blodwen-set-thread-data ty a)
   (blodwen-thread-data a))
```

...and test case to validate that thread-local storage is working.

### DIFF
--- a/support/chez/support.ss
+++ b/support/chez/support.ss
@@ -173,7 +173,7 @@
 (define (blodwen-get-thread-data ty)
   (blodwen-thread-data))
 
-(define (blodwen-set-thread-data a)
+(define (blodwen-set-thread-data ty a)
   (blodwen-thread-data a))
 
 (define (blodwen-mutex) (make-mutex))
--- a/support/gambit/support.scm
+++ b/support/gambit/support.scm
@@ -107,7 +107,7 @@
   (let ((data (thread-specific (current-thread))))
     (if (eq? data #!void) #f data)))
 
-(define (blodwen-set-thread-data a)
+(define (blodwen-set-thread-data ty a)
   (thread-specific-set! (current-thread) a))
 
 (define blodwen-mutex make-mutex)
--- a/support/racket/support.rkt
+++ b/support/racket/support.rkt
@@ -165,7 +165,7 @@
 (define (blodwen-get-thread-data ty)
   (thread-cell-ref blodwen-thread-data))
 
-(define (blodwen-set-thread-data a)
+(define (blodwen-set-thread-data ty a)
   (thread-cell-set! blodwen-thread-data a))
 
 (define (blodwen-mutex) (make-semaphore 1))
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -125,7 +125,7 @@ chezTests
       "chez007", "chez008", "chez009", "chez010", "chez011", "chez012",
       "chez013", "chez014", "chez015", "chez016", "chez017", "chez018",
       "chez019", "chez020", "chez021", "chez022", "chez023", "chez024",
-      "chez025", "chez026", "chez027",
+      "chez025", "chez026", "chez027", "chez028",
       "reg001"]
 
 nodeTests : List String
--- a/tests/chez/chez028/ThreadData.idr
+++ b/tests/chez/chez028/ThreadData.idr
@@ -1,0 +1,49 @@
+module Main
+
+import System.Concurrency.Raw
+
+
+child : Condition -> Mutex -> IO ()
+child cond mtx = do
+    mutexAcquire mtx
+    str <- getThreadData String
+    putStrLn $ "child data: " ++ (show str)
+
+    setThreadData 17
+    i <- getThreadData Int
+    putStrLn $ "child data now: " ++ (show i)
+
+    conditionSignal cond
+    conditionWait cond mtx
+
+    putStrLn $ "child exiting"
+
+    conditionSignal cond
+    mutexRelease mtx
+
+
+main : IO ()
+main = do
+    setThreadData 13
+    i <- getThreadData Int
+    putStrLn $ "parent data initialized to: " ++ (show i)
+
+    setThreadData "hello"
+    str <- getThreadData String
+    putStrLn $ "parent data now: " ++ (show str)
+
+    mtx <- makeMutex
+    cond <- makeCondition
+
+    mutexAcquire mtx
+    _ <- fork (child cond mtx)
+    conditionWait cond mtx
+
+    str2 <- getThreadData String
+    putStrLn $ "parent data still: " ++ (show str2)
+
+    conditionSignal cond
+    conditionWait cond mtx
+
+    putStrLn $ "parent exiting"
+    mutexRelease mtx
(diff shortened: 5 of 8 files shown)


gallais:
Wouldn't a better fix be to instead change the type to `{0 a : Type} ->`?
(Or to drop the `{a : Type} ->` entirely as this will have the same meaning.

lodi:
Yeah, I agree; I tried doing exactly that yesterday but I couldn't get it working for some reason...

I tried:

```idris
%foreign "scheme:blodwen-set-thread-data"
prim__setThreadData : {0 a : Type} -> a -> PrimIO ()
%foreign "scheme:blodwen-get-thread-data"
prim__getThreadData : (0 a : Type) -> PrimIO a

setThreadData : {0 a : Type} -> a -> IO ()
setThreadData val = primIO $ prim__setThreadData val

getThreadData : (0 a : Type) -> IO a
getThreadData a = primIO $ prim__getThreadData a
```

and

```idris
%foreign "scheme:blodwen-set-thread-data"
prim__setThreadData : a -> PrimIO ()
%foreign "scheme:blodwen-get-thread-data"
prim__getThreadData : PrimIO a

setThreadData : {0 a : Type} -> a -> IO ()
setThreadData val = primIO $ prim__setThreadData val

getThreadData : (0 a : Type) -> IO a
getThreadData a = primIO $ prim__getThreadData
```

and

```idris
%foreign "scheme:blodwen-set-thread-data"
prim__setThreadData : a -> PrimIO ()
%foreign "scheme:blodwen-get-thread-data"
prim__getThreadData : PrimIO a

setThreadData : a -> IO ()
setThreadData val = primIO $ prim__setThreadData val

getThreadData : (a : Type) -> IO a
getThreadData a = primIO $ prim__getThreadData
```

But I always get extra arguments generated at the callsites of `getThreadData` and `setThreadData` respectively.

```
Warning in compile: possible incorrect argument count in call (blodwen-set-thread-data farg-0 farg-1) at line 276, char 83 of /home/lodi/pr/Idris2/tests/chez/chez028/build/exec/_tmpchez_app/_tmpchez.ss

Warning in compile: possible incorrect argument count in call (blodwen-get-thread-data farg-0) at line 281, char 76 of /home/lodi/pr/Idris2/tests/chez/chez028/build/exec/_tmpchez_app/_tmpchez.ss
```

The generated file:

```scheme
(define Raw-Concurrency-System-setThreadData (lambda (arg-0 arg-1 ext-0) (Raw-Concurrency-System-prim__setThreadData 'erased arg-1 ext-0)))
(define Raw-Concurrency-System-getThreadData (lambda (arg-0 ext-0) (Raw-Concurrency-System-prim__getThreadData 'erased ext-0)))

(define Raw-Concurrency-System-prim__setThreadData (lambda (farg-0 farg-1 farg-2) (blodwen-set-thread-data farg-0 farg-1)))
(define Raw-Concurrency-System-prim__getThreadData (lambda (farg-0 farg-1) (blodwen-get-thread-data farg-0)))

(define blodwen-thread-data (make-thread-parameter #f))
(define (blodwen-get-thread-data)
  (blodwen-thread-data))
(define (blodwen-set-thread-data a)
  (blodwen-thread-data a))
```

I'm doing a `make clean && make bootstrap SCHEME=scheme` in between edits.

Russoul:
There is actually a third parameter that needs to be introduced here: `%World` . @gallais Don't scheme backends keep the erased parameters in function signatures ? I haven't found any code that gets rid of those... Also the value of type `%World` is passed as `#f` whatever that is. Proof that this is how it is (if I am right about the fact that erased parameters and `%World` are preserved in scheme code):

A bit of Idris-lua:
```
module Test

%foreign "lua:function(ty, value, world) print('" 
            ++ "arg1 = ' .. idris.inspect(ty) .. '\\n"
            ++ "arg2 = ' .. idris.inspect(value) .. '\\n"
            ++ "arg3 = ' .. world) end|inspect"
printTyImpl : a -> PrimIO ()

printTy : HasIO io => a -> io ()
printTy x = primIO (printTyImpl x)

main : IO ()
main = printTy (the (List Nat) [1, 2])
```
image

Russoul:
Edit: `%World` is present only in foreign definition but it won't get passed to `blodwen-get-thread-data` and `blodwen-set-thread-data`. But the point is, I don't think there is a way of doing what @lodi done using erased parameters as they are passed down to functions

lodi:
Sorry @Russoul, I still don't understand.  I've pared it down to this:

```idris
%foreign "scheme:blodwen-set-thread-data"
prim__setThreadData : a -> PrimIO ()

export
setThreadData : a -> IO ()
setThreadData val = primIO (prim__setThreadData val)
```

which is very similar to another declaration in that file:

```idris
%foreign "scheme:blodwen-lock"
prim__mutexAcquire : Mutex -> PrimIO ()

export
mutexAcquire : Mutex -> IO ()
mutexAcquire m = primIO (prim__mutexAcquire m)
```

but then in the generated scheme I get:

```scheme
(define (blodwen-set-thread-data a)
  (blodwen-thread-data a))

(define (blodwen-lock m) 
  (mutex-acquire m))

(define Raw-Concurrency-System-prim__setThreadData 
  (lambda (farg-0 farg-1 farg-2) 
    (blodwen-set-thread-data farg-0 farg-1)))

(define Raw-Concurrency-System-prim__mutexAcquire 
  (lambda (farg-0 farg-1) 
    (blodwen-lock farg-0)))
```

From what you wrote above, I understand that Idris generates a lambda with an extra `%World` parameter at the end.  But why does `setThreadData` still take an extra argument on top of that compared to something like `mutexAcquire`?

Russoul:
Because implicit `{0 a : Type}`, though erased, is still present in function signatures and is passed as some kind of dummy object

Russoul:
So I think your original fix works best here, let’s see if @gallais confirms

lodi:
Oh I see, okay thanks @Russoul.

lodi:
Hi @gallais.  With regard to the conversation above, do you have any concerns merging the original PR?

gallais:
I don't know enough about this. I would be more comfortable if someone
who actually understands that part of the system decides whether to merge.

edwinb:
If it helps unblock this (I'm just doing a quick pass through outstanding PRs...) the 0-multiplicity arguments are still passed, but with a null value. It is fiddly to erase them completely, and I did an experiment a while back that suggested it didn't actually affect performance. But I'll revisit that some day.

For the sake of making progress, perhaps the right thing to do at first is to add the argument in the scheme, and deal with the multiplicity problem afterwards.

edwinb:
I've brought this up to date and opened a new PR, so I'll close this one.

PR #543, opened 2020-08-10
Port Decidable.Order from Idris1

abailly:
I have found `Decidable.Order`'s `order` function to be very useful in `with` clauses and I use it in a few places of my code, so I would be interested in having it part of the standard libs.


libs/base/Data/Fun.idr
```diff
@@ -0,0 +1,20 @@
...
+%default total
+
+||| Build an n-ary function type from a Vect of Types and a result type
+public export
+Fun : Vect n Type -> Type -> Type
+Fun [] r = r
+Fun (t::ts) r = t -> Fun ts r
+
+public export
+chain : {ts : Vect n Type} -> Fun [r] r' -> Fun ts r -> Fun ts r'
```
gallais, line 14: Is this `map` for `Functor (Fun ts)`?

libs/base/Data/Rel.idr
```diff
@@ -0,0 +1,13 @@
...
+
+import Data.Fun
+
+||| Build an n-ary relation type from a Vect of Types
+public export
+Rel : Vect n Type -> Type
+Rel ts = Fun ts Type
+
+public export
+liftRel : (ts : Vect n Type) -> (p : Rel ts) -> (Type -> Type) -> Type
```
gallais, line 11: I think you ought to distinguish between `All : (ts : Vect n Type) -> Rel ts -> Type`
the function turning a `Rel` into a `Type` by universally quantifying over its domain
(its dual `Ex` would be useful too btw) and just using `map` on the return `Type` of
the relation.

libs/base/Decidable/Decidable.idr
```diff
@@ -0,0 +1,23 @@
...
+
+--------------------------------------------------------------------------------
+-- Interface for decidable n-ary Relations
+--------------------------------------------------------------------------------
+
+--  Note: Implementation resolution for Decidable is likely to not work in the REPL
+--        at the moment.
+public export
+interface Decidable (ts : Vect k Type) (p : Rel ts) where
+  total decide : liftRel ts p Dec
```
gallais, line 14: This would then become `All ts (map Dec p)`.

libs/base/base.ipkg
```diff
@@ -40,6 +40,7 @@ modules = Control.App,
           Debug.Trace,
```
gallais, line 40: Are you missing `Data.Fun` in this list?

libs/base/Decidable/Order.idr
```diff
@@ -0,0 +1,156 @@
...
+  total symmetric : (a : t) -> (b : t) -> eq a b -> eq b a
+
+interface (Equivalence t eq) => Congruence t (f : t -> t) (eq : t -> t -> Type) where
+  total congruent : (a : t) ->
+                    (b : t) ->
+                    eq a b ->
+                    eq (f a) (f b)
+
+minimum : (Ordered t to) => t -> t -> t
+minimum {to} x y with (order {to} x y)
```
gallais, line 40: You seem to consistently need to introduce `to` and apply it
explicitly which makes sense given that their can be multiple
interesting orders on a given type.

I think this is a sign `to` should be an explicit argument in
all of these definitions.

libs/base/Decidable/Order.idr
```diff
@@ -0,0 +1,156 @@
...
+  reflexive a = Refl
+
+implementation Equivalence t Equal where
+  symmetric a b prf = sym prf
+
+implementation Congruence t f Equal where
+  congruent a b eq = cong f eq
+
+--------------------------------------------------------------------------------
+-- Natural numbers
```
gallais, line 64: This is a bit of a surprising location for these. I would
expect them to be under `Data.Nat.Order` (for instance).

abailly:
Thanks for the insightful suggestions Ohad, I did not put anymore thoughts
in this than the minimum needed to make it compile on Idris2, I will do
what you suggest.

libs/base/Data/Rel.idr
abailly: I don't know what `Ex` type would be, what are you thinking of?

gallais: `Ex` would be the same type but a different function:

```idris
Ex : (ts : Vect n Type) -> (p : Rel ts) -> Type
Ex [] p = p
Ex (t :: ts) p = (x : t ** Ex ts (p x))
```

abailly: How is it a dual of All? Because All produces product types and Ex sum types?

abailly: Also, how would this be used?

gallais: One interesting result is generalised (un)currying:

`Ex ts p -> t` ~ `All ts (map (\ pxs -> pxs -> t) p)`

That is to say:

`(x_1 : t_1 ** ... ** x_n : t_n ** p x_1 ... x_n) -> t`
is equivalent to
`(x_1 : t_1) -> ... -> (x_n : t_n) -> p x_1 ... x_n -> t`

gallais:
I think you are now missing `Data.Rel` and `Decidable.Decidable` in the `.ipkg` file.

abailly:
@gallais  fixed missing declaration in package file

PR #548, opened 2020-08-10
add support for more casts from and to BitsN types

### LINKED ISSUE #545: No casts defined for BitsN types
When working with `Bits8` I noticed there is no way to convert a `Bits8` to any other numeric type.

# Steps to Reproduce

Create a file `casts.idr`

```idris
asNormalisedDouble : Bits8 -> Double
asNormalisedDouble b = (cast b) / 255.0

toBits16 : Bits8 -> Bits16
toBits16 b = cast b
```

Load the file into the REPL and see the errors. (See section "Observed Behavior")

# Expected Behavior

The file loads fine in the REPL and calling `asNormalisedDouble 255` should give the result `1.0`. Calling `toBits16 32` should give the result `32`.

# Observed Behavior

```
1/1: Building casts (casts.idr)
casts.idr:2:25--2:31:While processing right hand side of asNormalisedDouble at casts.idr:2:1--4:1:
Can't find an implementation for Cast Bits8 Double at:
2	asNormalisedDouble b = (cast b) / 255.0
	                        ^^^^^^

casts.idr:5:14--5:20:While processing right hand side of toBits16 at casts.idr:5:1--6:1:
Can't find an implementation for Cast Bits8 Bits16 at:
5	toBits16 b = cast b
	             ^^^^^^
```

# Solutions

Using `prim__cast_Bits8Integer` you can go from `Bits8` to `Integer` and from there to other types via `fromInteger`. This could be a stopgap for implementing `Cast` instances at first.

```idris
public export
Cast Double Bits8 where
    cast d = fromInteger $ the Integer (cast d)

public export
Cast Bits8 Double where
    cast b = fromInteger $ prim__cast_Bits8Integer b
```

For high computing workloads, going through `Integer` might have severe performance implications. "Native" casts from `BitsN` to all other `BitsN` types as well as `Int` and `Double` would be better for this as a codegen backend could use more optimised casts this way.

The functions for this should be `prim__cast_Bits8Bits16`, `prim__cast_Bits8Bits32`, ..., `prim__cast_Bits32Bits8` etc.

karroffel:
closes #545

Adds primitive casts between different `BitsN` types as well as for the
`Int` and `Integer` type.

`Cast` instances from and to `BitsN` types have been added to the `prelude`.

The scheme backends (Chez, Racket, Gambit) contain functions for
"narrowing" the bit width of integers, widening is a no-op.
The ES backend behaves similarly.


karroffel:
The `prelude` changes fail to compile because the bootstrap code does not support the primitive functions yet.

How should I proceed with this PR? Should I split it in two so that the primitive functions are supported but not yet in the prelude, then a second PR could add the `Cast` implementations when a new bootstrap compiler version is released? I don't know what's the intended process for this :slightly_smiling_face:

gallais:
The process (as far as I understand it) is as follows:

* Make a PR to edwinb/Idris2-boot adding the same functions to that prelude
* Use the idris -> idris2-boot -> idris2 way to build your candidate
* Grab the replacements for the files in `bootstrap/idris2_app` from `build/exec/idris2_app`
* Replace the value of `IdrisPaths-yprefix` in these files with `__PREFIX__`
* Commit these new bootstrap file and push them to your PR

karroffel:
Idris2-boot does not support `BitsN` types, so backporting the change in the compiler won't be possible. I could add the primitive cast functions to the prelude anyway and just fill them in like this:
```idris
prim__cast_Bits16Bits8 : a -> b
prim__cast_Bits16Bits8 _ = idris_crash "primitive cast Bits16->Bits8 unsupported in Idris2-boot"
```

Then the bootstrap compiler would be able to compile Idris2 which would then implement those operations properly. Is that about right?

karroffel:
I added the functionality to perform the casts in the REPL or in the type checker as well.

gallais:
The only thing you are missing is a couple of test cases
showing that widening does not change the value & that
shortening can either preserve the value if it is small
enough or modify it.

Ideally you would have the same test case in `idris2`,
`chez` and `node` (with a comment at the top pointing
that all 3 should yield the same results).

karroffel:
That's a good point, I will add some more tests for this.

karroffel:
I added test cases to check for lossless upcasting, wrapping behaviour for downcasts and casting of negative numbers.
If I should add some more tests please let me know!

gallais:
Lovely tests! I'm currently fixing the conflicts introduced by the merge of other PRs.

tests/chez/chez029/run
```diff
@@ -0,0 +1,3 @@
+$1 --no-banner BitCasts.idr < input
```
ShinKage, line 1: Tests should always have `--no-color --console-width 0` since they are too much dependent on the particular system where the tests are run.

gallais: Thanks for the tip!

tests/node/node022/run
```diff
@@ -0,0 +1,3 @@
+$1 --cg node --no-banner BitCasts.idr < input
```
ShinKage, line 1: Oops, missed this, sorry. It also needs `--no-color --console-width 0`.

PR #551, opened 2020-08-11
Split CFUnsigned into 8/16/32/64 bits versions

### LINKED ISSUE #549: FFI only reading up to 32 bit integers (e.g. no uint64_t -> Bits64)
Hi, I created a small example to test whether one can read/write `Bits64, Bits32` etc. through the C FFI:

https://github.com/neriglissar/idris2-int-test

I found that:
- Reading from C into Bits64 is capped at 32 bits
- Reading from C into Bits16 can exceed 16 bits
- Writing more than 32 bits to C crashes the runtime

Hope someone is able to fix this for the Scheme backend! Thanks.

# Steps to Reproduce
```
1/1: Building Integers (Integers.idr)
Integers> :exec run
```
```
1/1: Building Integers (Integers.idr)
Integers> :exec fun
```
# Expected Behavior
```
1/1: Building Integers (Integers.idr)
Integers> :exec run
====== Bits 64 ======
C:     18446744073709551615
Idris: 18446744073709551615
C:     18446744073709551615
Idris: 18446744073709551615
```

````
Integers> :exec fun
====== Bits 64 ======
Idris: 18446744073709551615
C:     18446744073709551615
````
# Observed Behavior
```
1/1: Building Integers (Integers.idr)
Integers> :exec run
====== Bits 64 ======
C:     18446744073709551615
Idris: 4294967295
C:     4294967295
Idris: 4294967295
```
and also
````
====== Bits 16 ======
C:     4294967295
Idris: 4294967295
C:     4294967295
Idris: 4294967295
`````
````
Integers> :exec fun
====== Bits 64 ======
Idris: 18446744073709551615
Exception in write__uint64: invalid foreign-procedure argument 18446744073709551615
````

SmiVan:
This should fix #549, as well as other potential problems caused by Bits8, Bits16 and Bits64 being treated as Bits32 for foreign functions.

Needs a review by someone who understands the bigger picture, particularly in relation to `TTC.idr` as it is entirely unclear to me what that does.

### DIFF
--- a/src/Compiler/CompileExpr.idr
+++ b/src/Compiler/CompileExpr.idr
@@ -506,10 +506,10 @@ getNArgs defs n args = pure $ User n args
 nfToCFType : {auto c : Ref Ctxt Defs} ->
              FC -> (inStruct : Bool) -> NF [] -> Core CFType
 nfToCFType _ _ (NPrimVal _ IntType) = pure CFInt
-nfToCFType _ _ (NPrimVal _ Bits8Type) = pure CFUnsigned
-nfToCFType _ _ (NPrimVal _ Bits16Type) = pure CFUnsigned
-nfToCFType _ _ (NPrimVal _ Bits32Type) = pure CFUnsigned
-nfToCFType _ _ (NPrimVal _ Bits64Type) = pure CFUnsigned
+nfToCFType _ _ (NPrimVal _ Bits8Type) = pure CFUnsigned8
+nfToCFType _ _ (NPrimVal _ Bits16Type) = pure CFUnsigned16
+nfToCFType _ _ (NPrimVal _ Bits32Type) = pure CFUnsigned32
+nfToCFType _ _ (NPrimVal _ Bits64Type) = pure CFUnsigned64
 nfToCFType _ False (NPrimVal _ StringType) = pure CFString
 nfToCFType fc True (NPrimVal _ StringType)
     = throw (GenericMsg fc "String not allowed in a foreign struct")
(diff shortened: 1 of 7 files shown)


SmiVan:
If the enumeration in `TTC.idr` needs to be preserved, then CFUnsigned should be replaced with CFUnsigned32 there and the other three could be appended at the end I guess.

ShinKage:
Since this PR makes changes to the TTC binary format, it should also bump the version in `Core.Binary.ttcVersion`.

SmiVan:
@ShinKage What is TTC, by the way?

ShinKage:
> @ShinKage By the way, what is TTC?
> I would appreciate it if you could give me some keywords to look up or at least the meaning of the abbreviation so I can figure out what it does.

Anybody with more precise information or experience, feel free to correct me if I'm wrong. In short, after processing a module for the first time the compiler dumps the processed information in a binary format to be read later already processed (i.e. installed modules are just these binary files). All the data that can be serialized must implement the TTC interface (see `Core.Binary` for the definition and the support functions and the `Core.TTC` for the interface implementations). You can check these generated binary files in the `build` (by default) directory in the directory of the ipkg file. There are also the .ttm files that are in the same binary format but contains other metadata for REPL and IDE mode, IIRC. You can find more information in the [implementation notes](https://idris2.readthedocs.io/en/latest/implementation/overview.html#ttc-format) and in the files mentioned before.

lodi:
@SmiVan, I did a bit of sleuthing and I figured out the etymology of 'TT'.  According to [this](http://docs.idris-lang.org/en/latest/reference/internals.html), "Idris is compiled to a simple, explicit core language. This core language is called TT because it looks a bit like a Π."  

Note that dependent types are called pi-types in the academic literature.  Why Pi?  From [wikipedia](https://en.wikipedia.org/wiki/Dependent_type#%7F'%22%60UNIQ--postMath-00000003-QINU%60%22'%7F_type), "The name 'pi-type' comes from the idea that these may be viewed as a Cartesian product of types."

melted:
This should cause no problem with bootstrapping, so I'll merge it. Thanks!

PR #552, opened 2020-08-12
rename parsePositive, refine docs

Russoul:
I don't know how bad it would be to change it now, but `parsePositive` is incorrect name ...
Also clarified what kind of numbers are meant in `parsePositive`, `parseInteger` and `parseDouble`

We could correct the behaviour of `parsePositive` and introduce `parseNatural` independently if changing names is too big of a deal.

Had to rename one use of `parsePositive` in compiler

### DIFF
--- a/libs/base/Data/Strings.idr
+++ b/libs/base/Data/Strings.idr
@@ -215,17 +215,20 @@ parseNumWithoutSign (c :: cs) acc =
   then parseNumWithoutSign cs ((acc * 10) + (cast ((ord c) - (ord '0'))))
   else Nothing
 
-||| Convert a positive number string to a Num.
+||| Convert a natural number string to a Num.
 |||
 ||| ```idris example
-||| parsePositive "123"
+||| parseNatural "123"
 ||| ```
 ||| ```idris example
-||| parsePositive {a=Int} " +123"
+||| parseNatural {a=Int} " +123"
+||| ```
+||| ```idris example
+||| parseNatural {a=Int} "0"
 ||| ```
 public export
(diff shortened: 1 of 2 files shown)


gallais:
I don't think a natural should ever be written with a `+` upfront
so this renaming does not make sense to me.

I guess `0` is considered valid so some people may want to call this
a non-negative rather than a positive but [this vocabulary is ambiguous](https://en.wikipedia.org/wiki/Sign_(mathematics)#non-negative_and_non-positive)
and positive also has the meaning of `in [0..)`.

Not convinced this is worth introducing a breaking change to `base` tbh.

Russoul:
I don’t know, the fact that it accepts `0` as a valid input was strange to me. I believe it’s more natural for a functional programmer to think of `0` as a neutral number, considering it’s role in structures like Monoid. Is there any practical reason to unify it with positives ?

And I believe the `+` sign is ok, because formally the set of Natural numbers is a union of `0` with a difference of the set of Integers and the set of negative Integers.
So Integer `+2` **is** a natural number.

I mean, even disregarding all above, ‘parsePositive’ is an ambiguous name.
‘parseNatural’ is not, because Idris defines Naturals starting from `0` and up.

But if it’s not worth the breaking change, I’ll live with it :)

edwinb:
I'm not sure this is worth the breaking change either, especially if it affects the compiler which is already a bit fiddly to keep consistent with the libraries since it's supposed to build with the previous version. I'm inclined to closet this one.

Russoul:
Sure, I’ll do it

PR #554, opened 2020-08-12
Input validation

Sventimir:
This module is a basis for framework aiming at checking runtime validity of user input. The idea is if we want to have our programs proven correct, at some point we have to make assumptions about user input. But the nature of that input is such that it comes as an unstructured text or even raw bytes. Therefore an automatic process is required that will check if the input adheres to programmer's assumptions and construct a proof thereof to later be used to prove the correctness of the whole program. On the other hand, if the input turns out to be invalid, the proof cannot be made and therefore further processing stops immediately with a nice error message.

The validation framework is realised as a monad transformer, so that primitive validators can be combined easily to build complex processes checking a lot of things. These processes come in two flavours. First we want to refine the type of input by making sure that it actually represents the data we expect. For instance we check that arguments expected to be numbers, are indeed numbers and not some rubbish and by doing so we refine the type of the input. This is pretty usual stuff. Then comes the part unique to Idris, which is proving theorems (assumptions) about thus refined input (for instance that a number is not greater than some upper bound), which returns a proof about input to be used later for calling functions that might require it. This is actually a lot harder to express in the type system and I am not at all sure if I have chosen the correct way, so any suggestions in this direction are welcome.

I have written a few primitive validators already as an example, but there's probably a lot more to be added later. I also attach a simple program computing GCDs as an example of how the framework is intended to be used.

[gcd.idr](https://github.com/idris-lang/Idris2/files/5064223/gcd.txt)


gallais:
This looks very cool. However I am not sure we need the distinction
`StructValidator` vs. `PropValidator`. It looks to me like we could
write exactly the same library with `ValidatorT m a b`  being defined
as `a -> Result m b`.

Sventimir:
@gallais, I believe it is necessary for the following reason. When you want to return a statement of the form `p x` from a `Validator`, then `x` must be an argument to the `Validator` constructor. Otherwise Idris considers it a free variable, which cannot be unified with anything. Or at least that was the problem I encountered when I initially wrote the `Validator` type with single constructor like you suggested. Perhaps it's possible to have the constructor:

```
PropValidator : {0 a, b : Type} -> {0 p : b -> Type} -> ((x : a) -> Result m (p x)) -> ValidatorT m a (p x)
```

I'm not sure if I tried that. Anyway it's still a different form that `StructValidator`. I will look into it further later today and write what I figured out.

libs/contrib/Control/Validation.idr
```diff
@@ -0,0 +1,193 @@
...
+equal x y = MkValidator dec
+    where
+    dec : z -> Result m (x = y)
+    dec _ = case decEq x y of
+        Yes prf => pure prf
+        No _ => fail "Values are not equal."
+
+||| Verify that a Nat is less than or equal to  certain bound.
+export
+lteNat : Monad m => {0 a : Type} -> (bound, n : Nat) -> ValidatorT m a (LTE n bound)
```
gallais, line 173: These look suspicious to me: the conclusion mentions an `a` that has absolutely
nothing to do with the validation at hand. And the `a` argument is indeed ignored
in `\_, x => isLTE x bound`.

A solution may be to have

```idris
data ValidatorT : (Type -> Type) -> (a : Type) -> (a -> Type) -> Type where
    MkValidator : ((x : a) -> Result m (p x)) -> ValidatorT m a p
```

and then we can have

```idris
lteNat : Monad m => (bound : Nat) -> ValidatorT m Nat (\ n => LTE n bound)
```

The simpler things like `natural` can be:

```idris
natural : Monad m => ValidatorT m String (\ _ => Nat)
```

Of course the problem is that it's not a functor / applicative / monad
anymore but I wonder whether we can recover useful combinators nonetheless.
You can still have `map` for instance:

```idris
map : (forall x. p x -> q x) -> ValidatorT m a p -> ValidatorT m a q
```

Sventimir: You're right and I thought about it as well. But when I realised that it's no longer a monad that way, I gave up. I think `do` notation is too valuable here to resign on it easily. And if it's not a monad, it can't be a monad transformer, so you won't be able to mix validation with an IO operation (for instance validating a large input file on the fly). Besides, I thought, maybe it's not so bad? It might happen from time to time, that having the original `a` for validating some of it's parts might be useful. Although admittedly I couldn't come up with a convincing example so far.

All in all I'm not really sure what should be done with it. What you point out is troubling indeed, but the ability to combine validators easily is essential here (it's precisely why I wrote this module in the first place). Without `ValidatorT` being a monad I think most of the value of this module would be lost.

gallais: It's in `contrib` so we can easily revisit this if we find a better design in
the future. In the meantime, I'd rather replace these `a` with `()`. It should
help with type inference & clarifies that we do not care about that `a`.

Sventimir: This would be nice, but I believe there is still the same problem as before, because having validators `ValidatorT m a b` and `ValidatorT m () (p x)`, you can't bind them together anymore. To maintain composability, we would have to have a wrapper, that would convert `Validator m () (p x)` to a suitable type, so that it could be composed. This is possible of course, but seems somewhat superfluous…

gallais: Oh I see. Looking at `ValidatorT` as a monad gives the `a` a `Reader`-flavour
rather than simply an input to the validator. That's... strange.

Maybe the right interface for composing `ValidatorT m` is more [`Arrow`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Control-Arrow.html) than `Monad`!
We are, after all, working in the Kleisli cat of the `ResultT m` monad.

Sventimir: Hmm, you're right. Arrow might be exactly what we want here. Making `ValdatorT` synonym of `Result` is not that great, because it makes you pass the input manually to all validators one after another. I consider this `Reader`-flavour quite handy in this context. Also I have never worked with arrows before, so it's a great opportunity to learn more. I'll read a little bit and try to remake `ValidatorT` as an arrow, because from the first lines [here](https://en.wikibooks.org/wiki/Haskell/Understanding_arrows) it already looks like it's exactly what we want. I suppose I'll come back in a few days with the solution remade.

Sventimir: So I have started porting `Control.Category` and `Control.Arrow` from Idris to Idris 2. However, I have a strange problem with interface resolution in the first of them:

```
module Control.Category

import Data.Morphisms


public export
interface Category (cat : k -> k -> Type) where
  id  : cat a a
  (.) : cat b c -> cat a b -> cat a c

public export
Category Morphism where
  id                = Mor id
  -- disambiguation needed below, because unification can now get further
  -- here with Category.(.) and it's only interface resolution that fails!
  (Mor f) . (Mor g) = Mor $ Basics.(.) f g

public export
Monad m => Category (Kleislimorphism m) where
  id                        = Kleisli (pure . id)
  (Kleisli f) . (Kleisli g) = Kleisli $ \a => g a >>= f

infixr 1 >>>

(>>>) : Category cat => cat a b -> cat b c -> cat a c
(>>>) {cat} f g = g . f
```
This fails to compile because:
```
While processing right hand side of >>> at Control/Category.idr:26:1--27:1:
Can't find an implementation for Category ?cat at:
26	(>>>) {cat} f g = g . f
	                  ^^^^^
```
It would seem that Idris either does not see that `?cat` is of type `cat` or ignores the requirement that `cat` should implement `Category` interface. Also I'm not sure, why instances of  `Category` have type `k -> k -> Type`. If I change it to `Type -> Type -> Type`, then it works.

nickdrozd:
Could you add your example file as a test case or something?

Sventimir:
@nickdrozd, correct me if I'm wrong, but I don't think there are any tests for the `libs`. Besides, it's not a test really, as to make a test case of it, you'd need to specify some inputs and expected outputs (i.e. whether they're valid inputs or not). It could be done if we had a testing framework in place, which I think is not the case.

Besides, the interface for gluing primitive validators together will likely change significantly before it gets merged.

However, if you'd like to compile it for yourself and play a little with it, it shouldn't be much of a problem. Probably it would be simplest to build Idris2 from my branch (I recommend Docker for making sure that different builds don't break each other down), so that it has the library in place. Alternatively you could copy modules added by this PR to some fresh project along with `gcd.idr` as Main, and it should work as well.

Sventimir:
@gallais , I have ported `Control.Category` and `Control.Arrow` from Idris1, but to no avail. Validator type we dreamt of cannot be made an instance of those for the same reason it can't be an instance of `Monad`. I retained the ported modules, so they can be merged by the way. However, there is one important difference compared to Idris1: we have:

```
inteface Category (cat : Type -> Type -> Type) where
```
rather than:
```
interface Category (cat : k -> k -> Type) where
```
because with the latter Idris didn't accept default implementations of functions in `Control.Arrow` interfaces, saying that `?cat does not implement Category` or something like that. This is very strange, probably a bug of some kind, so I don't know whether to port these modules after all or maybe delete them, since I didn't use them anyway.

As to the `Control.Validation` module, I introduced another type of `PropValidator`, which can be explicitly converted to a `ValidatorT` for any `a`. This way the value under validation gets ignored more explicitly and it's also mentioned in the docs. I don't know if anything better than that is even possible…

edwinb:
I see this has been stuck for a while, but just to observe, we've typically been fairly liberal with what to include in `contrib` so even if something isn't complete or polished, we can err on the side of merging. 
Eventually we probably do need to be a bit more organised about libraries. I'm not really sure what should be part of this repo and what should be elsewhere - there comes a point where there's too much to maintain here, but on the other hand, we don't yet have an agreed upon package manager so things have to live somewhere.

edwinb:
I'm trying to tidy up the backlog of PRs - I'll merge this if the conflicts can be resolved. I think the only thing is that `Data.Either` is now in `base`.

andrevidela:
@Sventimir I've merged the changes from master to fix the conflict but now it seem you have linearity errors, could you take a look so we can merge this PR?

Sventimir:
But this can't possibly be related to my changes, can it? I mean, I didn't touch this module nor anything it depends on, did I?

andrevidela:
it might be a consequence of #879 I'm afraid

Sventimir:
@andrevidela if so, then why it happens in this branch and not in master?

andrevidela:
No idea! Let me take a look :)

Sventimir:
Okay, I'll look into this as well.

andrevidela:
if I checkout from master and pick your `Validation` file, add it to `contrib.ipkg` and add its dependency `Trans.Either` everything works! Maybe something is wrong with `Category` and `Arrow`? That's where I left off, maybe I'll look into it more again tomorrow

Sventimir:
`Category` and `Arrow` aren't essential to the change. Since I failed to make `Validator` a `Monad`, I hoped it could at least be made an `Arrow`. Unfortunately this also is not the case, so these modules can be safely removed if they make trouble.

PR #558, opened 2020-08-15
Support package names with @, / and - in lambdaRequire, fix an error …

xoltar:
…message, and add an idris list to js array primitive


melted:
Looks good!

PR #559, opened 2020-08-15
contrib: Add a 2-holed congruence function

ohad:
For lack of a better place, I've put it in `Syntax.PreorderReasoning`

These equations are natural in equational reasoning, but less so when
rewriting, so that's why it's there


gallais:
`cong` is defined in the prelude with this type signature:

```idris
Prelude.cong : (0 f : (t -> u)) -> (1 _ : a = b) -> f a = f b
```

Should we make `cong2`'s signature similar?

ohad:
Let me check if this somehow affects frex (:P), I doubt it would.

Do you think `cong2` should also go into the prelude?

gallais:
It's common enough that it was needed for the Lecture 1 exercises of SPLV
so I'd say yeah put it in the prelude. :D

PR #560, opened 2020-08-16
Version increment

edwinb:


edwinb:
Well this is annoying. This is a guess, but it looks like the bootstrap fails when you increment the version number because it's looking in 0.2.0 because that's what's coded in the bootstrap scheme! I suppose I should be updating that with a new version anyway. I'll add it to the checklist...

PR #563, opened 2020-08-17
implement HVect

thomasdziedzic:
mostly copied from the HVect from idris1.

I had a hard time getting deleteAt and the Shows interface working. But I got help in the #idris channel on slack. Thanks to arnaud and ionathanch.

### DIFF
--- a/libs/contrib/Data/HVect.idr
+++ b/libs/contrib/Data/HVect.idr
@@ -32,7 +32,7 @@ index (FS j) (x :: xs) = index j xs
 ||| ["string"]
 ||| ```
 public export
-deleteAt : (1 i : Fin (S l)) -> HVect ts -> HVect (deleteAt i ts)
+deleteAt : (1 i : Fin (S l)) -> (1 _ : HVect ts) -> HVect (deleteAt i ts)
 deleteAt FZ (x :: xs) = xs
 deleteAt (FS FZ) (x :: (y :: xs)) = x :: xs
 deleteAt (FS (FS j)) (x :: (y :: xs)) = x :: deleteAt (FS j) (y :: xs)
@@ -44,7 +44,7 @@ deleteAt (FS (FS j)) (x :: (y :: xs)) = x :: deleteAt (FS j) (y :: xs)
 ||| ["firstString", "string"]
 ||| ```
 public export
-replaceAt : (1 i : Fin k) -> t -> HVect ts -> HVect (replaceAt i t ts)
+replaceAt : (1 i : Fin k) -> t -> (1 _ : HVect ts) -> HVect (replaceAt i t ts)
 replaceAt FZ y (x :: xs) = y :: xs
 replaceAt (FS j) y (x :: xs) = x :: replaceAt j y xs
 
@@ -55,7 +55,7 @@ replaceAt (FS j) y (x :: xs) = x :: replaceAt j y xs
 ||| [True, "string"]
 ||| ```
 public export
-updateAt : (1 i : Fin k) -> (index i ts -> t) -> HVect ts -> HVect (replaceAt i t ts)
+updateAt : (1 i : Fin k) -> (index i ts -> t) -> (1 _ : HVect ts) -> HVect (replaceAt i t ts)
 updateAt FZ f (x :: xs) = f x :: xs
 updateAt (FS j) f (x :: xs) = x :: updateAt j f xs
 
@@ -66,7 +66,7 @@ updateAt (FS j) f (x :: xs) = x :: updateAt j f xs
 ||| [1, "string"]
 ||| ```
 public export
-(++) : HVect ts -> HVect us -> HVect (ts ++ us)
+(++) : (1 _ : HVect ts) -> HVect us -> HVect (ts ++ us)
 (++) [] ys = ys
 (++) (x :: xs) ys = x :: (xs ++ ys)
 
@@ -79,11 +79,11 @@ public export
   (x :: xs) == (y :: ys) = x == y && xs == ys
 
 public export
-consInjective1 : {xs, ys: HVect ts} -> {x, y: a} -> x :: xs = y :: ys -> x = y
+consInjective1 : {xs, ys: HVect ts} -> {x, y: a} -> (1 _ : x :: xs = y :: ys) -> x = y
 consInjective1 Refl = Refl
 
 public export
-consInjective2 : {xs, ys: HVect ts} -> {x, y: a} -> x :: xs = y :: ys -> xs = ys
+consInjective2 : {xs, ys: HVect ts} -> {x, y: a} -> (1 _ : x :: xs = y :: ys) -> xs = ys
 consInjective2 Refl = Refl
 
 public export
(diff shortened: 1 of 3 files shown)


libs/contrib/Data/HVect.idr
```diff
@@ -0,0 +1,135 @@
...
+public export
+Eq (HVect []) where
+  [] == [] = True
+
+public export
+(Eq t, Eq (HVect ts)) => Eq (HVect (t :: ts)) where
+  (x :: xs) == (y :: ys) = x == y && xs == ys
+
+public export
+hvectInjective1 : {xs, ys: HVect ts} -> {x, y: a} -> x :: xs = y :: ys -> x = y
```
gallais, line 83: Can we call these `consInjective(1/2)`? It's not the type constructor
that is proven injective, it's the constructor `(::)`.

thomasdziedzic: renamed

libs/contrib/Data/HVect.idr
```diff
@@ -0,0 +1,135 @@
...
+index (FS j) (x :: xs) = index j xs
+
+||| Delete an element from an HVect.
+|||
+||| ```idris example
+||| > deleteAt 0 (the (HVect _) [1, "string"])
+||| ["string"]
+||| ```
+public export
+deleteAt : {l : Nat} -> {ts : Vect (S l) Type} -> (1 i : Fin (S l)) -> HVect ts -> HVect (deleteAt i ts)
```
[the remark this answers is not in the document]
thomasdziedzic: simplified

berewt:
A few suggestions:

I had this for `deleteAt`, I don't know which one you would prefer:

```
deleteAt FZ (x :: xs) = xs
deleteAt (FS n) [x] = absurd n
deleteAt (FS n) (x :: xs@(_::_)) = x :: deleteAt n xs
```

Also, for `get`, `put` and `update`, we can set a linearity 1 on the Elem proof. (and not sure you need to provide the proof explicitly for them on the rhs).

Great work!

thomasdziedzic:
> A few suggestions:
> 
> I had this for `deleteAt`, I don't know which one you would prefer:
> 
> ```
> deleteAt FZ (x :: xs) = xs
> deleteAt (FS n) [x] = absurd n
> deleteAt (FS n) (x :: xs@(_::_)) = x :: deleteAt n xs
> ```
> 

Do you know what the benefits are of one over the other? Or is it just a stylistic difference?

> Also, for `get`, `put` and `update`, we can set a linearity 1 on the Elem proof. (and not sure you need to provide the proof explicitly for them on the rhs).
> 

Added linearity on the Elem proof, and removed the explicit proofs on the rhs.

> Great work!

Thanks!

berewt:
> > I had this for `deleteAt`, I don't know which one you would prefer:
> > ```
> > deleteAt FZ (x :: xs) = xs
> > deleteAt (FS n) [x] = absurd n
> > deleteAt (FS n) (x :: xs@(_::_)) = x :: deleteAt n xs
> > ```
> 
> Do you know what the benefits are of one over the other? Or is it just a stylistic difference?

I think (TBC) that if you use it at the type level, this version is sligthly better as you don't have to decompose the index value in the `FS (FS x)` case.

thomasdziedzic:
Used the @berewt version of deleteAt in favor of simpler decomposition.

ShinKage:
> Do you know what the benefits are of one over the other? Or is it just a stylistic difference?

I don't know if there are performance differences, for sure since it's a `public export`, the stylistic choices for the `deleteAt` function automatically are enforced on each property that uses `deleteAt` at the type-level, for example:

```idris
-- deleteAt with absurd case
test : {i : _} -> {ts : _} -> (hv : HVect ts) -> S (length (deleteAt i hv)) = length hv
test {i = FZ} (x :: xs) = Refl
test {i = (FS z)} {ts = [t]} (x :: []) = absurd z
test {i = (FS z)} {ts = t :: (t' :: ts)} (x :: (y :: w)) = cong S $ test {i = z} (y :: w)

-- deleteAt with double FS decomposition
test : {i : _} -> {ts : _} -> (hv : HVect ts) -> S (length (deleteAt i hv)) = length hv
test {i = FZ} (x :: y) = Refl
test {i = (FS FZ)} (x :: (y :: z)) = Refl
test {i = (FS (FS z))} (x :: (y :: w)) = cong S $ test {i = FS z} (y :: w)
```

Admittedly, is a bit of a contrived example, but you can clearly see that the use of absurd in the `deleteAt` definition is followed by the use of absurd in proving properties, also the case splitting must be the same.

thomasdziedzic:
> Admittedly, is a bit of a contrived example, but you can clearly see that the use of absurd in the deleteAt definition is followed by the use of absurd in proving properties, also the case splitting must be the same.

Thanks @ShinKage for the explanation, after that example, I think I will prefer the deleteAt version without an absurd case.

libs/contrib/Data/HVect.idr
```diff
@@ -0,0 +1,134 @@
...
+
+||| Replace an element with the correct type.
+public export
+put : t -> HVect ts -> {auto 1 p : Elem t ts} -> HVect ts
+put y (x :: xs) {p = Here} = y :: xs
+put y (x :: xs) {p = (There p')} = x :: put y xs
+
+||| Update an element with the correct type.
+public export
+update : (t -> u) -> HVect ts -> {auto 1 p : Elem t ts} -> HVect (replaceByElem ts p u)
```
gallais, line 132: I find it strange that `put` is homogeneous but `update` is not.

More generally, it seems a bit strange to me to have both:
* a `Fin`-based interface (that justifies using `Vect` as the index for `HVect`)
* an `Elem`-based interface (that would be well-suited for a `List`-based `HVect` aka `All`)

berewt: We could have homogeneous and heterogeneous for both (but... naming is hard).

`Elem` as an implicit proof is very handy for heterogeneous lists, disregarding their implementation, as it allows you to refer to a specific position for a given type.

I also use the `Fin` based access in my codebase, I don't remember exactly when (and more importantly why) though.

thomasdziedzic: Do you think implementing homogeneous & heterogeneous versions of put and update in this would be beneficial, or is the current implementation of put & update good?

libs/contrib/Data/HVect.idr
```diff
@@ -115,20 +115,56 @@ public export
...
+htPut y (x :: xs) {p = There p'} = x :: htPut y xs
+
+||| Update an element with the correct type. (Homogeneous)
+|||
+||| ```idris example
+||| > update "hello world!" [1, "string"]
+||| [1, "hello world!"]
+||| ```
+public export
+update : t -> HVect ts -> {auto 1 p : Elem t ts} -> HVect ts
```
gallais, line 157: Note that this is just `put`. `update` should take a `t -> t` function as an argument.

thomasdziedzic: Thanks for catching this oversight, fixed!

gallais:
I see that `index` is declared linear in its `HVect` argument. My guess is that
a lot more of these functions can be made linear (e.g. `(++)` in its two arguments).

(Sorry I am thinking of more & more stuff each time I take another look at the PR
because I was pinged by github's notifications :sweat_smile:)

thomasdziedzic:
Ok, added linear annotations to as many functions as I could!

PR #569, opened 2020-08-18
[ new ] topics for logging levels

gallais:
Log levels are characterised by two things:
* a dot-separated path of ever finer topics of interest e.g. scope.let
* a natural number corresponding to the verbosity level e.g. 5

If the user asks for some logs by writing

    %log scope 5

they will get all of the logs whose path starts with `scope` and whose
verbosity level is less or equal to `5`. By combining different logging
directives, users can request information about everything (with a low
level of details) and at the same time focus on a particular subsystem
they want to get a lot of information about. For instance:

    %log 1
    %log scope.let 10

will deliver basic information about the various phases the compiler goes
through and deliver a lot of information about scope-checking let binders.


gallais:
Alright that should do it. The topic names may not be the best, the levels are the
ones that were there already. Whoever is using logging to debug can go back and change
these up to finely tune the report they're getting. Hopefully we will slowly converge
towards a nice coverage of the codebase (there is no drawback to adding more logging
anymore!) and a nice set of topics.

PR #570, opened 2020-08-18
Fix returned sexp for make-lemma command result in IDE mode

abailly:
the returned structure is inline with what Emacs idris-mode expects, so this fix allows using `C-c C-e` on a hole and get it lifted as a toplevel definition


edwinb:
Hmm, this has been sitting here for ages and nobody has looked at it. Sorry about that! Is it still needed? I assume it is, but I don't use the emacs mode myself...

ohad:
This works with `main`, and I've now updated `idris2-mode` to support `make-lemma`.

Since no one seem to want to take ownership of idris2-mode, I did, merging into it (in one way or another) all the existing PRs. The new hub for it is:
http://www.github.com/idris-community/idris2-mode
PRs are welcome.
The aim is for it to always be compatible with HEAD main, and in the long-term we should resolve the previous idris-mode (maintained by @jfdm ) to support multiple versions of the protocol, including HEAD main.

PR #571, opened 2020-08-18
Implement MonadReader and related types/instances

bigs:
This includes:
- ReaderT transformer
- Instances for Functor, Applicative, Monad, MonadTrans, HasIO,
  Alternative
- asks helper function
- Reader alias


gallais:
Looks good to me! Can we maybe also get `local` as that needs to be in the interface AFAICT.

```idris
||| `local f c` runs the computation `c` in an environment modified by `f`. 
local : MonadReader r m => (r -> r) -> m a -> m a
```

bigs:
Pushed that up @gallais!

PR #573, opened 2020-08-19
Add instructions about thread support in Chez Scheme to INSTALL.md

david-christiansen:
I encountered a difficult-to-diagnose error while bootstrapping Idris 2, and these instructions fixed it for me.


thomasdziedzic:
What if we automate a check in the makefile that errors out when it can't detect threading support enabled in chez/racket?

That way, we can present an error message to the user when their version of scheme isn't compiled with threading support, otherwise, it would print nothing.

david-christiansen:
That would also be nice!

melted:
Looks good!

PR #578, opened 2020-08-19
add `extraRuntime` option for Scheme backends

lodi:
This patch adds a new code generator directive to the three Scheme backends to allow the user to textually include additional backend-specific code to the generated output:

```shell
idris2 -o hello --directive extraRuntime=code.scm --directive extraRuntime=/home/lodi/other.scm Hello.idr
# will insert the contents of code.scm and other.scm into hello.ss after existing blodwen-* definitions  
```

In my case, I need this to interface with Chez-specific functionality that doesn't exist in the other backends.


gallais:
Could you add a test case?

I guess we could also use some documentation [somewhere here](https://github.com/idris-lang/Idris2/tree/master/docs/source/backends).

lodi:
@gallais Sure, I added a test and some docs.

tests/chez/chez030/run
```diff
@@ -0,0 +1,4 @@
+$1 --directive extraRuntime=extensions.scm -o chez030 ExtraRuntime.idr
```
ShinKage, line 1: Test must be run with the additional `--no-color --console-width 0` flags, in this particular case it may not matter, but without these flags compiler output may be dependent on the particular system, due to color and automatic detection of console width. (I should really find a place where to put this indication, sorry)

lodi: No problem.  I don't think there's any output in this case but I added those options anyway just to be on the safe side.

PR #579, opened 2020-08-20
refactor Data.String.Parser

clayrat:
* fix the order of parsing in `<*>`
* add totality annotations
* add `hchainl`/`hchainr` and `parens` combinators
* misc code golfing and formatting

### DIFF
--- a/libs/contrib/Data/String/Parser.idr
+++ b/libs/contrib/Data/String/Parser.idr
@@ -18,12 +18,11 @@ record State where
     maxPos : Int
 
 Show State where
-    show s = "(" ++ show s.input ++", " ++ show s.pos ++ ", " ++ show s.maxPos ++")"
+    show s = "(" ++ show s.input ++ ", " ++ show s.pos ++ ", " ++ show s.maxPos ++ ")"
 
 ||| Result of applying a parser
 public export
 data Result a = Fail Int String | OK a State
-
 
 public export
 record ParseT (m : Type -> Type) (a : Type) where
@@ -32,7 +31,7 @@ record ParseT (m : Type -> Type) (a : Type) where
 
 public export
 Parser : Type -> Type
-Parser a = ParseT Identity a
+Parser = ParseT Identity
 
 public export
 implementation Monad m => Functor (ParseT m) where
@@ -45,9 +44,9 @@ implementation Monad m => Functor (ParseT m) where
 public export
 Monad m => Applicative (ParseT m) where
     pure x = P $ \s => pure (OK x s)
-    f <*> x = P $ \s => case !(x.runParser s) of
-                            OK r s' => case !(f.runParser s') of
-                                            OK f' fs => pure (OK (f' r) fs)
+    f <*> x = P $ \s => case !(f.runParser s) of
+                            OK f' s' => case !(x.runParser s') of
+                                            OK r rs => pure (OK (f' r) rs)
                                             Fail i err => pure (Fail i err)
                             Fail i err => pure (Fail i err)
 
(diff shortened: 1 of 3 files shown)


libs/contrib/Data/String/Parser.idr
```diff
@@ -85,81 +85,114 @@ parseT p str = do res <- p.runParser (S str 0 (strLength str))
...
+hchainr parg pop pend = go id <*> pend
+  where
+  go : (end -> end) -> Parser (end -> end)
+  go f = (do arg <- parg
+             op <- pop
+             go $ f . op arg) <|> pure f
+
+||| Succeeds if the next char satisfies the predicate `f`
+export
+satisfy : Monad m => (Char -> Bool) -> ParseT m Char
```
gallais, line 165: This ought to be `total` (even if that means you need to use an `assert`)
given that the bound checks are performed.

libs/contrib/Data/String/Parser.idr
```diff
@@ -85,81 +85,114 @@ parseT p str = do res <- p.runParser (S str 0 (strLength str))
...
+                              if s.pos+len <= s.maxPos
+                                  then let head = strSubstr s.pos len s.input in
+                                       if head == str
+                                         then OK () (S s.input (s.pos + len) s.maxPos)
+                                         else Fail (s.pos) ("string " ++ show str)
+                                  else Fail (s.pos) ("string " ++ show str)
+
+||| Succeeds if the next char is `c`
+export
+char : Monad m => Char -> ParseT m ()
```
gallais, line 193: Same here re:totality.

libs/contrib/Data/String/Parser.idr
```diff
@@ -170,6 +203,11 @@ export
 spaces : Parser ()
 spaces = skip (many space) <?> "white space"
 
+||| Discards brackets around a matching parser
+export
+parens : Parser a -> Parser a
```
gallais, line 208: Is there a reason all of these are suddenly about `Parser` rather than `ParserT`?

tests/chez/chez028/ExpressionParser.idr
```diff
@@ -11,45 +11,45 @@ import Data.String.Parser.Expression
...
 table' : OperatorTable Integer
 table' =
-  [ [ Infix (do token "*"; pure (*) ) AssocLeft
-    , Infix (do token "/"; pure (div) ) AssocLeft
+  [ [ Infix (token "*" *> pure (*)   ) AssocLeft
+    , Infix (token "/" *> pure (div) ) AssocLeft
     ]
-  , [ Infix (do token "+"; pure (+) ) AssocLeft
-    , Infix (do token "-"; pure (-) ) AssocLeft
+  , [ Infix (token "+" *> pure (+) ) AssocLeft
```
gallais, line 24: May be worth adding `(<$)` to `base` (or `prelude`, wherever `(<$>)` is defined)?
I often reach for it only to find it is missing.

tests/chez/chez028/ExpressionParser.idr
```diff
@@ -11,45 +11,45 @@ import Data.String.Parser.Expression
...
-         <?> "simple expression"
+  term' = lexeme (integer <|> expr')
+          <?> "simple expression"
 
   expr' : Parser Integer
   expr' = buildExpressionParser Integer table' term'
 
-showRes : Show a => Either String (a, Int) -> IO ()
-showRes res = case res of
+run : Show a => Parser a -> String -> IO ()
```
gallais, line 45: This `run` seems useful. Should it be moved to the library?

The fact we are happily ignoring `rem` also makes me thing that we are in the business
for and `eof : ParserT m ()` that only succeeds if we have reached the end of the file!

gallais (commented):
Looks like good upgrades. I have added a couple of suggestions; hope you don't mind.

libs/contrib/Data/String/Parser.idr
alexhumphreys: Ah, that might be from when I added #526. Copied some parsers from my test into this file, and think I was using Parser rather than ParseT

PR #580, opened 2020-08-20
Suppress banner when running with --ide-mode-socket option

ska80:
At least on Emacs this fixes the following warning:

```
Warning from Idris:    / // __  / ___/ / ___/   __/ /     Version 0.2.1-
error in process filter: open-network-stream: make client process failed: Connection refused, :name, Idris IDE support, :buffer, *idris-connection*, :host, 127.0.0.1, :service, 310, :nowait, nil, :tls-parameters, nil
error in process filter: make client process failed: Connection refused, :name, Idris IDE support, :buffer, *idris-connection*, :host, 127.0.0.1, :service, 310, :nowait, nil, :tls-parameters, nil
```


melted:
Good fix! Thanks!

PR #581, opened 2020-08-20
Make IDEMode respond after executing a `consolewidth` instruction

ohad:
emacs's idris-mode seems to hang without a response


ohad:
I plan to merge this as soon as it passes tests to avoid a whole bunch of emacs programmers' session hanging.

ShinKage:
Vim does not currently depend on IDEMode, so it should not have an effect on it.

PR #590, opened 2020-08-24
add EitherT transformer

MarcelineVQ:
add EitherT transformer
add Foldable/Traversable implementations for Either

Filling in the missing zoo as needed.

### DIFF
--- a/libs/base/Control/Monad/Either.idr
+++ b/libs/base/Control/Monad/Either.idr
@@ -1,0 +1,157 @@
+module Control.Monad.Either
+
+-------------------------------------------------
+-- The monad transformer `EitherT e m a` equips a monad with the ability to
+-- return a choice of two values.
+
+-- Sequenced actions of `Either e m a` produce a value `a` only if none of the
+-- actions in the sequence returned `e`. Because returning `e` exits the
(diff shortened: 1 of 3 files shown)


gallais:
Don't forget to add the new module to the corresponding `ipkg` file.

The build failure seems unrelated to your changes. Looks like my tests for the
logging behaviour are pretty annoying because they reference internal name. I'll
try to add a bit of `sed` magic to clean up the output & hopefully they won't
break quite as many new PRs!

MarcelineVQ:
Oh yes how silly of me considering I've done just that on other projects

clayrat:
> The build failure seems unrelated to your changes. Looks like my tests for the
> logging behaviour are pretty annoying because they reference internal name. I'll
> try to add a bit of `sed` magic to clean up the output & hopefully they won't
> break quite as many new PRs!

Yeah, I also had to manually update `basic044` for #579

gallais:
Hoping #593 will be sufficiently portable to not fail on mac OS & windows.
A more long term solution would be to have a flag telling the printer not to
display internal names. But printing is pure atm I believe so that'd be quite
a big change.

PR #594, opened 2020-08-25
Update typedd.rst

witt3rd:
Added implicit argument to exercise 5 of section 4.2

### DIFF
--- a/docs/source/typedd/typedd.rst
+++ b/docs/source/typedd/typedd.rst
@@ -66,6 +66,12 @@ For the reasons described above:
 
     tryIndex : {n : _} -> Integer -> Vect n a -> Maybe a
 
++ In exercise 5 of 4.2, add an implicit argument:
+
+.. code-block:: idris
+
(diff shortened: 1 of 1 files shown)


gallais:
The book is in my former office and I don't have access to it
so I have no idea what this function is supposed to do. Do we
really need access to `n`? It seems to me that inspecting the
`Vec n a` arguments should give us enough information about
`n` to implement the function.

Then again it is true for `tryIndex` too but there we do need
`n` because it is implemented via a possibly failing cast from
`Integer` to `Fin n` first and then a safe lookup.

witt3rd:
Without it:
```
1/1: Building ex_4_2 (ex_4_2.idr)
ex_4_2.idr:11:50--11:51:While processing right hand side of sumEntries at ex_4_2.idr:11:1--17:1:
n is not accessible in this context at:
11      sumEntries {n} pos xs ys = case integerToFin pos n of
```

gallais:
Oh I see it's the same approach of cast the integer first &
then perform safe operations. Thanks for the explanation.

PR #595, opened 2020-08-25
Eliminate schemeCall

### LINKED ISSUE #446: Eliminate schemeCall in the library
Rewriting uses of `schemeCall` to use `%foreign` in `libs` will be a great help now that we have an official non-scheme backend as well as an API for other backends.

Russoul:
A follow-up to #281 

Converts the remaining uses of `schemeCall` to `%foreign` calls
Removes `schemeCall` and `prim__schemeCall` from the library
Removes all traces of `prim__schemeCall` from the backends
Tweaks the support library as needed

Closes #446

### DIFF
--- a/libs/base/Control/Monad/Either.idr
+++ b/libs/base/Control/Monad/Either.idr
@@ -1,0 +1,157 @@
+module Control.Monad.Either
+
+-------------------------------------------------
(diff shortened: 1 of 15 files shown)


melted:
It works, ship it!

closes #446

PR #598, opened 2020-08-26
[ refactor ] suggested during SPLV 

gallais:
Main change
===========

The main change is to the type of functions dealing with an untouched
segment of the local scope. e.g.

```idris
weak : {outer, vars : _} -> (ns : List Name) ->
       tm (outer ++ inner) -> tm (outer ++ ns ++ inner)
```

Instead we now write

```idris
weak : SizeOf outer -> SizeOf ns -> tm (outer ++ inner) -> tm (outer ++ ns ++ inner)
```

meaning that we do not need the values of `outer`, `inner` and `ns`
at runtime. Instead we only demand a `SizeOf outer` and `SizeOf ns`
which are essentially a `Nat` each together with an (erased) proof
that the list in question is of that length.

Other modifications
===================

Quadratic behaviour
-------------------

A side effect of this refactor is the removal of two sources of
quadratic behaviour. They typically arise in a situation where
work is done on a scope of the form

```idris
outer ++ done ++ ns ++ inner
```

When `ns` is non-empty, some work is performed and then the variable
is moved to the pile of things we are `done` with. This leads to
recursive calls of the form `f done` -> `f (done ++ [v])` leading
to a cost quadratic in the size of `ns`.

Now that we only care about `SizeOf done`, the recursive call is
(once all the runtime irrelevant content is erased) for the form
`f n` -> `f (S n)`!

More runtime irrelevance
------------------------

In some places we used to rely on a list of names `vars` being
available. However once we only care about the length of `vars`,
the fact it is not available is not a limitation.

For instance a `SizeOf vars` can be reconstructed from an environment
assigning values to `vars` even if `vars` is irrelevant. Indeed the
size of the environment is the same as that of `vars`.

Consequences
============

My non-scientific benchmarking gives me a ~9% improvement on the test
suite (before: 2m45s, after: 2m30s). I haven't tried to look at allocation
although this was one of the questions given that this replaces a `List`
with a `Nat`.

The run on the test suite is much less impressive: 1m41s for master,
1m40s for the refactoring. Not sure that 1% is significant.

### DIFF
--- a/src/Core/Coverage.idr
+++ b/src/Core/Coverage.idr
@@ -168,7 +168,7 @@ mkAlt : {vars : _} ->
         FC -> CaseTree vars -> (Name, Int, Nat) -> CaseAlt vars
 mkAlt fc sc (cn, t, ar)
     = ConCase cn t (map (MN "m") (take ar [0..]))
-              (weakenNs _ (emptyRHS fc sc))
+              (weakenNs (map take) (emptyRHS fc sc))
 
 altMatch : CaseAlt vars -> CaseAlt vars -> Bool
 altMatch _ (DefaultCase _) = True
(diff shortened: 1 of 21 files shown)


src/Core/Coverage.idr
```diff
@@ -323,17 +321,19 @@ buildArgs fc defs known not ps cs@(Case {name = var} idx el ty altsIn)
     buildArgAlt : KnownVars vars (List Int) ->
                   CaseAlt vars -> Core (List (List ClosedTerm))
     buildArgAlt not' (ConCase n t args sc)
-        = do let con = Ref fc (DataCon t (length args)) n
+        = do let l = mkSizeOf args
+             let con = Ref fc (DataCon t (length args)) n
```
gallais, line 325: Note to self: this `length args` could be taken from `l` directly.

src/Core/TT.idr
```diff
@@ -555,13 +655,14 @@ eqTerm _ _ = False
...
+  weakenNs : SizeOf ns -> tm vars -> tm (ns ++ vars)
 
-  weakenNs [] t = t
-  weakenNs (n :: ns) t = weaken (weakenNs ns t)
+  weakenNs p t = case sizedView p of
+    Z   => t
+    S p => weaken (weakenNs p t)
 
-  weaken = weakenNs [_]
+  weaken = weakenNs (MkSizeOf (S Z) (S Z))
```
gallais, line 665: Note to self: `suc zero`

PR #603, opened 2020-08-27
Add `dup`

Russoul:
I've noticed myself using them a lot, so hopefully it makes sense to add them to the library

### DIFF
--- a/libs/prelude/Prelude/Basics.idr
+++ b/libs/prelude/Prelude/Basics.idr
@@ -25,6 +25,11 @@ id : (1 x : a) -> a           -- Hopefully linearity annotation won't
                               -- break equality proofs involving id
 id x = x
 
+||| Function that duplicates its input
+public export
+dup : a -> (a, a)
+dup x = (x, x)
+
 ||| Constant function.  Ignores its second argument.
 public export %inline
 const : a -> b -> a


libs/base/Data/Maybe.idr
```diff
@@ -60,3 +60,10 @@ lowerMaybe (Just x) = x
 export
 raiseToMaybe : (Monoid a, Eq a) => a -> Maybe a
 raiseToMaybe x = if x == neutral then Nothing else Just x
+
+||| Unwraps the first argument in case of `Just`
+||| returns its second argument otherwise
+export
+orElse : Maybe a -> Lazy a -> a
```
gallais, line 67: This is `flip fromMaybe`. I am not sure we should add such variants of existing functions. :/

libs/prelude/Prelude/Basics.idr
```diff
@@ -25,6 +25,11 @@ id : (1 x : a) -> a           -- Hopefully linearity annotation won't
                               -- break equality proofs involving id
 id x = x
 
+||| Function that duplicates its input
+public export %inline
+dup : a -> (a, a)
```
gallais, line 30: Given this is duplicating its argument, I am not sure we should inline it aggressively.

libs/base/Data/Maybe.idr
Russoul: Oops, it is. I absolutely agree. What about dup ?

gallais: `dup` is a welcomed addition!

Russoul:
@gallais sorry for the mess, looks like someone is going to spend the weekend learning how to use git :)

libs/prelude/Prelude/Basics.idr
Russoul: removed %inline

gallais:
No problem.

Don't hesitate to push multiple patches to a PR even if that leads to a messy history:
when merging we can easily squash everything into a single clean commit that corresponds
to the resulting diff.

Russoul:
Got it !

Issue #553, opened 2020-08-12
Instantiating records using field names.


thomasdziedzic:
The way we currently create a record and instantiate it:

```idris
record Dog where
  constructor MkDog
  name : String
  age : Nat

myDog : Dog
myDog = MkDog "Idris" 2
```

The way I would like to instantiate it:

```idris
myDog : Dog
myDog = MkDog { name = "Idris", age = 2 }
```

The reason is for records with a large amount of fields, it makes it clearer what fields take on what values, rather than relying on the position of the arguments.

gallais:
One problem I can see with this approach is the interaction with implicit fields:

```idris
record Dog where
  constructor MkDog
  {name : String}
  age : Nat

myDog : Dog
myDog = MkDog {name = "Idris"} 2
```

The definition of `myDog` looks like it could be either:

* the usual constructor + implicit name + missing field
* the new constructor + assignment (with a missing field) applied to `2`

We could maybe do something like the following?

```idris2
record MkDog { ... }
```

gallais:
I believe this was fixed.

Issue #61, opened 2020-05-20
feature request: constructor only/wildcard pattern matches


edwinb:
<a href="https://github.com/MarcelineVQ"><img src="https://avatars3.githubusercontent.com/u/16345659?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [MarcelineVQ](https://github.com/MarcelineVQ)**
_Saturday Apr 25, 2020 at 00:09 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/320_

----

It would be useful to have a way, in patterns, to check for the presence of a constructor without regard to its arguments.
Currently given
```idris
data Tree : Type -> Type where
  Tip : Tree a
  Node : (v : a) -> (l : Tree a) -> (r : Tree a) -> Tree a
```
in a pattern match in Idris where we don't use the values in Node we would write `(Node _ _ _)` to match the Node constructor.
In Haskell one can do the same, but can also write `Node{}`.
`(Node _ _ _)` and `Node{}` are equivalent but the latter more strongly expresses a complete disinterest for any information other than that we have a Node.

This is particularly useful for places where you just want to satisfy some NonEmpty property for your value. For example given
```idris
data NonEmpty : Tree a -> Type where
  IsNonEmpty : NonEmpty (Node v l r)

maxView : (t : Tree a) -> {auto ok : NonEmpty t} -> (a, Tree a)
maxView (Node v l Tip) = (v,l)
maxView (Node v l r@(Node _ _ _)) =
  let (x,r') = maxView r in (x, Node v l r')
```
Currently if we want idris to infer a NonEmpty where we need it we might write
```idris
case t of
  Tip => ?foo
  n@(Node _ _ _) => maxView n
```
compared to
```idris
case t of
  Tip => ?foo
  n@Node{} => maxView n
```
and even maxView could make use of this since it also only needs to know that it has a Node, not what the Node contains
```idris
maxView (Node v l r@Node{}) =
```
It doesn't seem like a big addition but these savings add up due to making for more direct code, you are stating just one time to the reader and compiler "I don't make use of Node's values here" instead of stating it once for each field of Node.

edwin mentions on irc:
`<edwinb> I thought about that [this feature] a while ago, and the reason I didn't is that, in Haskell, it's consistent with the record syntax, but in Idris there's nothing for it to be consistent with`
`<edwinb> it might still be a reasonable special case. It is useful.`

[...]

PR #607, opened 2020-08-28
New syntax for named applications and "record" updates

### LINKED ISSUE #725: Senseless implicits in applications
# Steps to Reproduce
```idris
test : {auto x : String} -> String
test @{_} @{_} @{_} @{_} {blabla = _} = "actually not X"

test' : String
test' = test {x = "not used"} @{_} @{_} {blabla22 = _} 
```
# Expected Behavior
Idris should output an error telling that all `@{_}` (except the first one in `test`), `blabla`, `blabla22` are not valid implicits
# Observed Behavior
Typechecks without holes

Russoul:
This implementation proposal solves #553 

Implementation utilises syntax suggested by @gallais in this comment: https://github.com/idris-lang/Idris2/issues/553#issuecomment-674023749

`record <Constructor> {<field1> = <expr1>, <field2> = <expr2>, ..., <fieldN> = <exprN>}`,
where `{field}` is a complete sequence of **unordered** explicit fields of `<Constructor>`

Implementation details:

- Resolves constructor name ambiguities by looking at unique field names of the candidates

- After disambiguating a name, elaborates further as `<Constructor> <expr1'> <expr2'> ... <exprN'>`, where
   `{expr'}` is an ordered sequence of explicit arguments of the record

### Update:
Now the project follows the developments and ideas of #642  and #626.

**A brief summary of changes:**

- Currently general `{...}` is implemented.
- Syntax for `record updates` is renewed.
- Bind-all-explicits pattern is implemented #61
- Unnamed arguments always take priority over named.
- Interface constraints are now elaborated as auto implicits of its constructor,
previously as explicits.
- Also this PR fixes #725.


src/Idris/Parser.idr
```diff
@@ -642,6 +643,22 @@ mutual
...
+      parseFieldName True  = map fieldName <$> [| name :: many recFieldCompat |]
+      parseFieldName False = map fieldName <$> [| name :: pure [] |]
+
+  recordInstance : FileName -> IndentInfo -> Rule PTerm
+  recordInstance fname indents
+      = do b <- bounds (do keyword "record"
+                           n <- name
+                           symbol "{"
+                           commit
+                           fs <- sepBy1 (symbol ",") (field fname indents False)
```
gallais, line 656: I expect we can have record with no fields. Or only implicit fields that
can be inferred. So this should probably be a `sepBy` rather than `sepBy1`.
I have extended the test case with an example that ought to work.

src/TTImp/Elab/Record.idr
```diff
@@ -233,3 +234,84 @@ checkUpdate rig elabinfo nest env fc upds rec expected
...
+                Env Term vars ->
+                FC -> Name -> List IFieldUpdate ->
+                Core RawImp
+elabInstance env fc n fs
+    = do
+         defs <- get Ctxt
+         names@(_ :: _) <- getConNames defs n
+              | _ => throw (NotRecordType fc n)
+         names@(_ :: _) <- mapMaybe (sequenceM . bimap pure id)
+                              <$> (sequence $ map (sequenceCore . bimap pure (findFieldsExplicit defs) . dup) names)
```
gallais, line 253: This is really hard to understand. Can we not have something simpler using
`flip traverse names $ \ n => do (...)`?

src/TTImp/Elab/Record.idr
```diff
@@ -233,3 +234,84 @@ checkUpdate rig elabinfo nest env fc upds rec expected
...
+              | _ => throw (GenericMsg fc "No constructor satisfies provided fields.")
+         providedfields <- sequence $ map getName fs
```
gallais, line 255: `sequence . map` is `traverse`. The same pattern is used multiple times below.

ohad:
Sorry for jumping on board beyond the last minute, but is it a good idea to allow creation of records that do _not_ have a named constructor?

Perhaps replacing `<Constructor>` with the record type constructor, or adding additional (optional) syntax at the end like:
`record _ { field1 = val1 .... fieldn = valn} : <RecordType> arg1 ... argn`.

src/Idris/Parser.idr
Russoul: One issue here. 

- `elabInstance` disambiguates constructor names by looking at the fields of records of those constructors (if any). It does not look into the expected type. And it does not employ idris's disambiguation process (which I currently don't know much about)
- And it looks like there is no way (at least which I know of) to distinguish between a record constructor and a data constructor via context lookups.

This means in your added test case (`Unit`) it won't be able to disambiguate the constructor from `MkUnit` declared in the `Prelude`

Do we want `elabInstance` to not throw and postpone name resolution in case it's not able to resolve it with provided fields info ?

Russoul:
I also found something strange (tested on idris-lang/Idris2/master branch):
```
record Unit where
  constructor MkUnit
  {impl : String}

unit : Fld.Unit
unit = Fld.MkUnit {impl = "str"}

```

Throws:

```
Error: While processing type of impl. Ambiguous elaboration. Possible results:
    Fld.Unit
    Builtin.Unit

Fld.idr:51:1--55:5
 51 | record Unit where
 52 |   constructor MkUnit
 53 |   {impl : String}
 54 |
 55 | unit : Fld.Unit
```

When I remove the implicit argument it typechecks fine

src/Idris/Parser.idr
gallais: Type-directed disambiguation of record constructor works at the moment
so I think it should also work for this new syntax.

```idris
namespace R1

  public export
  record R1 where
    constructor MkR

namespace R2

  public export
  record R2 where
    constructor MkR

r1 : R1
r1 = MkR

r2 : R2
r2 = MkR
```

Russoul:
Also:
```
flip traverse fullnames $ 
        \(fullname, allfields) => do
               seqexp <- flip traverse allfields \x => lookupOrElse x fs' (throw (NotCoveredField fc x))
               pure $ apply (IVar fc fullname) seqexp
```
Throws:

```

Error: While processing right hand side of elabInstance. Ambiguous elaboration. Possible results:
    ?postpone [locals in scope: vars, fs, providedName, fc, env, e, u, m, c, defs, {_:2574}, {_:2573}, {_cn:2}, {names:1}, {_:2663}, {_:2662}, {_:0}, names, providedfields, fullnames, fs', fullname, allfields, {lamc:0}]
    ?postpone [locals in scope: vars, fs, providedName, fc, env, e, u, m, c, defs, {_:2574}, {_:2573}, {_cn:2}, {names:1}, {_:2663}, {_:2662}, {_:0}, names, providedfields, fullnames, fs', fullname, allfields, {lamc:0}]

src/TTImp/Elab/Record.idr:263:16--263:105
     |
 263 |                seqexp <- flip traverse allfields \x => lookupOrElse x fs' (throw (NotCoveredField fc x))
     |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Those two `?postpone`s look the same to me(but maybe that's irrelevant)

Without the `flip`:
```
traverse (\(fullname, allfields) => do
      seqexp <- flip traverse allfields \x => lookupOrElse x fs' (throw (NotCoveredField fc x))
      pure $ apply (IVar fc fullname) seqexp) fullnames
```
Typechecks

Just bothers me that I can't go with the first as it's cleaner

ShinKage:
Instead of using `flip traverse` (flip is problematic since is not dependent), mabe you can add a specialized `for` where `traverse` is declared for `Core`.

Russoul:
@ShinKage Good point !

Russoul:
> Sorry for jumping on board beyond the last minute, but is it a good idea to allow creation of records that do _not_ have a named constructor?
> 
> Perhaps replacing `<Constructor>` with the record type constructor, or adding additional (optional) syntax at the end like:
> `record _ { field1 = val1 .... fieldn = valn} : <RecordType> arg1 ... argn`.

It’s pretty powerful:
```
showMaybe : Show a => Show (Maybe a)
showMaybe = record _ {show = \mb => fromMaybe “Nothing” ((“Just ” ++) . show <$> mb)}
```
One could make local instances of interfaces 

I’ll try to add this feature. Specifying a constructor instead of a type constructor feels a bit more natural though, doesn’t it ? Or not ? But an implementation would undoubtedly be simpler with the later. Which way should we go ?

gallais:
I quite like the type constructor suggestion.

Russoul:
Implements @ohad's proposal. Please see the test file for more information. Still has unfinished parts, but mostly done. Just wanted to get some feedback.

Russoul:
Summary of additions/changes (excluding the original version of the PR):

- Interface constraints are elaborated as auto implicit arguments of a synthesised data(record) constructor(previously as explicit arguments with machine generated names thus inaccessible)
- Constructor name can be left blank and inferred by the typechecker
- It is possible to construct instances of interfaces (even without constructors) locally. This is a consequence of the first two points

ohad:
Wow! Thanks @Russoul !

1. The seemingly added expressiveness for defining interfaces is not actually more expressive. If you're happy with boilerplate, you can add constructors even to interfaces that haven't been defined with constructors with the existing interface mechanisms. E.g.:
```idris
interface Foo (a : Nat) where
  Baz : Nat
  
[MkFoo] {a : Nat} -> {baz : Nat} -> Foo a where
  Baz = baz

gnat : Foo 1
gnat = MkFoo {a = 1} {baz = 3}
```
lets you then run:
```
λΠ> Baz @{gnat}
3
```
For larger interfaces you can repeat the same game (with more code): 
a) add a named implementation that takes the implicit arguments all the parameters of the interface and its methods
b) call this named implementation whenever you need to dynamically construct an interface.
So the new syntax doesn't make interfaces more expressive than they are in terms of letting you construct interfaces on the fly.

I think it _does_ add expressiveness to records, though I don't have a proof. If that's true, it's weird that interfaces are _more_ expressive than records, considering that under the hood, interfaces elaborate to records :D.

2. I (weakly) vote against the `new {...}` syntax in the proposal: `new` is often an impure/effectful operation involving memory allocation. But I'm not going to argue more than this sentence about concrete syntax.

3. I agree that if type ascription (`the RecordType $ record _ {...}`) is enough for the type-checker, then it's better to find a more succinct syntax (i.e., without `_`).

gallais:
What's the status on this? We had made great progress and then stalled?
Is there any way we could get this wrapped up and consider the discussion
on syntactic sugar in #626 a stretch goal that can come later?

Russoul:
@gallais That RFC is what I am actually waiting for :) I'd be happy to continue.
The thing is, while thinking about the syntax, I came up with this: #642. It generalises the problem to function applications. The further course of development is, therefore, determined by our decision.

gallais:
I would rather have small, self-contained additions than "let's add syntax for
constructing record *and* also revolutionise the way we do function application".

Russoul:
I have an almost working version of the latest proposal (`{..., ...}` only). I'll push a new draft PR, so we can play and see if we like it. And I'll leave this PR untouched, so we can fall back to the self-contained modification if things go south ...

Russoul:
Basically it's fully functional, I'll do some code cleanup later.

All explicit arguments in named form have lower precedence than unnamed explicits. This way we don't have to introduce too many changes to Idris internals. Current implementation is not invasive.

`{...}` now allows any mixture of arguments of any type (explicit, implicit, auto)
`{x, y, _}` means bind `x`, `y` and any extra explicit arguments in any quantity, see #61.
`{}` means bind any extra explicit arguments in any quantity.
`{x := t, y $= t}` - new syntax for record updates (the old one still usable)

Russoul:
@gallais I'll do another round of checking. Probably add a bunch more tests. Then ask for review.

Russoul:
Looks like we found a bug in `PreorderReasoning` (test fails on in)

```
public export
data FastDerivation : (x : a) -> (y : b) -> Type where
  (|~) : (x : a) -> FastDerivation x x
  (~~) : FastDerivation x y -> (step : (z : c ** y ~=~ z)) -> FastDerivation x (fst step)
  
public export 
Calc : {x : a} -> {y : b} -> FastDerivation x y -> x ~=~ y
Calc (|~ x) = Refl
Calc ((~~) {z=_} {y=_} der (_ ** Refl)) = Calc der
```

`z` is indeed not a valid argument in `(~~)`

Russoul:
I updated the head comment of this PR to reflect all current changes and additions

edwinb:
This looks really nice. I'll take a closer look (just to make sure I understand it) but if the conflicts are fixed is this basically ready for merge now?

Russoul:
Yes :)

edwinb:
Great, thanks, looks good. There's two more things that need doing, but I will merge this as is and save them for later. They are:

- the documentation needs to be consistent with the new syntax (it's currently correct, but not complete)
- your name should be in the CONTRIBUTORS file!

I particularly like that this works well with the record update syntax. I'd still like a way of constructing records without a named constructor, but that can be resolved elsewhere. Maybe something like `record { foo = 1, bar = 2 }`

Russoul:
Will do ! Thanks @edwinb !
And thanks to everyone who participated in the discussions !

PR #609, opened 2020-08-28
inline dup

foss-mc:


gallais:
One thing that is not currently inlined and I don't understand why is `($)`.
As for `dup`, the fact it duplicates its argument is a strong argument against inlining.

PR #612, opened 2020-08-28
Add missing `Data.Stream.Extra` module to contrib.ipkg

foss-mc:
I forgot to add it


gallais:
I forget about this like half the time. :weary:

PR #617, opened 2020-08-30
[ fix #616 ] Namespaces are stored in reverse order

### LINKED ISSUE #616: Qualifying do with a namespace only typechecks when reversed
https://gist.github.com/redfish64/b5bbcea34a591474e3cf4b2c681a0078

```
namespace A
  namespace B
    export
    (>>=) : Nat -> (() -> Nat) -> Nat
    (>>=) x fy = x + (fy ())

test : Nat
test = B.A.do
         5
         6
         7
         
test2 : Nat
test2 = A.B.do
         5
         6
         7
```
# Steps to Reproduce

Typecheck

# Expected Behavior

test shouldn't typecheck, and test2 should.

# Observed Behavior

test typechecks, and test2 doesn't.

```
- + Errors (1)
 `-- SillyReverseBug.idr line 15 col 9:
     While
     processing
     right
     hand
     side
     of test2. Undefined
     name B.A.>>=. 
     
     SillyReverseBug.idr:15:10--15:11
         |
      15 |          5
         |          ^
     

```

Tested against d56b090c4c3bc5df91d769e5519b0f7b42a84245

Russoul:
Quickfix, closes #616

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -662,7 +662,7 @@ mutual
                 ("do" :: ns) =>
                    do actions <- bounds (block (doAct fname))
                       pure (PDoBlock (boundToFC fname (mergeBounds nsdo actions))
-                                     (Just (reverse ns)) (concat actions.val))
+                                     (Just ns) (concat actions.val))
                 _ => fail "Not a namespaced 'do'"
 
   lowerFirst : String -> Bool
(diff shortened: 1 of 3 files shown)


gallais:
Right, if we are to trust the `Show` instance in `Core Name`
then `NS` stores its list of namespaces in reverse order.
Given that `namespacedIdent` returns a reversed list already
there is no need to reverse it again.

PR #618, opened 2020-08-30
Preserve foreign declaration types in `Compiler.ES.ImperativeAst`

bamboo:
So backends based on it can expose a simpler FFI.


gallais:
Thanks!

PR #619, opened 2020-08-30
add binary literals

MarcelineVQ:
Written via "0b" in the manner of other literals. e.g. 0b111001 = 57
useful precursor to other bits things in the works


gallais:
Thanks!

PR #631, opened 2020-09-02
[ refactor ] Introducing Namespace and ModuleIdent

gallais:
Until now namespaces were stored as (reversed) lists of strings.
It led to:

* confusing code where we work on the representation rather than say
  what we mean (using `isSuffixOf` to mean `isParentOf`)

* potentially introducing errors by not respecting the invariant cf.
  bug report #616 (but also name generation in the scheme backend
  although that did not lead to bugs as it was self-consistent AFAICT)

* ad-hoc code to circumvent overlapping interface implementation when
  showing / pretty-printing namespaces

This PR introduces a Namespace newtype containing non-empty lists of
strings. Nested namespaces are still stored in reverse order but the
exposed interface aims to support programming by saying what we mean
(`isParentOf`, `isApproximationOf`, `X <.> Y` computes to `X.Y`, etc.)
irrespective of the underlying representation.

Edit: I should add that one inconvenient aspect of this is that we
cannot pattern-match on namespaces on the LHS anymore. Having guards
in the language would help mitigate this problem! In case anyone is
looking for a project. ;)


gallais:
That ought to do it. :]

Russoul:
> I should add that one inconvenient aspect of this is that we cannot pattern-match on namespaces on the LHS anymore. Having guards in the language would help mitigate this problem

But we have `with` clauses, they are verbose, but we could fix that when clauses are not dependent.
Or you mean something else ?

gallais:
> they are verbose

Yes. And also with guards you can write:

```idris
f x | p x = ?
    | q x = ?
    | r x = ?
    | otherwise = ?
```

but using `with` you would get 3 levels of nesting!

src/TTImp/WithClause.idr
```diff
@@ -20,7 +20,7 @@ mutual
   getMatch lhs (Implicit _ _) tm = pure []
 
   getMatch lhs (IVar _ (NS ns n)) (IVar loc (NS ns' n'))
-      = if n == n' && isSuffixOf ns' ns then pure [] else matchFail loc
+      = if n == n' && isApproximationOf ns ns' then pure [] else matchFail loc
```
gallais, line 23: This looks like a bug: should be `isParentOf ns' ns`

PR #638, opened 2020-09-04
[ refactor ] introducing Namespace

gallais:
Restarting the PR from scratch after reverting the merged commit.
Hopefully this will be easier to diagnose with the compound changes
visible in one place.

====================================================================

Until now namespaces were stored as (reversed) lists of strings.
It led to:

* confusing code where we work on the underlying representation of
  namespaces rather than say what we mean (using `isSuffixOf` to mean
  `isParentOf`)

* potentially introducing errors by not respecting the invariant cf.
  bug report #616 (but also name generation in the scheme backend
  although that did not lead to bugs as it was self-consistent AFAICT)

* ad-hoc code to circumvent overlapping interface implementation when
  showing / pretty-printing namespaces

This PR introduces a `Namespace` newtype containing a list of strings.
Nested namespaces are still stored in reverse order but the exposed
interface aims to support programming by saying what we mean
(`isParentOf`, `isApproximationOf`, `X <.> Y` computes to `X.Y`, etc.)
irrespective of the underlying representation.


src/Compiler/CompileExpr.idr
```diff
@@ -135,38 +135,73 @@ mkDropSubst i es rest (x :: xs)
...
-natHack (CCon fc (NS ["Types", "Prelude"] (UN "Z")) _ []) = CPrimVal fc (BI 0)
-natHack (CCon fc (NS ["Types", "Prelude"] (UN "S")) _ [k])
-    = CApp fc (CRef fc (UN "prim__add_Integer")) [CPrimVal fc (BI 1), k]
-natHack (CApp fc (CRef _ (NS ["Types", "Prelude"] (UN "natToInteger"))) [k]) = k
-natHack (CApp fc (CRef _ (NS ["Types", "Prelude"] (UN "integerToNat"))) [k]) = k
-natHack (CApp fc (CRef fc' (NS ["Types", "Prelude"] (UN "plus"))) args)
-    = CApp fc (CRef fc' (UN "prim__add_Integer")) args
-natHack (CApp fc (CRef fc' (NS ["Types", "Prelude"] (UN "mult"))) args)
-    = CApp fc (CRef fc' (UN "prim__mul_Integer")) args
-natHack (CApp fc (CRef fc' (NS ["Nat", "Data"] (UN "minus"))) args)
```
gallais, line 148: HUM. This one is in `Data.Nat`!

edwinb: Oh my...

We really need to find a better way of doing this, for this and other reasons. I suppose that is why I chose the name `natHack` but still, the magic strings aren't good.

Pattern synonyms would help. They're high on my list of things to do, but term is starting soon...

src/Compiler/CompileExpr.idr
```diff
@@ -135,38 +135,73 @@ mkDropSubst i es rest (x :: xs)
...
-    = CApp fc (CRef fc' (UN "prim__sub_Integer")) args
-natHack (CLam fc x exp) = CLam fc x (natHack exp)
-natHack t = t
+natHack = magic
+    [ MagicCCon typesNS "Z" 0
+         (\ fc, [] => CPrimVal fc (BI 0))
+    , MagicCCon typesNS "S" 1
+         (\ fc, [k] => CApp fc (CRef fc (UN "prim__add_Integer")) [CPrimVal fc (BI 1), k])
+    , MagicCRef typesNS "natToINteger" 1
```
gallais, line 174: Hmm. This **IN**teger looks suspicious

Issue #597, opened 2020-08-26
Injectivity proofs for constructors should not use their arguments


gallais:
I got bitten by this today, I had to edit `base/Data/Nat.idr` like so:

```diff
- succInjective : (left, right : Nat) -> S left = S right -> left = right
+ succInjective : (0 left, right : Nat) -> S left = S right -> left = right
```

because I wanted to peel off an `S` constructor in an equality proof about terms
that will not be here at runtime anymore. When we are proving that a constructor
is injective, we never actually need to have our hands on the constructor's
arguments.
	
It would be nice to fix this issue everywhere in the libs.

shaylew:
Would it be sensible to have a (disabled by default?) warning for arguments that are unrestricted in the signature but only used 0 or 1 time in the definition? I'm happy to comb the libs for this (I'm fresh out of SPLV and looking for starter issues), but also it looks like the kind of thing that'll crop up again and again without tooling assistance.

shaylew:
A couple of questions about the conventions for injectivity proofs:

- Almost all injectivity proofs in the libs take their left and right sides implicitly; `succInjective` seems to be an outlier. (There are still some that take implicit arguments at nonzero multiplicity, though.) Is there some rhyme or reason to this?
- Most injectivity proofs take their *proof* argument at unrestricted multiplicity... but a couple (`Data.HVect.consInjective1`) only ask for `1` of it. Is there a right answer for which to use?

I'm not sure what it even means to distinguish `0 _ : a = b`, `1 _ : a = b`, and `ω _ : a = b`. You can pattern match on `Refl` at multiplicity `0`, but e.g. `cong` takes its proof at `1` anyway. And it looks like -- because there's no subtyping for multiplicity 0 -- you stop being able to use ordinary `.` composition for injectivity proofs whose equality argument is at multiplicity `0`, which is quite painful. Is `ω` the right answer for now?

gallais:
> Is there some rhyme or reason to this?

I don't think there is. The libraries could use being a bit more consistent in their
naming schemes & the multiplicities being used. One problem is of course that it's
going to break people's code and we do try to avoid that in `base` in particular.

> Is there a right answer for which to use?

Equality proofs should not exist at runtime anyway so I think that being able to
insist that the input proof has quantity `0` is not that important here. Of course
some user could insist in their library that they want a `(0 _ : a = b)` and then
be annoyed that the ones we provide are not compatible... :/

shaylew:
[...]

PR #644, opened 2020-09-06
[re #597] make constructor injectivity proofs use 0 multiplicity

shaylew:
[ #597 ] Constructor injectivity proofs don't have to match on their arguments, so they might as well take them at quantity 0. This PR cleans up a couple of places in the libraries where an injectivity proof named its arguments explicitly but left them at multiplicity `ω`.


gallais:
Thanks!

Issue #625, opened 2020-08-31
Reflection out of sync with language


michaelmesser:
https://github.com/idris-lang/Idris2/blob/b449e5ae8a25fbf7d558528038b885210c0572c8/src/TTImp/TTImp.idr#L198
https://github.com/idris-lang/Idris2/blob/5487290499e59e45c3af0290848a6286231bc501/libs/base/Language/Reflection/TTImp.idr#L100

This causes reflection to fail when using `ForeignFn`.

gallais:
fixed by #646

PR #646, opened 2020-09-07
Fix how reify treats ForeignFn

MarcelineVQ:
Addresses https://github.com/idris-lang/Idris2/issues/625
I wasn't sure which the right choice was so you can cherrypick dd128f2 to address this by changing `ForeignFn : List String -> FnOpt` to `ForeignFn : List TTImp -> FnOpt` to match the RawImp, or the PR as a whole can be taken to fix this issue while keeping `libs/base/Language/Reflection/TTImp.idr` as `ForeignFn : List String -> FnOpt`

### DIFF
--- a/libs/base/Language/Reflection/TTImp.idr
+++ b/libs/base/Language/Reflection/TTImp.idr
@@ -97,7 +97,7 @@ mutual
        GlobalHint : Bool -> FnOpt
        ExternFn : FnOpt
        -- Defined externally, list calling conventions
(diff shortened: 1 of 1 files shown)


gallais:
I'd be in favour of the more structured solution with the `List TTImp` type.

MarcelineVQ:
Right on, works for me, on the unlikely chance anyone is already counting on ForeignFn taking `List String` they can simply `traverse quote` their `List String`

PR #648, opened 2020-09-07
Optional where

### LINKED ISSUE #620: Make "where" clause optional in empty interface declarations
In Idris 1 an "empty" interface has an optional "where" clause, i.e.

```idris
interface Something where
```
is equivalent to 

```idris
interface Something
```

In Idris 2 the "where" clause is always required for empty interfaces. Could the "where" be made optional on empty interfaces so we don't (visually) have a dangling where clause?

Russoul:
Make `where` keyword optional in empty `interface` and `record` declarations
Remove trailing `where` in `Void` declaration (Builtin.idr)

Closes #620

### DIFF
--- a/libs/prelude/Builtin.idr
+++ b/libs/prelude/Builtin.idr
@@ -104,7 +104,7 @@ namespace DPair
 ||| Use `void` or `absurd` to prove anything if you have a variable of type
 ||| `Void` in scope.
 public export
-data Void : Type where
+data Void : Type
 
 -- Equality
 
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1250,6 +1250,7 @@ implDecl fname indents
                                              (map (collectDefs . concat) body)))
          atEnd indents
          pure (b.val (boundToFC fname b))
+
 fieldDecl : FileName -> IndentInfo -> Rule (List PField)
 fieldDecl fname indents
       = do doc <- option "" documentation
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -78,6 +78,12 @@ whereBlock fname col
          ds <- blockAfter col (topDecl fname)
          pure (collectDefs (concat ds))
 
+ctor : IndentInfo -> Rule Name
+ctor idt = do exactIdent "constructor"
+              n <- name
+              atEnd idt
+              pure n
+
 -- Expect a keyword, but if we get anything else it's a fatal error
 commitKeyword : IndentInfo -> String -> Rule ()
 commitKeyword indents req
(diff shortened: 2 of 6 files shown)


gallais:
What kind of error message do you get if you forget to add `where`
in a normal data / interface / record declaration or an implementation?

Russoul:
Full file:
```
interface A a
  method : a

```
Error:
```
     ____    __     _         ___                                           
    / // __  / ___/ / ___/   __/ /     Version 0.2.1-b5c1f3c60

1/1: Building Test2 (Test2.idr)
Error: Parse error at line 2:10:
Expected end of input (next tokens: [symbol :, identifier a, end of input])
Main> Bye for now!

```
Idris thinks `method` is a parameter of `A`, then it encounters `:` and fails.
The same error I get for record declaration with missing `where`

Full file:
```
data A
   Constructor : A
```
Error:
```
     ____    __     _         ___                                           
    / // __  / ___/ / ___/   __/ /     Version 0.2.1-b5c1f3c60

1/1: Building Test2 (Test2.idr)
Error: Parse error at line 1:1:
Couldn't parse declaration (next
tokens: [data, identifier A, identifier Constructor, symbol :, identifier A, end of input])
Main> Bye for now!
```

Russoul:
In case of an implementation: if a top level term starts with the `implementation` keyword I get a message similar to what I'd get from `data`. In case the `implementation` keyword is omitted: I get a result similar to the `interface` case.

Russoul:
Edit: forgot ` : Type` in data type declaration in one of the examples above. Fixed. Message doesn't change much

MarcelineVQ:
re: Void
A datatype without a where is a forward declaration. Letting you write code like
```idris
data Biz : Type

data Fing : Type where
  Fom : Biz -> Fing

data Biz : Type where
  Foo : Fing -> Biz
```
We have the claim that Biz exists first, so that code that follows it can refer to it, and then the definition of Biz later on where it's able to make use of the things defined before it.
It's an alternative to mutual blocks, which (I'm guessing) first pass over the block to collect all the claims and then pass again for the definitions. All of that is to say that Void without `where` is just a claim that some type called Void exists, and not a definition of Void existing and being empty.
Is that different with your parser changes, and if it is does that impact anything about forward declaration?

Russoul:
Ah, that is unfortunate. I've done nothing to how `data` is parsed. And I have no idea why all checks have passed with the forward declared `Void`. Maybe it is not used in Idris and the libraries. Very sad. Also due to the presence of the trailing `where` in an empty data definition to mark it as complete, It would probably be counterintuitive to allow omitting `where` in empty records or interfaces.

Russoul:
That said, maybe it would be right to leave the syntax as is.
@MarcelineVQ Thanks for your comment

Issue #654, opened 2020-09-09
Partial functions do not allow non-covering functions in `with` blocks with %default total


ohad:
# Steps to Reproduce
```idris
%default total

partial
foo : Nat -> Nat
foo Z = Z

partial
bar : Nat -> Nat
bar n with (0)
 bar n | k = foo n
```

# Expected Behavior
Type-check fine.

# Observed Behavior
```
with block in bar is not covering.
     
     :9:1--12:1
      09 | bar n with (0)
      10 |  bar n | k = foo n
      11 | 
     
     Calls non covering function Main.foo
```

ohad:
Fixed in 0600a9ba113a7f8a50496cec0ea63aea69c06ffb

PR #656, opened 2020-09-09
Fix bug #654

ohad:
Auxiliary functions introduced in elaboration (e.g., through case splits and with clauses) now
have the same totality annotation as the function they're defined in.

Moved auxiliary function `findSetTotal` into `Context.idr` since it's
now used by `ProcessDef.idr` too.

Added a totality requirement argument to `checkClause` so that the
with-clause case could propagate it to the functions it generates in
elaboration.

Dyanmically rebind the default totality annotation in rhs elaboration 
in pattern matches to the current function's totality annotation, 
restoring the previous default afterwards.

When synthesizing clauses, allow synthesised clauses to be
partial.

It's an arbitrary choice, the other choice can be the default
option. But if the place in which we're synthesising has a non-default
choice, then we're synthesising under the wrong option too.

### DIFF
--- a/src/TTImp/ProcessDef.idr
+++ b/src/TTImp/ProcessDef.idr
@@ -712,16 +712,13 @@ processDef opts nest env fc n_in cs_in
                        else linear
          nidx <- resolveName n
          
-         -- Set the default totality option to the current one, saving the global one
-         defaultTotality <- getDefaultTotalityOption
-         let treq = fromMaybe defaultTotality (findSetTotal (flags gdef))
-         setDefaultTotalityOption treq
-         
-         cs <- traverse (checkClause mult (visibility gdef) treq
+         -- Dynamically rebind default totality requirement to this function's totality requirement
+         -- and use this requirement when processing `with` blocks
+         let treq = fromMaybe !getDefaultTotalityOption (findSetTotal (flags gdef))
+         cs <- withTotality treq $ 
+               traverse (checkClause mult (visibility gdef) treq
                                      hashit nidx opts nest env) cs_in
-         -- Restore global default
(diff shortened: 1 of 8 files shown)


src/TTImp/ProcessDef.idr
```diff
@@ -708,8 +711,17 @@ processDef opts nest env fc n_in cs_in
...
+         
+         -- Set the default totality option to the current one, saving the global one
+         defaultTotality <- getDefaultTotalityOption
+         let treq = fromMaybe defaultTotality (findSetTotal (flags gdef))
+         setDefaultTotalityOption treq
+         
+         cs <- traverse (checkClause mult (visibility gdef) treq
                                      hashit nidx opts nest env) cs_in
+         -- Restore global default
+         setDefaultTotalityOption defaultTotality
```
gallais, line 723: A common technique to do this in a clean way is to define a

```idris
withTotality : TotalReq -> Lazy (Core a) -> Core a
withTotality tot c = do
         -- Set the default totality option to the current one, saving the global one
         defaultTotality <- getDefaultTotalityOption
         setDefaultTotalityOption tot
         x <- c
         setDefaultTotalityOption defaultTotality
         pure x
```

and then use that function in your code.

If you want to be totally (ha!) safe, this should probably even wrap the
call to `c` in a `catch` so that you may reset the default even if an error
is thrown (before re-throwing it).

ohad: Thanks, that's a good suggestion. Errors was a concern for me, and this is an elegant way to do it.
(I thought this would be done more nicely with dynamically bound state + dynamic unwind, but now we're actually doing it like that.)

ohad:
I'll merge this once it passes the tests

PR #659, opened 2020-09-10
Auto-implicit `__con` now added before implicits.

### LINKED ISSUE #658: Interface constraint variable unavailable in implicit arguments to interface methods
# Steps to Reproduce

Try to type check

    interface A where
      x : Nat
      f : {pf : x = 0} -> ()

# Expected Behavior

Should type-check just fine (does so in Idris 1).

# Observed Behavior

I get the error

    Error: While processing type of f. Undefined name __con. 

# Remarks

The problem is that in `src/Idris/Elab/Interface.idr` get auto-implicit variable `__con` gets inserted after *all* implicit variables, not just the interface variables.

I've added a pull request: #659

mr-infty:
The interface constraint variable `__con` was added after all implicit
variables before during interface elaboration, resulting in the variable
being unavailable in implicit arguments to interface methods.

For example, when type-checking

    interface A where
      x : Nat
      f : {pf : x = 0} -> ()

the type-checker would complain with

    Error: While processing type of f. Undefined name __con.

, whereas

    interface A where
      x : Nat
      f : (pf : x = 0) -> ()

would type-check just fine.

Note that the first example still doesn't type-check, but that's because
of another error.

**Remark**: This fixes #658.

### DIFF
--- a/src/Idris/Elab/Interface.idr
+++ b/src/Idris/Elab/Interface.idr
@@ -103,13 +103,9 @@ getMethDecl {vars} env nest params mnames (fc, c, opts, n, (d, ty))
              else IPi fc r p mn arg (stripParams ps ret)
     stripParams ps ty = ty
 
--- bind the auto implicit for the interface - put it after all the other
--- implicits
+-- bind the auto implicit for the interface - put it first, as it may be needed
+-- in other method variables, including implicit variables
 bindIFace : FC -> RawImp -> RawImp -> RawImp
-bindIFace _ ity (IPi fc rig Implicit n ty sc)
-       = IPi fc rig Implicit n ty (bindIFace fc ity sc)
-bindIFace _ ity (IPi fc rig AutoImplicit n ty sc)
-       = IPi fc rig AutoImplicit n ty (bindIFace fc ity sc)
 bindIFace fc ity sc = IPi fc top AutoImplicit (Just (UN "__con")) ity sc
 
 -- Get the top level function for implementing a method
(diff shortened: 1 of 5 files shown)


ziman:
Makes sense, thanks!

Could you add a test to raise an alarm if someone makes an incompatible change in the future?

mr-infty:
> Makes sense, thanks!
> 
> Could you add a test to raise an alarm if someone makes an incompatible change in the future?

Thanks for the review Matúš! Where do you think should I add the test? Under interface tests or under regressions?

ziman:
Thanks, I'd go for `regXXX`.

src/Idris/Elab/Interface.idr
```diff
@@ -128,8 +124,8 @@ getMethToplevel {vars} env vis iname cname constraints allmeths params
          -- Make the constraint application explicit for any method names
          -- which appear in other method types
          let ty_constr =
-             bindPs params $ substNames vars (map applyCon allmeths) ty
-         ty_imp <- bindTypeNames [] vars (bindIFace fc ity ty_constr)
+             substNames vars (map applyCon allmeths) ty
+         ty_imp <- bindTypeNames [] vars (bindPs params $ bindIFace fc ity ty_constr)
```
ziman, line 128: BTW, is `bindIface` used anywhere else? Would it be easier to inline it here, now that it's no longer a pattern-matching recursive definition?

mr-infty: `bindIface` isn't used anywhere else. I just left it in there because I thought it's better to change the code as little as possible, and because inlining the definition makes the code look a bit messy. On the other hand, it looks a bit odd to have such a trivial function, so I'm happy to inline it if you want.

gallais:
Now that I actually understand what is going on in this Elab phase
because of the work on #808, I'm happy to merge it. Thanks for the fix!

Issue #649, opened 2020-09-07
Error: INTERNAL ERROR: $m is not a function application


gallais:
# Steps to Reproduce

```idris
test : Nat
test = let foo : (m, n : Nat) -> m === n -> Nat
           foo .m m Refl = Z
       in foo Z Z Refl
```

# Observed Behavior

```
1/1: Building letdot (letdot.idr)
Error: INTERNAL ERROR: $m is not a function application
Error(s) building file letdot.idr
```

ziman:
The parser allows only the parenthesised form `foo .(m) m Refl = Z` for forced patterns. That's currently parsed as postfix application but #662 is a proposal to fix that.

ziman:
I believe that this can be closed, given that we have [a test for it](https://github.com/idris-lang/Idris2/pull/662/files#diff-9342bf0e02a81ea3b2fa13bc7ac0faf7a19c60cacea0026510f8aaa559136195R80).

Issue #441, opened 2020-07-09
No longer allowed function parameter names that clash with record field names if accessing that field within the function


jakeelituv:
# Steps to reproduce
```
record TestRecord where
  constructor MkRec
  n : Int

testFunc : Int -> TestRecord -> Bool
testFunc n testRecord =
    n == testRecord.n
```
results in: 
```
Mismatch between:
	Int
and
	TestRecord -> ?ty
```

If instead I rename the parameter in the function to m it compiles
```
record TestRecord where
  constructor MkRec
  n : Int

testFunc : Int -> TestRecord -> Bool
testFunc m testRecord =
    m == testRecord.n
```

Alternatively, if I don't ever call the ".n" function the following compiles, even if my function uses n as a parameter:

```
record TestRecord where
  constructor MkRec
  n : Int

testFunc : Int -> TestRecord -> Bool
testFunc n testRecord =
    n == 2
```

# Expected Behavior

This didn't happen until a commit from a few days ago so I assume this isn't supposed to happen.

ziman:
Indeed, a recent merge removed the distinction between function projections and other functions.

I'm surprised that disambiguation does not consider the global `n`. This should be fixed.

ziman:
It seems that the disambiguator will always pick the local variable if there's one.

Maybe `.field` should not consider locals at all and you'd have to write `.(f)` to apply the local `f`?

ziman:
I can see a few ways to address this. I think that the common theme is that we want `x.field` to behave differently than `field x`, namely in the presence of a local variable named `field`.

* Put `RF` back in. Then `x.field` would parse as `App (RF "field") (UN "x")`, and `x.(field)` would parse as `App (UN "field") (UN "x")` because the bracketed thing is an ordinary expression. You would not be able to say `"string".length` anymore, though.
* Extend TTImp with projections. Since desugaring of `x.field` happens in Idris, TTImp cannot tell the difference between `x.field` and `field x` when disambiguating identifiers later. We'd have to extend the AST of TTImp to preserve the information that the disambiguation of `field` in `x.field` should not prioritise the locals.
* Make disambiguation not prefer locals. Then both `field x` and `x.field` would behave exactly the same, and try both globals and locals. That's probably undesirable and might be annoying.

Maybe there are better solutions?

ziman:
If we put `RF` back in, we'd have to write `"string".(length)`. On the other hand, you could define
```idris
.length : String -> Nat
.length = length
```
[...]

PR #662, opened 2020-09-10
Revert arbitrary postfix application

ziman:
The [arbitrary postfix application `x.(arbitrary expression)` patch](https://github.com/idris-lang/Idris2/pull/343) merged a few months ago caused two issues:

1. It [stole the syntax of forced patterns](https://github.com/idris-lang/Idris2/issues/649), which also uses `.(...)` but was unreachable in the modified parser. We could change the syntax of forced patterns but I thought it might be better to take the conservative solution.

2. It [made local identifiers shadow record projections](https://github.com/idris-lang/Idris2/issues/441). Strictly speaking, this is not a problem of arbitrary postfix applications but was broken in that patch. The patch removed the `RF` constructor for `Name`s, and also changed the approach to the dot syntax: it made `.foo` stand for the postfix application of `foo`, while originally, `.foo` was a special name that you could give to a function to make it a postfix operator. (Record declarations would generate such postfix operators automatically but you could define them standalone, too.)

This pull request restores the original behaviour: `.(...)` now stands for forced patterns again and `.foo` is a special kind of name used for postfix operators. I believe that most of the utility of the arbitrary `.(...)` syntax can be recovered using `(&) : a -> (a -> b) -> b`, and the above two issues would be fixed by merging this pull request.

Putting arbitrary postfix applications back into the language at some later point would be fine, of course -- there doesn't seem to be anything wrong about them per se -- but we'd have to sort out their interaction with the two issues first.


andrevidela:
I don't see tests checking for #649 or #441, did I miss them?

ziman:
Good point, I forgot. Thanks!

ziman:
Any objections? I'd merge this soon.

PR #663, opened 2020-09-11
add ability to target scheme flavor in foreign specifiers

MarcelineVQ:
There's some missing flexibility in how foreign specifiers can be used with scheme that is addressed here with minimal changes to how scheme specifiers are read. This opens up uses for users that they otherwise would have had to modify the compiler's support files to accomplish.

### DIFF
--- a/tests/chez/chez031/expected
+++ b/tests/chez/chez031/expected
@@ -1,0 +1,1 @@
+Uncaught error: Specifiers.idr:18:1--24:5:No recognised foreign calling convention
--- a/tests/chez/chez031/expected
+++ b/tests/chez/chez031/expected
@@ -1,1 +1,27 @@
-Uncaught error: Specifiers.idr:18:1--24:5:No recognised foreign calling convention
+The given specifier was not accepted by any backend. Available backends:
+  chez, racket, node, javascript, gambit
+Some backends have additional specifier rules, refer to their documentation.
+
+Specifiers.idr:28:1--34:5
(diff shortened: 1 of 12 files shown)


tests/chez/chez031/expected
```diff
@@ -0,0 +1 @@
+Uncaught error: Specifiers.idr:18:1--24:5:No recognised foreign calling convention
```
gallais, line 1: Seeing "Uncaught error" is a bit scary. Is there any way to print
something nicer? Would be cool if the compiler could also say something
like "I was hoping to find ... because option ... was passed".

MarcelineVQ: That's the normal error you'd get if you wrote a specifier that didn't match any backends, e.g.
```idris
%foreign "fizz:bop,pop"
fizz : Int -> Int
```
since there's no fizz backend.
I'm not clear what you'd prefer to see but I am interested, could you elaborate?

gallais: When we have a type error we pretty-print an error message with some
location, explanation, etc. we don't just let an "uncaught error"
escape.

Could we do the same here? If the specifier did not match any of the
existing backends, we could for instance list them so that they are
easier to discover.

MarcelineVQ: Sure, sounds like a good idea. I'm not sure how specific the error should be but this should give the user some leads and is no longer uncaught.

gallais:
Thanks! <3

PR #667, opened 2020-09-13
 [contrib] Add misc libraries to contrib

ohad:
Expose some `private` function in libs/base that I needed, and seem like
their visibility was forgotten

I'd appreciate a code review, especially to tell me I'm
re-implementing something that's already elsewhere in the library

Mostly extending existing functionality:
* `Data/Void.idr`: add some utility functions for manipulating absurdity.
* `Decidable/Decidable/Extra.idr`: add support for double negation elimination in decidable relations
* `Data/Fun/Extra.idr`:
  + add `application` (total and partil) for n-ary functions
  + add (slightly) dependent versions of these operations
  + add n-ary function precomposition
* `Decidable/Order/Strict.idr`: a strict preorder is what you get when
  you remove the diagonal from a pre-order. For example, `<` is the
  associated preorder for `<=` over `Nat`.
  Analogous to `Decidable.Order`. The proof search mechanism struggled
  a bit, so I had to hack it --- sorry.

### DIFF
--- a/libs/contrib/Data/Fun/Extra.idr
+++ b/libs/contrib/Data/Fun/Extra.idr
@@ -1,54 +1,87 @@
 module Data.Fun.Extra
 
 import Data.Fun
+import Data.Rel
 import Data.HVect
 
-||| Apply an n-ary function to an n-ary tuple of inputs
-public export
-apply : {ts : Vect n Type} -> {cod : Type} -> Fun ts cod -> HVect ts -> cod
-apply f [] = f
-apply f (t::ts) = apply (f t) ts
+%default total
 
 ||| Apply an n-ary function to an n-ary tuple of inputs
 public export
-applyPartially : {ts : Vect n Type} -> {ss : Vect m Type} -> {rs : Vect k Type} 
-               -> {auto 0 fordShape : rs = ts ++ ss}
-               -> {cod : Type} -> Fun rs cod -> HVect ts -> Fun ss cod
-applyPartially {fordShape = Refl} f [] = f
-applyPartially {fordShape = Refl} f (x::xs) = applyPartially (f x) xs
+uncurry : {ts : Vect n Type} -> {cod : Type} -> Fun ts cod -> HVect ts -> cod
+uncurry f [] = f
+uncurry f (t::ts) = uncurry (f t) ts
+
+||| Apply an n-ary function to an n-ary tuple of inputs
+public export
+curry : {ts : Vect n Type} -> {cod : Type} -> (HVect ts -> cod) -> Fun ts cod 
+curry {ts = []     } f = f []
+curry {ts = t :: ts} f = \x => curry (\xs => f (x :: xs))
+
+{- 
+
+The higher kind Type -> Type has a monoid structure given by
+composition and the identity (Cayley). The type (n : Nat ** Vect n a)
+has a monoid structure given by `(n, rs) * (m, ss) := (n + m, rs +
+ss)` and `(0,[])`. 
+
+  `Fun' : (n : Nat ** Vect n Type) -> Type -> Type` 
+  
+is then a monoid homomorphism between them. I guess this is some
+instance of Cayley's theorem, but because of extensionality we can't
+show we have an isomorphism.
+-}
+public export
+homoFunNeut_ext : Fun [] cod -> id cod
+homoFunNeut_ext x = x
+
+public export
+homoFunMult_ext : {rs : Vect n Type} -> Fun (rs ++ ss) cod -> (Fun rs . Fun ss) cod
+homoFunMult_ext {rs = []     }  gs = gs
+homoFunMult_ext {rs = t :: ts} fgs = \x => homoFunMult_ext (fgs x)
+
+public export
+homoFunNeut_inv : id cod -> Fun [] cod
+homoFunNeut_inv x = x
+
+public export
+homoFunMult_inv : {rs : Vect n Type} -> (Fun rs . Fun ss) cod -> Fun (rs ++ ss) cod
+homoFunMult_inv {rs = []     } gs = gs
+homoFunMult_inv {rs = t :: ts} fgs = \x => homoFunMult_inv (fgs x)
+
+
+||| Apply an n-ary function to an n-ary tuple of inputs
+public export
+applyPartially : {ts : Vect n Type} -> {ss : Vect m Type} 
+               -> {cod : Type} -> Fun (ts ++ ss) cod -> (HVect ts -> Fun ss cod)
+applyPartially fgs = uncurry {ts} {cod = Fun ss cod} (homoFunMult_ext {rs=ts} {ss} {cod} fgs)
 
 
 {- -------- (slightly) dependent versions of the above ---------------
    As usual, type dependencies make everything complicated          -}
 
-
-||| Build an n-ary dependent function type from a Vect of Types and a dependent result type
-||| NB: the types in the tuple do not depend on each other.
-public export
-GenFun : (ts : Vect n Type) -> (cod : Fun ts Type) -> Type
-GenFun [] cod = cod
-GenFun (t :: ts) cod = (x : t) -> GenFun ts (cod x )
-
 ||| Apply an n-ary dependent function to its tuple of inputs (given by an HVect)
 public export
-applyGen : {ts : Vect n Type} -> {cod : Fun ts Type} 
-        -> GenFun ts cod -> (xs : HVect ts) -> apply cod xs
-applyGen f [] = f
-applyGen {ts = t :: ts} {cod} f (x :: xs) = applyGen {cod= cod x} (f x) xs
+uncurryAll : {ts : Vect n Type} -> {cod : Fun ts Type} 
+        -> All ts cod -> (xs : HVect ts) -> uncurry cod xs
+uncurryAll f [] = f
+uncurryAll {ts = t :: ts} {cod} f (x :: xs) = uncurryAll {cod= cod x} (f x) xs
+
+public export
+curryAll : {ts : Vect n Type} -> {cod : Fun ts Type}
+        -> ((xs : HVect ts) -> uncurry cod xs)
+        -> All ts cod
+curryAll {ts = []     } f = f []
+curryAll {ts = t :: ts} f = \x => curryAll (\ xs => f (x:: xs))
 
 chainGenUncurried : {ts : Vect n Type} -> {cod,cod' : Fun ts Type} -> 
-           ((xs : HVect ts) -> apply cod xs -> apply cod' xs) ->  
-           GenFun ts cod -> GenFun ts cod'
+           ((xs : HVect ts) -> uncurry cod xs -> uncurry cod' xs) ->  
+           All ts cod -> All ts cod'
 chainGenUncurried {ts = []} f gs = f [] gs
 chainGenUncurried {ts = (t :: ts)} f gs = \x => chainGenUncurried (\u => f (x :: u)) (gs x)
 
-||| Partially apply an n-ary dependent function to a tuple of its inputs
-||| Yes, the type is disgusting
-applyPartiallyGen : {ts : Vect n Type} -> {ss : Vect m Type} -> {rs : Vect k Type} 
-                -> {cod : Fun rs Type} 
-                -> {auto 0 fordShape : rs = ts ++ ss}
-                -> GenFun rs cod -> (xs : HVect ts) -> 
-                   GenFun ss $ applyPartially {ts} {ss} {rs} {cod = Type} cod xs
-applyPartiallyGen {ts = []     } {fordShape = Refl} f [] = f
-applyPartiallyGen {ts = t :: ts} {fordShape = Refl} f (x :: xs) = applyPartiallyGen (f x) xs
-  
+public export
+homoAllNeut_ext : Fun [] cod -> id cod
+homoAllNeut_ext x = x
+
+-- Not sure it's worth it getting the rest of Cayley's theorem to work
--- a/libs/contrib/Data/Fun/Extra.idr
+++ b/libs/contrib/Data/Fun/Extra.idr
@@ -105,11 +105,12 @@ introduceWitness (w :: witness) f = (w ** introduceWitness witness f)
 
 ----------------------------------------------------------------------
 public export
-data FunVect : (ts, ss : Vect n Type) -> Type where
-  Nil  : FunVect [] []
-  (::) : (f : t -> s) -> FunVect ts ss -> FunVect (t::ts) (s::ss)
+data Pointwise : (r : a -> b -> Type) -> (ts : Vect n a) -> (ss : Vect n b) -> Type where
+  Nil  : Pointwise r [] []
+  (::) : {0 ss, ts : Vect n Type} -> 
+         (f : r t s) -> Pointwise r ts ss -> Pointwise r (t::ts) (s::ss)
 
 public export
-precompose : FunVect ts ss -> Fun ss cod -> Fun ts cod
+precompose : Pointwise (\a,b => a -> b) ts ss -> Fun ss cod -> Fun ts cod
 precompose [] h = h
 precompose (f :: fs) h = \x => precompose fs (h (f x))
--- a/libs/contrib/Data/Fun/Extra.idr
+++ b/libs/contrib/Data/Fun/Extra.idr
@@ -8,15 +8,15 @@ import Data.HVect
 
 ||| Apply an n-ary function to an n-ary tuple of inputs
 public export
-uncurry : {ts : Vect n Type} -> {cod : Type} -> Fun ts cod -> HVect ts -> cod
-uncurry f [] = f
-uncurry f (t::ts) = uncurry (f t) ts
+uncurry : {n : Nat} -> {0 ts : Vect n Type} -> Fun ts cod -> HVect ts -> cod
+uncurry {n = 0  } f [] = f
+uncurry {n = S n} f (x::xs) = uncurry (f x) xs
 
 ||| Apply an n-ary function to an n-ary tuple of inputs
 public export
-curry : {ts : Vect n Type} -> {cod : Type} -> (HVect ts -> cod) -> Fun ts cod 
-curry {ts = []     } f = f []
-curry {ts = t :: ts} f = \x => curry (\xs => f (x :: xs))
+curry : {n : Nat} -> {0 ts : Vect n Type} -> (HVect ts -> cod) -> Fun ts cod 
+curry {ts = []    } f = f []
+curry {ts = _ :: _} f = \x => curry (\xs => f (x :: xs))
 
 {- 
 
@@ -36,7 +36,7 @@ homoFunNeut_ext : Fun [] cod -> id cod
 homoFunNeut_ext x = x
 
 public export
-homoFunMult_ext : {rs : Vect n Type} -> Fun (rs ++ ss) cod -> (Fun rs . Fun ss) cod
+homoFunMult_ext : {n : Nat} -> {0 rs : Vect n Type} -> Fun (rs ++ ss) cod -> (Fun rs . Fun ss) cod
 homoFunMult_ext {rs = []     }  gs = gs
 homoFunMult_ext {rs = t :: ts} fgs = \x => homoFunMult_ext (fgs x)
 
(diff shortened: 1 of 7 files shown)


libs/contrib/Data/Fun/Extra.idr
```diff
@@ -0,0 +1,54 @@
...
+
+
+{- -------- (slightly) dependent versions of the above ---------------
+   As usual, type dependencies make everything complicated          -}
+
+
+||| Build an n-ary dependent function type from a Vect of Types and a dependent result type
+||| NB: the types in the tuple do not depend on each other.
+public export
+GenFun : (ts : Vect n Type) -> (cod : Fun ts Type) -> Type
```
gallais, line 28: This is `Data.Rel`'s `All`. I need to think harder about the stuff that
comes after that to understand what it corresponds to.

libs/contrib/Data/Fun/Extra.idr
```diff
@@ -0,0 +1,54 @@
+module Data.Fun.Extra
+
+import Data.Fun
+import Data.HVect
+
+||| Apply an n-ary function to an n-ary tuple of inputs
+public export
+apply : {ts : Vect n Type} -> {cod : Type} -> Fun ts cod -> HVect ts -> cod
```
gallais, line 8: I would call this `curry` and also define `uncurry`. Note that both `ts` and
`cod` can probably be given a `0` annotation (we only need to be able to match on `n`).

```idris
uncurry : {ts : Vect n Type} -> {cod : Type} -> (HVect ts -> cod) -> Fun ts cod
```

libs/contrib/Data/Fun/Extra.idr
```diff
@@ -0,0 +1,54 @@
...
+apply f [] = f
+apply f (t::ts) = apply (f t) ts
+
+||| Apply an n-ary function to an n-ary tuple of inputs
+public export
+applyPartially : {ts : Vect n Type} -> {ss : Vect m Type} -> {rs : Vect k Type}
```
gallais, line 14: This is the composition of `curry` and `Fun` seen as a homomorphism from
the monoid `HVect` to `Type -> Type`. Probably worth implementing the homo
parts independently:

```idris
homoNeutral : Fun [] cod -> id cod
homoMult    : Fun (rs ++ ss) cod -> (Fun rs . Fun ss) cod
```

libs/contrib/Data/Void.idr
```diff
@@ -0,0 +1,13 @@
+module Data.Void
+
+export
+persistentVoid : (0 x : Void) -> Void
```
gallais, line 4: This is a specialised version of `void`

libs/contrib/Data/Void.idr
```diff
@@ -0,0 +1,13 @@
...
+persistentVoid _ impossible
+
+export
+absurdity : Uninhabited t => (0 _ : t) -> s
+absurdity x = absurd (persistentVoid (uninhabited x))
```
gallais, line 9: This is `void (uninhabited x)`

libs/contrib/Data/Void.idr
```diff
@@ -0,0 +1,13 @@
...
+
+export
+absurdSince : (Uninhabited t) => (0 _ : x -> t) -> (x -> s)
+absurdSince since x = absurdity (since x)
```
gallais, line 13: I would probably call this `contradiction` as it's essentially
`(¬ x /\ x) -> y`. Note that `x` could be made `0`.

libs/contrib/Decidable/Decidable/Extra.idr
```diff
@@ -0,0 +1,25 @@
...
+
+public export
+NotNot : {ts : Vect n Type} -> (r : Rel ts) -> Rel ts
+NotNot r = map @{Nary} (Not . Not) r 
+
+[DecidablePartialApplication] {x : t} -> (tts : Decidable (t :: ts) r) => Decidable ts (r x) where
+  decide = decide @{tts} x
+
+public export
+doubleNegation : {ts : Vect n Type} -> {r : Rel ts} -> Decidable ts r =>
```
gallais, line 17: We should probably use a different name because we ought to
be able to prove something similar for `All ts (NotNot r)`.

libs/contrib/Decidable/Order/Strict.idr
```diff
@@ -0,0 +1,84 @@
+||| An strict preorder (sometimes known as a quasi-order, or an
+||| ordering) is what you get when you remove the diagonal `{(a,b) | a
+||| r b , b r a}` from a preorder. For example a < b is an ordering.
+||| This module extends base's Decidable.Order with the strict versions.
+||| The interface system seems to struggle a bit with some of the constructions,
+||| so I hacked them a bit. Sorry.
+module Decidable.Order.Strict
+
```
gallais, line 8: Not too fond of the `Hack` terminology. Is this not a free completion?

libs/contrib/Data/Fun/Extra.idr
ohad: I see, thanks! I would really like to have a fully dependently typed `GenFun`, but I've never managed to work out how to best represent telescopes (rather than `Vect n Type`).

I've renamed it, it seems weird to `import Data.Rel`, but I guess that's OK.

libs/contrib/Data/Fun/Extra.idr
ohad: Done, though I would call them `uncurry` and `curry` instead (opposite of what you wrote).

ohad: Re. annotation: I'm never sure what one should match on. 
Should vectors carry their length and be packed? 
Should vectors be linked lists and we should only calculate their length if we want to?
Should we be able to pass constraints like `{wn n : Nat} -> {wts ts : Vect n Type} -> {0 auto atLeastOne : NonZero [wn, wt]}` that allows the type-checker to match on the pattern:
```idris
{n = 0} {ts : []}
{n = S n} {ts : x :: xs}
```
because each invocation is guaranteed to have a case tree (but we don't know right now which one it should be)?

gallais: A `Vect n Type` is essentially a linked list of erased stuff.
It's a lot cheaper to just carry the `n` around hence my suggestion.

ohad: At some point someone needs to compute this `n`. Merely having a `Vect n` doesn't give you it.

libs/contrib/Data/Fun/Extra.idr
ohad: I see. I guess it's a special case of Cayley's theorem. Cool, I've added it. I've dropped the partial application from the dependent type, it's not really worth it. (It would be more interesting to generalise it to full telescopes.)

libs/contrib/Data/Void.idr
ohad: Thanks, removed it.

libs/contrib/Data/Void.idr
ohad: I know it's petty, but it's annoying that we have `absurd` but write `void (uninhabited)` when we're irrelevant.
I propose we either keep `absurdity`, or change `absurd` to take an irrelevant argument. 
(I'd prefer the latter, but that would mean changing `base`.)

libs/contrib/Data/Void.idr
ohad: Renamed.

> Note that `x` could be made `0`.

Noted, but since `Not` is defined to be `a -> Void`, we can't use `contradiction` in place of a `Not a`.

We could change the library so that `Not a` takes an irrelevant `a`.

libs/contrib/Data/Fun/Extra.idr
gallais: Yes. I am arguing in favour of carrying the `n` around instead
of carrying the `Vect n Type` around. One Integer vs. a linked
list of erased things.

libs/contrib/Data/Void.idr
gallais: I think the problem with that proposal is that `No absurd` will
not typecheck anymore because `No` expects a proof-relevant
function. So you'd need to manually go expand all of the existing
`No absurd` to `No (\ 0 x => absurd x)`.

libs/contrib/Data/Fun/Extra.idr
ohad: Not quite, if you're not careful you'll be carrying one integer for every cons-cell.

libs/contrib/Data/Void.idr
ohad: I'll keep absurdity then.

libs/contrib/Data/Fun/Extra.idr
gallais: Even more reasons to make it `{0 ts : Vect n Type}`.

ohad: This is frustrating. This means there's something I'm not getting, perhaps an assumption I'm not even realising I'm making.

At runtime, is a `Nat` represented in binary or as a linked list of `Succ` cells?

libs/contrib/Decidable/Order/Strict.idr
ohad: What do you mean by `a free completion`?

libs/contrib/Decidable/Decidable/Extra.idr
ohad: Good point! Thanks!
I've now refactored it into: 
```idris
doubleNegationElimination : ... (witness : HVect ts) -> uncurry (NotNot {ts} r) witness -> uncurry              r  witness
doubleNegationForall : ... All ts (NotNot {ts} r) -> All ts r
doubleNegationExists : ...Ex ts (NotNot {ts} r) -> Ex ts r
```
and added introduction and elimination rules for `Ex` in `Data.Fun.Extra`:
```idris
extractWitness : {ts : Vect n Type} -> {r : Rel ts} -> Ex ts r -> HVect ts
extractWitnessCorrect : {ts : Vect n Type} -> {r : Rel ts} -> (f : Ex ts r) -> uncurry {ts} r (extractWitness {r} f)
introduceWitness : {ts : Vect n Type} -> {r : Rel ts} -> (witness : HVect ts) -> uncurry {ts} r witness -> Ex ts r
```

libs/contrib/Data/Fun/Extra.idr
gallais: At runtime a `Nat` is an `Integer` i.e. it's stored in binary.

libs/contrib/Decidable/Order/Strict.idr
gallais: I mean that you are throwing in exactly the missing proofs to
get a partial order from a strict preorder.

I would call that [a completion](https://ncatlab.org/nlab/show/completion) of the strict preorder into a partial one.

And my hunch it's free comes from the "exactly the missing" and
not anything more.

ohad: Not quite: each strict preorder has an associated poset via a coreflection (adding the missing diagonal) and each preorder has a strict preorder via a reflection (throwing away the diagonal). I think this adjunction restricts to an equivalence (actually isomorphism) when you restrict to posets only.

In any case the 'hack' isn't because this notion isn't canonical. It's because the type system can't seem to work out the correct instances and I need to use the 'hack' to construct them explicitly.

libs/contrib/Data/Void.idr
ohad: I'm in favour of keeping it as is `contradiction (Uninhabited t) => (0 _ : x -> t) -> (x -> s)`  for the same reason we keep `absurd` as is.

libs/contrib/Data/Fun/Extra.idr
ohad: OK, so that's not my implicit assumption.

So we have a vector, i.e., linked list, somewhere. The linked list is not carrying its own length.
If I require library functions to have the length as relevant, then we'd be forced to calculate the length of the vector, on top of having the vector already, every time we call the library function.

gallais: I don't agree with the assumption that we necessarily are carrying around
the linked list. It does not contain anything of runtime interest (it's only
`Type`s!) except for its length which is implicit in the structure.

So we may as well pass the length around and let the vector be erased!

ohad: The vector isn't going to be erased, it contains the data we care about (in other places in the code-base).
It's just that if we make it irrelevant, we won't pass around a pointer for its different cells, and we'll need to somehow get its length.
Which means either calculating it for a long vector, or amortizing the calculation by allocating a length for each cell.

ohad: OK, I think I understand. Your point is that we don't need the types at runtime, just the number of arguments.

ohad: Got it. Changed it now. 
Thanks for the patience, I appreciate it a lot.

libs/contrib/Data/Fun/Extra.idr
```diff
@@ -8,15 +8,15 @@ import Data.HVect
 
 ||| Apply an n-ary function to an n-ary tuple of inputs
 public export
-uncurry : {ts : Vect n Type} -> {cod : Type} -> Fun ts cod -> HVect ts -> cod
-uncurry f [] = f
-uncurry f (t::ts) = uncurry (f t) ts
+uncurry : {n : Nat} -> {0 ts : Vect n Type} -> Fun ts cod -> HVect ts -> cod
```
gallais, line 11: Now that I think about it, the `n` can *also* be `0` here because the
`HVect` argument gives you enough information! In `curry` however you
really cannot get away with not having it.

libs/contrib/Data/Fun/Extra.idr
```diff
@@ -45,14 +45,14 @@ homoFunNeut_inv : id cod -> Fun [] cod
...
-homoFunMult_inv : {rs : Vect n Type} -> (Fun rs . Fun ss) cod -> Fun (rs ++ ss) cod
+homoFunMult_inv : {n : Nat} -> {0 rs : Vect n Type} -> (Fun rs . Fun ss) cod -> Fun (rs ++ ss) cod
 homoFunMult_inv {rs = []     } gs = gs
 homoFunMult_inv {rs = t :: ts} fgs = \x => homoFunMult_inv (fgs x)
 
 
 ||| Apply an n-ary function to an n-ary tuple of inputs
 public export
-applyPartially : {ts : Vect n Type} -> {ss : Vect m Type} 
+applyPartially : {n : Nat} -> {0 ts : Vect n Type}
```
gallais, line 55: That's funny: the `HVect` argument gives you enough of a handle
on `n` to not need it but because the function is defined by
composing `homoFunMult_ext` we cannot remove it!

Note that `cod` could be made `0` though.

libs/contrib/Data/Fun/Extra.idr
```diff
@@ -62,19 +62,19 @@ applyPartially fgs = uncurry {ts} {cod = Fun ss cod} (homoFunMult_ext {rs=ts} {s
 
 ||| Apply an n-ary dependent function to its tuple of inputs (given by an HVect)
 public export
-uncurryAll : {ts : Vect n Type} -> {cod : Fun ts Type} 
+uncurryAll : {n : Nat} -> {0 ts : Vect n Type} -> {0 cod : Fun ts Type}
```
gallais, line 65: Same here wrt. to `HVect` giving you a handle on `n`.

libs/contrib/Data/Fun/Extra.idr
gallais: I really like that QTT forces us to think harder about all of this!

libs/contrib/Data/Fun/Extra.idr
ohad: Yes, I noticed it elsewhere, but missed it here. Thanks!

libs/contrib/Data/Fun/Extra.idr
ohad: It is funny :).
Someone showed me another instance of this phenomena in more 'traditional' mathematics:
|x|² = x²
but if we're only deducing differentiability compositionally, the LHS is non-differentiable at 0 while the RHS is.

libs/contrib/Data/Fun/Extra.idr
ohad: Done.

libs/contrib/Data/Fun/Extra.idr
ohad: I should give credit where it's due, that someone was @jasigal

libs/contrib/Decidable/Order/Strict.idr
ohad: OK, I've now removed the term 'hack' from the identifiers and called them `MkPreorder` and `InferPoset`, for example

ohad:
Once we resolve the 'hacking' terminology and this passes the tests, I plan to rebase and merge this into master.

gallais:
Maybe worth squashing rather than mere rebase?

ohad:
Sure, happy to squash.

ohad:
Actually, I've just added n-ary function precomposition to `Data.Fun.Extra` (ended up using it somewhere).

Also, have we resolved the 'hack' issue?

libs/contrib/Data/Fun/Extra.idr
```diff
@@ -102,3 +102,14 @@ introduceWitness : {0 r : Rel ts} -> (witness : HVect ts) ->
                    uncurry {ts} r witness -> Ex ts r
 introduceWitness []             f = f
 introduceWitness (w :: witness) f = (w ** introduceWitness witness f)
+
+----------------------------------------------------------------------
+public export
+data FunVect : (ts, ss : Vect n Type) -> Type where
+  Nil  : FunVect [] []
+  (::) : (f : t -> s) -> FunVect ts ss -> FunVect (t::ts) (s::ss)
+
```
gallais, line 111: Isn't this `HVect (zipWith (\ a, b => a -> b) ts ss)`?

ohad: For example:
```idris
export
Poset Nat (<<=) where
  antisymmetric a b = precompose [LTEbyPlus a b, LTEbyPlus b a]
                               $ antisymmetric a b
```
gives:
```

```
     While processing right hand side of antisymmetric. Can't solve constraint
     between: [(a <<= b -> LTE a b), (b <<= a -> LTE b a)] and zipWith (\a, b => a -> b) ?ts ?ss.
     
     Data/Nat/Order/Strict.idr:108:34--108:64
          |
      108 |   antisymmetric a b = precompose [LTEbyPlus a b, LTEbyPlus b a]
```
indexing by computations is something I really interested in, but this much indexing might be a little too interesting at the moment.

Comment (gallais) [reply libs/contrib/Data/Fun/Extra.idr:111]: In this case what about using:

```idris
data Pointwise : (a -> b -> Type) -> (ts : Vect n a) -> (ss : Vect n b) -> Type where
  Nil  : Pointwise [] []
  (::) : (f : r t s) -> Pointwise r ts ss -> Pointwise r (t::ts) (s::ss)
```

It would be nice not to have to rely on too many ad-hoc data structures
(e.g. how `HVect` is essentially `Data.Vect.All id`...)

ohad: I see. 
I get different errors this time:
```idris
public export
data Pointwise : (r : a -> b -> Type) -> (ts : Vect n a) -> (ss : Vect n b) -> Type where
  Nil  : Pointwise r [] []
  (::) : (f : r t s) -> Pointwise r ts ss -> Pointwise r (t::ts) (s::ss)

public export
Arrow : Type -> Type -> Type
Arrow a b = a -> b

public export
precompose : Pointwise Arrow ts ss -> Fun ss cod -> Fun ts cod
precompose [] h = h
precompose (f :: fs) h = \x => precompose fs (h ?hole)
```

```
Data/Nat/Order/Strict.idr line 109 col 33:
     While processing right hand side of antisymmetric. Can't find an implementation for Poset Nat ?po.
     
     Data/Nat/Order/Strict.idr:109:34--109:51
          |
      109 |                                $ antisymmetric a b
          |                                  ^^^^^^^^^^^^^^^^^
```
It's entirely possible there's another instability elsewhere in the system.

ohad: Also, it's clear to me that neither `Fun` nor `All ts cod` are the final say --- we should one day be able to deal with telescopes. So whatever we're currently doing is already 'ad-hoc'

ohad: Thanks, that works now. I have to include the annotation on `(::)`, otherwise the type-checker fails on the second clause of `precompose`. Unification error?

```idris
public export
data Pointwise : (r : a -> b -> Type) -> (ts : Vect n a) -> (ss : Vect n b) -> Type where
  Nil  : Pointwise r [] []
  (::) : {0 ss, ts : Vect n Type} -> 
         (f : r t s) -> Pointwise r ts ss -> Pointwise r (t::ts) (s::ss)

public export
precompose : Pointwise (\a,b => a -> b) ts ss -> Fun ss cod -> Fun ts cod
precompose [] h = h
precompose (f :: fs) h = \x => precompose fs (h (f x))
```

ohad: @edwinb  just in case

gallais: There's no hope of getting good inference for telescopes in their full
generality but stuff like `Fun` is very amenable to it. Cf. [self-promotion](http://gallais.github.io/pdf/tyde19.pdf).

ohad: I know the self-promoted paper: I wish we had a telescopic version of it.

Putting inference aside, it would be nice to have telescopes in full generality with type ascription. E.g., I'm happy to annotate with the telescope, if it gives me the rest for free. It's how to encode telescopes that I'm interested in.

gallais: It's not too hard. There is quite a bit of wiggle room depending on whether you
want them left or right nested or whether you want full trees. If you want full
trees then you need IR (cf. [this example in Agda](https://github.com/liamoc/dddp/pull/1)).

Anyway, we should probably move this discussion somewhere else. :D

ohad: Happy to! Is this final issue resolved?

gallais: I think so! Eventually we should move `Pointwise` to `Data.Vect.Quantifiers` in `base`
but we don't have any interesting results to attach to it at the moment so it's not
urgent.

PR #668, opened 2020-09-13
add specifier check for cli compilation

MarcelineVQ:
Missed a case in #663 which is covered now by checking for exceptions where postOptions is used

### DIFF
--- a/src/Idris/Driver.idr
+++ b/src/Idris/Driver.idr
@@ -183,7 +183,8 @@ stMain cgs opts
                                   displayErrors res
                                   pure res
 
-                 doRepl <- postOptions result opts
+                 doRepl <- catch (postOptions result opts)
+                                 (\err => emitError err *> pure False)
                  if doRepl then
                    if ide || ideSocket then
                      if not ideSocket
--- a/src/Core/Core.idr
+++ b/src/Core/Core.idr
@@ -462,6 +462,14 @@ export
 (<*>) : Core (a -> b) -> Core a -> Core b
 (<*>) (MkCore f) (MkCore a) = MkCore [| f <*> a |]
 
+export
+(*>) : Core a -> Core b -> Core b
+(*>) (MkCore a) (MkCore b) = MkCore [| a *> b |]
+
+export
+(<*) : Core a -> Core b -> Core a
+(<*) (MkCore a) (MkCore b) = MkCore [| a <* b |]
+
 export %inline
 when : Bool -> Lazy (Core ()) -> Core ()
 when True f = f
--- a/src/Idris/SetOptions.idr
+++ b/src/Idris/SetOptions.idr
@@ -8,7 +8,6 @@ import Core.Unify
 import Utils.Path
 
 import Idris.CommandLine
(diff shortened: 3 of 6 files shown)


src/Idris/Driver.idr
```diff
@@ -182,7 +182,9 @@ stMain cgs opts
                                   displayErrors res
                                   pure res
 
-                 doRepl <- postOptions result opts
```
gallais, line 185: Prior to this change, `doRepl` would become `False` and we would
wound up taking the `else` branch below. Which means that in
addition to displaying the error & exiting we would `showTimeRecord`.

Could we *not* bypass that if-then-else so that we keep executing the
same block of code rather than duplicating the logic? I guess something
like

```
doRepl <- catch (postOptions result opts)
                (\ err => pure False)
```

ought to work. If it does not then we may need to register that `err`
so that it is noticed by `errorLine` down below. But I would expect
that to already be done at the point of emission.

MarcelineVQ: Oops yes I'll add `showTimeRecord` into there

errorLine is a little more bare than you could be thinking, it just stores the line number where an error occurs and is only updated in one or two places (via updateErrorLine). There isn't actually error displaying happening in the original code, just an `exitWith (ExitFailure 1)` if errorLine is Just. My guess was that error reporting was expected to happen before that point which is why I did it at `doRepl`.

There are a couple options to explore if we want to report errors, such as bundling the error with doRepl to print later when we exit:
```idris
                 (doRepl,err) <- catch (pure (!(postOptions result opts), Nothing))
                                       (\err => pure (False, Just err))
```
Or to change errorLine to store the error in-addition-to/instead-of just the line.

Both of these are more involved than just repeating the call to showTimeRecord in the catch though. Thoughts?

gallais: I guess we already have a duplicated `showTimeRecord` after the call to
`repl` in a different branch so we may as well do the same here.

src/Idris/Driver.idr
```diff
@@ -184,7 +184,7 @@ stMain cgs opts
                                   pure res
 
                  doRepl <- catch (postOptions result opts)
-                                 (\err => emitError err *> pure False)
+                                 (\err => False <$ showTimeRecord <* emitError err)
```
MarcelineVQ, line 187: Because we continue on with False instead of quitting immediately, `showTimeRecord` will be done twice with this change as it's done at line 211. We can quit here or let the later showTimeRecord do the work

I think if I ever revisit this section this if-chain will be refactored into a sum type, it wouldn't save any space but should make what the ifs branches are doing more direct. Or maybe by then we'll have multi-way-if.

gallais: Oh shoot you had already pushed a fix? I thought the PR was going stale and
needed just one small change to be mergeable. That's why I hate force pushes:
we completely lose the information of what happened when and what the new
changes are.

MarcelineVQ: I apologize, forgot the basic rule of don't force-push things other people can touch.

PR #671, opened 2020-09-15
State param order

MarcelineVQ:
Nipping this historical artifact in the bud before it roots. It's quite often useful to be able to `map` directly to the result of a StateT computation and due to how Functor works this is made harder when the tuple is `(a,state)` vs `(state,a)`
Specifically that `map f (x,y)` applies `f` to `y`.

It's also rare that a person's `state` argument takes up more space than their `StateT s a` argument in `runState`. e.g.
```
foo = runState (do multiline
                   statet
                   thing)
               thestartingstate
```
and so one has to immediately decide if it's worth the extra parens and formatting involved or if they should take the time to pull out `multiline_statet_thing` into a separate definition.

Let's avoid this hassle by providing the state upfront instead and allowing users to write their bulkier code where it will fit better.
```
foo = runState thestartingstate $ do multiline
    statet
    thing
```
This makes for much nicer in-progress code, things can be moved out later on if they grow too large.

Note that while one can write
```idris
(`runState` thestartingstate) $ do ...`
```
this pr is about setting the default. How often is the code making up your `StateT s a` larger than your `state` code? Should users have to write this operator section if putting `StateT s a` last is the majority use case?

### DIFF
--- a/docs/source/typedd/typedd.rst
+++ b/docs/source/typedd/typedd.rst
@@ -346,6 +346,8 @@ arguments ``div`` and ``rem``.
 In ``ArithState.idr``, since ``(>>=)`` is ``export``, ``Command`` and ``ConsoleIO``
 also have to be ``export``.
 
+evalState from Control.Monad.State now takes the ``stateType`` argument first.
+
 Chapter 13
 ----------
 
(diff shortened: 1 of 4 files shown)


gallais:
Yes please! :heart:

Issue #676, opened 2020-09-16
pair ( , ) seems to forget to use <*> in idiom brackets


MarcelineVQ:
# Steps to Reproduce
```idris
fez : IO Int
fez = pure 1

fine : IO (Int, Maybe Int)
fine = [| MkPair (fez) (pure Nothing) |]

bad : IO (Int, Maybe Int)
bad = [| ( (fez) , (pure Nothing) ) |]
```
# Expected Behavior
no difference
# Observed Behavior
```
While processing right hand side of bad. Can't find an implementation for Applicative ?f.

src/Foo.idr:12:7--12:39
    |
 8  | bad = [| ( (fez) , (pure Nothing) ) |]
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This seems to occur from selecting alternatives for what (,) is, somehow a step is missed and <*> never comes into play.
The raw for these forms is as follows, with FC's removed for readability, and a version using + at the end to show that it's not an operator problem. Though note that the + version wouldn't typecheck anyway, it's just to show specifically (,) is at issue.

```
`([| MkPair (fez) (pure Nothing) |])
IApp
  (IApp
    (IVar (UN "<*>"))
    (IApp
      (IApp
        (IVar (UN "<*>"))
        (IApp
          (IVar (UN "pure"))
          (IVar (UN "MkPair"))))
      (IVar (UN "fez"))))
  (IApp
    (IVar (UN "pure"))
    (IVar (UN "Nothing")))


`([| ( (fez) , (pure Nothing) ) |])
IApp
  (IVar (UN "pure"))
  (IAlternative
    (UniqueDefault
      (IApp
        (IApp
          (IVar (NS (MkNS ["Builtin"]) (UN "MkPair")))
          (IVar (UN "fez")))
        (IApp
          (IVar (UN "pure"))
          (IVar (UN "Nothing")))))
    [IApp
      (IApp
        (IVar (NS (MkNS ["Builtin"]) (UN "Pair")))
        (IVar (UN "fez")))
      (IApp
        (IVar (UN "pure"))
        (IVar (UN "Nothing")))
    ,IApp
      (IApp
        (IVar (NS (MkNS ["Builtin"]) (UN "MkPair")))
        (IVar (UN "fez")))
      (IApp
        (IVar (UN "pure"))
        (IVar (UN "Nothing")))
    ])


`([| (fez) + (pure Nothing) |])
IApp
  (IApp
[...]

PR #678, opened 2020-09-17
fix idiom brackets to account for IAlternative

MarcelineVQ:
Things like (,) () aren't straightforward IVar's but are IAlternative's
which present options about how the term should resolve. [| |] was not
accounting for this.

Addressing #676

### DIFF
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -118,6 +118,8 @@ bindBangs ((n, fc, btm) :: bs) tm
                                 (Implicit fc False) tm)
 
 idiomise : FC -> RawImp -> RawImp
+idiomise fc (IAlternative afc u alts)
+  = IAlternative afc (mapAltType (idiomise afc) u) (idiomise afc <$> alts)
 idiomise fc (IApp afc f a)
     = IApp fc (IApp fc (IVar fc (UN "<*>"))
                     (idiomise afc f))
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -285,7 +285,10 @@ mutual
            pure (IVar fc bn)
   desugarB side ps (PIdiom fc term)
       = do itm <- desugarB side ps term
-           pure (idiomise fc itm)
+           logRaw "desugar.idiom" 10 "Desugaring idiom for" itm
+           let val = idiomise fc itm
+           logRaw "desugar.idiom" 10 "Desugared to" val
+           pure val
   desugarB side ps (PList fc args)
       = expandList side ps fc args
   desugarB side ps (PPair fc l r)
--- a/tests/idris2/basic032/Idiom.idr
+++ b/tests/idris2/basic032/Idiom.idr
@@ -12,4 +12,3 @@ export
 
 addm : Maybe Int -> Maybe Int -> Maybe Int
 addm x y = [| x + y |]
-
--- a/src/TTImp/BindImplicits.idr
+++ b/src/TTImp/BindImplicits.idr
@@ -105,11 +105,7 @@ doBind ns (IQuote fc tm)
 doBind ns (IUnquote fc tm)
     = IUnquote fc (doBind ns tm)
 doBind ns (IAlternative fc u alts)
-    = IAlternative fc (doBindAlt u) (map (doBind ns) alts)
-  where
-    doBindAlt : AltType -> AltType
-    doBindAlt (UniqueDefault t) = UniqueDefault (doBind ns t)
-    doBindAlt u = u
+    = IAlternative fc (mapAltType (doBind ns) u) (map (doBind ns) alts)
 doBind ns tm = tm
 
 export
--- a/src/TTImp/TTImp.idr
+++ b/src/TTImp/TTImp.idr
@@ -4,6 +4,7 @@ import Core.Binary
 import Core.Context
 import Core.Env
 import Core.Normalise
+import Core.Options
 import Core.Options.Log
 import Core.TT
 import Core.TTC
@@ -1010,3 +1011,16 @@ mutual
                8 => do n <- fromBuf b
                        pure (ILog n)
                _ => corrupt "ImpDecl"
+
+
+-- Log message with a RawImp
+export
+logRaw : {auto c : Ref Ctxt Defs} ->
+         String -> Nat -> Lazy String -> RawImp -> Core ()
+logRaw str n msg tm
+    = do opts <- getSession
+         let lvl = mkLogLevel str n
+         if keepLog lvl (logLevel opts)
+            then do coreLift $ putStrLn $ "LOG " ++ show lvl ++ ": " ++ msg
+                                          ++ ": " ++ show tm
+            else pure ()
--- a/src/TTImp/TTImp.idr
+++ b/src/TTImp/TTImp.idr
@@ -380,6 +380,11 @@ data ImpREPL : Type where
      Quit : ImpREPL
 
 export
+mapAltType : (RawImp -> RawImp) -> AltType -> AltType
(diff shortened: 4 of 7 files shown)


src/Idris/Desugar.idr
```diff
@@ -118,6 +118,8 @@ bindBangs ((n, fc, btm) :: bs) tm
                                 (Implicit fc False) tm)
 
 idiomise : FC -> RawImp -> RawImp
+idiomise fc (IAlternative afc u alts)
+  = IAlternative afc (mapAltType (idiomise afc) u) (idiomise afc <$> alts)
 idiomise fc (IApp afc f a)
```
gallais, line 123: Should the `IApp` case also commute with `IAlternative`?
i.e. something like

```suggestion
idiomise fc (IApp afc f a)
-    = IApp fc (IApp fc (IVar fc (UN "<*>"))
-                    (idiomise afc f))
-                    a
+    = let app = \ fn => IApp fc (IVar fc (UN "<*>")) fn a in
+      case idiomise afc f of
+        IAlternative loc alt ts => IAlternative loc alt (map app ts)
+        tm => app tm

File: tests/idris2/basic032/Idiom.idr [hunk 2]
```diff
@@ -12,4 +12,3 @@ export
 
 addm : Maybe Int -> Maybe Int -> Maybe Int
 addm x y = [| x + y |]
-
```
gallais, line 15: If you add an example like

```idris
pairing : Applicative f => f a -> f b -> f (a, b)
pairing a b = [| (a, b) |]
```

at the bottom here, you get an ambiguity error. I wonder whether
it is solved if you make `IAlternative` and `<*>` commute like I
suggested above.

src/Idris/Desugar.idr
gallais: (forgot the `mapAltType` in the `IAlternative` branch)

tests/idris2/basic032/Idiom.idr
MarcelineVQ: There's more going on here than I expected at first in regards to IAlternative and how idris typechecks them.
What we (roughly) get for `pairing : Applicative f => f a -> f b -> f (a, b)` with the current fix is:
```idris
IAlternative fc
  (UniqueDefault
    (IApp fc
      (IApp fc
        (IVar fc (UN "<*>"))
        (IApp fc
          (IApp fc
            (IVar fc (UN "<*>"))
            (IApp fc
              (IVar fc (UN "pure"))
              (IVar fc (NS (MkNS ["Builtin"]) (UN "MkPair")))))
          (IVar fc (UN "x"))))
      (IVar fc (UN "y"))))
    [IApp fc
      (IApp fc
        (IVar fc (UN "<*>"))
        (IApp fc
          (IApp fc
            (IVar fc (UN "<*>"))
            (IApp fc
              (IVar fc (UN "pure"))
              (IVar fc (NS (MkNS ["Builtin"]) (UN "Pair")))))
          (IVar fc (UN "x"))))
      (IVar fc (UN "y"))
     ,IApp fc
       (IApp fc
         (IVar fc (UN "<*>"))
         (IApp fc
           (IApp fc
             (IVar fc (UN "<*>"))
             (IApp fc
               (IVar fc (UN "pure"))
                 (IVar fc (NS (MkNS ["Builtin"]) (UN "MkPair")))))
                 (IVar fc (UN "x"))))
           (IVar fc (UN "y"))]
```
And what would actually work is if instead we could produce:
```idris
(IApp fc
  (IApp fc 
    (IVar fc (UN "<*>"))
    (IApp fc
      (IApp fc
        (IVar fc (UN "<*>"))
        (IApp fc
          (IVar fc (UN "pure"))
          (IAlternative fc (UniqueDefault (IVar fc (NS (MkNS ["Builtin"]) (UN "MkPair"))))
             [(IVar fc (NS (MkNS ["Builtin"]) (UN "MkPair"))), (IVar fc (NS (MkNS ["Builtin"]) (UN "Pair")))])))
      (IVar fc (UN "x"))))
  (IVar fc (UN "y")))
```
It's unclear to me why the first version is ambiguous when the type of `pairing` should inform us we're after MkPair where we're using `[| (a, b) |]`. The second version was accepted as the body for `pairing` via feeding it to `%runElab check` however.

I wasn't quite sure how to write your given code to see if it helped because it's just a little off, `app` in particular, but I tried some variations without luck.

gallais: I have tried to do something like this:

```diff
+idiomisePure : FC -> RawImp -> RawImp
+idiomisePure fc fn = IApp fc (IVar fc (UN "pure")) fn
+
 idiomise : FC -> RawImp -> RawImp
-idiomise fc (IAlternative afc u alts)
-  = IAlternative afc (mapAltType (idiomise afc) u) (idiomise afc <$> alts)
+idiomise fc fn@(IAlternative afc u alts)
+  = if any (isJust . isIApp) alts
+   -- if we have at least an `App` then we cannot do any better than idiomising
+   -- under the alternative
+    then IAlternative afc (mapAltType (idiomise afc) u) (idiomise afc <$> alts)
+    -- if there is no `App` however we know all of these will be caught by the
+    -- pure case and we can put the pure layer on top of the `IAlternative` one.
+    -- This means the applicative can get disambiguated first, and the other ambiguity
+    -- second!
+    else idiomisePure fc fn
 idiomise fc (IApp afc f a)
     = IApp fc (IApp fc (IVar fc (UN "<*>"))
                     (idiomise afc f))
-                    a
-idiomise fc fn = IApp fc (IVar fc (UN "pure")) fn
+              a
+idiomise fc fn = idiomisePure fc fn
```

and then desugaring a `PPair` to (an `IAlternative` of the two constructors) applied
to the arguments rather than an `IAlternative` of (the two constructors applied to the
arguments):

```diff
   desugarB side ps (PPair fc l r)
       = do l' <- desugarB side ps l
            r' <- desugarB side ps r
-           let pval = apply (IVar fc mkpairname) [l', r']
-           pure $ IAlternative fc (UniqueDefault pval)
-                  [apply (IVar fc pairname) [l', r'], pval]
+           let def = IVar fc mkpairname
+           pure $ apply (IAlternative fc (UniqueDefault def) [IVar fc pairname, def])
+                        [l', r']
```

However this last change breaks *a lot* of things as we pick an `IAlternative`
based on its content only & then commit to it. So `IAlternative` between two
vars will always succeed with the first one AFAICT!

I don't really see a good way to fix this problem so that `pairing` works.
I would be in favour of merging as-is as it's already an improvement on the
current state.

gallais:
Oh wait, I do want to try my `IApp`-`IAlternative` commuting trick.

MarcelineVQ:
I'm not sure why the racket test output is wrong on the checks here, I was not able to reproduce that

gallais:
Is it because `exec` is trying to run the thing using the scheme backend?

MarcelineVQ:
Even removing chez from my system I couldn't reproduce racket's failure to print in the test.
I've modified the test so it won't matter, it's still testing this change properly, just not quite as thoroughly.

clayrat:
It's still a bit annoying to write `[| (foo, [| (bar, [| (baz, fnord) |] ) |] ) |]` for n-ary tuples. 

I wonder if we should be able to write `[| (foo, bar, baz, fnord) |]` and auto-expand that?

gallais:
Note that should be able to write it prefix: `[| (,,,) foo bar baz fnord |]`

clayrat:
Ah thanks! Do you think having a (more) sugared form is useful?

gallais:
Yeah it would be nice. But given that tuples and their sections are
currently desugared into nested pairs directly in the parser, it is
probably quite a bit of work to make this structure explicit and
observable later in the pipeline so that the desugarer can take
advantage of it.

Issue #645, opened 2020-09-06
Totality checker is slow to figure out some functions aren't total (which also are total in Idris 1)


0xd34df00d:
The following example ( https://gist.github.com/0xd34df00d/33b82447f11bc7761744f556fb01ba58 , in particular, the mutually recursive implementation of `Sized`) takes about 30 seconds to typecheck on my machine. Changing `%default total` to `%default covering` brings this down to "almost zero" number of seconds.

I also believe that the totality checker is overly pessimistic here, as this seems to be total (and I'd expect a totality checker to figure out this is total). Also, Idris 1's totality checker accepts this as total (with irrelevant modifications like the import list). But, perhaps, that's a separate issue.

I also have examples where the totality checker takes 10-20 _minutes_ to conclude something probably isn't total, but those examples are huge, and the more I strip off, the faster the (negative) totality check becomes.

# Steps to Reproduce

1. Copy the contents of the above gist over to a file, say, `Test.idr`.
2. Run `idris2 --check Test.idr`.

# Expected Behavior

Instantaneous type check. Ideally, resulting in "your stuff is total".

# Observed Behavior

Dozens of seconds spent in the totality checker, resulting in
```
Error: size is not total, possibly not terminating due to recursive
path Main.size -> Main.Sized implementation at Test.idr:70:3--74:17 -> Main.size -> Main.Sized implementation at Test.idr:74:3--84:17 -> Main.size -> Main.Sized implementation at Test.idr:70:3--74:17 -> Main.size -> Main.Sized implementation at Test.idr:74:3--84:17 -> Main.size -> Main.Sized implementation at Test.idr:84:3--90:1 -> Main.size -> Main.Sized implementation at Test.idr:70:3--74:17 -> Main.size -> Main.Sized implementation at Test.idr:74:3--84:17 -> Main.size -> Main.Sized implementation at Test.idr:84:3--90:1 -> Main.size -> Main.size -> Main.Sized implementation at Test.idr:74:3--84:17 -> Main.size -> Main.Sized implementation at Test.idr:84:3--90:1 -> Main.size -> Main.Sized implementation at Test.idr:70:3--74:17 -> Main.size

Test.idr:84:3--90:1
 84 |   implementation Sized (T g e t) where
 85 |     size eprf = S $ case eprf of
 86 |                          T_Unit gok => size gok
 87 |                          T_Var gok _ => size gok
 88 |                          T_Abs arrTy bodyTy => size arrTy + size bodyTy
[...]

PR #679, opened 2020-09-18
Added some logging information to the totality checker.

jfdm:
This will hopefully make debugging of the totality/termination checker that little bit easier, especially given #645.

I need someone to check the levels, as I think I have sett them to high...

### DIFF
--- a/src/Core/Termination.idr
+++ b/src/Core/Termination.idr
@@ -43,7 +43,7 @@ export
 checkIfGuarded : {auto c : Ref Ctxt Defs} ->
                  FC -> Name -> Core ()
 checkIfGuarded fc n
-    = do log "termination.guarded" 6 $ "Check if Guarded: " ++ show n
+    = do log "totality.guarded" 6 $ "Check if Guarded: " ++ show n
          defs <- get Ctxt
          Just (PMDef _ _ _ _ pats) <- lookupDefExact n (gamma defs)
               | _ => pure ()
@@ -363,7 +363,7 @@ mutual
       = do Just gdef <- lookupCtxtExact fn_in (gamma defs)
                 | Nothing => throw (UndefinedName fc fn_in)
            let fn = fullname gdef
-           log "termination.termination.sizechange" 10 $ "Looking under " ++ show fn
+           log "totality.termination.sizechange" 10 $ "Looking under " ++ show fn
            aSmaller <- resolved (gamma defs) (NS builtinNS (UN "assert_smaller"))
            cond [(fn == NS builtinNS (UN "assert_total"), pure []),
               (caseFn fn,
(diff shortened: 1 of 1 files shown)


src/Core/Termination.idr
```diff
@@ -362,7 +363,7 @@ mutual
       = do Just gdef <- lookupCtxtExact fn_in (gamma defs)
                 | Nothing => throw (UndefinedName fc fn_in)
            let fn = fullname gdef
-           log "termination" 10 $ "Looking under " ++ show fn
+           log "termination.termination.sizechange" 10 $ "Looking under " ++ show fn
```
gallais, line 366: `termination.termination`?

jfdm: i was carrying on from the existing topic! Maybe we should change the lead topic to be `totality`?

gallais: I see. I think `totality` would indeed be clearer. Or just not having `termination` and
`positivity` under the same heading (we can always use multiple topics anyways).

jfdm: TBH I think what is also happening is that I assume that topics are presented hierarchically i.e. "grandparent.parent.child". This is probably where our mental models differ. Regardless the change has been made.

src/Core/Termination.idr
```diff
@@ -410,7 +410,7 @@ export
 calculateSizeChange : {auto c : Ref Ctxt Defs} ->
                       FC -> Name -> Core (List SCCall)
 calculateSizeChange loc n
-    = do log "termination.sizechange" 5 $ "Calculating Size Change: " ++ show n
+    = do log "totality.sizechange" 5 $ "Calculating Size Change: " ++ show n
```
gallais, line 413: This should probably be `totality.termination.sizechange` for uniformity
or is it an unrelated pass (in which case a different name would be clearer)?

jfdm: I think a different name is required, same for the one for guarded. They don't appear as subcalls to existing functions in the module and are both exported.

gallais: Tbh I tend to think of these in terms of topics, not call graphs.

jfdm: I've made the appropriate changes, and think that his is good to go.

src/Core/Termination.idr
```diff
@@ -43,7 +43,7 @@ export
 checkIfGuarded : {auto c : Ref Ctxt Defs} ->
                  FC -> Name -> Core ()
 checkIfGuarded fc n
-    = do log "totality.guarded" 6 $ "Check if Guarded: " ++ show n
+    = do log "totality..termination.guarded" 6 $ "Check if Guarded: " ++ show n
```
gallais, line 46: two dots

jfdm: urgh

PR #681, opened 2020-09-19
WIP: More backend-independence

ziman:
This patch is a potpourri of things required to self-host Idris2 on the Malfunction backend, and I expect that it will be useful for other non-Scheme backends, too (such as the OCaml backend).

In patch order, the changes are:
* Use `fastUnpack` in the lexer. (The original way was too slow.)
* Extend `contrib/Data.SortedSet` with `singleton` and `null`. Needed in the code of the Malfunction backend.
* Create `contrib/Data.String.Iterator` for backend-independent iteration over characters. Required for `src/Core.Hash`, which used index-based random access (!) into `String`s. This may work for UTF-32-encoded Chez strings but does not work for UTF-8-encoded strings, which I suspect would be a popular choice for backends other than Malfunction, too.
* Extend `src/Data.StringMap` with `adjust`. Needed in the code of the Malfunction backend.
* Install C headers into `${libdir}/include/*.h` to match the existing `${libdir}/lib/*.{so,a}`. Set appropriate file permissions. Needed for any backend that compiles C against `libidris2_support`.
* Fix the C buffer code and make the types more explicit. Needed for binary compatibility with ML-based binary primitives: TTC files written by Chez can now be read by ML and vice versa.
* Implement the new foreign functions in the Scheme support code, too.


ziman:
There's something strange going on with buffers.

I'll open another PR once I sort that out.

PR #683, opened 2020-09-19
More backend independence

ziman:
This patch is a potpourri of things required to self-host Idris2 on the Malfunction backend, and I expect that it will be useful for other non-Scheme backends, too (such as the OCaml backend).

In patch order, the changes are:
* Use `fastUnpack` in the lexer. (The original way was too slow.)
* Extend `contrib/Data.SortedSet` with `singleton` and `null`. Needed in the code of the Malfunction backend.
* Create `contrib/Data.String.Iterator` for backend-independent iteration over characters. Required for `src/Core.Hash`, which used index-based random access (!) into `String`s. This may work for UTF-32-encoded Chez strings but does not work for UTF-8-encoded strings, which I suspect would be a popular choice for backends other than Malfunction, too.
* Extend `src/Data.StringMap` with `adjust`. Needed in the code of the Malfunction backend.
* Install C headers into `${libdir}/include/*.h` to match the existing `${libdir}/lib/*.{so,a}`. Set appropriate file permissions. Needed for any backend that compiles C against `libidris2_support`.
* Fix the C buffer code and make the types more explicit. Needed for binary compatibility with ML-based binary primitives: TTC files written by Chez can now be read by ML and vice versa.
* Implement the new foreign functions in the Scheme support code, too.


libs/base/Data/Strings.idr
```diff
@@ -16,11 +16,16 @@ partial
 foldl1 : (a -> a -> a) -> List a -> a
 foldl1 f (x::xs) = foldl f x xs
 
+%foreign
+  "scheme:string-unpack"
+export
+fastUnpack : String -> List Char
+
```
gallais, line 23: I think we could use a comment explaining when / why to pick
`fastUnpack` over `unpack`.

ziman: Thanks, I added one.

chrrasmussen:
I welcome the changes to improve the support for UTF-8 strings (as that is how strings are represented in Erlang). I do, however, have one question regarding how `StringIterator` currently works: Would it be okay to change `StringIterator` it to not be based on an offset, but rather give the backend an opportunity to choose what data are passed to each step?

The reason I am asking is that Erlang does not have an easy way to access characters at a certain offset. Instead, the way I would do it is to ask for the [`string:next_grapheme/1`](https://erlang.org/doc/man/string.html#next_grapheme-1) (or [`string:next_codepoint/1`](https://erlang.org/doc/man/string.html#next_codepoint-1)) which returns a grapheme/codepoint (the head) and the rest of the string (the tail). So, in the case of Erlang, I would pass "the rest of the string" as the payload for the next call to `prim__readChar`.

<details>
  <summary>Full code</summary>

```idris
module Data.String.Iterator

export
data StringIterator : Type where [external]

private
%foreign "..."
prim__initStringIterator : String -> StringIterator

export
fromString : String -> StringIterator
fromString = prim__initStringIterator

private
data ReadResult
  = EOF
  | Character Char StringIterator

-- Runs in O(1) time.
private
%foreign "scheme:read-string-char"
prim__readChar : StringIterator -> ReadResult

export
uncons : StringIterator -> Maybe (Char, StringIterator)
uncons strIter =
  case prim__readChar strIter of
    EOF => Nothing
    Character ch nextStrIter => Just (ch, nextStrIter)

export
foldl : (a -> Char -> a) -> a -> String -> a
foldl f acc s = loop (prim__initStringIterator s) acc
  where
    loop : StringIterator -> a -> a
    loop strIter acc =
      case prim__readChar strIter of
        EOF => acc
        Character ch nextStrIter => loop nextStrIter (f acc ch)
```
</details>

<details>
  <summary>The diff</summary>

```diff
 module Data.String.Iterator
 
 export
-record StringIterator where
-  constructor MkSI
-  string : String
+data StringIterator : Type where [external]
 
-  -- backend-dependent offset into the string
-  -- see prim__readChar below
-  offset : Int
+private
+%foreign "..."
+prim__initStringIterator : String -> StringIterator
 
 export
 fromString : String -> StringIterator
-fromString s = MkSI s 0
+fromString = prim__initStringIterator
 
 private
 data ReadResult
   = EOF
-  | Character Char Int  -- character + width in backend-dependent units
+  | Character Char StringIterator
 
 -- Runs in O(1) time.
--- Takes a backend-dependent offset into the string.
--- On ML-based backends, this is in bytes;
--- in Scheme, this is in codepoints.
 private
 %foreign "scheme:read-string-char"
-prim__readChar : Int -> String -> ReadResult
+prim__readChar : StringIterator -> ReadResult
 
 export
 uncons : StringIterator -> Maybe (Char, StringIterator)
-uncons (MkSI s ofs) =
-  case prim__readChar ofs s of
+uncons strIter =
+  case prim__readChar strIter of
     EOF => Nothing
-    Character ch width => Just (ch, MkSI s (ofs + width))
+    Character ch nextStrIter => Just (ch, nextStrIter)
 
 export
 foldl : (a -> Char -> a) -> a -> String -> a
-foldl f acc s = loop 0 acc
+foldl f acc s = loop (prim__initStringIterator s) acc
   where
-    loop : Int -> a -> a
-    loop ofs acc =
-      case prim__readChar ofs s of
+    loop : StringIterator -> a -> a
+    loop strIter acc =
+      case prim__readChar strIter of
         EOF => acc
-        Character ch width => loop (ofs + width) (f acc ch)
+        Character ch nextStrIter => loop nextStrIter (f acc ch)
```
</details>

To make this code work with Scheme (+ ML and others) they would implement the function `prim__initStringIterator`, returning a reference to the string and the the offset `0`. Also, `prim__readChar` would itself calculate the next offset. These two steps where previously done by Idris.

If this change is okay for you, @ziman, and the other maintainers of Idris 2, I can flesh out the implementation for Scheme and JavaScript.

Comments about the code:
- `foldl` is currently not total (It wasn't originally either). I can fix this.
- The type `ReadResult` is more or less the same type as `Maybe (Char, StringIterator)`, but I am not sure what is preferrable for the backend. `ReadResult` requires the backend to know about 1 data type, while `Maybe (Char, StringIterator)` requires the backend to know about 2 data types (`Maybe` and `Pair`).

ziman:
Yes, that's a very good idea, I think! I simplified the code a lot by making `StringIterator` abstract, ditching `ReadResult` and the `prim__*` functions. I also tested the iterator on the three Scheme backends.

I could use help with the JS implementation: from a brief web search, it's not clear to me how to implement it efficiently.

Not sure about that `foldl`. Do we just assert its totality? Are there cases where it would not terminate but it's not a bug?

ziman:
Would it be better to replace `StringIterator` with a lazy list, maybe?

ziman:
[A version with a lazy list](https://github.com/ziman/idris2/commit/f73fa550758e4cbc48f95614f65a5cfb23c5bfa9).

ziman:
Another thing is that `ReadResult` unboxed the tuple into the `Character` constructor. The representation of `Just (char, (string, next_offset))` is `(1, (0, char, (string, next_offset)))`, while the representation of `Character char (string, next_offset)` would be `(1, char, (string, next_offset))`, so you save one pointer indirection in something that could quite likely be a hot loop.

I haven't benchmarked it, though.

chrrasmussen:
Thanks! 😃 The use of `LazyList` looks neat.

Apparently, JavaScript represents strings as UTF-16. (Source: https://mathiasbynens.be/notes/javascript-unicode)

I think the following example can be adapted for the JavaScript codegen: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/@@iterator. I am looking into the JavaScript implementation now.

I see what you mean regarding the performance of `ReadResult` vs. `Maybe`/`Pair` in `foldl`. I also don't know if it is significant or not.

gallais:
> Do we just assert its totality?

If you have primitives reading a whole file as a string then you have
the threat of infinite strings existing (e.g. `/dev/urandom`).

chrrasmussen:
I just now realized that using the JavaScript string iterator is a bad idea. It is mutating the iterator object. I will try to find another approach.

@gallais: In that case, wouldn't the problem be with the read file function? I would think that the read file function will not return. Looking at `System.File.readFile` it is not considered total.

chrrasmussen:
Here is an implementation for the JavaScript code generator: https://github.com/chrrasmussen/Idris2/commit/84f398a25d58e854de4afac0c3c85cf4e9f82348

The code is using `str.codePointAt()`. The lookup is fast, but it turns out it only handles code *point* characters that fit inside a single UTF-16 code *unit*. A code *point* can require up to 21 bits, which means that they need two UTF-16 code *units*. The JavaScript code generator is already using `codePointAt()`, so maybe it is okay for now.

I did also try `Array.from()` which handles code points in a better way, but there was a significant delay when converting the string to an array. In the following example, the code that uses `Array.from()` is 10x slower.
```javascript
var arr = 'a'.repeat(1_000_000);
var arr = Array.from('a'.repeat(1_000_000));
```

@ziman: Feel free to cherry-pick the commit or just copy/paste the code into your branch. If you prefer, I can also create a pull request after this one is merged.

ziman:
So what about [this](https://github.com/idris-lang/Idris2/commit/d26a9c55bf293cbb3dbe416a2939a377dae12083)? It allows mutating iterators.

chrrasmussen:
Very cool! Here is an implementation for JavaScript using the string iterator: https://github.com/chrrasmussen/Idris2/commit/ad5e77e5ff7851a0ac8cbbd88ab1c3bb8e256843

ziman:
Any objections? I'd merge this soon.

PR #687, opened 2020-09-20
Fix String->Int cast on Scheme backends

Russoul:
Numbers in Scheme have an internal flag: exact/inexact.
This fix flags numbers outputted from `floor` when casting between `String` and `Int` as exact.

Fixes `cast {to = Int} "12.3"` reducing to `12.0` (internally)

I have tested it on my machine with Chez as a backend. I can't properly install Racket for some reason.
So if someone reading this has Racket in working order, could you please check this ?

Not sure how to write a test in such case, as the fix only concerns the internal representation, nor that a test is really needed.

### DIFF
--- a/support/chez/support.ss
+++ b/support/chez/support.ss
@@ -46,18 +46,18 @@
       ((equal? x "") "")
       ((equal? (string-ref x 0) #\#) "")
       (else x))))
+(define exact-floor
+  (lambda (x)
+    (inexact->exact (floor x))))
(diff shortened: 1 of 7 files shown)


gallais:
Seems to work when compiled with racket on my machine.
The added test should make sure it works with all the builds we have.

Russoul:
Thanks @gallais :) Silly me, I thought is was hard to test...

PR #691, opened 2020-09-22
[ fix #63 ] interleaving let binders and local declarations

### LINKED ISSUE #63: one type annotation in let means all other definitions must have type annotation

gallais:


src/Utils/String.idr
```diff
@@ -11,3 +11,8 @@ dotSep (x :: xs) = x ++ concat ["." ++ y | y <- xs]
 export
 stripQuotes : (str : String) -> String
 stripQuotes str = substr 1 (length str `minus` 2) str
+
+export
+lowerFirst : String -> Bool
```
Russoul, line 16: There is a duplicate definition of this function at [src/TTImp/Utils.idr](https://github.com/idris-lang/Idris2/blob/d105dd11a7e76dae28c34a43f931608b0e08c95e/src/TTImp/Utils.idr#L12)

gallais: Nice catch!

PR #692, opened 2020-09-22
Remove FArgList

chrrasmussen:
The PR #690 reminded me about `FArgList`, which is not needed anymore (since #595).


gallais:
Thanks!

PR #695, opened 2020-09-23
Division theorem

ohad:
Division Theorem. For every natural number `x` and positive natural
number `n`, there is a unique decomposition:
`x = q*n + r`
with `q`,`r` natural and `r` < `n`.

`q` is the quotient when dividing `x` by `n`
`r` is the remainder when dividing `x` by `n`.

This commit adds a (very hacky, for now) proof for this fact, in case
we want to reason about modular arithmetic (for example, when dealing
with binary representations).

In the process, it bulks up the stdlib with:
+ a generic `PreorderReasoning` module for arbitrary preorders,
analogous for the equational reasoning module
+ some missing facts about `Nat` operations.

I'll cleaning this code up before merging. In the meanwhile, I'd be grateful for another pair of eyes looking for:
+ things that should go elsewhere
+ inconsistent naming conventions
+ different ways to organise this


libs/base/Data/Nat.idr
```diff
@@ -170,37 +170,45 @@ export
...
+mod' Z        centre right = centre
+mod' (S fuel) centre right =
       if lte centre right then
         centre
       else
-        mod' left (minus centre (S right)) right
+        mod' fuel (minus centre (S right)) right
+
+
+public export total
```
gallais, line 186: The default is already `total`

libs/contrib/Data/Nat/Division.idr
```diff
@@ -0,0 +1,340 @@
...
+import Decidable.Equality
+import Decidable.Order
+import Decidable.Order.Strict
+
+%default total
+
+-- Need a few more facts about order
+-- The converse to lteIsLTE:
+LteIslte : (a, b : Nat) -> a `LTE` b -> a `lte` b = True
+                          -- DYI Syntex.PreorderReasoning from contrib
```
gallais, line 20: Typo: `Synt(s/e/a)x`

gallais (commented):
I haven't looked at all of the functions in detail, one important question first:
would it make sense to have a `divmod'` that does *both* `div'` and `mod'` at the
same time?

libs/base/Data/Nat.idr
ohad: Fixed, thanks.

libs/contrib/Data/Nat/Division.idr
ohad: Fixed, thanks!

ohad:
> I haven't looked at all of the functions in detail, one important question first:
> would it make sense to have a `divmod'` that does _both_ `div'` and `mod'` at the
> same time?

It would make sense, and I'm happy to push it through all the proofs. 
Do we really want to compute 'div' whenever we compute 'mod' and the other way around?

For example, I could instead:
1. have a `divmod'` function as you suggest
2. prove that `div'` and `mod'` are extensionally equal to its components,
3. do all the proofs w.r.t. `divmod'`
4. export the theorem using the extensionality lemmata of step 2.

Is that better?

One advantage is that the proof becomes more decoupled from the implementation in `Data.Nat.idr`: so long as the extensionality lemmata hold, the proof should stay the same.

gallais:
Sounds like a good game plan.

ohad:
> Sounds like a good game plan.

Done.

More comments welcome!

clayrat:
A few years I've made a port of Coq's binary arithmetics which included a few variants of division theorem, dunno if this might be helpful here? 

* https://github.com/sbp/idris-bi/blob/master/src/Data/Bin/DivMod.idr 
* https://github.com/sbp/idris-bi/blob/master/src/Data/Biz/DivMod.idr
* https://github.com/sbp/idris-bi/blob/master/src/Data/BizMod2/DivMod.idr

ohad:
Thanks!
That's a lot of rewrites! I usually move to equational reasoning at that point, because I stop following what's happening with all the rewriting.

I skimmed through this quickly --- I don't see a uniqueness theorem for the decomposition. 
Did I miss it? Perhaps that follows from the `gcd` properties?

libs/base/Data/Nat.idr
```diff
@@ -586,6 +627,18 @@ sucMinR : (l : Nat) -> minimum l (S l) = l
 sucMinR Z = Refl
 sucMinR (S l) = cong S $ sucMinR l
 
+-- Proofs on lte -----------------------
+
```
gallais, line 631: Longer term (i.e. in a different PR), it is probably a good idea to think about introducing

```idris
data Reflects : Type -> Bool -> Type where
  True : p -> Reflects p True
  False : Not p -> Reflects p false
```

This way with very similar code we should be able to prove both that
`lte a b = True -> LTE a b` but also that `lte a b = False -> Not (LTE a b)`.
And have generic results on `Reflects` that makes proving these kind of properties
easier (e.g. the fact that `Not (LTE a b) -> LT b a` should be enough to get
`notlteIsLT` from `lteIsLTE`).

libs/contrib/Data/Nat/Division.idr
```diff
@@ -0,0 +1,448 @@
...
+ LteIslte a b a_lt_b | (True  ** prf) = prf
+ LteIslte a b a_lt_b | (False ** prf) = void $ irreflexive {spo = Data.Nat.LT} a 
+   $ CalcWith {leq = LTE} $
+   |~ 1 + a
+   <~ 1 + b  ...(plusLteMonotoneLeft 1 a b a_lt_b)
+   <~ a      ...(notlteIsLT _ _ prf)
+
+-- The converse to lteIsLTE:
+GTIsnotlte : (a, b : Nat) -> b `LT` a -> a `lte` b = False
+                          -- DYI Syntex.PreorderReasoning from contrib
```
gallais, line 31: Typo: synt(e/a)x

libs/contrib/Data/Nat/Division.idr
```diff
@@ -0,0 +1,448 @@
...
+                     multLteMonotoneRight k    a     b   a_lte_b)
+  ~~ (1 + k) * b ...(Refl)
+
+multLteMonotoneLeft : (a, b, r : Nat) -> a `LTE` b -> a*r `LTE` b*r
+multLteMonotoneLeft a b r a_lt_b = 
+  rewrite multCommutative a r in
+  rewrite multCommutative b r in
+  multLteMonotoneRight r a b a_lt_b
+
+
```
gallais, line 83: All of these properties about `LTE` should probably be in a different module
e.g. `Data.Nat.Order.Properties`.

libs/base/Data/Nat.idr
ziman: > `False : Not p -> Reflects p false`

This might be slightly dangerous ;)

gallais: Ugh. I tried to be thorough in my renaming of the Agda code I copy/pasted :(

clayrat:
> Thanks!
> That's a lot of rewrites! I usually move to equational reasoning at that point, because I stop following what's happening with all the rewriting.
> 
> I skimmed through this quickly --- I don't see a uniqueness theorem for the decomposition.
> Did I miss it? Perhaps that follows from the `gcd` properties?

It only exists for `Biz`: https://github.com/sbp/idris-bi/blob/master/src/Data/Biz/DivMod.idr#L520, though I think it should be pretty straightforward to extract versions for `Bip` and `Bin`.

libs/contrib/Data/Nat/Division.idr
ohad: Argh, sorry. Have grepped for this now.

libs/base/Data/Nat.idr
ohad: Done now, pending PR #722

ohad:
> It only exists for `Biz`: https://github.com/sbp/idris-bi/blob/master/src/Data/Biz/DivMod.idr#L520, though I think it should be pretty straightforward to extract versions for `Bip` and `Bin`.

Ah, then I did miss it. Sorry!

libs/contrib/Data/Nat/Order/Properties.idr
```diff
@@ -0,0 +1,100 @@
...
+                       \prf' =>
+                         (invert $ replace {p = Reflects (S a `LTE` S b)} prf 
+                                 $ lteReflection (S a) (S b)) prf'
+  
+
+
+
+-- The converse to lteIsLTE:
+export
+LteIslte : (a, b : Nat) -> a `LTE` b -> a `lte` b = True
```
gallais, line 45: This is a direct consequence of the `Reflects` proof.
You should be able to prove a general theorem:

```idris
Reflects p b -> if c then p else Not p -> b = c
````

giving you both `LTE a b -> lte a b = True` and `Not (LTE a b) -> lte a b = False`

**Review by gallais — COMMENTED** [review]
A last refactoring suggestion.

File: libs/contrib/Data/Nat/Order/Properties.idr (hunk 6 above)
Comment (ohad) [reply libs/contrib/Data/Nat/Order/Properties.idr:43]: Done, thanks!

File: libs/contrib/Data/Nat/Order/Properties.idr [hunk 7]
```diff
@@ -38,31 +38,26 @@ notlteIsLT a b prf = notLTImpliesGTE
...
 LteIslte : (a, b : Nat) -> a `LTE` b -> a `lte` b = True
-                          -- DYI Syntax.PreorderReasoning from contrib
-LteIslte  a b a_lt_b with (the (x : Bool ** a `lte` b = x) (a `lte` b ** Refl))
- LteIslte a b a_lt_b | (True  ** prf) = prf
- LteIslte a b a_lt_b | (False ** prf) = void $ irreflexive {spo = Data.Nat.LT} a 
-   $ CalcWith {leq = LTE} $
-   |~ 1 + a
-   <~ 1 + b  ...(plusLteMonotoneLeft 1 a b a_lt_b)
-   <~ a      ...(Properties.notlteIsLT _ _ prf)
+LteIslte  a b a_lt_b = let u = reflect {c = True} (lteReflection a b) a_lt_b in ?hole
```
gallais, line 44: You've left a `?hole` :D

Also: if you systematically need to pass `c` explicitly then it's probably
worth to make it explicit in the original proof.

ohad: Sorry about that!

ohad: It's a new pattern for me, so I had to play around with it first.

ohad:
If everyone's comfortable with this, I'll merge it after it passes the tests.

PR #697, opened 2020-09-23
Use collect-rendezvous to trigger a collection

melted:
`collect` requires that only one thread is active, and will
throw an exception otherwise. This is hard to ensure outside
collect-request-handler. `collect-rendezvous` will handle that that and
run the request handler.

This way we can't pass a generation number to collect.. This could be done
a parameter saying if we're exiting if needed.

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -84,7 +84,7 @@ schHeader chez libs
     "(let ()\n"
 
 schFooter : String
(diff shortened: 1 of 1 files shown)


wenkokke:
I just built this pull request. All tests pass, and it solves the problems I've been having writing concurrent Idris2.

andrevidela:
If that's ok with everyone I'm going to close this since it's gone stale. This is jsut for tiding up the PR tracker, feel free to reopen if you want to work on it again! :)

PR #701, opened 2020-09-24
Add Bifunctor interface

Russoul:
Adds `Bifunctor` interface to `Prelude`
Implements Bifunctor for `Pair`, `Either` and `These`
Adds `mapSnd` and `mapHom` to `Prelude`
Renames `bimap` to `mapHom` in Idris/Error
Fixes spacing a bit

I've chosen the names for functions that act only on first or second object of a "pair" to be `fmap` and `smap`. I think they convey the meaning better than `first` and `second` used in haskell.

Tell me if I shall rename something


MarcelineVQ:
Good stuff! Names are the hardest part of programming.
I don't mind first/second and imo they are so named because they're less likely to clash with other short identifiers people come up with on the spot in their code.

I had a bit to say about whether this lives in `base` or `prelude` but I think it's fine being added to `prelude` here, having thought about it. I've wanted my choice of what to map over often enough that having it in scope always (i.e. in Prelude) seems appropriate to me.  
There is a conversation to have about organizing `prelude` vs `base` vs `Prelude` and how each should relate but it is its own conversation.

This could be a good opportunity to move `These` to base as well for public use. No pressure there, it'll happen anyway eventually. 

I probably wouldn't redefine things that case constructors like `Functor (Pair a)` in terms of `smap` without some performance testing to see if the indirection hurts. Without doing that myself I'm not sure it matters either though! Only mentioning it because idris by itself does not yet do much optimizing that I'm aware of.

This is an aside and not really a consideration/issue for this PR but I've also a habit of %inline for things that are merely constructors or merely aliased function calls. But haven't tested if this matters, so grain of salt. I'm only mentioning them in case it sparks an interest in people to test performance:
```idris
%inline
mapFst : (a -> c) -> (a, b) -> (c, b)
mapFst = fmap

-- not quite as applicable, just an example of what I mean by merely constructors.
%inline
left : a -> Either a b
left = Left
```

melted:
I don't think we should add new stuff directly to the prelude, only the absolutely most common things should go there. Is there any reason this has to be there?

Russoul:
@MarcelineVQ Yeah, `%inline` is definitely a must:

```idris
use : (a -> b) -> (b -> a) -> (a, b, c, d) -> (b, a, c, d)
use f g tuple4 = let q  = Main.mapFst f tuple4
                     q' = Main.mapSnd (Main.mapFst g) q
                  in q' 
```

and its casetree when `%inline` flag is off vs on:
```idris
   [{arg:0}, {arg:1}, {arg:2}, {arg:3}, {arg:4}, {arg:5}, {arg:6}]: (%let q (Main.mapFst [___, _
__, ___, !{arg:4}, !{arg:6}]) (%let q' (Main.mapSnd [___, ___, ___, (%lam {eta:0} (Main.mapFst [
___, ___, ___, !{arg:5}, !{eta:0}])), !q]) !q'))
```

```idris
   [{arg:0}, {arg:1}, {arg:2}, {arg:3}, {arg:4}, {arg:5}, {arg:6}]: (%let q (Main.fmap [___, ___
, ___, !{arg:4}, !{arg:6}]) (%let q' (Main.smap [___, ___, ___, (%lam {eta:0} (Main.fmap [___, _
__, ___, !{arg:5}, !{eta:0}])), !q]) !q'))
```

Now with concrete types:

```idris
test : (String, Nat) -> (Maybe String, String)
test = Main.mapSnd show . Main.mapFst Just
```

inlined:

```
[{ext:0}]: (%case (%case !{ext:0} [(%concase Builtin.MkPair Just 0 [{e:2}, {e:3}] (%con Builtin.M
kPair Just 0 [(%con Prelude.Types.Just Just 1 [!{e:2}]), !{e:3}]))] Nothing) [(%concase Builtin.MkPa
ir Just 0 [{e:2}, {e:3}] (%con Builtin.MkPair Just 0 [!{e:2}, (Prelude.Show.show [!{e:3}])]))] Nothi
ng)
```

I'll push an update with a bunch of inline pragmas where they seem appropriate.

@melted I asked @gallais before starting this PR. He felt like putting the interface into the Prelude. I am personally not predisposed to any choice of its residence.

@gallais What do you think about `These` ? Shall I move it to `base` ?

MarcelineVQ:
How are you viewing that case tree btw. logging? yaffle?

Russoul:
Logging. I made a Lua backend and have an option to print the trees of all global names on demand there. Very useful. That should probably be a part of Idris. I’ll look into how we can add a logger domain where we print compiled casetrees

gallais:
Moving `These` to base seems sensible.

libs/prelude/Prelude/Interfaces.idr
```diff
@@ -105,6 +105,37 @@ public export
...
+  bimap : (a -> c) -> (b -> d) -> f a b -> f c d
+  bimap f g = fmap f . smap g
+
+  ||| The action of the Bifunctor on morphisms pertaining to the first object
+  |||
+  ||| ````idris example
+  ||| fmap (\x => x + 1) (1, "hello") == (2, "hello")
+  ||| ````
+  |||
+  fmap : (a -> c) -> f a b -> f c b
```
gallais, line 127: This is confusing as `fmap` is usually a `Functor` method in Haskell.
I would prefer e.g. `mapLeft` and `mapRight` respectively.
In Agda's stdlib we use `map₁` and `map₂` but the explicit reference to
sides might be clearer (& is also a direct generalisation of the corresponding
functions on `Either`).

I would also be happy with `first` and `second` as they are established
names in Haskell AFAIK.

gallais:
Looking at the latest changes: if we were already using `mapFst` and `mapSnd` for
`Pair` and `These` then we should probably go with that.

Russoul:
No problem, I also thought about that

libs/prelude/Prelude/Types.idr
```diff
@@ -85,13 +85,19 @@ natToInteger (S k) = 1 + natToInteger k
...
+%inline
 public export
 Functor (Pair a) where
-  map f (x, y) = (x, f y)
+  map = mapSnd
 
 public export
-mapFst : (a -> c) -> (a, b) -> (c, b)
-mapFst f (x, y) = (f x, y)
+mapHom : (a -> b) -> (a, a) -> (b, b)
```
gallais, line 99: I am not convinced by this function. Why not have it for all `Bifunctor` or not at all?

Russoul: I don't know, overlooked it

PR #703, opened 2020-09-27
Preliminary datatypes for telescopes

ohad:
A telescope is the variable context in a dependently typed program.
Dependent types with free varialbes  are really only well-defined in a telescope.

A segment of a telescope is a tuple of types in that
telescope. These are usually implicit in formal developments, but
seem to be crucial to get a compositional account.
 
I've learned this from @pigworker on an [SPLV](http://www.macs.hw.ac.uk/splv/)'19 bus ride. 
A literary reference would be welcome. This paper is a start:
[Unifiers as Equivalences: Proof-Relevant Unification of Dependently Typed Data.](https://dl.acm.org/doi/10.1145/2951913.2951917)
Jesper Cockx, Dominique Devriese, Frank Piessens, ICFP'16.
but they don't seem to have segments and their right action on contexts (`|++` in the PR).

This PR so far also includes a preliminary library for n-ary congruence construct.
A key ingredient is n-ary dependent functions, compare with `base/Data.Fun` and:
[Generic level polymorphic n-ary functions](https://dl.acm.org/doi/10.1145/3331554.3342604). Guillaume Allais. TyDe 2019.

This is all based on discussions with @gallais , so even if this is new, it's likely something he already worked out somewhere but never got around to implement in Idris.

Here's an example of n-ary congruence. I included the no-go's to illustrate why we need dependent congruence.
```idris
module Main

import Data.Telescope

import Data.Fin
import Data.Nat
import Data.Vect
import Syntax.PreorderReasoning


||| Weakening is functorial
func : (i : Fin n) -> weakenN k (weakenN m i) ~=~ weakenN (m + k) i
-- func           FZ = Refl
-- --                  ^^^^  Can't solve constraint
-- --                        between: k and plus k m.
-- func           FZ = rewrite plusAssociative n m k in 
--                     Refl
-- --                  ^^^^ Can't solve constraint
-- --                       between: k and plus k m.
-- --                      
-- -- This does work: 
-- func {n = S n} FZ = rewrite plusAssociative n m k in Refl
-- -- But it doesn't scale: it becomes harder and harder
-- -- to guess in more involved cases. Idris struggles to work out the correct
-- -- implicit arguments and where to do the rewriting (see next case).

func {n = S n} FZ = keep $ Calc $ 
  |~ FZ {k = (n + m) + k}
  ~~ FZ {k = (n +(m  + k))} 
         -- We can't simply use FZ, we need a dependent congruence     
         -- ...(cong FZ (sym $ plusAssociative n m k))
         -- --       ^^ Mismatch between: Fin (S ?k) and ?t -> ?u.
         ...(Congruence.cong 
                       {delta}
                     (\u => FZ {k = u}) 
                    $ sym $ plusAssociative n m k)
  where
    delta : Segment 1 []
    delta = [const Nat]
             

-- func (FS i) = rewrite func i in ?hole
-- --            ^^^^^^^^^^^^^^^^ Rewriting by ?_ m k k i = ?x did not change
-- --                             type weakenN k (weakenN m (FS i)) = weakenN (m + k) (FS i).

func {n = S n} (FS i) = keep $ Calc $
  |~ FS {k =(n +  m)+ k } (weakenN k (weakenN m i))
  ~~ FS {k = n + (m + k)} (weakenN (m + k) i)    
            ...(Congruence.cong {n = 2} {delta=Delta}
                         {sy}
                         (\n, i => FS {k=n} i)
                         (sym $ plusAssociative n m k)
                         (func i))
  where
    Delta : Segment 2 []
    Delta = [const Nat, \case Empty =. n => Fin n]
    sy : SimpleFun Empty Delta Type
    sy n i = Fin (S n)
```
It's still not as sleek as I'd like it to be --- I'd like to get rid of the `where` clauses --- but it looks like we're in a good-enough zone, so a good place to share this.


libs/contrib/Data/Telescope/Telescope.idr
```diff
@@ -0,0 +1,55 @@
...
+  ||| A type with dependencies in the given context
+  public export
+  TypeIn : Telescope k -> Type
+  TypeIn gamma = (env : Environment gamma) -> Type
+  
+  ||| A tuple of values of each type in the telescope
+  public export
+  data Environment : Telescope k -> Type where
+    Empty : Environment []
+    (=.)  : Environment gamma -> ty env -> Environment (gamma -. ty)
```
gallais, line 35:
```suggestion
    (=.)  : (env : Environment gamma) -> ty env -> Environment (gamma -. ty)
```

I am actually surprised anything works given this issue?!

ohad: Oh wow, thanks! I'm surprised as h/well. Another inconsistency?

gallais:
Let's merge this as is. `contrib` content is experimental anyway so
it's fine if we break things when we change them in the future.

PR #708, opened 2020-09-28
[ new ] do not implicitly generalise names in scope

gallais:
The fact that

```
mapId : (xs : List a) -> map id xs === xs
```

used to mean

```
mapId : {0 id : a -> b} -> (xs : List a) -> map id xs === xs
```

depiste the fact a function `id : a -> a` is in scope was very confusing
for newcomers (and in obscure cases, experienced users too). With this
patch we only generalise over the variables that are not visible in the
current scope.


gallais:
Marked as draft because I have an issue when the function that should not be
generalised over is defined in a namespace at the moment.

ziman:
If I remember correctly, this was the behaviour of Idris 1 before @edwinb reverted it to the current behaviour due to problems it caused. I don't remember exactly what the issues were but I vaguely remember Edwin saying that now the rules may not be super convenient but are at least they are clear and local.

I am also a bit worried that with this patch, we may break user code by merely adding stuff to the standard library. Which is probably quite an inconvenient property to have in a programming language.

Of course, this probably needs solving somehow...

ziman:
[Here's an example](https://github.com/idris-lang/Idris-dev/pull/2232#issuecomment-99395039). This particular issue would go away with [my PR](https://github.com/idris-lang/Idris2/pull/662), though.

gallais:
> now the rules may not be super convenient but at least they are clear and local.

I am more concerned about the fact they seem to be wildly counter-intuitive as
we get about one new issue a month about it.

> I am also a bit worried that with this patch, we may break user code by merely
  adding stuff to the standard library. Which is probably quite an inconvenient
  property to have in a programming language.

This is already the case given that we may introduce ambiguities that cannot be resolved.
That being said, this rule is relying on the *scope* so as soon as people can have
explicit import lists, they can program defensively against such unintended scope
expansions!

edwinb:
The rules are a bit counter-intuitive but there are problems whatever we decide to do given the decision from years ago about implicitly binding names, which is hard to change now!

I think what I'd prefer is for the system to give a warning if it's about to implicitly bind an existing name, which Idris 1 tried to do, although it failed sometimes for reasons I never worked out. Then at least you'd know you needed to do something about it.

PR #713, opened 2020-09-30
contrib.Data.String.Parser updates

clayrat:
reexport identity monad, usability fixes


libs/contrib/Data/String/Parser.idr
```diff
@@ -209,22 +209,28 @@ covering
...
 parens : Monad m => ParseT m a -> ParseT m a
 parens p = char '(' *> p <* char ')'
 
-||| Discards whitespace after a matching parser
+||| Discards optional whitespace after a matching parser
 export
 covering
 lexeme : Monad m => ParseT m a -> ParseT m a
-lexeme p = p <* spaces
+lexeme p = p <* option () spaces
```
gallais, line 233: If you look at e.g. `token` the intention is really to have a non-empty amount of
space after the keyword so that e.g. `import` is a reserved keyword but `important`
is not recognised as `import`+`ant`.

libs/contrib/Data/String/Parser.idr
```diff
@@ -209,22 +209,28 @@ covering
 takeWhile : Monad m => (Char -> Bool) -> ParseT m String
 takeWhile f = pack <$> many (satisfy f)
 
+||| Similar to `takeWhile` but fails if the resulting string is empty.
+export
+covering
+takeWhile1 : Monad m => (Char -> Bool) -> ParseT m String
+takeWhile1 f = pack <$> some (satisfy f)
+
 ||| Parses one or more space characters
```
gallais, line 218: Similarly here we want *one* or more spaces.
You can always define `spaces?` as `optional spaces`

clayrat: Yes, and that is exactly what the `many` -> `some` change is about.

libs/contrib/Data/String/Parser.idr
clayrat: Makes sense in theory, but then the tests break. https://github.com/idris-lang/Idris2/blob/master/tests/chez/chez028/ExpressionParser.idr uses `lexeme integer` to parse `1+2`. I guess we could change either the test or the definition of `token` separately?

alexhumphreys: I originally wrote that test. `term` originally used `<* spaces`, and a future PR changed it to use lexeme. Can't remember for sure, but I believe my intention was to be able to parse `1+2` or `1 + 2`, so probably for the test it makes sense to use `<* option spaces`. But then yeah I guess lexeme should behave as @gallais said. Just my opinion, feel free to ignore it 😅

libs/contrib/Data/String/Parser.idr
gallais: Oh sorry, I will never not get confused by these two.

libs/contrib/Data/String/Parser.idr
clayrat: This only deals with spaces after numbers, but spaces after operators are still implied by the usage of `token` in these tables: https://github.com/idris-lang/Idris2/blob/master/tests/chez/chez028/ExpressionParser.idr#L14 

I'm not really sure what's the best design here, do we just introduce separate `lexeme0` and `token0` combinators?

alexhumphreys: Ah I was copying a bunch of stuff from Lightyear, and looks like `spaces` matches zero or more spaces: https://github.com/ziman/lightyear/blob/f737e25a09c1fe7c5fff063c53bd7458be232cc8/Lightyear/Char.idr#L47


And `Lexeme` there uses `spaces`
https://github.com/ziman/lightyear/blob/f737e25a09c1fe7c5fff063c53bd7458be232cc8/Lightyear/Strings.idr#L91 so `token` doesn't _require_ spaces after.

Not sure if that's common behaviour for a token parser.

clayrat: @gallais so what's your opinion - do we roll back to zero-or-more behaviour of `spaces`, introduce a separate set of `spaces0`/`lexeme0`/`token0` combinators or follow yet another road?

clayrat: Ok I've reverted the changes to the initial zero-or-more combinators.

clayrat:
Amidst all the bikeshedding I actually forgot that one of the main reasons for this PR was to add a combinator for one-or-more spaces :)

PR #716, opened 2020-10-01
Retrieve network address family preprocessor constants from C runtime code

sysvinit:
The current implementation of `ToCode SocketFamily` uses hardcoded constants, which are correct on Linux but not all correct on other systems (though I've only checked OpenBSD, as that's what I have to hand). For portability's sake, it would be a better idea to read the magic constants from the definitions in the system C library, so we don't need to special case every platform we want Idris to run on. I've implemented this by adding some getter functions to the C runtime support library (one for each constant we want to know about) and then calling the appropriate getter to decode each symbolic constant on the Idris side.


ziman:
Thanks!

PR #717, opened 2020-10-01
Suggest adding a Jupyter kernel for Idris

zenntenn:
After all, Agda has one ;) https://github.com/jupyter/jupyter/wiki/Jupyter-kernels


ohad:
Do we know what this involves doing?
Can we write the integration code in Idris? (I recall Jupyter doc saying this is possible)

ziman:
Looks like we can scratch off the preceding two items (OCaml, Node/JS backends), too. :)

PR #719, opened 2020-10-02
Make Idris2 test harness available for the many and not the few.

jfdm:
I have taken the liberty to copy the core functionality from Idris2 test harness into `contrib`.

The test harness provides the core functionality to run a *single* golden file test.
This allows the developer freedom to design the rest of the test harness.

Please see Idris2 test harness for example usage.

Once accepted we can update the bootstrap and remove the shared code from Idris2's own harness.


libs/contrib/Test/Golden.idr
```diff
@@ -0,0 +1,194 @@
...
+|||
+||| # Usage
+|||
+||| When compiled to an executable the expected usage is:
+|||
+|||```sh
+|||runtests <path to executable> [--interactive] [--only [NAMES]]
+|||```
+|||
+||| assuming that the test runner is compiler to an executable named `runtests`.
```
gallais, line 37: typo: compiler -> compiled

libs/contrib/Test/Golden.idr
```diff
@@ -0,0 +1,194 @@
...
+||| + `idris2` the path of the executable we are testing.
+||| + `onlyNames` the list of tests to run relative to the generated executable.
+||| + `interactive` Should we update the expected file or not.
+|||
+||| # Usage
+|||
+||| When compiled to an executable the expected usage is:
+|||
+|||```sh
+|||runtests <path to executable> [--interactive] [--only [NAMES]]
```
gallais, line 34: `<path to idris2>` maybe? The latest mention of an "executable" is talking
about `runtests` itself which confused me.

libs/contrib/Test/Golden.idr
```diff
@@ -0,0 +1,194 @@
...
+
+||| Run the specified Golden test with the supplied options.
+|||
+||| See the module documentation for more information.
+export
+runTest : Options -> String -> IO Bool
+runTest opts testPath
+    = do changeDir testPath
+         isSuccess <- runTest'
+         changeDir "../.."
```
gallais: This hardcodes the fact that the target paths are 2 levels of nesting deep.
And the fact that the golden test does not change that fact. Should we
do `pwd <- currentDir` (from `base`'s `System.Directory`) before `changeDir`
and change back to the saved `pwd` instead?

libs/contrib/Test/Golden.idr
```diff
@@ -0,0 +1,194 @@
...
+    = do Right ok <- openFile f Read
+             | Left err => pure False
+         closeFile ok
+         pure True
+
+firstExists : List String -> IO (Maybe String)
+firstExists [] = pure Nothing
+firstExists (x :: xs) = if !(exists x) then pure (Just x) else firstExists xs
+
+||| Find the first occurance of an executable on `PATH`.
```
gallais, line 189: typo: occurance -> occurrence

libs/contrib/Test/Golden.idr
```diff
@@ -0,0 +1,194 @@
...
+export
+pathLookup : List String -> IO (Maybe String)
```
gallais, line 191: These functions are not used in the library. And neither is `onlyNames`.
Perhaps we need a function that abstract away the job done by `main` in
Idris' test runner?

A `TestTree` could be:

* a list of requirements (e.g. the `"CHEZ"`, `"NODE"`, etc. dependencies)
* a list of FileNames

and the golden runner would take a `List TestTree` and for each one of them:

* check whether filtering by `onlyNames` leaves any test
* if so then check the requirements are available
* if so then traverse the list of filtered tests with `runTest`

gallais (commented):
Yes please! I have added a few comments.

libs/contrib/Test/Golden.idr
```diff
@@ -0,0 +1,194 @@
...
+|||
+||| This was originally used as part of Idris2's own test suite and
+||| the core functionality is useful for the many and not the few.
+||| Please see Idris2 test harness for example usage.
+|||
+||| # Test Structure
+|||
+||| This harness works from the assumption that each individual golden test comprises of a directory with the following structure:
+|||
+||| + `run.sh` a *shell* script that runs the test
```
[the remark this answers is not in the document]
jfdm: fixed in next commit.

libs/contrib/Test/Golden.idr
```diff
@@ -0,0 +1,194 @@
...
+||| + `expected` the expected output of `run.sh`
+|||
+||| During testing, the test harness will generate an artefact named `output` and use `git diff` to examine the two files.
+||| This assumes that the expected has been checked into `git`.
```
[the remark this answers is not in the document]
jfdm: fixed in next commit.

libs/contrib/Test/Golden.idr
```diff
@@ -0,0 +1,194 @@
...
+||| If `git diff` failes then the runner will simply present the expected and 'given' files.
```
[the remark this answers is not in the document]
jfdm: fixed in next commit.

libs/contrib/Test/Golden.idr
jfdm: changed to `path to executable under test`.

libs/contrib/Test/Golden.idr
jfdm: i've changed the test runner to take in a string rather than "../.." this allows the user to decide the behaviour rather than us.

libs/contrib/Test/Golden.idr
jfdm: Yes, I do not disagree but all this can be done in a later PR.

libs/contrib/Test/Golden.idr
jfdm: fixed in next commit.

libs/contrib/Test/Golden.idr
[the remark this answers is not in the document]
gallais: There are multiple mentions of `run.sh`. I only mentioned the first one my review
but all of them should be changed.

libs/contrib/Test/Golden.idr
gallais: I can add the changes if you want. It's so small (and justifies the presence
of currently unused features) it feels silly to split it in a different PR
altogether.

jfdm: go ahead!

libs/contrib/Test/Golden.idr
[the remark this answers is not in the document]
jfdm: that has now been fixed.

jfdm:
I am using this in anger now, and I have added some fixups and new features that arose from using it.

gallais:
I have pushed a wee thing but haven't tested it.

Maybe I should copy/paste this file as `Lib.hs` in `tests/` and see whether
I can reimplement `tests/Main.idr` using it. That sounds like a decent challenge.

jfdm:
Ah excellent! I think a local use of this for Idris2 would be healthy. All the other stuff was working fine for me here. I have not checked your specific changes.

gallais:
Seems to work like a charm on my machine!

gallais:
The new `runner` is all you need to run a test suite. In `tests/Main.idr` we now use:

```idris
main : IO ()
main = runner
  [ testPaths "ttimp" ttimpTests
  , testPaths "idris2" idrisTests
  , testPaths "typedd-book" typeddTests
  , testPaths "ideMode" ideModeTests
  , testPaths "prelude" preludeTests
  , testPaths "chez" chezTests
  , testPaths "node" nodeTests
  ] where

    testPaths : String -> TestPool -> TestPool
    testPaths dir = record { testCases $= map ((dir ++ "/") ++) }
```

PR #722, opened 2020-10-02
[contrib] Add a `Reflects` datatype

ohad:
As discussed in PR #695


ohad:
I'm expecting the controversial feature to be the module name.
Happy to move it elsewhere, just say where.

gallais (approved):

ohad:
So not controversial :-)

Squashing and merging.

gallais:
> So not controversial :-)

I don't have a good name for it either. And it's contribs so if we think
of something better at a later stage it's not a big deal to break things
by moving them around :D

PR #727, opened 2020-10-03
Remove invalid implicit in PreorderReasoning

Russoul:


ohad:
Thanks! I fixed it in one of my other branches, but I have several PRs in flight right now, and this should probably go on its own PR.

ohad (approved):

gallais:
How come Idris did not complain that `z` was not a valid name?

ohad:
Good point, I'll open a new bug.

PR #729, opened 2020-10-03
Revert addition of (>>) operator.

sysvinit:
See the GitHub comments on 7d5ec53b53d2, and the IRC conversation in the commit message. `(*>)` exists in `Applicative` already, so `(>>)` is redundant.

Thanks to @MarcelineVQ for pointing this out.

### DIFF
--- a/libs/prelude/Prelude/Interfaces.idr
+++ b/libs/prelude/Prelude/Interfaces.idr
@@ -176,10 +176,6 @@ interface Applicative m => Monad m where
 
 %allow_overloads (>>=)
 
-public export
-(>>) : (Monad m) => m a -> m b -> m b
-a >> b = a >>= \_ => b
-
 ||| `guard a` is `pure ()` if `a` is `True` and `empty` if `a` is `False`.
 public export
 guard : Alternative f => Bool -> f ()
--- a/libs/prelude/Prelude/Ops.idr
+++ b/libs/prelude/Prelude/Ops.idr
@@ -14,7 +14,7 @@ infixr 4 ||
 infixr 7 ::, ++
 
 -- Functor/Applicative/Monad/Algebra operators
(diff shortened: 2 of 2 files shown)


gallais:
I mean... yeah it is redundant but it's a very common function
in Haskell. Users may be reaching for it out of habit and it does
not cost us much to give it to them.

I experience a similar issue almost daily when I reach for `mapM`
and get a scope error before remembering that I can use `traverse`
instead to mean the same thing.

Personally I'd keep the changes.

MarcelineVQ:
People learning modern haskell are incentivised to use, *>, traverse, sequenceA, etc so I think it's a little silly to cater idris to haskell users, and to legacy haskell users in particular. haskell is a great model for a variety of things but it's a model to improve from not solely to emulate.
That's merely my opinion though, adding >> is also not such a big deal that it would harm anyone :>
If >> is later removed it's a very direct change for code maintainers at least.

gallais:
My reasoning is that I'd rather newcomers get tripped over things that are intrinsically
new and foreign in Idris rather than some stylistic issues about e.g. `*>` vs. `>>`.
When we have a linter, I would not oppose adding a rule saying "prefer `*>` to `>>`".

andrevidela:
One way this could be addressed would be to have a "haskell compatibility" library which mimics the expected Prelude from Haskell. I have not written any haskell in a long time so I do not know how much there would be to this library beyond `>>` and `mapM`. Would that be an interesting middle ground?

MarcelineVQ:
One could do that but it wouldn't address @gallais concern about easing in new people since they'd have to opt-in to that alternate Prelude. In the end this isn't an issue to spend too much collective effort on so I'm fine with closing this pr or otherwise leaving >> as it is for that reason :>

PR #730, opened 2020-10-04
add multi-way-if syntax sugar

MarcelineVQ:
Adds `if` of the form:
```idris
if | pred1 => expr1
   | pred2 => expr2
```
Refer to added documentation and tests to see its use.

Open questions/task still needing solving are:
How to emit warnings at desugar-time when we have unreachable cases, e.g.
```idris
earlyStop : Int -> Char
earlyStop i = if | i > 10 => '7'
                 | _ => '0'
                 | i == 2 => 'X'
                 | _ => 'Z'
```
How to commit the same level of reduction we do in types for the predicates in a multi-way if, allowing the following to be considered total even without a _ catchall case when combined with a check that at least one branch evaluates to True.
```idris
otherwise : Bool
otherwise = True

constTest : Int -> Char
constTest i = if | i > 0 => 'P'
                 | i == 0 => 'Z'
                 | otherwise = 'X'
```

### DIFF
--- a/docs/source/tutorial/typesfuns.rst
+++ b/docs/source/tutorial/typesfuns.rst
@@ -1254,6 +1254,38 @@ we get a default value:
     *UsefulTypes> lookup_default 4 [3,4,5,6] (-1)
     -1 : Integer
 
+Multi-way ``if`` expressions
+----------------------------
+
+Too many nested ``if`` can make code hard to read so Idris provides an
+alternative notation for reducing nesting and making choices easier to
+follow.
+
+The following function ``iffy`` would become quite messy if we want to
+include more choices of what to do with ``i``. It's also a lot of effort to
+insert choices at arbitrary points without repetitive indentation changes.
+Multi-way Ifs solve this in ``okay`` by allowing you to state your choices
+and what to do about them in turn, we're also able to insert a case for ``i
+== 0`` with ease:
+
+.. code-block:: idris
+
+    iffy : Int -> Char
+    iffy i = if i + 3 == 7 then '7'
+               else if i < 0 then 'N'
+                 else 'X'
+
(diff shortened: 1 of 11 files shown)


Russoul:
The syntax is pleasantly lightweight and probably its biggest selling point is aligned term arrangement. Which is great, as nested if-then-else statements can become an indentation nightmare if used aggressively.

But another point has to be made regarding the use of if-then-else expressions in general in dependent setting.
They are not used that much, as the language is expressive enough that hand coding 'destructors' for any data is practical and often preferred over relying on primitive constructs like if-then-else.

But undoubtedly multi-way-if will find its uses.

 I'd love to see similar syntax for simplified 'with' expressions (basically like guards in Haskell, but far more powerful), where destructed term doesn't depend on function arguments. Currently it is so bloated ...

edwinb:
I think I like this. I agree with @Russoul that we shouldn't encourage if-then-else but sometimes this sort of thing is really handy. I'm not sure how best to deal with totality checking it though. Maybe requring an `else` clause at the end? Hmm.

gallais:
> How to emit warnings at desugar-time when we have unreachable cases, e.g.

Currently you won't get any warning about unreachable branches if you are using
cascading `if`s so the fact this feature does not provide these warnings is not
a problem preventing inclusion IMO.

edwinb:
Since this is still a draft and discussion seems to have stopped, I'm going to close this, but that doesn't mean it should be rejected forever, so if we reach consensus please feel free to reopen.

I'm generally inclined to resist adding more sugar, but there is still probably value in this or something like it. Actually I think what I really want is something like Haskell's pattern guards, since they aren't restricted to Bool and have the nice fall through behaviour.

PR #736, opened 2020-10-08
[base, contrib] Propose re-organisation of the order (strict, lax) interfaces

ohad:
The issue I'm addressing is that the order relation (`po`, or `spo`, etc) cannot be
inferred from the carrier of the preorder, and needs to be passed explicitly.

I propose to add two interfaces:
* `Order.ComparisonRelation t` for order relations (like `Data.Nat.LTE`); and
* `Strict.StrictComparison t` for strict order relations (like `Data.Nat.LT`);

We need two interfaces because we sometimes want to use both at the
same time, and at the moment this ends up being very confusing for
interface resolution. We can revisit this decision once #607 is merged,
since then we'll have more flexibility in manipulating interfaces.

Thanks to @berewt for discussing this proposal!


ohad:
Outdated, and I'm not convinced it's a good way anyway.

PR #739, opened 2020-10-11
Experimental C backend with reference counting

edwinb:
Thanks to @vfrinken for putting this together - all I've done is merge it in and get it to invoke the C compiler.

There's still plenty to do here, particularly performance where there's a lot of low hanging fruit but the basics work. The main goal of this is portability - especially, being able to run Idris programs on unusual or constrained platforms. Generated code also has much better startup time than the scheme back ends!


melted:
I pushed a PR for Windows to @edwinb branch. It will show up here automatically if he merges it.

PR #740, opened 2020-10-12
Add NameAt to Ide mode

Russoul:
New IDE mode command that is used to implement `Go to definition` mechanics in Vim, see [the plugin](https://github.com/Russoul/Idris2-Vim-Ide).

The command for the given partial name of a possible global definition returns a list of full names of found global definitions in scope and the file name for each.

The command will also support searching for the definition by a partial name *and* a position of its use. But `FC` data that contains exact info about locations is a bit odd right now.


Russoul:
@redfish64 Does this patch fulfil the requirements to implement jump-to-def in emacs ?

@ziman @gallais Also can we merge it if there are no objections ?

I guess this one is now used both in vim and emacs.

redfish64:
Yes, I switched over the emacs idris2-mode to use this as well

Ruslan Feizerahmanov <notifications@github.com> 於 2020年10月26日 週一 23:44 寫道：

andrevidela:
This looks good, is there anything blocking it?

edwinb:
Looks fine to me. Probably the only blocker is me not having looked for a while, sorry!

PR #744, opened 2020-10-14
[ contrib ] add McBride's General monad

gallais:
Starting my experiment by reimplementing the state of the art.


clayrat:
I'd advise against using `Future` because that term usually means an [asynchronous primitive](https://en.wikipedia.org/wiki/Futures_and_promises).

andrevidela:
`Eventually` would probably be a better name. `Delayed` or `Delay` too but it has already been taken by lazy evaluation

gallais:
It's usually called `Delay` but that is a reserved keyword AFAICT
because I was getting parse errors when trying to call it `Delay`.

First I was thinking about calling it `Promise` but that is an async
thing in js so I opted for `Future`. Did not realise it was also a
common name for the same thing. :(

I am not a fan of `Eventually` as there is no guarantee that the
constructor `Now` is ever used. (But `Eventually` would be a good
name for the inductive type proving that a "`Future`" indeed has a
bounded depth).

`Postponed`? `Slow`?

ziman:
Given that we're probably not going to use both `Future` and `Promise` for async things, we could use one of them for this? (Such as... `Future`? :) )

I don't think that async should usurp the whole namespace of delayed computations.

mattpolzin:
Hi, new here (but couldn't resist helping paint the bike shed).

As someone with exposure to several languages where promises and futures are both used, I would definitely assume (incorrectly in this case) we were talking about "that thing I already know really well" regardless of which of those names was used.

Is `Deferred` appropriate here?

clayrat:
`Stalled` maybe? :)

gallais:
I am not a huge fan of the past participle as the common name is `Delay`, not `Delayed`.
Just before going to bed yesterday, it hit me: what about `Late`?

mattpolzin:
I like `Late` as a parallel to `Lazy`. Here's one that might be too clever by a half: `ToBe`

PR #752, opened 2020-10-22
Added ability for IDEMode to report location of definition of name

redfish64:
This is to create similar functionality of emacs' command 'haskell-mode-jump-to-def
I implemented the clientside portion here: https://github.com/redfish64/idris2-mode


redfish64:
Just realized this is a copy of https://github.com/idris-lang/Idris2/pull/740/ , closing.

PR #753, opened 2020-10-23
Implement Futures as a Parallelism Primitive

fabianhjr:
This is a simple Chez implementation of a Future Primitive; Since many systems are multicore this might be a good low hanging fruit in improving the Idris2 Compiler Performance.

The idea of adopting Future as a threading primitive rather than a Thread is to facilitate parallelism in the case of pure functions. (Since it makes returning values from a thread easier without needing to worry about mutation/synchronization or other interproccess communication techniques) This is implemented as an OS Thread per future in Chez. So for the time being it is intended to launch tasks in parallel that currently take some time. (Parsing for example)

### DIFF
--- a/libs/contrib/System/Future.idr
+++ b/libs/contrib/System/Future.idr
@@ -17,8 +17,8 @@ fork : Lazy a -> Future a
 fork = prim__makeFuture
 
 export
-forkIO : IO a -> IO (Future a)
-forkIO a = io_pure $ fork $ unsafePerformIO a
+forkIO : HasIO io => IO a -> io (Future a)
+forkIO a = primIO $ prim__io_pure $ fork $ unsafePerformIO a
 
 export
 await : Future a -> a
--- a/libs/contrib/System/Future.idr
+++ b/libs/contrib/System/Future.idr
@@ -17,10 +17,6 @@ fork : Lazy a -> Future a
 fork = prim__makeFuture
 
 export
-forkIO : HasIO io => IO a -> io (Future a)
-forkIO a = primIO $ prim__io_pure $ fork $ unsafePerformIO a
-
-export
 await : Future a -> a
 await f = prim__awaitFuture f
 
@@ -37,3 +33,11 @@ public export
 Monad Future where
   join = map await
   v >>= func = join . fork $ func (await v)
+
+export
+performFutureIO : HasIO io => Future (IO a) -> io (Future a)
+performFutureIO = primIO . prim__io_pure . map unsafePerformIO
+
+export
+forkIO : HasIO io => IO a -> io (Future a)
+forkIO a = performFutureIO $ fork a
--- a/libs/contrib/System/Future.idr
+++ b/libs/contrib/System/Future.idr
@@ -1,0 +1,44 @@
+module System.Future
+
+%default total
+
+-- Futures based Concurrency and Parallelism
+
+export
+data Future : Type -> Type where [external]
+
+%extern prim__makeFuture : {0 a : Type} -> Lazy a -> Future a
+%foreign "racket:blodwen-await-future"
+         "scheme:blodwen-await-future"
+prim__awaitFuture : {0 a : Type} -> Future a -> a
+
+export
+fork : Lazy a -> Future a
+fork = prim__makeFuture
+
+export
+forkIO : IO a -> IO (Future a)
+forkIO a = io_pure $ fork $ unsafePerformIO a
+
+export
+await : Future a -> a
+await f = prim__awaitFuture f
+
+public export
+Functor Future where
+  map func future = fork $
+    let result = await future in func result
+
+public export
+Applicative Future where
+  pure v = fork v
+  funcF <*> v = fork $
+    let result = await v
+        func   = await funcF in
+        func result
+
+public export
+Monad Future where
+  join = map await
+  v >>= func = join $ fork $
+    let result = await v in func result
--- a/libs/contrib/System/Future.idr
+++ b/libs/contrib/System/Future.idr
@@ -26,19 +26,14 @@ await f = prim__awaitFuture f
 
 public export
 Functor Future where
-  map func future = fork $
-    let result = await future in func result
+  map func future = fork $ func (await future)
 
 public export
 Applicative Future where
   pure v = fork v
-  funcF <*> v = fork $
-    let result = await v
-        func   = await funcF in
-        func result
+  funcF <*> v = fork $ (await funcF) (await v)
 
 public export
 Monad Future where
   join = map await
-  v >>= func = join $ fork $
-    let result = await v in func result
+  v >>= func = join . fork $ func (await v)
--- a/libs/contrib/contrib.ipkg
+++ b/libs/contrib/contrib.ipkg
@@ -98,6 +98,7 @@ modules = Control.ANSI,
           Text.PrettyPrint.Prettyprinter.Render.String,
           Text.PrettyPrint.Prettyprinter.Render.Terminal,
 
+          System.Future,
           System.Random,
           System.Path,
           Syntax.WithProof,
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -156,6 +156,9 @@ mutual
       = do p' <- schExp chezExtPrim chezString 0 p
            c' <- schExp chezExtPrim chezString 0 c
            pure $ mkWorld $ "(blodwen-register-object " ++ p' ++ " " ++ c' ++ ")"
+  chezExtPrim i MakeFuture [_, work]
+      = do work' <- schExp chezExtPrim chezString 0 work
+           pure $ "(blodwen-make-future " ++ work' ++ ")"
   chezExtPrim i prim args
       = schExtCommon chezExtPrim chezString i prim args
 
--- a/src/Compiler/Scheme/Common.idr
+++ b/src/Compiler/Scheme/Common.idr
@@ -206,6 +206,7 @@ data ExtPrim = NewIORef | ReadIORef | WriteIORef
              | SysOS | SysCodegen
              | OnCollect
              | OnCollectAny
+             | MakeFuture
              | Unknown Name
 
 export
@@ -223,6 +224,7 @@ Show ExtPrim where
   show SysCodegen = "SysCodegen"
   show OnCollect = "OnCollect"
   show OnCollectAny = "OnCollectAny"
+  show MakeFuture = "MakeFuture"
   show (Unknown n) = "Unknown " ++ show n
 
 ||| Match on a user given name to get the scheme primitive
@@ -241,7 +243,8 @@ toPrim pn@(NS _ n)
             (n == UN "prim__os", SysOS),
             (n == UN "prim__codegen", SysCodegen),
             (n == UN "prim__onCollect", OnCollect),
-            (n == UN "prim__onCollectAny", OnCollectAny)
+            (n == UN "prim__onCollectAny", OnCollectAny),
+            (n == UN "prim__makeFuture", MakeFuture)
             ]
            (Unknown pn)
 toPrim pn = Unknown pn
--- a/src/Compiler/Scheme/Racket.idr
+++ b/src/Compiler/Scheme/Racket.idr
@@ -42,6 +42,7 @@ schHeader : String -> String
 schHeader libs
   = "#lang racket/base\n" ++
     "; @generated\n" ++
+    "(require racket/future)\n" ++ -- for parallelism/concurrency
     "(require racket/math)\n" ++ -- for math ops
     "(require racket/system)\n" ++ -- for system
     "(require rnrs/bytevectors-6)\n" ++ -- for buffers
@@ -95,6 +96,9 @@ mutual
       = do p' <- schExp racketPrim racketString 0 p
            c' <- schExp racketPrim racketString 0 c
            pure $ mkWorld $ "(blodwen-register-object " ++ p' ++ " " ++ c' ++ ")"
+  racketPrim i MakeFuture [_, work]
+      = do work' <- schExp racketPrim racketString 0 work
+           pure $ mkWorld $ "(blodwen-make-future " ++ work' ++ ")"
   racketPrim i prim args
       = schExtCommon racketPrim racketString i prim args
 
--- a/support/chez/support.ss
+++ b/support/chez/support.ss
@@ -217,6 +217,23 @@
 (define (blodwen-condition-signal c) (condition-signal c))
 (define (blodwen-condition-broadcast c) (condition-broadcast c))
 
+(define-record future-internal (result ready mutex signal))
+(define (blodwen-make-future work)
+  (let ([future (make-future-internal #f #f (make-mutex) (make-condition))])
+    (fork-thread (lambda ()
+      (let ([result (work)])
+        (with-mutex (future-internal-mutex future)
+          (set-future-internal-result! future result)
+          (set-future-internal-ready! future #t)
+          (condition-broadcast (future-internal-signal future))))))
+    future))
+(define (blodwen-await-future ty future)
+  (let ([mutex (future-internal-mutex future)])
+    (with-mutex mutex
+      (if (not (future-internal-ready future))
+          (condition-wait (future-internal-signal future) mutex))
+      (future-internal-result future))))
+
 (define (blodwen-sleep s) (sleep (make-time 'time-duration 0 s)))
 (define (blodwen-usleep s)
   (let ((sec (div s 1000000))
--- a/support/racket/support.rkt
+++ b/support/racket/support.rkt
@@ -212,6 +212,7 @@
   (channel-put c 'ready))
 
 (define (blodwen-sleep s) (sleep s))
+(define (blodwen-usleep us) (sleep (* 0.000001 us)))
 
 (define (blodwen-time) (current-seconds))
 
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -137,6 +137,7 @@ chezTests = MkTestPool [Chez]
       "chez019", "chez020", "chez021", "chez022", "chez023", "chez024",
       "chez025", "chez026", "chez027", "chez028", "chez029", "chez030",
       "chez031",
+      "concurrency001",
       "perf001",
       "reg001"]
 
--- a/tests/chez/concurrency001/Futures.idr
+++ b/tests/chez/concurrency001/Futures.idr
@@ -45,3 +45,14 @@ erasureAndNonbindTest = do
   let _ = nonbind
   let n = nonbind
   macsleep 2000 -- Even if not explicitly awaited, we should let threads finish before exiting
+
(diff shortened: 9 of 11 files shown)


andrevidela:
This looks great! I have a high-level question about this because I am not familiar with traditional implementations. Could this be implemented purely as a library using MVars? I assumed MVars were the most primitive concept from which concurrent APIs could be implemented in functional languages.

fabianhjr:
@andrevidela, there are two overall approaches: os-threads or green-threads.

OS Threads use the operating system threads, thread scheduler, mutexes, and other primitives. In this case it is using pthread in linux. (Chez scheme uses pthread for these primitives in Linux)

Green Threads would be an implementation in the runtime system of "virtual threads" but that would require to implement at least a thread pool, virtual threads with context non-os context switching, a thread scheduler, mutexes, conditions, and other primitives.

As far as I know that advantages of green threads can only be archived by writing a lot of code that is already part of the OS so rather than duplicating that I would prefer to just use the OS concurrency primitives.

andrevidela:
As far as i understand, the API for Futures would remain the same, whether we use green threads of Posix threads, wouldn't it?

The question is, isn't there [already a more fundamental concurrent primitive](https://www.microsoft.com/en-us/research/wp-content/uploads/1996/01/concurrent-haskell.pdf) that allows us to implement both futures, channels and protocols, as library code rather than foreign calls? If that's the case, shouldn't we be using that instead?

I'm curious to know what other people think about this because I do not have a lot of experience with concurrent programs and I know that channels (a-la golang) or async-await (a-la javascript) are features that lots of people are extremely fond of and expect in modern programming languages.

fabianhjr:
The main issue I was having was that, unlike haskell, Idris is eager. So even with the same support code but without the specific chez compiler changes, the call to `prim__makeFuture` would be done as a value (already evaluated) instead of a lazy/delayed reference (that cannot be marshaled by `%foreign`).

We could implement Futures as a ThreadPool and MVars and use them in `src/Compiler/Scheme` since the compiler is in Idris, however it would be a lot of extra complexity from the get go and as you mentioned the API would likely remain the same or very similar.

ShinKage:
It's great that we can have Futures with little modifications to the compiler. However (and I guess it is what @andrevidela is also referring to) is Future too much of a specific concurrency model to warrant a compiler primitive? Implementing primitives for more general constructs like MVar or even STM (https://www.microsoft.com/en-us/research/wp-content/uploads/2005/01/2005-ppopp-composable.pdf) that would allow to write more abstractions as library, like futures or channels (e.g  https://hackage.haskell.org/package/async-2.2.2 which highly abstracts over MVar and STM). Admittedly, I'm not well-versed in concurrency primitives or low-level libraries and this could be a good start for some kind of high level concurrency features, until we get more general stuff.

fabianhjr:
> Future too much of a specific concurrency model to warrant a compiler primitive

I would say _yes_, however the need for the compiler primitive came from the need to delay the evaluation rather than the usage of pthread. All the used primitives by `blodwen-future` are already exposed in the same module `System.Concurrency.Raw`.

```
Main> awaitFuture . makeFuture $ "a"
prim__awaitFuture (prim__makeFuture Delay (fromString "a"))
```

fabianhjr:
Rephrasing the previous point,

This merge requests implements something akin of:

```
fork: Lazy a -> ForkedLazy a
join: ForkedLazy a -> a
```

Instead of the normal fork-thread primitive. I haven't tried it but I figure `unsafePerformIO` would play well with `fork` / `makeFuture`. Though as far as naming goes I wasn't sure if it was best to desemphasize this change being modeled after a future by calling it fork or to have to create IO to make a thread by implementing a fork `IO a -> IO ThreadID` that always ran `unsafePerformIO` in the other thread.

fabianhjr:
There is an issue with `sleep` / `usleep` on ChezScheme for macOS.

https://github.com/Homebrew/homebrew-core/pull/10159 disabled some clock related stuff and it is currently like that in master. They raised an issue on upstream that has been idle since then https://github.com/cisco/ChezScheme/issues/146

gallais:
> since you both have worked on the action/workflow of macOS.

Have I? I have never used macOS in my life. IIRC I did modify all of the
workflows to set an `IDRIS2_TESTS_CG` environment variable but that's about it.

Russoul:
Why are the new primitives, namely `makeFuture` and `awaitFuture`, pure ?

I am not sure if I'd want any explicit process manipulation in pure code. What are the use cases ?

I've always thought about pure code as a subject of aggressive compiler optimizations. The order of execution of its subexpressions or whether they are to be executed at all is undermined. The compiler is actually free to rewrite pure code to be executed in parallel if it sees fit, that is a quite far-reaching statement though.

fabianhjr:
> Why are the new primitives, namely makeFuture and awaitFuture, pure ?

It doesn't matter if they are run in parallel or serially. If you have some `expensiveFunction` you need to map in a lot of data you could do `makeFuture . expensiveFunction <$> listOfData`. Pure functions, being referentially transparent, can be run in parallel for free without having to worry about synchronization.

The test case uses unsafePerformIO because otherwise we wouldn't be able to tell. (Other than monitoring resource usage)

fabianhjr:
> Have I? I have never used macOS in my life. IIRC I did modify all of the workflows to set an IDRIS2_TESTS_CG environment variable but that's about it.

Ah sorry, my mistake. I only took a quick look at the contributors of the workflow in github. Well, hopefully a macOS dev will be able to help us out.

fabianhjr:
Renamed everything to a more familiar/neutral join/fork.

fabianhjr:
Have a small demo with performance issues (sequential code would be slightly faster) but multithreaded nontheless at #756.

```
src/ > time idris2 -p contrib Idris/ConcurrentModuleCompiler.idr --exec testCase
[..] All transitive modules required by Idris.Main
Executed in    5.69 secs   fish           external 
   usr time    8.33 secs  203.00 micros    8.33 secs 
   sys time    1.30 secs   45.00 micros    1.30 secs 
```

Inspecting the systems monitor shows 20% usage across 16 threads (8-core /w SMT)

melted:
We shouldn't disable the test on macOS, there's no reason the feature can't work there, so we should fix it to work there before merging.

Eyeballing it, there are rather short sleep periods in the test, maybe the resolution of the timers aren't reliable enough, what happens if those are multiplied by 10 say?

fabianhjr:
> what happens if those are multiplied by 10 say?

Already tried up to 1000x the current timings :c

melted:
Could you try adding a fflush stdout after the printing in the test?

fabianhjr:
Ok, tried the suggestion but fulshing/not flushing stdout wasn't the problem.

Russoul:
This is quite strange. I've just tested the patch on my mac, it works all the time.
 (Did a few `make test only=concurrency001` spams)

fabianhjr:
Hi @Russoul, do you recall how you installed The?

Russoul:
I am almost sure it was [brew](https://formulae.brew.sh/formula/chezscheme)

fabianhjr:
Hi @melted / @Russoul, the issue seems to be related to usleep; I changed the code to sleep on macos/darwin and it passes green now however the test will now take ~3 seconds to complete instead of 0.4 seconds on other platforms.

The only remaining issue is that the bootstrap step for racket installs the support first before bootstrap and a specific change to the compiler has to be made to import futures for racket. Should I update the bootstrap code? (I think it would be necessary regardless if people want to use Futures in the Idris2 codebase)

gallais:
> Should I update the bootstrap code? 

Probably not. Last time we did [caused issues](https://github.com/idris-lang/Idris2/pull/576#issuecomment-704974174) with my (not finished yet) attempt
to make Idris2 compilable with the previously released version of Idris2.

fabianhjr:
Ok @gallais, then what I think would be best is to only fix the racket bootstrapping. What I did was copying the generated racket file from `idris2 --build idris2.pkg --cg racket` into the respective bootstrap file (`bootstrap/idris2_app/idris2.rkt` and then only commit the changes related to this PR which are a couple of lines that can be easily reviewed and are additions not used elsewhere. That should limit the potential damage.

fabianhjr:
Hi @gallais, I cleaned up the commits a bit and removed the Chez support code to avoid  a bootstrapping issue. This should be ready to review/merge.

chrrasmussen:
Regarding @Russoul's comment about the functions being pure: I also think that `fork`/`forkIO` and `await` should include `IO` (or `HasIO`) in their type signatures.


### Breaking referential transparency

It is currently possible to write the following code:
```idris
import System.Concurrency
import System.Random

getRandomInt : Int
getRandomInt = await (forkIO randomIO)

main : IO ()
main = do
  printLn getRandomInt
  printLn getRandomInt
  printLn getRandomInt
```

Running this code (`idris2 -p contrib --exec main Main.idr`) will output something similar to:
```
887071849
362116491
1536377306
```


### Behaviour of using `forkIO` without `await`

Another strange behaviour I noticed:
```idris
import System.Concurrency

main : IO ()
main = do
  putStrLn "Started"
  let _ = forkIO (putStrLn "This line is printed")            -- #1
  let notUsed = forkIO (putStrLn "This line is NOT printed")  -- #2
  putStrLn "Done"
```

Running this code (`idris2 --exec main Main.idr`) will output something similar to:
```
Started
Done
This line is printed
Exception in collect: cannot collect when multiple threads are active
```

The results of the calls to `forkIO` are not used later in the code, but in the case of `#1` the side-effects are still executed. I have not tried to figure out why this happens, but I think this behaviour also indicate that `forkIO` should run in the `IO` context.

Running `forkIO` without `await` may cause the error `Exception in collect: cannot collect when multiple threads are active` (as seen above). This error may show up even if we change the type signature to run in `IO` context. I am able to reproduce this error when using [`Prelude.fork`](https://github.com/idris-lang/Idris2/blob/master/libs/prelude/Prelude/IO.idr#L110-L112) as well. I seem to recall that @melted found a solution for this error, but that the solution required a recent version of Chez?

---

I did some experimentation in my own branch (where I changed the type signatures): https://github.com/chrrasmussen/Idris2/commit/c77013338f23d6f4e75983d34a66bbb064a0c797

I removed the implementations for `Functor`, `Applicative` and `Monad`. Looking at Haskell's [async library](https://hackage.haskell.org/package/async) they include a [`Functor` instance for their `Async` type](https://hackage.haskell.org/package/async-2.2.2/docs/Control-Concurrent-Async.html#g:1). I think that can be useful, but I want to avoid calling `await` (especially since it is now running in the `IO` context), but I don't know how to implement this in the Chez code.

fabianhjr:
Hi @chrrasmussen, `forkIO` should definitely have a `HasIO m => m a -> m (Future a)` to allow for better IO composition and reasoning.

However, as for referential transparency example it seems like Idris2 is revaluating / recalculating values each time they are needed and should be investigated further.

```
import System.Random

some : Int
some = unsafePerformIO $ randomIO

main : IO Unit
main = do
  printLn some
  printLn some
  printLn some
```

results in:

```
idris2 reval.idr -p contrib --exec main
-879496324
1599532704
-729511928
```

Which is similar to the issue you describe regarding forking.

Regarding your second example, I will investigate. Though I think the runtime will have to block the exit/shutdown of runtime until all forked threads finish. (Which wouldn't necessarily happen if the forked code isn't total)

fabianhjr:
Thanks to @ziman for clearing up my misunderstanding of how IO works in the signature of `forkIO` is a more appropriate `IO a -> IO (Future a)`

fabianhjr:
I can use HasIO for the return type but will need some interface to provide an `unliftIO` to accept something more general as an argument to forkIO.

fabianhjr:
@chrrasmussen your example adapted to the new type signatures now produces:

```
Started
Done
This line is printed
Exception in collect: cannot collect when multiple threads are active
This line is NOT printed
```

Continuing to investigate the collection error and will add it as a test case when done.

fabianhjr:
@chrrasmussen the collection error is part of the garbage collection though I am inclined to leave it as is since it would warn about unawaited Futures. (Chez code in question: https://github.com/cisco/ChezScheme/blob/0cc825d8a046b19320976875989717c85a44edab/s/7.ss#L733-L734 )

I modified your example a bit to ensure that this new version of forkIO behaves well with erasure/non-usage:

```
import System.Concurrency
import System

nonbind : IO (Future ())
nonbind = forkIO $ putStrLn "This shouldn't print"

main : IO ()
main = do
  putStrLn "Started"
  _ <- forkIO $ do
    putStrLn "This line is printed"
  notUsed <- forkIO $ do
    usleep 1000
    putStrLn "This line is also printed"
  let _ = nonbind
  let n = nonbind
  putStrLn "Done"
  usleep 2000
```

produces:

chrrasmussen:
@fabianhjr Thanks for considering my feedback!


### `await` should be run in `IO` context

I still believe that `await` should be run in the `IO` context. As we saw previously, Idris 2 handles `let notUsed = ?val` and `let _ = ?val` differently. This is probably a bug, but it should not affect the behaviour of the program.

The fact that `await` may be optimized away can cause problems like the following:

```idris
import Data.IORef
import System
import System.Concurrency

main : IO ()
main = do
  putStrLn "Start"
  ref <- newIORef "Initial value"
  future1 <- forkIO $ do
    sleep 1
    putStrLn "#1"
    writeIORef ref "Updated value"
  let notUsed = await future1
  -- Expect future1 to have completed at this point
  putStrLn "#2"
  storedValue <- readIORef ref
  putStrLn storedValue
  sleep 3
```

Expected output:
```
Start
#1
#2
Updated value
```

Currently outputs:
```
Start
#2
Initial value
#1
```

This example can be "fixed" by changing `let notUsed = await future1` to `let _ = await future1`.


### The implementation of `Functor`, `Applicative` and `Monad` may cause surprising behaviour

Since the implementation of `Functor`, `Applicative` and `Monad` calls `await`, they may cause waiting at locations that are unintended by the user.

In the following example I intend to create two futures that start immediately, but I "accidentally" call `await` (via the `map` function) before starting the second future.

```idris
import System
import System.Concurrency

main : IO ()
main = do
  putStrLn "Start"
  future1 <- forkIO $ do
    sleep 1
    putStrLn "#2"
  let _ = map (const ()) future1
  future2 <- forkIO $ do
    putStrLn "#1"
  sleep 3
```

Expected output:
```
Start
#1
#2
```

Currently outputs:
```
Start
#2
#1
```

Removing the line `let _ = map (const ()) future1` "fixes" the problem.


### Summary

In both examples I would argue that the compiler is free to remove the calls to `map` and `await`, because the assigned variables in the `let`-bindings are not used later in the program. Even if the compiler (currently) does not perform this optimization, I, as a developer, should be free to do this, without changing the behaviour of the program. Changing `await` to be run in the `IO` context makes it clear that ordering matters, and that the call to `await` should be kept even when the return value is not used.

Generally, making the functions run in the `IO` context will encourage their use closer to the root of the program (usually the `main` function). In my opinion, even the `fork` function (the lazy variant) should be run in the `IO` context.

I think that taking inspiration from Haskell's [Async library](https://hackage.haskell.org/package/async-2.2.2/docs/Control-Concurrent-Async.html) is a good idea. For example, they only provide a `Functor` implementation for their `Async` type. In JavaScript the `Functor` implementation could wrap an existing future like this: `future.then()` (See example below). In Chez I think it should be possible to save each function passed to `map`, and then call the resulting function from `await`.


#### Example in JavaScript

```idris
future <- forkIO $ do
  putStrLn "test"
  pure 42
let mappedFuture = map (* 2) future
result <- await mappedFuture
printLn result
```

Would be translated into something like:
```javascript
async function main() {
    var future = new Promise(function (resolve, reject) {
        console.log("test");
        resolve(42);
    });
    var mappedFuture = future.then(function (value) { return value * 2; });
    var result = await mappedFuture;
    console.log(result);
}

main();
```

I will do some more experimentation later this week.

fabianhjr:
> The implementation of Functor, Applicative and Monad may cause surprising behaviour

Yes, that is not what I intended, thanks to your other pointers on let I noticed that the Functor implementation gets fixed by removing the let. Seems like Idirs will eagrly evaluate the let before passing it.

```idris
map : IO ()
map = do
  future1 <- forkIO $ do
    macsleep 1000
    putStrLn "#2"
  let future3 = map (const "#3") future1
  future2 <- forkIO $ do
    putStrLn "#1"
  pure $ await future2
  putStrLn (await future3)
```

This now prints 1,2,3 and doesn't block on the map.

fabianhjr:
> The fact that await may be optimized away can cause problems like the following

```
import Data.IORef
import System
import System.Concurrency

main : IO ()
main = do
  putStrLn "Start"
  ref <- newIORef "Initial value"
  future1 <- forkIO $ do
    sleep 1
    putStrLn "#1"
    writeIORef ref "Updated value"
  let notUsed = await future1
  -- Expect future1 to have completed at this point
  putStrLn "#2"
  storedValue <- readIORef ref
  putStrLn storedValue
  sleep 3
```

I am more inclined to normal usage being without await but rather if a comuputation depends on a future it should be mapped or binded as with my Scala Futures background. Though, I think what makes this counterintuitive is the signatures of the interfaces.


At least as an pseudocode example of how I imagine the usage of this interface:

```idris
do
  file <- forkIO $ readFile "example"
  processedFile = process <$> file
  transmitFile <- processedFile >>= transmit
```

This would be an `IO (Future a)` with `process : Text -> ProcessedFile` and `transmit : ProcessedPile -> IO (Future a)` which wouldn't await when called.

edwinb:
This looks like an interesting approach, and I think it does make sense for the primitives to have the types they do since that enables parallel evaluation in code that doesn't use `%World`, so not in `IO`. (Note I'm careful not to say 'pure' because everything is pure! It's just about whether they take the `%World` as an argument or not).

The main thing is that, since this is clearly a new experiment, it shouldn't be in `base`, which is for things that have become stable. I'd prefer it to be in `contrib` for now, and perhaps the module should be called `Futures` rather than `Concurrency` since it's not just about concurrency.

In general (not just talking about this PR) I'd prefer to welcome this sort of experiment, and err on the side of merging as long as nothing existing gets broken. Idris 2 is mostly a vehicle for research after all, and the best way of finding out if something works is to try it for a bit. For the more extreme experiments, we have `%language` pragmas.

fabianhjr:
As requested moved the module from System.Concurrency in base to System.Future in contrib.

gallais:
I have now seen 2 or 3 instances of the test suite failing on windows because the
"Hello Idris", "Hello World", etc. come in the wrong order. Just like we have a
fix for MacOS' sleep, do we need one for windows?

fabianhjr:
Hi @gallais, I don't have a windows machine available so my guess would be to test 10x the timings if it is a rare occurrence first,  Maybe the overhead of creating a thread is higher on Windows. (And `/ 10` the macSleep to avoid that jumping to become a 30 seconds test)

When picking those timings I wanted to balance the total runtime with reproducibility.

fabianhjr:
Ok, noticed you were referring specifically to the CI runs in which 2 runs on master out of about o dozen have failed due to timing. I created a PR #828 to address that flakyness.

Issue #16, opened 2020-05-20
Unexpected run-time complexity


edwinb:
<a href="https://github.com/nicolabotta"><img src="https://avatars3.githubusercontent.com/u/2726946?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [nicolabotta](https://github.com/nicolabotta)**
_Thursday Aug 01, 2019 at 16:20 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/71_

----

The program
```
import Data.Fin
import Data.Vect

tail : {0 A : Type} -> {n : Nat} -> (Fin (S n) -> A) -> (Fin n -> A)
tail f = f . FS

toVectFun : {0 A : Type} -> {n : Nat} -> (Fin n -> A) -> Vect n A
toVectFun {n =   Z} _ = Nil
toVectFun {n = S m} f = (f FZ) :: (toVectFun (tail f))

N : Nat
N = ...

xs : Vect (S N) Nat
xs = toVectFun {n = S N} finToNat

main : IO ()
main = putStrLn ("xs(" ++ show N ++ ") = " ++ show (index (fromInteger (natToInteger N)) xs))
```
type checks and executes in more than quadratic time in `N` in idris 2:
```
N =  20, user time =   0.78s  
N =  40, user time =   2.57s 
N =  80, user time =  14.49s
N = 160, user time =  segfault
N = 320, user time =  segfault
```
For "large" values of `N`, it segfaults. In idris 1, the correspondent program also does not run in linear time
```
N =  100, user time =   3.3s  
N =  200, user time =   4.9s 
N =  400, user time =  10.1s
N =  800, user time =  36.4s
N = 1600, user time = 197.6s 
```
, but it does not segfault and is faster. What is the expected complexity of the program? Shouldn't building the vector and reading its last component be linear operations in `N`?

edwinb:
<a href="https://github.com/edwinb"><img src="https://avatars1.githubusercontent.com/u/255926?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [edwinb](https://github.com/edwinb)**
_Thursday Aug 01, 2019 at 18:00 GMT_

----

At least at compile time, it's almost certainly down to the `fromInteger` which has to compute a proof that the Fin is within bounds. This is a compile time thing. Are you compiling and executing things too? And if so, what happens? Can you be precise about where these numbers come from?

edwinb:
[...]

PR #754, opened 2020-10-23
 Fix usleep type signature

fabianhjr:
Type was erroneously `(x : Int) -> So (x >= 0) => io ()`, normalized to Int without proofs

### DIFF
--- a/libs/base/System.idr
+++ b/libs/base/System.idr
@@ -17,11 +17,11 @@ prim__usleep : Int -> PrimIO ()
 
 export
 sleep : HasIO io => Int -> io ()
-sleep sec = primIO (prim__sleep sec)
+sleep = primIO . prim__sleep
 
 export
-usleep : HasIO io => (x : Int) -> So (x >= 0) => io ()
-usleep sec = primIO (prim__usleep sec)
+usleep : HasIO io => Int -> io ()
+usleep = primIO . prim__usleep
 
 -- This one is going to vary for different back ends. Probably needs a
 -- better convention. Will revisit...


gallais:
What happens in the `cast` if the `Nat` is too big?
Wrap-around to a negative `Int` thus breaking the invariant?

fabianhjr:
Just tried:

```
-usleep : HasIO io => Nat -> io ()
+usleep : HasIO io => Fin 1000000 -> io ()
```

after your observation of the cast overflow. However this is currently impacted by #16. (Takes a very long time to type-check/compile) I agree that what you pointed out in your comment is an issue; however I wanted to at least retain the non-negative check from usleep.

Should I leave the non-primitive interface as an Int without a check? (or leave the `So ()` one? Though I think this `So` is _very_ cumbersome for usage)

gallais:
What about using one of the `BitsN`?

fabianhjr:
There might still be an issue, Bits32 can overflow by 1 bit (since Int32 is signed and Bits32 is unsigned) so the last non-overflowing one would be Bits16 but that wouldnt be enough to cover from 0 to 1,000,000 us (1 sec) on the usleep function, only up to 65,535.

fabianhjr:
Ah nvm, forgot that auto proofs on nats are easy enough to restrict the domain. This should address the original concern on overflows and underflows (bellow zero)

gallais:
This `{auto ok: us `LT` (power 2 31)}` is a very bad idea: `LT m n` proofs
are linear in the size of `m` so you'd potentially need to build a proof of
size `2 ^ 31 - 1`.

ziman:
I'd say that `Nat` is very wrong for this purpose. It has all the disadvantages you've already discovered and no advantage for this use case. `Nat` is generally quite inefficient and its only advantage is its unary structure, which makes the inefficiency of `Nat` worth it when you're using it to measure and reason about the size of inductive data structures, such as vectors — but that's not the case here.

`Nat` is not a general-purpose numeric type and there is no reason to use it by everywhere / by default.

I'd say `BitsN` or the current `Int` with a proof are the appropriate solutions here. Maybe you could `0`-quantify the proof, though.

fabianhjr:
> Maybe you could 0-quantify the proof, though.

Wouldn't that be still required though?

PR #755, opened 2020-10-24
Add `replaceWhen` for lists.

mattpolzin:
I've implemented replacement on lists by some criteria other than equality a few times now so I figured I would see if this was the kind of thing worth including in the base lib.

There does seem to be quite a bit of precedent for this naming but I'm honestly just trying to get my footing here so let me know! Part of me thinks this _must_ be defined on `Functor` somewhere, but I don't see it there.

Thanks.


gallais:
Sorry about the delay. The `*By` functions tend to take an `a -> a -> Bool` function
instead of the `Eq a` constraint. I think the name is confusing here as it's not what
is happening.

What about `replaceWhen`?

mattpolzin:
That's a good point. I like your suggestion to name it `replaceWhen`. Will do.

gallais:
Thanks for the PR!

PR #766, opened 2020-10-31
Gambit cg

mokshasoft:
I'am working on porting my Idris1 bare-metal/FreeRTOS/seL4 projects to Idris2. And this is the first step.


edwinb:
Thanks for the contribution! This looks fine, but could you also add a note to the Gambit CG docs to say what the option is please? Otherwise the docs might get more out of control than they already are.

mokshasoft:
Yes, I'll do that no problem. Good with updated docs!

edwinb:
Great, thanks!

PR #771, opened 2020-11-09
Fix for Issue 74

redfish64:
The problem is that if the function is defined in a sub module, then the current namespace (accessed by calling getNS) differs from the function namespace, therefore it is not considered visible by TTImp.Elab.App.checkVisibleNS

Fixed by using "setAllPublic". Not sure if this is the best way to do this.

### DIFF
--- a/src/Idris/IDEMode/Commands.idr
+++ b/src/Idris/IDEMode/Commands.idr
@@ -23,6 +23,7 @@ public export
 data IDECommand
      = Interpret String
      | LoadFile String (Maybe Integer)
+     | DefLoc String
      | TypeOf String (Maybe (Integer, Integer))
      | CaseSplit Integer Integer String
      | AddClause Integer String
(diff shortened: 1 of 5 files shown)


fabianhjr:
Hi, your PR contains  2 unrelated commits that cancel each other out. Could you drop those two (ce2f0b7, 1514e7b) from your merge request?

redfish64:
Ok, sorry about that. #772

buzden:
FYI, you could do this in the same PR by just force-pushing your git branch. So, generally, there's no need to create new PR for this.

redfish64:
I actually used a different branch in my fork because I wanted to keep that
history (I created a new branch and forked it from idris-lang/Idris2, then
cherry picked the commit). But thanks for letting me know.

Denis Buzdalov <notifications@github.com> 於 2020年11月9日 週一 17:16 寫道：

Issue #74, opened 2020-05-20
case-split fails on private definition in nested namespace


edwinb:
<a href="https://github.com/MarcelineVQ"><img src="https://avatars3.githubusercontent.com/u/16345659?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [MarcelineVQ](https://github.com/MarcelineVQ)**
_Sunday May 10, 2020 at 05:21 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/363_

----

# Steps to Reproduce
```idris
module Main

namespace Foo1
  foo : Nat -> ()
  foo k1 = ?foo1_rhs
  
namespace Foo2
  export
  foo2 : Nat -> ()
  foo2 k2 = ?foo2_rhs

foo3 : Nat -> ()
foo3 k3 = ?foo3_rhs
```
Split `k1` with any method, e.g.  :cs 5 7 k1
# Expected Behavior
Expect k1 k2 k3 can all be split to Z (S k) cases.
# Observed Behavior
An error of `No valid case splits` but the underlying cause is a hidden error of `:1:1--1:1:Name Main.Foo1.foo is private`
This error comes from `handleUnify` in `mkCase` called from `src/TTImp/Interactive/CaseSplit.idr:getSplitsLHS`

This issue does not occur on private definitions that are at the top level, only in nested namespaces.
This issue also appears for the :gd command.

supachai19:
ไม่มีอะไรให้ดูตัวอย่าง

jfdm:
I've recently been bitten by this bug, but have noticed an interesting aspect.

Given the following snippet:

```idris
module Bug

namespace BazzFoo
  public export
  data BazzFoo : Type where
    MkPair : Int -> BazzFoo

  getInt : BazzFoo -> Int
  getInt x = ?getInt_rhs

getInt : BazzFoo -> Int
getInt x = ?getInt_rhs_a
```
Case splitting on the pattern variable `x` in `Bug.BazzFoo.getInt` results in `No valid case splits'. However, case splitting on `x` in `Bug.getInt` works as expected.

gallais:
Fixed by 72af040

PR #772, opened 2020-11-09
Fixes Issue #74. 

redfish64:
The problem is that if the function is defined in a submodule, then the current namespace (accessed by calling getNS) differs from the function namespace, therefore it is not considered visible by TTImp.Elab.App.checkVisibleNS

### DIFF
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -336,8 +336,11 @@ record Context where
     -- access in a program - in all other cases, we'll assume everything is
     -- visible
     visibleNS : List Namespace
-    allPublic : Bool -- treat everything as public. This is only intended
+    allPublic : Bool -- treat everything as public. This is intended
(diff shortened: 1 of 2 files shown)


gallais:
Could you change the comment attached to `allPublic` in `Core.Context`
to mention this new use case?

redfish64:
Is this ok? I'm not sure what to put.

``` 
 -- treat everything as public. This is intended
 -- for checking partially evaluated definitions
 -- or for use outside of the main compilation
 -- process.
```

gallais:
My aim with this update is that anyone touching that bit of code will be
reminded of the potential extent of their modification. So I'd had an

```
-- process (e.g. when implementing interactive features such as case splitting)
```

redfish64:
ok, I added the comment.

gallais:
Thanks!

PR #780, opened 2020-11-20
Add extra parsers

alexhumphreys:
The combinators are mostly those found in [Lightyear](https://github.com/ziman/lightyear) that I was missing when porting a parser to `Data.String.Parser`. I'm not sure that I implemented `requireFailure` 100% correctly, so would be best to take a close look at that.

This also changes the return type of `char` and `string`. They previously returned `()`, they now return `Char` and `String` repectively. I can change it back if this is not desired behaviour, but this seems to be the normal return type for `char` and `string` as found in [Lightyear](https://github.com/ziman/lightyear/blob/master/Lightyear/Char.idr#L23), [attoparsec](http://hackage.haskell.org/package/attoparsec-0.8.1.0/docs/Data-Attoparsec-Char8.html#v:char) and [parsec](https://hackage.haskell.org/package/parsec-3.1.11/docs/Text-Parsec-Char.html#v:char). Ran into that issue when trying to parse `(char '+') <|> (char '-')` and realizing I couldn't inspect the result usefully. 

Signed-off-by: Alex Humphreys <alex.humphreys@here.com>

### DIFF
--- a/libs/contrib/Data/String/Parser.idr
+++ b/libs/contrib/Data/String/Parser.idr
@@ -6,6 +6,8 @@ import Control.Monad.Trans
 import Data.Strings
 import Data.Fin
 import Data.List
+import Data.List1
+import Data.Vect
 
 %default total
 
@@ -121,6 +123,17 @@ export
 optional : Functor m => ParseT m a -> ParseT m (Maybe a)
 optional = optionMap Nothing Just
 
+||| Succeeds if and only if the argument parser fails.
+|||
+||| In Parsec, this combinator is called `notFollowedBy`.
+export
+requireFailure : Functor m => ParseT m a -> ParseT m ()
+requireFailure (P runParser) = P $ \s => reverseResult s <$> runParser s
+where
+  reverseResult : State -> Result a -> Result ()
+  reverseResult s (Fail _ _) = OK () s
+  reverseResult s (OK _ _)   = Fail (pos s) "Purposefully changed OK to Fail"
+
 ||| Fail with some error message
 export
 fail : Applicative m => String -> ParseT m a
@@ -138,12 +151,12 @@ satisfy f = P $ \s => pure $ if s.pos < s.maxPos
 
 ||| Succeeds if the string `str` follows.
 export
-string : Applicative m => String -> ParseT m ()
+string : Applicative m => String -> ParseT m String
 string str = P $ \s => pure $ let len = strLength str in
                               if s.pos+len <= s.maxPos
                                   then let head = strSubstr s.pos len s.input in
                                        if head == str
-                                         then OK () (S s.input (s.pos + len) s.maxPos)
+                                         then OK str (S s.input (s.pos + len) s.maxPos)
                                          else Fail s.pos ("string " ++ show str)
                                   else Fail s.pos ("string " ++ show str)
 
@@ -156,13 +169,25 @@ eos = P $ \s => pure $ if s.pos == s.maxPos
 
 ||| Succeeds if the next char is `c`
 export
-char : Applicative m => Char -> ParseT m ()
-char c = skip $ satisfy (== c) <?> "char " ++ show c
+char : Applicative m => Char -> ParseT m Char
+char c = satisfy (== c) <?> "char " ++ show c
 
 ||| Parses a space character
 export
 space : Applicative m => ParseT m Char
 space = satisfy isSpace <?> "space"
+
+||| Parses a letter or digit (a character between \'0\' and \'9\').
+||| Returns the parsed character.
+export
+alphaNum : Applicative m => ParseT m Char
+alphaNum = satisfy isAlphaNum <?> "letter or digit"
+
+||| Parses a letter (an upper case or lower case character). Returns the
+||| parsed character.
+export
+letter : Applicative m => ParseT m Char
+letter = satisfy isAlpha <?> "letter"
 
 mutual
     ||| Succeeds if `p` succeeds, will continue to match `p` until it fails
@@ -290,3 +315,47 @@ integer : Monad m => ParseT m Integer
 integer = do minus <- succeeds (char '-')
              x <- some digit
              pure $ if minus then (intFromDigits x)*(-1) else intFromDigits x
+
+
+||| Parse repeated instances of at least one `p`, separated by `s`,
+||| returning a list of successes.
+|||
+||| @ p the parser for items
+||| @ s the parser for separators
+export
+covering
+sepBy1 : Monad m => (p : ParseT m a)
+                 -> (s : ParseT m b)
+                 -> ParseT m (List1 a)
+sepBy1 p s = [| p ::: many (s *> p) |]
+
+||| Parse zero or more `p`s, separated by `s`s, returning a list of
+||| successes.
+|||
+||| @ p the parser for items
+||| @ s the parser for separators
+export
+covering
+sepBy : Monad m => (p : ParseT m a)
+                -> (s : ParseT m b)
+                -> ParseT m (List a)
+sepBy p s = optionMap [] forget (p `sepBy1` s)
+
+||| Parses /one/ or more occurrences of `p` separated by `comma`.
+export
+covering
+commaSep1 : Monad m => ParseT m a -> ParseT m (List1 a)
+commaSep1 p = p `sepBy1` (char ',')
+
+||| Parses /zero/ or more occurrences of `p` separated by `comma`.
+export
+covering
+commaSep : Monad m => ParseT m a -> ParseT m (List a)
+commaSep p = p `sepBy` (char ',')
+
+||| Run the specified parser precisely `n` times, returning a vector
+||| of successes.
+export
+ntimes : Monad m => (n : Nat) -> ParseT m a -> ParseT m (Vect n a)
+ntimes    Z  p = pure Vect.Nil
+ntimes (S n) p = [| p :: (ntimes n p) |]
--- a/tests/chez/chez027/StringParser.idr
+++ b/tests/chez/chez027/StringParser.idr
@@ -4,6 +4,7 @@ import Control.Monad.Identity
 import Control.Monad.Trans
 
 import Data.Maybe
+import Data.Vect
 import Data.String.Parser
 
 %default partial
@@ -44,7 +45,7 @@ main = do
     res <- parseT (string "hi") "hiyaaaaaa"
     case res of
         Left err => putStrLn "NOOOOOOO!"
-        Right ((), i) => printLn i
+        Right (_, i) => printLn i
     bad <- parseT (satisfy isDigit) "a"
     showRes bad
     bad2 <- parseT (string "good" <?> "Not good") "bad bad bad"
@@ -61,5 +62,13 @@ main = do
     res <- parseT maybeParser "abcdef"
     showRes res
     res <- parseT maybeParser "def"
+    showRes res
+    res <- parseT (commaSep alphaNum) "a,1,b,2"
+    showRes res
+    res <- parseT (ntimes 4 letter) "abcd"
+    showRes res
+    res <- parseT (requireFailure letter) "1"
+    showRes res
+    res <- parseT (requireFailure letter) "a" -- Should error
     showRes res
     pure ()
--- a/tests/chez/chez027/expected
+++ b/tests/chez/chez027/expected
@@ -9,5 +9,9 @@ Parse failed at position 0: Not good
 ""
 True
 False
+['a', '1', 'b', '2']
+['a', 'b', 'c', 'd']
+()
+Parse failed at position 0: Purposefully changed OK to Fail
 1/1: Building StringParser (StringParser.idr)
 Main> Main> Bye for now!


gallais:
> Ran into that issue when trying to parse `(char '+') <|> (char '-')` and realizing I couldn't inspect the result usefully.

Haven't had a look at the PR yet but FYI this is best practice:

```idris
  p <- (Plus <$ char '+') <|> (Minus <$ char '-')
  case p of
    Plus => ...
    Minus => ...
```

as oppose to the dodgy

```idris
  p <- char '+' <|> char '-'
  case p of
    '+' => ...
    '-' => ...
    _ => ?hole -- now what? catchall?
               -- What if I add `<|> char '*'` above?
               -- I won't get a warning telling me to handle this new case
```

Essentially: if you care about which side of the disjunction the value came
from, you should make sure the *parsed* data returns that information rather
than re-analysing the underlying value of some literal.

Note that you have `choose : Parser a -> Parser b -> Parser (Either a b)`
in case you want an informative alternative to `(<|>)` but don't care enough
to introduce your own auxiliary definition to tag all possibilities.

alexhumphreys:
Oh damn that does look better than what I was doing 😅 Thanks!

libs/contrib/Data/String/Parser.idr
```diff
@@ -290,3 +291,75 @@ integer : Monad m => ParseT m Integer
...
+||| parsed character.
+export
+letter : Monad m => ParseT m Char
+letter = satisfy isAlpha <?> "letter"
+
+||| Succeeds if and only if the argument parser fails.
+|||
+||| In Parsec, this combinator is called `notFollowedBy`.
+export
+requireFailure : Monad m => (p : ParseT m a)
```
clayrat, line 355:
```suggestion
requireFailure : Functor m => ParseT m a -> ParseT m ()
requireFailure (P runParser) = P $ \s => reverseResult s <$> runParser s
where
  reverseResult : State -> Result a -> Result ()
  reverseResult  s             (Fail _ _) = OK () s
  reverseResult (S _ _ maxPos) (OK _ _)   = Fail maxPos "Purposefully changed OK to Fail" -- not sure about maxPos
```

libs/contrib/Data/String/Parser.idr
```diff
@@ -290,3 +291,75 @@ integer : Monad m => ParseT m Integer
...
+||| Parses a letter or digit (a character between \'0\' and \'9\').
+||| Returns the parsed character.
+export
+alphaNum : Monad m => ParseT m Char
+alphaNum = satisfy isAlphaNum <?> "letter or digit"
+
+||| Parses a letter (an upper case or lower case character). Returns the
+||| parsed character.
+export
+letter : Monad m => ParseT m Char
```
clayrat, line 348:
```suggestion
letter : Applicative m => ParseT m Char
```

libs/contrib/Data/String/Parser.idr
```diff
@@ -290,3 +291,75 @@ integer : Monad m => ParseT m Integer
...
+||| of successes.
+export
+ntimes : Monad m => (n : Nat) -> ParseT m a -> ParseT m (Vect n a)
+ntimes    Z  p = pure Vect.Nil
+ntimes (S n) p = [| p :: (ntimes n p) |]
+
+||| Parses a letter or digit (a character between \'0\' and \'9\').
+||| Returns the parsed character.
+export
+alphaNum : Monad m => ParseT m Char
```
clayrat, line 342:
```suggestion
alphaNum : Applicative m => ParseT m Char
```

libs/contrib/Data/String/Parser.idr
clayrat: Um, dunno how to make a multiline suggestion..

clayrat (commented):
It'd be also nice to regroup `requireFailure` with other `Functor`-tier combinators and ditto for `letter` and `alphaNum`

tests/chez/chez027/StringParser.idr
```diff
@@ -38,13 +39,28 @@ maybeParser = do res <- optional (string "abc")
                  string "def"
                  pure $ isJust res
 
+-- test commaSep
+commaSepParser : ParseT IO (List Char)
+commaSepParser = do res <- commaSep alphaNum
```
clayrat, line 44: Can we just return `commaSep alphaNum` directly without the rewrapping here and below?

alexhumphreys: Updated.

libs/contrib/Data/String/Parser.idr
alexhumphreys: Updated.

libs/contrib/Data/String/Parser.idr
alexhumphreys: Updated.

libs/contrib/Data/String/Parser.idr
alexhumphreys: This version looks much better! Updated.

alexhumphreys:
@clayrat thanks for the review, I added all the feedback. Just for my own learning so I know for future, what makes some parsers `Applicative` vs `Functor` vs `Mondad`-tier? If it'd take too long to explain then no worries, but I am just curious.

clayrat:
It basically boils down to which primitive combinators you're using. `Functor`-tier stuff can only blindly work on existing data (`skip`/`optionMap`). `Applicative` is a bit weird - the `Applicative` instance for `Parser` itself requires a `Monad` on `m` because we need to thread the parse state through `<*>`, so we can only make primitive matchers with unwrapping the definition and using `pure` directly. And `Monad` is, well, for everything else where you have any sort of context awareness.

clayrat:
As an aside, I wonder if there exist useful [Selective](https://github.com/clayrat/idris-selective/)-tier combinators?

alexhumphreys:
Ok, think i get the functor/monad distinction, will have to closely look at the applicative ones to see if I grok it.

And that selective stuff looks cool, will have a read of the references 👍

libs/contrib/Data/String/Parser.idr
```diff
@@ -121,6 +122,17 @@ export
...
+||| Succeeds if and only if the argument parser fails.
+|||
+||| In Parsec, this combinator is called `notFollowedBy`.
+export
+requireFailure : Functor m => ParseT m a -> ParseT m ()
+requireFailure (P runParser) = P $ \s => reverseResult s <$> runParser s
+where
+  reverseResult : State -> Result a -> Result ()
+  reverseResult  s             (Fail _ _) = OK () s
+  reverseResult (S _ _ maxPos) (OK _ _)   = Fail maxPos "Purposefully changed OK to Fail" -- not sure about maxPos
```
gallais, line 134: I think you should use `pos` here: the failure is at the given position, not
the end of the file. I would also write this as

```suggestion
  reverseResult s (OK _ _)   = Fail (pos s) "Purposefully changed OK to Fail"maxPos
```

so that it's robust to us changing the definition of `State`.

libs/contrib/Data/String/Parser.idr
```diff
@@ -290,3 +314,47 @@ integer : Monad m => ParseT m Integer
...
+||| Parse repeated instances of at least one `p`, separated by `s`,
+||| returning a list of successes.
+|||
+||| @ p the parser for items
+||| @ s the parser for separators
+export
+covering
+sepBy1 : Monad m => (p : ParseT m a)
+                 -> (s : ParseT m b)
+                 -> ParseT m (List a)
```
gallais, line 328: Consider returning a `List1` given that `sepBy1` returns a non-empty list.

libs/contrib/Data/String/Parser.idr
clayrat: Or even `s.pos`

gallais: Sure. Given that some of the postfix syntax changes have been reverted, I am
not sure what parses and what does not so I was overly cautious in my suggestion.

alexhumphreys: I went with `(pos s)` just to avoid any postfix issues, plus it's what I'm used to let. Let me know if I should change it.

libs/contrib/Data/String/Parser.idr
alexhumphreys: Good call, updated. It made `sepBy` a bit of fun to implement as i had to call `forget`, but I think I got it right.

libs/contrib/Data/String/Parser.idr
```diff
@@ -290,3 +315,48 @@ integer : Monad m => ParseT m Integer
...
+|||
+||| @ p the parser for items
+||| @ s the parser for separators
+export
+covering
+sepBy : Monad m => (p : ParseT m a)
+                -> (s : ParseT m b)
+                -> ParseT m (List a)
+sepBy p s = (do a <- (p `sepBy1` s)
+                pure $ forget a) <|> pure []
```
clayrat, line 343: You can also implement it with ```optionMap [] forget (p `sepBy1` s)```

clayrat: Come to think of it, all instances of `bar <|> pure foo` here can be rewritten as `optional foo bar` too, not sure if it's beneficial performance-wise though.

clayrat: Thinking about it a bit, does this combinator make sense in general? I.e., do you really want to get `[]` out of parsing `"abc"` with ```integer `sepBy` char ','```?

alexhumphreys: Yeah I found that a little weird too, but I just copied it from [Lightyear](https://github.com/ziman/lightyear/blob/f737e25a09c1fe7c5fff063c53bd7458be232cc8/Lightyear/Combinators.idr#L77) and it seems [parsec](http://hackage.haskell.org/package/parsec-3.0.0/docs/Text-Parsec-Combinator.html#v:sepBy) behaves the same.

gallais: > do you really want to get [] out of parsing "abc" with integer `sepBy` char ','?

Yeah, why not? If you want to insist on getting at least one value then you should use `sepBy1`.

alexhumphreys: updated to use:
```
optionMap [] forget (p `sepBy1` s)
```
because that looks much better.

As regards what this parser _should_ be, good question. It seems like the ideal behaviour would be something like:
```
sepBy p s = (do pure $ forget !(p `sepBy1` s)) <|> (spaces *> pure [])
```
so there's a difference between the `sepBy1` parsing failing in that case that the text is empty vs the case that the text is garbage. There may be a reason the other implementations didn't do it this way that I'm missing, eg. I can imagine that `spaces` might be a bad assumption.

gallais: I would expect

```idris
id <$ char '[' <*> sepBy p (char ',') <* char ']'
```

to successfully parse lists of `p` values, including the empty list "[]".

gallais: Also it's probably a bad idea to arbitrarily munch some of the space:
someone may actually care very much about consuming space themselves
(e.g. because their syntax is indentation-based)

clayrat:
@alexhumphreys Btw, one other thing you could do by limiting yourself to `Applicative` without running into the issue of threading the parse state, would be to define specialized versions of `<*`/`*>`, where you run some computation only for its effect, e.g.:
```
(<**) : Applicative m => ParseT m a -> m b -> ParseT m a
(<**) pf ma = P $ \s => (pf.runParser s) <* ma
```
However I think this would pollute the design space without winning you much.

libs/contrib/Data/String/Parser.idr
alexhumphreys: @gallais so the example you gave surrounded by `[]` seems to behave as you expect (I slightly changed it below, hopefully not in important ways), it's the one that @clayrat suggested that's a bit strange:

```
Main> parse (char '[' *> sepBy letter (char ',') <* char ']') "[a,b]"
Right (['a', 'b'], 5)  -- ✅
Main> parse (char '[' *> sepBy letter (char ',') <* char ']') "[a]"
Right (['a'], 3) -- ✅
Main> parse (char '[' *> sepBy letter (char ',') <* char ']') "[]"
Right ([], 2) -- ✅
Main> parse (char '[' *> sepBy letter (char ',') <* char ']') "[1]"
Left (fastConcat ["Parse failed at position ", "1", ": ", "char ']'"]) -- ✅
Main> parse (sepBy letter (char ',')) "a"
Right (['a'], 1) -- ✅
Main> parse (sepBy letter (char ',')) "a,b"
Right (['a', 'b'], 3) -- ✅
Main> parse (sepBy letter (char ',')) "1"
Right ([], 0) -- 🤨
Main> parse (sepBy letter (char ',')) "]]]"
Right ([], 0) -- 🤨
```

I would have expected the last 2 to be parse errors and not empty lists. It seems lightyear and parsec behave this way, so I'm ok with it, it's just a little surprising to me.

clayrat: Something like `parse (commaSep integer) "[a,b,c]"` might be even more puzzling potentially

gallais: The pointer into the string returned at the end is still at `0` which
indicates nothing has been consumed. If you meant for something to only
succeed if it parses a whole input then you should have a parser of the
form `p <* eof`!

alexhumphreys: Oh! I hadn't spotted that `0`, makes much more sense now 😄

gallais:
Thanks!

PR #781, opened 2020-11-20
[ contrib ] adding Data.Container

gallais:
The last proof was quite tricky.


libs/contrib/Data/Container.idr
```diff
@@ -312,7 +312,7 @@ namespace Derivative
   export
   fromCompose : Extension (Derivative (Compose c d)) x ->
                 Extension (Pair (Derivative d) (Compose (Derivative c) d)) x
-  fromCompose (MkExtension (MkExtension shp1 shp2 ** p1 ** p2) chld)
+  fromCompose (MkExtension (MkExtension shp1 shp2 ** (p1 ** p2)) chld)
```
buzden, line 315: I thought that `**` is already right-associative. Why are these parenthesis needed?

gallais: My guess it that this change was introduced after we released the version of Idris
that is currently used for the bootstrapping process. Similarly `let x : t := e in b`
is valid in the dev version but not in the bootstrap one so we cannot use this form
yet in either the libs or the compiler.

PR #782, opened 2020-11-21
prim void to js codegen

rbarreiro:
avoid the error below on the javascript codegen
Error: INTERNAL ERROR: prim not implemented: Prelude.Uninhabited.void

### DIFF
--- a/src/Compiler/ES/ES.idr
+++ b/src/Compiler/ES/ES.idr
@@ -308,7 +308,7 @@ jsOp (Cast Bits64Type Bits16Type) [x] = boundedUInt 16 x
 jsOp (Cast Bits64Type Bits32Type) [x] = boundedUInt 32 x
 
 jsOp (Cast ty StringType) [x] = pure $ "(''+" ++ x ++ ")"
-jsOp (Cast ty ty2) [x] = jsCrashExp $ "invalid cast: + " ++ show ty ++ " + ' -> ' + " ++ show ty2
+jsOp (Cast ty ty2) [x] = jsCrashExp $ jsString $ "invalid cast: + " ++ show ty ++ " + ' -> ' + " ++ show ty2
 jsOp BelieveMe [_,_,x] = pure x
 jsOp (Crash) [_, msg] = jsCrashExp msg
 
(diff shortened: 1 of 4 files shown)


gallais (changes requested):
Thanks for the fix! Can you please rename `void` to `prim__void` to
match the convention we use for external commands? You would then
introduce a new `void` that is simply an alias to `prim__void` but,
unlike the `prim__` one is exported.

Cf. [`System.Info`](https://github.com/idris-lang/Idris2/blob/f33feecbea82aee09883eed0f50b0f68216085ab/libs/base/System/Info.idr#L3) for instance for .

src/Compiler/ES/ES.idr
```diff
@@ -377,6 +377,7 @@ jsPrim (NS _ (UN "prim__os")) [] =
     let oscalc = "(o => o === 'linux'?'unix':o==='win32'?'windows':o)"
     sysos <- addConstToPreamble "sysos" (oscalc ++ "(" ++ os ++ ".platform())")
     pure sysos
+jsPrim (NS _ (UN "void")) [_, _] = jsCrashExp $ jsString $ "Error: Executed 'void'"
```
rbarreiro, line 380: Thank you for the review :)
```suggestion
jsPrim (NS _ (UN "void")) [_, _] = jsCrashExp $ jsString $ "Error: Executed 'prim__void'"
```
is this the change you meant?

rbarreiro: https://github.com/idris-lang/Idris2/blob/f33feecbea82aee09883eed0f50b0f68216085ab/src/Compiler/Scheme/Common.idr#L448
I got the inspiration from here, should both be changed?

gallais: No I meant changing `prelude` to match the convention of calling `%extern` functions
`prim__SOMETHING`. You'd then have to change the left-hand side here to match on
`jsPrim (NS _ (UN "prim__void"))`  instead of `void` but the RHS can stay the same.

rbarreiro: Humm, that way I would have to change Prelude, no?
https://github.com/idris-lang/Idris2/blob/e93fba3952649762d8d165124c0888756547dacf/libs/prelude/Prelude/Uninhabited.idr#L15-L18

rbarreiro: I need some help, I get this error when I boostrap:
Uncaught error: INTERNAL ERROR: Can't compile unknown external primitive Prelude.Uninhabited.prim__void

How should I go about this?

gallais: I have pushed a patch. Let's see if it works.

gallais:
Right, we can't rename the primitives without changing the bootstrapping code.
So let's give up on that renaming. :(

rbarreiro:
maybe we can have the backends support the two forms, and change the prelude only when the bootsrap is updated

gallais:
Good point.

gallais (approved):

Issue #785, opened 2020-11-22
Parsing literate files


Yu-zh:
Idris2 was running into issues with the literate file: https://gist.github.com/Yu-zh/f0053ef856b40d336f2af6ad599346c3. This file comes from [IEdit.org](tests/idris2/literate012/IEdit.org). I added an extra line at the beginning and adjusted the input file to: https://gist.github.com/Yu-zh/a0f7cf4923d2229645ed577ddd2a5c40. Then it doesn't work well.

# Steps to Reproduce

```
idris2 --no-color --console-width 0 --no-banner IEdit.org < input
```

# Expected Behavior

```
Main> append {n = 0} [] ys = ?foo_1
append {n = (S k)} (x :: xs) ys = ?foo_2
Main> vadd [] [] = ?bar_1
Main> vadd (x :: xs) (y :: ys) = ?baz_1
Main> suc x x Refl = ?quux_1
Main> suc' {x = y} {y = y} Refl = ?quuz_1
Main> Bye for now!
```

# Observed Behavior

```
1/1: Building IEdit (IEdit.org)
Main> No clause to split here
Main> No clause to split here
Main> No clause to split here
Main> No clause to split here
Main> No clause to split here
Main> Bye for now!
```

PR #786, opened 2020-11-22
Improving error messages

### LINKED ISSUE #784: Location information in error messages
Below is the output by running the test case [idris2/total007/partial007.idr](https://github.com/idris-lang/Idris2/blob/22bfa909710da92f15e832e5c66fbc68e6e017c3/tests/idris2/total007/partial.idr). It seems that the reported locations of the problematic functions wrap around to the next line. This happens in other cases, too. Is this considered expected?

# Observed Behavior

```
1/1: Building partial (partial.idr)
Error: foo is not covering.

partial.idr:5:1--7:4
 5 | total
 6 | foo : Maybe a -> a
 7 | foo (Just x) = x

Missing cases:
    foo Nothing

Error: qsortBad is not total, possibly not terminating due to recursive path Main.qsortBad -> Main.qsortBad -> Main.qsortBad

partial.idr:13:1--15:9
 13 | total
 14 | qsortBad : Ord a => List a -> List a
 15 | qsortBad [] = []
```

Yu-zh:
This fixes #784 and #785. First time contributor, please let me know if you have any suggestions or if my changes are actually unwanted.


src/Idris/Parser.idr
```diff
@@ -1390,13 +1390,19 @@ topDecl fname indents
...
-    = let (cs', rest) = spanBy isClause ds in
-          PDef annot (cs ++ concat cs') :: assert_total (collectDefs rest)
+    = let (csWithFC, rest) = spanBy isClause ds 
+          cs' = cs ++ concat (map snd csWithFC)
+          annot' = foldr mergeFC annot (map fst csWithFC)
+      in
+          PDef annot' cs' :: assert_total (collectDefs rest)
   where
-    isClause : PDecl -> Maybe (List PClause)
+    isClause : PDecl -> Maybe (FC, List PClause)
```
gallais, line 1399: Now that `isClause` returns exactly all of the payloads of the `PDef`
constructor, I would lift it as a toplevel function in `Idris.Syntax`
and call it `isPDef`.

src/Idris/Parser.idr
```diff
@@ -1390,13 +1390,19 @@ topDecl fname indents
...
     isClause (PDef annot cs)
-        = Just cs
+        = Just (annot, cs)
     isClause _ = Nothing
+    mergeFC : FC -> FC -> FC
```
gallais, line 1403: `mergeFC` should go `Core.FC`. We should probably define the more general:
`mergeFC : FC -> FC -> Maybe FC` that fails in case the two input `FC` are
incompatible (if they reference different file names or if one is empty but
not the other).

We can then use `fromMaybe EmptyFC (mergeFC x y)` here.

src/Parser/Rule/Common.idr
```diff
@@ -16,19 +16,19 @@ EmptyRule token ty = Grammar token False ty
 export
 location : {token : _} -> EmptyRule token (Int, Int)
 location
-    = do tok <- bounds peek
+    = do tok <- weakBounds peek
```
gallais, line 19: Not sure what the point of `weakBounds` is here: we're extracting
the `Int`s anyway and throwing the `WithBounds` away?

Worth noting btw that:

* `(tok.startLine, tok.startCol)` can be refactored as `start tok`
* `(tok.endLine, tok.endCol)` can be refactored as `end tok`

tests/idris2/coverage010/expected
```diff
@@ -1,8 +1,9 @@
...
 Error: main is not covering.
 
-casetot.idr:12:1--13:5
+casetot.idr:12:1--12:13
+    |
  12 | main : IO ()
- 13 | main = do
+    | ^^^^^^^^^^^^
 
 Calls non covering function Main.case block in case block in main
```
gallais, line 9: Any chance we can have the error point at the faulty case rather than
the toplevel function enclosing it?

tests/idris2/error004/expected
```diff
@@ -27,5 +27,5 @@ Error2.idr:16:9--16:34
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Possible correct results:
-    Show implementation at Error2.idr:11:1--15:6
-    Show implementation at Error2.idr:7:1--11:5
+    Show implementation at Error2.idr:11:1--13:45
+    Show implementation at Error2.idr:7:1--9:45
```
gallais, line 31: These ranges are much better. Thanks!

tests/idris2/interface016/expected
```diff
@@ -9,4 +9,4 @@ TwoNum.idr:4:7--4:8
 
 Possible correct results:
     conArg (implicitly bound at TwoNum.idr:4:3--4:8)
-    conArg (implicitly bound at TwoNum.idr:2:1--5:1)
+    conArg (implicitly bound at TwoNum.idr:2:1--4:8)
```
gallais, line 12: I would rather have `2:1--2:N` but this is an improvement nonetheless.

gallais (commented):
Thanks for the fixes! I have added some small refactoring suggestions
and questions for parts of the code I did not understand.

I have also added some requests on the off chance you immediately see
how they can be fixed now that you're steeped into ranges and positions
but they are not needed for this PR to be accepted. ;)

src/Parser/Rule/Common.idr
Yu-zh: Yes, I introduced `weakBounds` because I didn't find a way to throw away position information in the existing code. If we use `bounds`, the position information will always stick around. That's the reason why some error messages wrap around to the next line, because the position information of the next token get merged with what has been just parsed.

Yu-zh:
@gallais Thanks for the suggestions. I have just refactored the code. I don't have solutions to further improve the error information off the top of my head, but I will keep them mind.

src/Text/Parser/Core.idr
```diff
@@ -84,6 +86,7 @@ Functor (Grammar tok c) where
   map f (SeqEmpty act next)
       = SeqEmpty act (\val => map f (next val))
   map {c} f (Bounds act) = rewrite sym $ orFalseNeutral c in SeqEmpty (Bounds act) (Empty . f) -- Bounds (map f act)
+  map {c} f (WeakBounds act) = rewrite sym $ orFalseNeutral c in SeqEmpty (Bounds act) (Empty . f) -- Bounds (map f act)
```
gallais, line 89: Should this preserve `WeakBounds`?

src/Text/Parser/Core.idr
```diff
@@ -152,6 +155,7 @@ mapToken f (SeqEat act next) = SeqEat (mapToken f act) (\x => mapToken f (next x
 mapToken f (SeqEmpty act next) = SeqEmpty (mapToken f act) (\x => mapToken f (next x))
 mapToken f (Alt x y) = Alt (mapToken f x) (mapToken f y)
 mapToken f (Bounds act) = Bounds (mapToken f act)
+mapToken f (WeakBounds act) = Bounds (mapToken f act)
```
gallais, line 158: Same question here.

gallais (commented):
I think the map functions may be broken.

src/Text/Parser/Core.idr
Yu-zh: Thanks for pointing this out. It was my bad that I just copied and pasted the code from the `Bounds` case. Actually, after reading how `map` is implemented, I realized `map` is exactly what I need to get rid of the position information in some cases. So I just removed the whole `WeakBound` thing and use `map` to get the job done.

src/Parser/Rule/Common.idr
```diff
@@ -16,20 +16,20 @@ EmptyRule token ty = Grammar token False ty
...
+    = do tok <- removeIrrelevance <$> bounds peek
+         pure $ start tok
 
 export
 endLocation : {token : _} -> EmptyRule token (Int, Int)
 endLocation
-    = do tok <- bounds peek
-         pure (tok.endLine, tok.endCol)
+    = do tok <- removeIrrelevance <$> bounds peek
+         pure $ end tok
```
gallais, line 26: I still don't understand the point of this `removeIrrelevance <$>`:
the value returned by the parser is `end tok` and it does not contain
this irrelevance flag as it's just a pair of two `Int`s.

Yu-zh: https://github.com/idris-lang/Idris2/blob/a7cb2745bcc7cf6b920ff00eecb18372835c4a06/src/Text/Parser/Core.idr#L261
The link points to how values returned by two parsers are composed. Here we are actually composing `bounds peek` and `pure $ end tok`; the former carries bounds whereas the latter doesn't. In this case, bounds from `bounds peek` will be carried over to the ultimate value returned by the parser. 

By adding `removeIrrelevance` I am actually trying to void the bounds carried by `bounds peek` so that the returned value of the composed parser doesn't carry any bounds. I hope this makes sense.

gallais:
Thanks!

PR #787, opened 2020-11-22
Local hints

ohad:
Edit: Better support for this is partly implemented in PR #846 , but the translation into a `let` binding is restrictive.

This WIP-PR explores using AutoSearch `hint`s locally, for example (taken from the test [`idris2/interface018`](https://github.com/ohad/Idris2/blob/27b3d3e0007a6a1d5b0e18b42e9eae05825bb52b/tests/idris2/interface018/LocalHints.idr#L1)):
```idris
findA : {auto foo : Foo} -> String
findA {foo = A} = "Found an A"
findA {foo = _} = "Failed to find an A"

Baz : String -> Type
Baz s = s = "Found an A" 

baz : (s : String ** Baz s)
baz = let %hint arg : Foo
          arg = A
      in (findA ** Refl)
```
This is already possible if we use `let`-binding, with the difference that `let`-bound values are kept abstract in types, so the above code really needs a local definition, and won't work with a `let`-binding. This kind of type-level visibility is crucial for programs that do type-level computation.

I can see this kind of feature useful when we need to manipulate sets of notation. For example (taken from the test [`idris2/interface019`](https://github.com/ohad/Idris2/blob/27b3d3e0007a6a1d5b0e18b42e9eae05825bb52b/tests/idris2/interface019/AlgebraLocalHints.idr#L87), in the [Eckmann-Hilton argument](https://en.wikipedia.org/wiki/Eckmann%E2%80%93Hilton_argument), we manipulate two different monoid structures over the same carrier, and with the local hints mechanism we can easily re-map the monoid operations:
```idris
EckmannHilton ma mb prf = 
  let %hint first : Additive a
      first = AdditiveStruct ma
      %hint second : Additive2 a
      second = AdditiveStruct2 mb in
  let SameUnits : (the a O === O2)
      SameUnits = Calc $
        |~ O
        ~~  O         .+.         O  ...(sym $ lftUnit ma O)
        ~~ (O :+: O2) .+. (O2 :+: O) ...(sym $ cong2 (.+.) 
                                             (rgtUnit mb O)
                                             (lftUnit mb O))
    ...
```

This PR is still WP because local hints end up being global still:
```idris
data Foo : Type where [noHints]
  A : Foo
  B : Foo

findFoo : {auto ambient : Foo} -> Foo
findFoo @{foo} = foo

foo1 : Nat -> Foo
foo1 n = let %hint aFoo : Foo
             aFoo = A
         in findFoo

foo2 : Nat -> Foo -- Should fail!
foo2 n = let _ = let %hint bFoo : Foo
                     bFoo = B
                 in ()
         in findFoo         

foo3 : Nat -> Foo
foo3 n = findFoo -- Should fail!
```
Example `foo1` should type-check fine because of the local hint. But both `foo2` and `foo3` should fail because the hints should be localised to their scope. I'll explain below why this happens, but the tl;dr is that currently we never erase hints.
(The reason we don't get an ambiguity error is more subtle.)

The reason this is difficult is that local definitions are not stored as variables in the current environment. Instead, they are lifted to the top-level, where they are type-checked as a top-level definition, and all the variables that reference them from the scope are substituted by a `Ref` to this top-level definition. There are technical and historical reasons why this is the case, and changing it right now is going to be a substantial rewrite.
The reason this complicates auto search is that, unlike variables which are explicit in the source code, we don't quite know where in auto search we'll need these local definitions at the time we elaborate the local definition. By the time we get there, we're somewhere deeper in the scope of the local definition, and we need to do extra work to recover the definition in the form it was before lifting.

We need to undo this lifting because we won't find the correct arguments using auto search, and so auto search won't apply the local hint. For example, in the Eckmann-Hilton argument above, we internally lifted the hint:
```idris
%hint first : (0 a : Type) -> (ma, mb : Monoid a) -> (prf : ...) -> Additive a
```
Since we have two variables of the same type `ma, mb : Monoid a`, auto search will fail to find them as arguments to `first`. Looking for these arguments with auto search is silly, since we already knew what they were before we lifted the definition. So we need to implement a `lower` mechanism to undo the lifting and start auto search from where it's needed.

The way we implement local hints here then is to add an extra argument to the global dictionary that stores hints: the depth in the environment where the local definition was defined. When we leave the scope of this local definition, we will delete the hint from the global dictionary. During auto search, if the local definition is still in the global dictionary, we'll lower it back to the current environment.

What's currently implemented is wrong, because we don't remove local hints when we leave their scope. That's because unification sometimes delay elaboration to a later stage, after some more constraints have been solved. For example, the following line in the [product monoid](https://github.com/ohad/Idris2/blob/27b3d3e0007a6a1d5b0e18b42e9eae05825bb52b/tests/idris2/interface019/AlgebraLocalHints.idr#L65):
```idris
     (\(x1,y1), (x2, y2)  => (x1 .+. x2, y1 .+. y2))
``` 
is delayed until after we leave the scope, probably because we need to synthesise types for the arguments `x1,y1,x2,y2`.

TODO: Whatever we implement should also be replicated in interactive expression search (why not?).

I'm making this PR to get some comments about the mechanism, and suggestions (esp. @gallais and @edwinb ) on how to best implement it without changing the existing Idris2 architecture too much.

### DIFF
--- a/idris2api.ipkg
+++ b/idris2api.ipkg
@@ -45,6 +45,7 @@ modules =
     Core.Hash,
     Core.InitPrimitives,
     Core.LinearCheck,
+    Core.Lower,
     Core.Metadata,
     Core.Name,
     Core.Name.Namespace,
--- a/libs/contrib/Data/List/Equalities.idr
+++ b/libs/contrib/Data/List/Equalities.idr
@@ -1,5 +1,6 @@
 module Data.List.Equalities
 
+import Data.Nat
 import Data.List
 
 %default total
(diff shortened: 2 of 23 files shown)


ohad:
Implemented properly in #1153

PR #789, opened 2020-11-23
Add dockerfile

alexhumphreys:
I'm not sure if you're interested in Docker, seeing as there was a Dockerfile for idris1 that [got removed](https://github.com/idris-lang/Idris-dev/commit/bdf5666708afc031fc588abff0ff49e02952ffe4), but I'm using a docker image built from this to [run tests](https://github.com/alexhumphreys/idrall/blob/master/.github/workflows/blank.yml#L19), so figured I'd open a PR and see if ye are interested. I can always close it if ye are not :) 

Can think later if you want to build/push it to Dockerhub, using github actions or Dockerhub builds, but for this PR it's just about the Dockerfile.

Signed-off-by: Alex Humphreys <alex.humphreys@here.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+FROM ubuntu:20.04 AS build
+
+RUN apt update && apt install gcc make chezscheme -y
+
+WORKDIR /opt/idris2
+
+ADD Makefile .
+ADD libs .
+ADD src .
+ADD support .
+
+ADD Release/ Release/
+ADD config.mk .
+ADD idris2api.ipkg .
+ADD support/ support/
+ADD LICENSE .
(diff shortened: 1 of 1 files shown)


edwinb:
I can see why people might want this sort of thing, but adding more things gives me more things to have to know about and maintain, so I'd prefer this sort of thing to be managed externally. The same would go for all the other possible ways of distributing Idris. Sorry!

alexhumphreys:
That's cool. I can try think of some way to automate the docker stuff external of this repo for my own usage, and just keep publishing my [`snazzybucket/idris2` image](https://hub.docker.com/repository/docker/snazzybucket/idris2).

PR #793, opened 2020-11-25
Apostrophies in modules were made to work

buzden:
Apostrophies are allowed in modules however you can get errors during code generation and when calling for the editor in the REPL. I tried to fix this.

By the way, I added a test only for chez since not all of codegens support the `:exec` command and different codegens can give different outputs for `:compile` command. However all the codegens had the original problem. Is it a good idea to try to organize separate testing target "execable" where to put all such tests and call it with all codegens?


gallais:
> By the way, I added a test only for chez since not all of codegens support the :exec command and different codegens can give different outputs for :compile command.

You can use `--cg ${IDRIS2_TESTS_CG}`.

buzden:
> You can use `--cg ${IDRIS2_TESTS_CG}`.

I'm afraid I don't understand you. Well, I know that I can run Idris with different codegens but it won't help me to write a test that would run for users with one codegen and won't run for those with another. This situation is like when "chez" tests are not run when you are using racket. So, "chez" tests have a precondition "codegen is chez". I'm talking about tests that have a precondition "codegen supports ':exec' command".

gallais:
We already have a test running `--exec` on all backens using this option:
https://github.com/idris-lang/Idris2/blob/master/tests/idris2/evaluator004/run#L2

buzden:
> We already have a test running `--exec` on all backends using this option

Oh, I see, we actually assume that all codegens support ":exec" and thus we don't need special class of tests for this. Let me change my test.

gallais:
> Oh, I see, we actually assume that all codegens support ":exec" and thus we don't need special class of tests for this. Let me change my test.

At least the one we are running continuous integration for. When the tests
are run by a user, unless the user specifies what they want we try a series
of defaults we know should work. Cf. [this bit of code](https://github.com/idris-lang/Idris2/blob/master/tests/Lib.idr#L254)

PR #799, opened 2020-11-27
[ fix #775 ] integerToNat is not, in fact, id

### LINKED ISSUE #775: converting negative Integer to Nat fails to return Z
When casting a negative `Integer` to `Nat` using `fromInteger` at runtime, the result remains negative.

# Steps to Reproduce
```idris
main : IO ()
main = do
  n <- map (the Nat . fromInteger . cast) getLine
  printLn n
```
Pass `-1` on standard input.

# Expected Behavior
This should print `0`, as similar code does in the REPL.

# Observed Behavior
This prints `-1`.  This seems to be true for all codegens.

I think this already shows what the problem is, but we can also show that the compiler indeed believes `n` to be a valid `Nat`:
```idris
%default total

incrNonzero : (n : Nat) -> S n = Z -> Void
incrNonzero _ Refl impossible

showIncrNonzero : Nat -> String
showIncrNonzero n = snz (S n) (incrNonzero n) where
  snz : (sn : Nat) -> (sn = Z -> Void) -> String
  snz (S n) _ = "nonzero"
  snz Z pf = void (pf Refl)

main : IO ()
main = do
  n <- map (the Nat . fromInteger . cast) getLine
  printLn n
  printLn (S n)
  printLn (showIncrNonzero n)
```
On input `-1`, all codegens seem to fail trying to execute `void`.

gallais:
This is only a partial solution. We need to add some magic to make
this cast efficient by replacing `integerToNat` with essentially

```idris2
integerToNat' i = if i < 0 then 0 else i
```

Edit: I think I have a solution. Via a *defined* """primitive""" :D


src/Compiler/Common.idr
```diff
@@ -180,8 +180,7 @@ natHackNames
     = [UN "prim__add_Integer",
        UN "prim__sub_Integer",
        UN "prim__mul_Integer",
-       NS preludeNS (UN "natToInteger"),
```
mattpolzin, line 183: Did you mean to remove `natToInteger` here? Sorry, not familiar with the machinery here yet but not sure how replacing `integerToNat` would remove the need for the inverse here and I noticed `natToInteger` is still handled in `natHack`

gallais: Yep: `natToInteger` is not used (it is removed by [this magic](https://github.com/idris-lang/Idris2/pull/799/files#diff-0fc453ae86601d6a1ec44d86a6d4c4c4d6d23ac8e74ab78797931f4bc8dff852R174)) so there is no point
in making sure it is part of the generated code. My guess is that this got out of sync
a long time ago.

I only noticed it because my `prim_integerToNat` was not getting compiled. So I
started looking for the code that computes a definition's dependencies to decide
what to include in the executable and found this.

mattpolzin: Ohhh. Yeah. That makes sense. I had it backwards in my head; it’s actually that `natHack` begets `natHackNames`. Thanks for the explanation!

gallais: Thanks for the review!

PR #807, opened 2020-12-01
Added minimal readline support

414owen:
Hoping to get the ball rolling on REPL readline support.
This is a very minimal change, with bindings stolen from the `samples/` folder.

It enables rebindable line editing, and session-local history.

It also addresses this comment in `Contributing.md`:
> it'd be nice if Ctrl-C didn't quit the whole system!

Well, now it seems to take two presses of Ctrl-C to exit the system.

Note that when compiled with readline support, the tests fail, due to the prompt not going to stdout.
There doesn't seem to be an easy way around this.
It might be worth creating a REPL monad, which can be interpreted as IO, or tested. I'm hoping this is out of the scope of this PR.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -5,6 +5,10 @@ PREFIX ?= $(HOME)/.idris2
 # For Windows targets. Set to 1 to support Windows 7.
 OLD_WIN ?= 0
 
+# Set to 1 to enable line editing and history functions in the Idris2 REPL,
+# and in System.Readline
+READLINE ?= 0
+
 ##################################################################
 
 RANLIB ?= ranlib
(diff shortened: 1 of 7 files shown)


edwinb:
Hi, thanks for the contribution. Sorry to be a pain, though, but I'd much prefer to do anything at all fancy in the REPL (including readline support) in a separate system, interacting with Idris via the IDE protocol, and keep what we have here using the existing minimal interface. This is partly about keeping dependencies down, and partly about keeping the maintenance burden down.

I started some readline bindings externally here: https://github.com/edwinb/idris2-clibs - maybe this would be the place to add your library support? Then it could be used as the basis for a much fancier REPL.

414owen:
Sounds good.
I'd love to help improve the REPL experience, but won't be able to look into an IDE protocol version for a while, unfortunately.

PR #808, opened 2020-12-01
[ fix #657 ] RigCount for interface parameters

### LINKED ISSUE #657: default interface methods are missing information they need to typecheck
# Steps to Reproduce
typecheck:
```idris
import Data.Buffer

export
data ForeignPtr : Type -> Type where
  MkFP : Buffer -> ForeignPtr a

public export
interface Storable a where
  size : Int

  peekByteOff : HasIO io => ForeignPtr a -> Int -> io a

  peekElemOff : HasIO io => ForeignPtr a -> Int -> io a
  peekElemOff fp off = peekByteOff fp (off * size {a})

Storable Bits8 where
  size = 8
  peekByteOff (MkFP b) off = getBits8 b off
```
# Expected Behavior
No issues
# Observed Behavior
I'd like to have a value in the interface that can be referred to by both user-made and default implementations but as it stands I'm unable to figure out how to achieve this for default ones.
The given codes causes:
```
Error(s) building file src/Foo.idr: While processing right hand side of peekElemOff. Undefined name a. 
src/Foo.idr:16:1--19:1
 16 | Storable Bits8 where
 17 |   size = 8
 18 |   peekByteOff (MkFP b) off = getBits8 b off
```
If we instead remove `{a}` from peekElemOff's `size {a}` thinking that `a` can be inferred from the types at hand we get:
```
Error(s) building file src/Foo.idr: While processing right hand side of {Default implementation of peekElemOff:0}.
Can't find an implementation for Storable ?a.
```
Changing `peekElemOff`'s type to `peekElemOff : Storable a => HasIO io => ForeignPtr a -> Int -> io a`, which I wouldn't have expected to work anyway, gives us:
```
Error(s) building file src/Foo.idr: While processing right hand side of {Default implementation of peekElemOff:0}.
Multiple solutions found in search of:
    Storable a
```
Which is kind of interesting because between the last two errors we go from "Can't find implementation" to 'Multiple solutions", though I think this last error isn't really important since writing `Storable a => ...` inside the `Storable a` interface is a little silly.

gallais:
With this change, interface parameters can be assigned a quantity.

* parameters with a name but no type annotation are considered erased
* parameters with a name & a type need to come with a quantity annoation

So:

```idris
interface Functor f where (...)
interface Functor (0 f : Type -> Type) where (...)
```

mean the same thing. But

```idris
interface Functor (f : Type -> Type) where (...)
```

means that `f` is in scope at an unrestricted quantity for the default definitions
in the block. See `idris2/interface018` for examples such as:

```idris
public export
interface Singleton (n : Nat) where
  sing : (m : Nat ** m === n)
  sing = (n ** Refl)
--        ^---- n is the interface's parameter. It gets specialised in implementations.
```


gallais:
Ughhh. `substNamesClause` captures. :(

gallais:
I am planning to merge this on Monday next week. Leaving it up a bit
longer because it's a breaking change if you were adding a type annotation
to your `interface`'s parameters (they will have an unrestricted quantity
annotation instead of `erased`).

Russoul:
Oops, sorry, the conflicts come from my PR. Shouldn't be a big deal though, the change is small:
interface constraints are now elaborated to auto implicit arguments of its constructor, previously - to explicit ones.

gallais:
Yours came first so if anything I am the one creating conflicts.
Same for the other PR I have just merged changing the spot in
which the IFace constraint is bound.

PR #815, opened 2020-12-04
Suggestion of alternative list index function

buzden:
This PR is somewhat controversial because for addition of simple function, well-known and commonly used `Decidable.Equality` module was split on two parts. However, this split does not influence on regular users of this module.

This split was done to avoid circular dependency of module imports since `Decidable.Equality` imports lots of `Data.*` modules for those types that are defined in the `prelude`, thus preventing, e.g. using `Data.Fin` module in the `Data.List` one.

### DIFF
--- a/libs/base/Data/Fin.idr
+++ b/libs/base/Data/Fin.idr
@@ -2,7 +2,7 @@ module Data.Fin
 
 import public Data.Maybe
 import Data.Nat
-import Decidable.Equality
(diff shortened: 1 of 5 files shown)


gallais:
Given that people are not supposed to import `Decidable.Equality.Def`,
I would probably call it `.Core` or `.Internal`.

buzden:
> I would probably call it `.Core`

Let it be so.

PR #816, opened 2020-12-04
Add javascript FFI for `fastUnpack`

andylokandy:


andylokandy:
PTAL @ziman

shmish111:
@andylokandy according to https://stackoverflow.com/questions/4547609/how-do-you-get-a-string-to-a-character-array-in-javascript it might be safer to use `Array.from` I think.

gallais:
Thanks for the PR and @shmish111 thanks for the review!

PR #819, opened 2020-12-05
Have lambda-case available everywhere lambda is

buzden:
You can use lambda functions without parenthesis and `$`-operator as the last argument:

```idris
f : (Int -> Bool) -> Int

il : Int
il = f \x => x > 0
```

But at the moment you can't do this for the lambda case.

```idris
lc : Int
lc = f $ \case 0 => True ; _ => False -- ok

ilc : Int
ilc = f \case 0 => True; _ => False -- does not parse
```

This PR fixes this.

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -742,6 +742,7 @@ mutual
     <|> implicitPi fname indents
     <|> explicitPi fname indents
     <|> lam fname indents
+    <|> lambdaCase fname indents
 
   typeExpr : ParseOpts -> FileName -> IndentInfo -> Rule PTerm
   typeExpr q fname indents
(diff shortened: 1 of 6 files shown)


gallais:
Is there a reason you are not putting `lambdaCase` in `binder`
so that it may be used everywhere a `lambda` may be used rather
than this ad-hoc fix for arguments only?

buzden:
> Is there a reason you are not putting `lambdaCase` in `binder`
> so that it may be used everywhere a `lambda` may be used rather
> than this ad-hoc fix for arguments only?

No, I think there is no such reason. This is done this way more because of lack of my knowledge of internals. I'll redo the way you are saying.

PR #820, opened 2020-12-06
[ doc ] new application syntax

Russoul:
I have set up automatic deletion of trailing whitespace on my Vim, that's why you see quite a few changes in CONTRIBUTORS.
Hopefully, that's okay.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1,46 +1,47 @@
 Thanks to the following for their help and contributions to Idris 2:
 
 Abdelhakim Qbaich
-Alex Gryzlov 
-Alex Silva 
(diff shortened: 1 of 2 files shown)


gallais:
It would be good to document what happens if two `Pi` types have the same name.
E.g. a (contrived) example:

```idris
module Shadowing
 
%default total
 
0 TypeOf : a -> Type
TypeOf {a} _ = a
 
weirdId : {0 a : Type} -> (1 a : a) -> TypeOf a
weirdId a = a
 
zero : Nat
-- zero = weirdId { a = Z }
--                      ^
-- Type error: expect Type, got Nat
-- But this works:
zero = weirdId { a = Nat, a = Z }
```

PR #822, opened 2020-12-06
export a function to parse ipkg files

shmish111:
I want to parse an ipkg file in some tooling I am working on


shmish111:
I also want to be able to use the PkgDesc once I have parsed it so I've exported the type and fields as well as the Show instance.

gallais:
Seems reasonable. :+1:

PR #823, opened 2020-12-06
Refactor System.Path and Utils.Path

andylokandy:
* Rename `startWith` to `isBaseOf`
* Rename `stripPrefix` to `dropBase`
* Add `dropExtension`
* Add `splitPath`
* Improve document

### DIFF
--- a/libs/contrib/System/Path.idr
+++ b/libs/contrib/System/Path.idr
@@ -109,7 +109,7 @@ Show Volume where
   show (UNC server share) = "\\\\" ++ server ++ "\\" ++ share
   show (Disk disk) = Strings.singleton disk ++ ":"
 
-||| Display the path in the format of this platform.
+||| Displays the path in the format of this platform.
 export
 Show Path where
   show path = 
@@ -245,7 +245,7 @@ parsePath =
                 (x::xs) => x :: delete CurDir xs
     pure $ MkPath vol (isJust root) body (isJust trailSep)
 
-||| Parse a String into Path.
+||| Parses a String into Path.
 |||
 ||| The string is parsed as much as possible from left to right, and the invalid 
 ||| parts on the right is ignored.
@@ -390,7 +390,7 @@ export
 (</>) : (left : String) -> (right : String) -> String
 (</>) left right = show $ append' (parse left) (parse right)
 
-||| Join path components into one.
+||| Joins path components into one.
 |||
 ||| ```idris example
 ||| joinPath ["/usr", "local/etc"] == "/usr/local/etc"
@@ -399,7 +399,7 @@ export
 joinPath : List String -> String
 joinPath xs = foldl (</>) "" xs
 
-||| Split path into components.
+||| Splits path into components.
 |||
 ||| ```idris example
 ||| splitPath "/usr/local/etc" == ["/", "usr", "local", "etc"]
@@ -409,7 +409,7 @@ export
 splitPath : String -> List String
 splitPath = map show . splitPath' . parse
 
-||| Split the path into parent and child.
+||| Splits the path into parent and child.
 |||
 ||| ```idris example
 ||| splitParent "/usr/local/etc" == Just ("/usr/local", "etc")
@@ -421,14 +421,14 @@ splitParent path =
     (parent, child) <- splitParent' (parse path)
     pure (show parent, show child)
 
-||| Return the path without its final component, if there is one.
+||| Returns the path without its final component, if there is one.
 |||
 ||| Returns Nothing if the path terminates by a root or volume.
 export
 parent : String -> Maybe String
 parent = map show . parent' . parse 
 
-||| Return the list of all parents of the path, longest first, self included.
+||| Returns the list of all parents of the path, longest first, self included.
 |||
 ||| ```idris example
 ||| parents "/etc/kernel" == ["/etc/kernel", "/etc", "/"]
@@ -437,7 +437,7 @@ export
 parents : String -> List String
 parents = map show . List.iterate parent' . parse
 
-||| Determine whether the base is one of the parents of target.
+||| Determines whether the base is one of the parents of target.
 |||
 ||| Trailing separator is ignored.
 |||
@@ -455,7 +455,7 @@ isBaseOf base target =
     && baseRoot == targetRoot
     && (baseBody `isPrefixOf` targetBody)
 
-||| Return a path that, when appended to base, yields target.
+||| Returns a path that, when appended to base, yields target.
 |||
 ||| Returns Nothing if the base is not a prefix of the target.
 export
@@ -483,7 +483,7 @@ export
 fileName : String -> Maybe String
 fileName  = fileName' . parse
 
-||| Extract the file name in the path without extension.
+||| Extracts the file name in the path without extension.
 |||
 ||| The stem is:
 |||
@@ -495,7 +495,7 @@ export
 fileStem : String -> Maybe String
 fileStem path = pure $ fst $ splitFileName !(fileName path)
 
-||| Extract the extension of the file name in the path.
+||| Extracts the extension of the file name in the path.
 |||
 ||| The extension is:
 |||
@@ -507,7 +507,7 @@ export
 extension : String -> Maybe String
 extension path = pure $ snd $ splitFileName !(fileName path)
 
-||| Update the file name in the path.
+||| Updates the file name in the path.
 |||
 ||| If there is no file name, it appends the name to the path;
 ||| otherwise, it appends the name to the parent of the path.
@@ -515,7 +515,7 @@ export
 setFileName : (name : String) -> String -> String
 setFileName name path = show $ setFileName' name (parse path)
 
-||| Append a extension to the path.
+||| Appends a extension to the path.
 |||
 ||| If there is no file name, the path will not change;
 ||| if the path has no extension, the extension will be appended;
@@ -539,7 +539,7 @@ export
           show $ setFileName' (stem ++ ext) path'
       Nothing => path
 
-||| Drop the extension of the path.
+||| Drops the extension of the path.
 export
 dropExtension : String -> String
 dropExtension path = path <.> ""
--- a/libs/contrib/System/Path.idr
+++ b/libs/contrib/System/Path.idr
@@ -29,7 +29,8 @@ pathSeparator = if isWindows then ';' else ':'
 
 ||| Windows path prefix.
 public export
-data Volume = 
+data Volume
+  =
   ||| Windows Uniform Naming Convention, consisting of server name and share
   ||| directory.
   |||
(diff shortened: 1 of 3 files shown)


andylokandy:
@gallais PTAL XD

andrevidela:
ping @andylokandy, is this still needed?

andylokandy:
yes, i think. The new function is needed in other pr.

libs/contrib/System/Path.idr
```diff
@@ -147,203 +151,244 @@ pathTokenMap = toTokenMap $
...
+    splitRoot : Path -> (Maybe Path, Path)
+    splitRoot path@(MkPath Nothing False _ _) = (Nothing, path)
+    splitRoot (MkPath vol root xs trailSep) =
+      (Just $ MkPath vol root [] False, MkPath Nothing False xs trailSep)
+
+    iterateBody : List Body -> (trailSep : Bool) -> List Path
+    iterateBody [] _ = []
+    iterateBody [x] trailSep = [MkPath Nothing False [x] trailSep]
+    iterateBody (x::y::xs) trailSep =
+      (MkPath Nothing False [x] False) :: iterateBody (y::xs) trailSep
```
andrevidela, line 313: isn't this `map (\x => MkPath Nothing False [x] False trailSep)` ?

libs/contrib/System/Path.idr
```diff
@@ -352,115 +397,147 @@ export
...
+splitParent path =
+  do
+    (parent, child) <- splitParent' (parse path)
+    pure (show parent, show child)
 
-||| Returns the path without its final component, if there is one.
+||| Return the path without its final component, if there is one.
 |||
-||| Returns Nothing if the path terminates in a root or volume.
+||| Returns Nothing if the path terminates by a root or volume.
```
andrevidela, line 424:
```suggestion
||| Return Nothing if the path terminates by a root or volume.
```

Its using the imperative tense just above :)

libs/contrib/System/Path.idr
```diff
@@ -352,115 +397,147 @@ export
...
 export
 parent : String -> Maybe String
-parent p = map show $ parent' (parse p)
+parent path = map show $ parent' (parse path)
```
andrevidela, line 427: Avoid the renaming altogether?

```suggestion
parent = map show . parent' . parse 
```

libs/contrib/System/Path.idr
```diff
@@ -352,115 +397,147 @@ export
...
-||| self included.
+||| Return the list of all parents of the path, longest first, self included.
 |||
 ||| ```idris example
 ||| parents "/etc/kernel" == ["/etc/kernel", "/etc", "/"]
 ||| ```
 export
 parents : String -> List String
-parents p = map show $ iterate parent' (parse p)
+parents path = map show $ List.iterate parent' (parse path)
```
andrevidela, line 436: Same here

```suggestion
parents = map show . List.iterate parent' . parse
```

libs/contrib/System/Path.idr
```diff
@@ -352,115 +397,147 @@ export
...
+isBaseOf base target =
+  let 
+    MkPath baseVol baseRoot baseBody _ = parse base
+    MkPath targetVol targetRoot targetBody _ = parse target
+  in
+       baseVol == targetVol
+    && baseRoot == targetRoot
+    && (baseBody `isPrefixOf` targetBody)
+
+||| Return a path that, when appends to base, yields target.
```
andrevidela, line 456: "when appends" ? did you mean "when appended"? 
```suggestion
||| Return a path that, when appended to base, yields target.
```

libs/contrib/System/Path.idr
```diff
@@ -352,115 +397,147 @@ export
...
+||| Returns Nothing if the base is not a prefix of the target.
 export
-stripPrefix : (base : String) -> (full : String) -> Maybe String
-stripPrefix base full
-    = do let MkPath vol1 root1 body1 _ = parse base
-         let MkPath vol2 root2 body2 trialSep = parse full
-         if vol1 == vol2 && root1 == root2 then Just () else Nothing
-         body <- stripBody body1 body2
-         pure $ show $ MkPath Nothing False body trialSep
+dropBase : (base : String) -> (target : String) -> Maybe String
```
andrevidela, line 460: Since this is an export, is the breaking renaming necessary? What is wrong with the previous name?

libs/contrib/System/Path.idr
```diff
@@ -352,115 +397,147 @@ export
...
+    dropBody : (base : List Body) -> (target : List Body) -> Maybe (List Body)
+    dropBody [] ys = Just ys
+    dropBody xs [] = Nothing
+    dropBody (x::xs) (y::ys) = if x == y then dropBody xs ys else Nothing
 
-||| Returns the final body of the path, if there is one.
-|||
-||| If the path is a normal file, this is the file name. If it's the
-||| path of a directory, this is the directory name.
+||| Returns the last body of the path.
```
andrevidela, line 474:
```suggestion
||| Return the last body of the path.
```

imperative tense, this this change has been carried through the entire file :)

libs/contrib/System/Path.idr
```diff
@@ -352,115 +397,147 @@ export
...
 |||
-||| Returns Nothing if the final body is "..".
+||| If the last body is a file, this is the file name; 
+||| if it's a directory, this is the directory name; 
+||| if it's ".", it recurses on the previous body;
+||| if it's "..", returns Nothing.
 export
 fileName : String -> Maybe String
-fileName p = fileName' (parse p)
+fileName path = fileName' (parse path)
```
andrevidela, line 482: Again, the name can be dropped altogether.

```suggestion
fileName  = fileName' . parse
```

src/Utils/Path.idr
```diff
@@ -153,196 +159,234 @@ lexPath str = let (tokens, _, _, _) = lex pathTokenMap str in tokens
...
-||| - On Windows, a path is absolute if it has a volume and starts
-|||   with the root. e.g., `c:\\windows` is absolute, while `c:temp`
-|||   and `\temp` are not. In addition, a path with UNC volume is absolute.
+||| - On Windows, a path is absolute if it starts with a disk and has root or
+|||   starts with UNC. For example, `c:\\windows` is absolute, while `c:temp`
+|||   and `\temp` are not.
 export
 isAbsolute : String -> Bool
-isAbsolute p = isAbsolute' (parse p)
+isAbsolute path = isAbsolute' (parse path)
```
andrevidela, line 365: You can remove the renaming with point-free 

```suggestion
isAbsolute = isAbsolute' . parse

```

andrevidela (changes requested):
Alright I went through it, it looks like a big diff but most of the changes are syntactic or whitespace. Did you use a linter for those?

Before we merge I'd also like to know the motivation for the breaking changes. Not like we can't have breaking changes, but I couldn't guess by looking at the code why they were necessary.

Good work otherwise :)

src/Utils/Path.idr
```diff
@@ -153,196 +159,234 @@ lexPath str = let (tokens, _, _, _) = lex pathTokenMap str in tokens
...
-                                       hasTrailSep = r.hasTrailSep } l
+append' left right =
+  if isAbsolute' right || isJust right.volume then
+    right
+  else if hasRoot right then
+    record { volume = left.volume } right
+  else 
+    record { body = left.body ++ right.body, hasTrailSep = right.hasTrailSep } left
+
+splitPath' : Path -> List Path
```
andrevidela, line 296: Since the only use of this function is in `splitPath`, why not return `List String` instead of going through `map show` to get the components? Semantically it feels very odd to return a list of paths which are all singleton paths

libs/contrib/System/Path.idr
andylokandy: In the case of `iterateBody (x::y::xs) trailSep`, the trailSep should be false. Do we have another way to express that?

libs/contrib/System/Path.idr
andylokandy: This is adopted from Rust documents. https://doc.rust-lang.org/std/path/struct.Path.html#method.is_relative

I think it is an abbr for "*It* returns ..."

libs/contrib/System/Path.idr
andylokandy: Sure!

libs/contrib/System/Path.idr
andylokandy: I was trying to introduce `isPrefixOf` for Path, just like List. But later I realized that we already have `isPrefixOf` for `String`, which will cause ambiguity, so I renamed prefix to base. And I personally think `base` is more appropriate for paths.

libs/contrib/System/Path.idr
andylokandy: ditto

libs/contrib/System/Path.idr
andrevidela: Yes, but in the rest of this document, you replace present tense by imperative tense by removing the `s` on lots of verbs. The change is fine, but it should be consistent, line 422 does this change by replacing `Returns` by `Return`, so the change should be carried over to this line too.

libs/contrib/System/Path.idr
andrevidela: Ah, I noticed I commended on the wrong file (since Utils.Path is duplicated). my confusion stemmed from this comment https://github.com/idris-lang/Idris2/pull/823/files/9f72a514f1ac523136423caceba725f975a37df5#diff-5621839bbc877aa5fbbbc07903cf1cdf7a3036ecf85836d96de83f873ca1d5a6R296

src/Utils/Path.idr
andylokandy: It's originally for avoiding to repeat the login in `Show`, i.e. showing the root and give trailSep to the last body.

andylokandy:
Thanks for looking into my PR!

> Alright I went through it, it looks like a big diff but most of the changes are syntactic or whitespace. Did you use a linter for those?

No, I tweaked the spaces manually.

> Before we merge I'd also like to know the motivation for the breaking changes. Not like we can't have breaking changes, but I couldn't guess by looking at the code why they were necessary.

Since the only breaking change is the `stripePrefix`, I believe that it's not widely used in other projects. And the change is to distinguish with `isPrefixOf`.

src/Utils/Path.idr
andrevidela: Alright! I've taken a look and I couldn't come up with something significantly better. If something's wrong with it we can always fix it in a subsequent pull request

libs/contrib/System/Path.idr
andrevidela: Sounds Good!

andrevidela (approved):

PR #825, opened 2020-12-07
Copy Vect elem + documentation into List library

AdamHarries:
This PR adds the `elem` function to the `List` library, implemented in the same was as in the `Vect` library. Additionally, it updates the existing documentation for the `elemBy` function to match that in the `Vect` library.


AdamHarries:
My mistake, it looks like this is already in the Prelude - closing!

gallais:
`elemBy` could be generalised to `elemBy : (a -> b -> Bool) -> a -> List b -> Bool` though.

PR #828, opened 2020-12-07
Increase timings of concurrency tests due to flaky windows runs

fabianhjr:


Alvaro-Santos:
Originally I was just going to submit a pull request, but since this one is so recent it might be better to discuss it first: even after this change I'm still getting flaky, non-deterministic behaviour on the chez/concurrency001 test when trying to compile on windows (with mingw+msys). The only reason the test fails is due to "Hello World" and "Hello Idris" sometimes appearing swapped; doubling the timers (to 60000 for "World" and 40000 for "Idris") on those 2 strings makes my tests pass every time.

Would the maintainers be ok with just increasing the timers even further?

fabianhjr:
I don't mind and I apologize for the windows issues, i tested it by stressing my cpu with another program and running these suit. I have no idea why windows specifically is having flakyness. UnU

My motherboard is currently bricked so I can only interact via my cellphone. Feel free to open a new pr to increase the timings.

fabianhjr:
Btw, i would rather you increase by an order of magnitude rather than doubling it.

PR #830, opened 2020-12-08
[ contrib ] Arithmetic on Fin

buzden:
There are three more or less separate changes (that correspond to separate commits):

- improvement in `Data.Fin.Extra` module's sectioning -- or else it becomes a mess;
- `Data.Fin.Extra` contains a function proving properties of relation between `finToNat` and `weaken`, I propose similar functions for `finToNat` regarding to `weakenN` and `shift`;
- addition of functions that realise somewhat a linearity property of `Fin`:
  - going from `Fin` of sum to (type) sum of `Fin`s: `Fin (a+b) -> Either (Fin a) (Fin b)`;
  - going from `Fin` of product to (type) product of `Fin`s: `Fin (a*b) -> (Fin a, Fin b)`.


libs/contrib/Data/Fin/Extra.idr
```diff
@@ -97,3 +123,49 @@ divMod {ok=_} (S n) (S d) =
...
+
+-------------------------------------------------
+--- Splitting operations and their properties ---
+-------------------------------------------------
+
+export
+splitSumFin : {a : Nat} -> Fin (a + b) -> Either (Fin a) (Fin b)
+splitSumFin {a=Z}   x      = Right x
+splitSumFin {a=S k} FZ     = Left FZ
+splitSumFin {a=S k} (FS x) = bimap FS id $ splitSumFin x
```
gallais, line 135: `bimap f id` is `mapFst f`

libs/contrib/Data/Fin/Extra.idr
```diff
@@ -97,3 +123,49 @@ divMod {ok=_} (S n) (S d) =
...
+
+export
+splitProdFin : {a, b : Nat} -> Fin (a * b) -> (Fin a, Fin b)
+splitProdFin {a=S _} x = case splitSumFin x of
+  Left  y => (FZ, y)
+  Right y => bimap FS id $ splitProdFin y
```
gallais, line 141: Same here.

libs/contrib/Data/Fin/Extra.idr
```diff
@@ -97,3 +123,49 @@ divMod {ok=_} (S n) (S d) =
...
+splitSumFin_correctness {a=S k} FZ = Refl
+splitSumFin_correctness {a=S k} (FS x) with (splitSumFin_correctness x)
+  splitSumFin_correctness {a=S k} (FS x) | subcorr with (splitSumFin x)
+    splitSumFin_correctness {a=S k} (FS x) | subcorr | Left  y = rewrite subcorr in Refl
+    splitSumFin_correctness {a=S k} (FS x) | subcorr | Right y = rewrite subcorr in Refl
+
+export
+splitProdFin_correctness : {a, b : Nat} -> (x : Fin $ a * b) ->
+                           let (o, i) = splitProdFin {a} {b} x in
+                           finToNat x = finToNat o * b + finToNat i
```
gallais, line 160: Can we not define the other direction `Fin a -> Fin b -> Fin (a * b)`
rather than going via `finToNat` for the correctness lemma?

libs/contrib/Data/Fin/Extra.idr
buzden: I believed that this function exists, but I failed to find it, shame on me

libs/contrib/Data/Fin/Extra.idr
buzden: I'm sure yes, let me poke a bit

buzden: Well, after implementing an appropriate (as I thought) function `Fin a -> Fin b -> Fin (a * b)` (which I dared to add to this PR), I realized that `splitProdFin` is actually a `divMod` function for division of `Fin (a * b)` by `(b : Nat)`.

So, there exists a function that does reverse, i.e. multiplying `Fin a` by `(b : Nat)` adding `Fin b` as a remainder, resulting to `Fin (a * b)`, but its correctness shall be proven also through `finToNat` because it acts on the border of `Fin`s and `Nat`s.

buzden: Well, I understand now, that we actually can try to reformulate correctness property not through `finToNat`, but using some `Nat` to `Fin` operations, like `last`. That is, instead of `finToNat x = finToNat o * b + finToNat i` we could write something like `x = i + o * last` having appropriate operations `+` and `*` upon `Fin`s (like those that I added recently). (I'm not sure there're no off-by-one errors in the example)

Do you think that this kind of proof has more value than `finToNat`-based?

buzden: Well, that was tough, but possible: https://github.com/idris-lang/Idris2/pull/830/commits/6ddbb6da3d1d1f6a23a4a7378b3086f4482b96c4. Is this approximately what you were originally talking about?

gallais:
I think quite a lot of the complexity you are facing may come from the fact your
functions are not as general as possible. I expect this ought to work a bit better?

```idris
public export
(+) : Fin a -> Fin b -> Fin (a + b)
(+) FZ y = rewrite plusCommutative a b in weakenN a y
(+) (FS x) y = FS (x + y)

public export
(*) : Fin a -> Fin b -> Fin (a * b)
(*) FZ     FZ     = FZ
(*) FZ     (FS _) = FZ
(*) (FS x) y      = y + x * y
```

gallais:
Pretty sure these are the more convenient isos:

```idris
public export
pad : (n : Nat) -> Fin m -> Fin (n + m)
pad Z     k = k
pad (S n) k = FS (pad n k)

public export
sum : {a : Nat} -> Either (Fin a) (Fin b) -> Fin (a + b)
sum (Left x)  = weakenN b x
sum (Right x) = pad a x

public export
mus : {a : Nat} -> Fin (a + b) -> Either (Fin a) (Fin b)
mus {a = Z}   k      = Right k
mus {a = S n} FZ     = Left FZ
mus {a = S n} (FS l) = mapFst FS (mus l)

public export
prod : {b : Nat} -> Fin a -> Fin b -> Fin (a * b)
prod FZ     y = weakenN (mult (pred a) b) y
prod (FS x) y = pad b (prod x y)

public export
dorp : Fin (a * b) -> (Fin a, Fin b)
dorp = ? {- ...divmod... -}
```

buzden:
> Pretty sure these are the more convenient isos:

I agree with you and I rewrote correctness proofs of split functions as an inverse of appropriate functions.

However, I couldn't manage to call them as you suggest because there could be two different pairs of notions for sum and product for `Fin`s:
  - the sum and product of `Fin`s as bounded `Nat`s
  - sum and product of `Fin`s as indices.

These sums and products have different properties.

What I was originally talked about is the index-sum and index-product (this is what you name `sum` and `prod` in your example). If it has the signature of `f : {...} -> Either (Fin a) (Fin b) -> Fin (a + b)` then it has nice property of `f (Right last)` is `last`. The same for product. If index-product has signature of `g : {...} -> Fin a -> Fin b -> Fin (a * b)`, then `g last last` is `last`. It's great.

When I was at first trying to express properties of my split functions through sum and product of `Fin`s-as-bounded-`Nat`s (I'll call then number-sum and number-product), I realised that signatures as above are not precise enough. E.g., if the signature of number-sum and number-product would be
```idris
(+) : {...} -> Fin a -> Fin b -> Fin (a + b)
(*) : {...} -> Fin a -> Fin b -> Fin (a * b)
```

it would lead to the following: `the (Fin 4) 3 + the (Fin 3) 2` would be `5` of type `Fin 7` and `the (Fin 4) 3 * the (Fin 3) 2` would be `6` of type `Fin 12`. I.e. `last + last` wouldn't be `last` and the same for `*`.

Thus, I refined the signatures of number-sum and number-product to

```idris
(+) : {...} -> Fin (S a) -> Fin (S b) -> Fin (S $ a + b)
(*) : {...} -> Fin (S a) -> Fin (S b) -> Fin (S $ a * b)
```

With these signatures, examples from above would be nicer: `the (Fin 4) 3 + the (Fin 3) 2` is `5` of type `Fin 6` and `the (Fin 4) 3 * the (Fin 3) 2` would be `6` of type `Fin 7`. I.e. signatures are precise, sum and product of `last`s are `last`s. But this leads to more complex signatures since they requires `S ...` in them.

---

Well, the question is the following. Actually, the original split functions and their properties now are expressed without number-sum and number-product. But the functions the these operations, and some properties of them are still in this PR. I feel myself that they can be useful sometime, but I'm not sure. So, should they leave or should I remove them?

buzden:
Rebased over changes of #857.

buzden:
I rebased again and added documentation comments to main functions.

gallais:
This was fairly painful but I think the proof have cleaner types and are more maintainable now.

PR #831, opened 2020-12-08
Remove Lexer.Core and Parser.Core packages.

msmorgan:
The contents of these packages has been moved into their parent.
These were originally split off in Idris 1 due to long compile
times. Idris 2 processes these modules much faster, so there is
no need to split it up.


gallais:
This has diverged so much from the main branch that I think I can close this.
I am not sure the original idea applies anymore either now that we have `Tokenizer` too.

msmorgan:
Ok, whatever.

When I originally split them up it was because the Core packages took multiple minutes to compile so it was impossible to iterate on Lexer and Parser themselves. I checked recently and they compile much faster now, so the packages don't need to be separate anymore.

gallais:
> I didn't know I was expected to keep this up to date with master
  despite receiving no comments or interest whatsoever.

I don't think you were, the PR just fell by the wayside.

I did not act on it initially because I was not convinced that packing
everything into a single module made the library easier to explore. And
now that we have additional submodules because of the nice compositional
lexer that Andy added recently, it's even less clear that merging these
together makes sense.

Rather than leaving this PR open indefinitely with no plan to bring it
up to speed or merge it, I closed it.

PR #832, opened 2020-12-09
[ new ] null method in Foldable

alissa42:
Adding functions `null` for Foldable structures and `isNullString` to test whether the structure / string is empty, respectively.

pr1
pr2

### DIFF
--- a/libs/prelude/Prelude/Interfaces.idr
+++ b/libs/prelude/Prelude/Interfaces.idr
@@ -217,6 +217,9 @@ interface Foldable (t : Type -> Type) where
   foldl : (func : acc -> elem -> acc) -> (init : acc) -> (input : t elem) -> acc
   foldl f z t = foldr (flip (.) . flip f) id t z
 
+  ||| Test whether the structure is empty.
+  null : t elem -> Bool
+
 ||| Similar to `foldl`, but uses a function wrapping its result in a `Monad`.
 ||| Consequently, the final value is wrapped in the same `Monad`.
 public export
--- a/libs/prelude/Prelude/Types.idr
+++ b/libs/prelude/Prelude/Types.idr
@@ -180,6 +180,8 @@ public export
 Foldable Maybe where
   foldr _ z Nothing  = z
   foldr f z (Just x) = f x z
+  null Nothing = True
+  null (Just _) = False
 
 public export
 Traversable Maybe where
@@ -272,6 +274,8 @@ public export
 Foldable (Either e) where
   foldr f acc (Left _) = acc
   foldr f acc (Right x) = f x acc
+  null (Left _) = True
+  null (Right _) = False
 
 public export
 Traversable (Either e) where
@@ -340,6 +344,9 @@ Foldable List where
 
   foldl f q [] = q
   foldl f q (x::xs) = foldl f (f q x) xs
+
+  null [] = True
+  null (_::_) = False
 
 public export
 Applicative List where
--- a/libs/base/Control/Monad/Either.idr
+++ b/libs/base/Control/Monad/Either.idr
@@ -112,6 +112,8 @@ Foldable m => Foldable (EitherT e m) where
   foldr f acc (MkEitherT e)
     = foldr (\x,xs => either (const acc) (`f` xs) x) acc e
 
+  null (MkEitherT e) = null e
+
 public export
 Traversable m => Traversable (EitherT e m) where
   traverse f (MkEitherT x)
--- a/libs/base/Data/List1.idr
+++ b/libs/base/Data/List1.idr
@@ -117,6 +117,7 @@ Monad List1 where
 export
 Foldable List1 where
   foldr c n (x ::: xs) = c x (foldr c n xs)
+  null _ = False
 
 export
 Show a => Show (List1 a) where
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -407,6 +407,9 @@ public export
 implementation Foldable (Vect n) where
   foldr f e xs = foldrImpl f e id xs
 
+  null [] = True
+  null _ = False
+
 --------------------------------------------------------------------------------
 -- Special folds
 --------------------------------------------------------------------------------
--- a/libs/contrib/Data/SortedMap.idr
+++ b/libs/contrib/Data/SortedMap.idr
@@ -254,11 +254,6 @@ export
 values : SortedMap k v -> List v
 values = map snd . toList
 
-export
-null : SortedMap k v -> Bool
-null Empty = True
-null (M _ _) = False
-
 treeMap : (a -> b) -> Tree n k a o -> Tree n k b o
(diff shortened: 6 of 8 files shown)


libs/prelude/Prelude/Interfaces.idr
```diff
@@ -234,6 +234,11 @@ public export
 concatMap : (Foldable t, Monoid m) => (a -> m) -> t a -> m
 concatMap f = foldr ((<+>) . f) neutral
 
+|||Test whether the structure is empty.
+public export
+null : Foldable t => t a -> Bool
+null = foldr (\_,_ => False) True
```
gallais, line 240: Because Idris is not lazy, it's probably a bad idea to write it this way:
`foldr c n (x :: xs) = c n (foldr c n xs)` means that the recursive call
will be fully evaluated before `c` is called.

This means that `null` is linear in the size of the `t a` input.
It's probably better to add it to the `Foldable` class and let users
implement it efficiently for each data type.

gallais (changes requested):

libs/prelude/Prelude/Interfaces.idr
alissa42: I am sorry for not being considerate and thanks for your patience.

gallais: I did not say that the PR would not get accepted. I just said `null` should
be moved to the definition of `interface Foldable` and then each implementation
should provide its own efficient instance of `null`.

This is a welcomed addition!

alissa42: Wow! Thank You. I had done some fix about them …… testing XD

alissa42: 图片
图片
The following code can be found at this [site](http://fars.ee/h8Qo).
Sorry to bother. What I had just done is to
- Add Foldable implementations for `List`, `Maybe` and `Either`, respectively.
- Add `null` in the `Foldable` interface.
- Add a default definition for `null` for compatibility and a flag `%allow_overloads null`.
I am wondering how to write the code correspondence to `Foldable Either`. As presentations in the pics, I got confusing.
Feeling sorry...
:(

alissa42: Maybe the Evaluation strategy?
Since the type of
`Left 42` : Either Integer ?b
`Right 42` : Either ?a Integer

ice1000: Note that it's a convention to see `Either a` (given `a` arbitrary type variable) as a functor, but we can also see that as a consequence of GHC's lack of type lambda support. If we only see `Either` as a bifunctor this won't be a problem but there is another convention, that is to use `Either` as the error monad, and that requires the `Either a` functor (as well as the `Either a` foldable).

libs/base/Data/Strings.idr
```diff
@@ -199,6 +199,10 @@ export partial
 strTail : String -> String
 strTail = prim__strTail
 
+export
+total isNullStr : String -> Bool
+isNullStr = (== "")
```
ice1000, line 204: What differs this from `null`?

alissa42: For String isn't a member of Foldable interface. I was attempt to overload it to null letter if possible.

libs/base/Data/Either.idr
```diff
@@ -31,6 +31,52 @@ isRight : Either a b -> Bool
 isRight (Left _)  = False
 isRight (Right _) = True
 
+--------------------------------------------------------------------------------
+-- `Either` used to a instance of interface `Foldable`.
```
gallais, line 35: You should implement `Foldable (Either e)`, not copy/paste the list of
definitions. Cf. the existing `Foldable` implementations in the prelude.

gallais (changes requested):

libs/base/Data/Either.idr
gallais: `Foldable (Either e)` is already defined in `Prelude.Types`.
There is no need to add all of these definitions again.
There is no need to turn on all of these `%allow_overloads`.

gallais, line 35: There is already an implementation of `Foldable (Either e)` in `Prelude.Types`.
There is no need to add all of these definitions again.
There is no need to add all the `%allow_overloads` you have inserted.

gallais (changes requested):

libs/prelude/Prelude/Types.idr
```diff
@@ -268,16 +271,6 @@ Monad (Either e) where
   (Left n) >>= _ = Left n
   (Right r) >>= f = f r
 
-public export
-Foldable (Either e) where
```
gallais, line 272: Don't remove that!

libs/base/Data/Either.idr
alissa42: As in [1](https://github.com/idris-lang/Idris2/pull/832#discussion_r539712524), I am wondering how to define the implementation of `null` for `(Either e)` and still don't know yet

libs/prelude/Prelude/Types.idr
alissa42: Sorry, got it

libs/base/Data/Either.idr
alissa42: How does people actually use `null (x : Either a b)` as `isLeft`? Intuitively it can not be null for the name either :thinking:

gallais: `null` for `t a` refers to whether or not `t a` contains an `a`.
With `Either e a`, `Left e` contains an `e` but not an `a`.

alissa42: Got that, thx u

gallais (commented):
I have added some missing instances, let's see if it builds now.

alissa42:
Although the implementation of `null` for `Either` can successfully be type checked and build, when using `null (Right _)` directly it must cause errors for not have given type information directly.
Although explicitly give its type by `the` works, it is a error at not dual type-check nor compile-time, but the runtime which I think can be dangerous. I haven't see such issues like this until yesterday (UTC +8), and have no idea to deal with it.
I feel sorry for the wasting CI and your patience.
I am still wondering what's the method to solve problems like this and how to avoid or solve it during my daily programming.
Sincerely apologize.

gallais:
Calling `null (Right _)` is fairly uncommon though and completely useless:
you already statically know that the expression is not null.

In practice you have an expression `e` that you know has type `Either a b`
and you run `null` on it. No inference issue there. Or, even more likely,
you are writing a program with a constraint `Foldable t => ...` where `t`
is just a variable and calling `null` on a `t a` value. Again no inference
issue.

alissa42:
Oh... that indeed can only live in REPL. Thanks for interpret  it for me. Anyway I won't use REPL to eval a expression to test anymore 0.0;;
Did we need a re-request review?
A S7ZOHP G(_S O@%%ZE8YL

alissa42 (commented):
I had learnt a lot, thanks.

PR #838, opened 2020-12-11
Double closing brace was made to not be treated specially

buzden:
This was done to enable occasional usage of `}}` in non-quote code to not to require a space between two braces.

For example, consider this code:

```idris
f : {a : Nat} -> Nat
f {a} = a + 1

g : {b : Nat} -> Nat
g {b} = b + 2

h : Nat -> Nat
h c = g {b = f {a=c}}
```

This PR makes this `g {b = f {a=c}}` to be parsed as expected without requiring a space between closing braces and without clashing with syntax for quotes
```
`{{...}}
```


buzden:
The bad thing, however, is that this becomes parsable:

```idris
import Language.Reflection

n : Name
n = `{{f} }
```

However, it's hard to imagine where '} }' can really be needed inside the name quote, so this shouldn't lead to some errors.

Russoul:
I think almost everyone has been caught by this one unawares, so I’d love to see it accepted, in spite of the syntax of name quoting being a bit compromised by the change.

buzden:
By the way, is there any reason to have name quotes to be just
```
`{name}
```
but not
```
`{{name}}
```
?

buzden:
Rebased due to changed in the test system.

gallais:
This was much smaller than I expected based on the comments on changing the way quotes
work. I should have had a look at the patch earlier. Thanks for the fix!

PR #842, opened 2020-12-12
Overhaul of concurrency primitives

wenkokke:
This pull requests contains an overhaul of the concurrency primitives for Idris2.
My main concerns with the previous primitives was:

- there was no way to wait for a thread to finish;
- there were only mutexes and condition variables, and these were only supported by the Chez Scheme RTS;
- the provisional support in the Racket RTS was deeply flawed.

I've made the following breaking changes:

  - Renamed `System.Concurrency.Raw` to `System.Concurrency`.

  - Modified the implementation of `Prelude.IO.fork` in the Chez Scheme RTS, which
    now returns a semaphore instead of a thread object. This allows the main
    thread to wait for the child thread to finish (see next bullet). The Racket
    implementation already returned a thread descriptor, which could be used to
    wait for the thread to finish.

  - Added `Prelude.IO.threadWait` which waits for a thread, identified by a
    `ThreadID`, to finish. This operation is supported by both the Chez Scheme and
    the Racket RTS'es.

  - Added semaphores to `System.Concurrency`, supported by both the Chez Scheme
    and Racket RTS'es.

  - Added barriers to `System.Concurrency`, supported by both the Chez Scheme
    and Racket RTS'es.

  - Added synchronous channels to `System.Concurrency`, supported by both the Chez
    Scheme and Racket RTS'es.

  - Fixed the support for mutexes in the Racket RTS. Formerly, they were
    implemented with semaphores, and calling`mutexRelease` multiple times would
    increment the internal counter multiple times, allowing multiple concurrent
    `mutexAcquire` operations to succeed simultaneously. Currently, `mutexRelease`
    fails when called on a mutex which isn't owned. (However, `mutexRelease` does
    not check whether the mutex is in fact owned by the current thread, which may
    be a bug.)

  - Modified the support for condition variables in the Racket RTS. Formerly,
    they were implemented using synchronous channels, meaning that: 
      + `conditionSignal` was a blocking operation; and 
      + calling `conditionSignal` on a condition variable on which no thread
        was waiting would wake the next thread to call `conditionWait`, whereas
        condition variables are supposed to be stateless, and only wake threads
        already in the queue. 
    The implementation was replaced with an implementation based on asynchronous
    channels and mutexes, based on the following paper:
    https://www.microsoft.com/en-us/research/wp-content/uploads/2004/12/ImplementingCVs.pdf

  - Removed `threadID` and `blodwen-thisthread`. Formerly, in the Chez Scheme
    backend, this function returned "the thread id of the current thread" as a
    value of type `ThreadID`. However, `fork` returned a "thread object" as a
    value of type `ThreadID`. These are *different kinds of values* in Chez
    Scheme. As there was nothing one could do with a value of type `ThreadID`, I
    chose to remove `threadID`, as it allowed me to implement `threadWait` more
    easily.

  - Renamed `blodwen-lock` and `blodwen-unlock` to `blodwen-mutex-acquire` and
    `blodwen-mutex-release` for consistency, as these functions are referred to
    with acquire and release both in Chez Scheme and in the Idris2 concurrency
    module.

I've also moved the tests for Futures from `concurrency001` to `futures001` to avoid confusing with the *other* concurrency tests I placed under `semaphoresXXX`, `channelsXXX` and `forkjoinXXX`.


gallais:
Note that you don't need the `input` files if you are not using them.

src/Compiler/Scheme/Racket.idr
```diff
@@ -42,6 +42,7 @@ schHeader : String -> String
 schHeader libs
   = "#lang racket/base\n" ++
     "; @generated\n" ++
+    "(require racket/async-channel)\n" ++ -- for asynchronous channels
```
fabianhjr, line 45: just a heads up, adding requires will need to wait until next bootstrap release to work. :C

(That is why I didn't use future in the racket support code right away)

andrevidela:
If that's ok with everyone I'm going to close this since it's gone stale. This is jsut for tiding up the PR tracker, feel free to reopen if you want to work on it again! :)

wenkokke:
@andrevidela Yes, this was ready for review.

PR #844, opened 2020-12-13
Added assert_linear.

wenkokke:
I've added `assert_linear` to the builtins module, which asserts that a function uses its argument linearly:

```idris
assert_linear : (1 f : a -> b) -> (1 val : a) -> b
```


gallais:
Thanks!

andrevidela:
🙏

PR #847, opened 2020-12-14
Allow `HasIO` to not depend on `Monad`.

jfdm:
The definition of `HasIO` depends on `Monad` to provide sequencing of `HasIO` like things.
A strict dependency on `Monad` makes it difficult to provide correct implementatons of `HasIO` for indexed computation contexts in which there are type-level operations.
Idris2's standard interface definitions are, unfortunatly, too restrictive and will pin type-level values unnecessarily as such making the wrong assumptions when defining Applicative, Monad, or even Show implementations.
Moreover, the type required by the interface is not the last type-level parameter.

Here we remove the explicit dependency, and provide an alias (`HasMonadIO`) to regain the dependency and thus allow one to regain 'Do-notation' by overloading `(>>=)` and `pure`.

### DIFF
--- a/libs/base/Data/Buffer.idr
+++ b/libs/base/Data/Buffer.idr
@@ -20,7 +20,7 @@ data Buffer : Type where [external]
 prim__bufferSize : Buffer -> Int
 
 export
-rawSize : HasIO io => Buffer -> io Int
+rawSize : HasMonadIO io => Buffer -> io Int
 rawSize buf = pure (prim__bufferSize buf)
 
 %foreign "scheme:blodwen-new-buffer"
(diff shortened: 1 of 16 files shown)


edwinb:
We did have this split initially, but decided not to do it in the end. The relevant patch is here:
https://github.com/idris-lang/Idris2/commit/dbdf7dab3dd56703428d2c540ea720f6c26f1616

Of course, that patch did involve a lot of "ifs" so if you have a compelling use case, especially one that can't be done another way, we could revisit it.

jfdm:
> We did have this split initially, but decided not to do it in the end. The relevant patch is here:
> [dbdf7da](https://github.com/idris-lang/Idris2/commit/dbdf7dab3dd56703428d2c540ea720f6c26f1616)

Thanks for linking to the 'patch'. It makes sense after a quick read. I will cancel this PR to brew on the matter some more.
 
> Of course, that patch did involve a lot of "ifs" so if you have a compelling use case, especially one that can't be done another way, we could revisit it.

Currently brewing...

PR #848, opened 2020-12-15
Update stringOps.c

vfrinken:
Bugfix in tail, instead of the last char a string with everything expect the head.


support/refc/stringOps.c
```diff
@@ -14,14 +14,16 @@ Value *head(Value *str)
...
-    Value_String *s = (Value_String *)str;
+    Value_String *tailStr = (Value_String *)newValue();
+    tailStr->header.tag = STRING_TAG;
+    Value_String *s = (Value_String *)input;
     int l = strlen(s->str);
-    c->c = s->str[l - 1];
-    return (Value *)c;
+    tailStr->str = malloc(l);
+    memset(tailStr->str, 0, l);
+    memcpy(tailStr->str, s->str + 1, l - 1);
```
gallais, line 25: What if `l` is 0?

vfrinken: it would lead to a segFault, so that needs to be fixed.

gallais:
Ok, I'm glad that my summary understanding of the C code was not completely off the mark.
Any chance we can add a test case exercising this part of the support code? i.e. a compiled
test using tail on a string and checking it works as expected on empty & non-empty strings.

gallais:
Ok, that should do it.

I am a bit concerned because I was getting segfaults because the test suite
was still relying on the old support library. Once I had installed the updated
version, everything worked just fine. :)

We'll see whether the CI works or whether we'll have to wait for a new minor
release before we can turn the test case on.

PR #853, opened 2020-12-17
WIP: Concurrent idris compilation

fabianhjr:
This is a work in progress to compile non-mutually depended upon modules concurrently.

Doesn't completely work right now, need to investigate more on Context and provide a way to merge two separate Contexts.

### DIFF
--- a/src/Core/Core.idr
+++ b/src/Core/Core.idr
@@ -389,8 +389,8 @@ export
 getWarningLoc : Warning -> Maybe FC
 getWarningLoc (UnreachableClause fc _ _) = Just fc
 
--- Core is a wrapper around IO that is specialised for efficiency.
-export
+||| Core is a wrapper around IO that is specialised for efficiency.
+public export
 record Core t where
   constructor MkCore
   runCore : IO (Either Error t)
@@ -419,6 +419,11 @@ export
 coreLift : IO a -> Core a
 coreLift op = MkCore (do op' <- op
                          pure (Right op'))
+
+export
+%inline
+coreUnlift : Core a -> IO (Either Error a)
+coreUnlift (MkCore a) = a
 
 {- Monad, Applicative, Traversable are specialised by hand for Core.
 In theory, this shouldn't be necessary, but it turns out that Idris 1 doesn't
--- a/src/Core/AutoSearch.idr
+++ b/src/Core/AutoSearch.idr
@@ -476,7 +476,7 @@ searchNames fc rigc defaults trying depth defining topty env ambig (n :: ns) tar
             then anyOne fc env topty elabs
             else exactlyOne fc env topty target elabs
   where
-    visible : Context ->
+    visible : GlobalContext ->
               List Namespace -> Name -> Core (Maybe (Name, GlobalDef))
     visible gam nspace n
         = do Just def <- lookupCtxtExact n gam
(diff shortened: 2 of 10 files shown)


fabianhjr:
There are two diverging approaches I am considering; either get the contents of the second context and add it to the first through a monadic traversal or create a pure Context -> Context -> Context merge.

By the looks of the record there is a Ref of an IOArray I would need to merge so I feel more inclined to do the first; Another issue is that it appears I will have to block while adding to the global context so that might slow the concurrency advantage a bit.

fabianhjr:
This version compiles and works(tm) sometimes. (Flaky depending on Ctxt updates)

However seems like there is almost no performance gains. Will also investigate splitting the context to a local and global one.

gallais:
> provide a way to merge two separate Contexts.

I expect it would make more sense to keep a single IOArray and
have a main thread owning it. The workers can then request fresh
pointers in an atomic way if they need to declare a new toplevel
definition.

Metas assignment should also be atomic and either succeed or return
the value already stored (so that you can compare it with the one you
wanted to put in and make sure they do match up).

libs/prelude/Prelude/Interfaces.idr
```diff
@@ -174,12 +174,24 @@ interface Applicative m => Monad m where
   (>>=) x f = join (f <$> x)
   join x = x >>= id
 
-%allow_overloads (>>=)
+public export
+(=<<) : Monad m => (a -> m b) -> m a -> m b
```
Russoul, line 178: This one is already defined at: contrib/Control/Monad/Syntax.idr

libs/prelude/Prelude/Ops.idr
```diff
@@ -15,6 +15,7 @@ infixr 7 ::, ++
 
 -- Functor/Applicative/Monad/Algebra operators
 infixl 1 >>=, >>
+infixr 1 =<<, <<
```
Russoul, line 18: (=<<)

src/Core/Core.idr
```diff
@@ -418,70 +418,37 @@ export
 %inline
 coreLift : IO a -> Core a
 coreLift op = MkCore (do op' <- op
-                         pure (Right op'))
-
-{- Monad, Applicative, Traversable are specialised by hand for Core.
```
Russoul, line 423: Are we there yet ? I know that the partial evaluator is a bit broken. But, on the other hand, there is machinery that transforms parametric method calls to specialised versions. The question is: does it work well ? Better ask @edwinb :)

libs/prelude/Prelude/Interfaces.idr
fabianhjr: Thanks for the heads up, added `<<` to the same Control.Monad.Syntax file and removed both flipped additions on prelude but moved this to #856 to separate the despecialization from the concurrency bit.

src/Core/Core.idr
fabianhjr: On testing it went from make of 114 seconds to 115 seconds. Separating this commit into #856)

andrevidela:
Ping, @fabianhjr are you still working on this?

fabianhjr:
Yes, but I went down a huge rabbit hole splitting Context unu

PR #854, opened 2020-12-17
Some functions, mostly for lazy lists

buzden:
There are a lot of functions for usual lists but almost no for lazy lists. I propose to add several functions for lazy lists in `contrib`, especially for their creation and cutting.

Also, I propose to add one function to `Data.List` in `base` and to add `%default total` there.

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -83,7 +83,7 @@ iterate f x  = x :: case f x of
 public export
 iterateN : Nat -> (a -> a) -> a -> List a
 iterateN Z     _ _ = []
-iterateN (S n) f x = x :: iterateN n f x
+iterateN (S n) f x = x :: iterateN n f (f x)
 
 public export
 takeWhile : (p : a -> Bool) -> List a -> List a
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -4,6 +4,8 @@ import Data.Nat
 import Data.List1
 import Data.Fin
 
+%default total
+
 public export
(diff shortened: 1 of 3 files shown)


libs/base/Data/List.idr
```diff
@@ -71,12 +73,18 @@ index' (_::xs) (FS i) = index' xs i
...
 public export
 iterate : (f : a -> Maybe a) -> (x : a) -> List a
 iterate f x  = x :: case f x of
   Nothing => []
   Just y => iterate f y
 
+public export
+iterateN : Nat -> (a -> a) -> a -> List a
+iterateN Z     _ _ = []
+iterateN (S n) f x = x :: iterateN n f x
```
gallais, line 86:
```suggestion
iterateN (S n) f x = x :: iterateN n f (f x)
```

libs/contrib/Data/List/Lazy.idr
```diff
@@ -53,3 +55,59 @@ public export
...
+public export
+iterate : (f : a -> Maybe a) -> (x : a) -> LazyList a
+iterate f x = x :: case f x of
+  Nothing => []
+  Just y  => iterate f y
+
+public export
+iterateN : Nat -> (a -> a) -> a -> LazyList a
+iterateN Z     _ _ = []
+iterateN (S n) f x = x :: iterateN n f x
```
gallais, line 80:
```suggestion
iterateN (S n) f x = x :: iterateN n f (f x)
```

gallais (changes requested):

PR #856, opened 2020-12-17
Refactor Core 

fabianhjr:
- Despecialize Core's implementation of Functor, Applicative, and Monad
- Add Core's HasIO implementation and an unliftCore (`Core a -> IO (Either Error a)`)

Part of #853.

### DIFF
--- a/src/Idris/ProcessIdr.idr
+++ b/src/Idris/ProcessIdr.idr
@@ -235,7 +235,7 @@ processMod : {auto c : Ref Ctxt Defs} ->
              (sourcecode : String) ->
              Core (Maybe (List Error))
 processMod srcf ttcf msg sourcecode
-    = catch (do
+    = assert_total $ catch (do -- TODO: Check totality
         setCurrentElabSource sourcecode
         -- Just read the header to start with (this is to get the imports and
         -- see if we can avoid rebuilding if none have changed)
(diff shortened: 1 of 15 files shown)


src/Idris/ProcessIdr.idr
```diff
@@ -235,7 +235,7 @@ processMod : {auto c : Ref Ctxt Defs} ->
              (sourcecode : String) ->
              Core (Maybe (List Error))
 processMod srcf ttcf msg sourcecode
-    = catch (do
+    = assert_total $ catch (do -- TODO: Check totality
```
fabianhjr, line 238: I added this due to a change in totality checking after despecializing in the concurrent-idris branch. Will investigate if this assert can somehow be removed.

fabianhjr:
`make` time after a `make clean` went from ~114 to ~120 seconds on my computer, a ~5% increase.

src/Idris/ProcessIdr.idr
fabianhjr: Couldn't figure out a way to remove the assert :/

fabianhjr:
Well, changed the concurrent bit of the other PR from Core to IO so this change is no longer needed.

Will leave the PR open in case it turns out desirable to despecialize Core.

edwinb:
It's nice that it's only a small performance hit (it used to be a pretty big one!) but unless there is a pressing need elsewhere, I'd rather leave this as it is for now. Lots of hits of 5% do add up after all.

PR #857, opened 2020-12-17
Some utilities for Fin, relations and decidability

JoeyEremondi:
Adds:
1. A total function to convert a Nat to a Fin with the proper LTE proof, with a proof that this is the inverse of `finToNat`
2. A `Data.Rel.Extra` module with a notion of the complement of a relation
3. The module `Decidable.Finite.Fin` which shows that if a predicate is decidable for any member of `Fin n`, then it's decidable whether it holds for any/all elements of `Fin n`.

My eventual plan is to have `Decidable.Finite` which uses `Rel` and the `Decidable` typeclass to show that anything that has an injection into the `Fin n` has the same Any/All decidablity, but what I've got is a reasonable start towards this.


libs/contrib/Data/Rel/Extra.idr
```diff
@@ -0,0 +1,37 @@
...
+import Data.Rel
+import Data.Fun
+import Data.Fun.Extra
+import Data.HVect
+
+%default total
+
+||| The logical complement of a relation.
+public export
+complement : {ts : Vect n Type} -> (p : Rel ts) -> Rel ts
```
gallais, line 12: This is an instance of [`map`](https://github.com/JoeyEremondi/Idris2/blob/finiteDec/libs/base/Data/Fun.idr#L18)

libs/contrib/Data/Rel/Extra.idr
```diff
@@ -0,0 +1,37 @@
...
+complement {ts = []} p = Not p
+complement {ts = t :: ts} p = \x => complement (p x)
+
+||| The negation of a relation for some elements
+||| implies that they are in the complement of the relation.
+public export
+notToComplement :
```
gallais, line 19: This looks like a generic result for `map`.

libs/contrib/Data/Rel/Extra.idr
```diff
@@ -0,0 +1,37 @@
...
+  -> Not (uncurry p elems)
+  -> uncurry (complement {ts = ts} p) elems
+notToComplement p [] npf = npf
+notToComplement p (x :: xs) npf = notToComplement (p x) xs npf
+
+||| Elements being in the complement of a relation
+||| imply that they are not in the relation.
+public export
+complementToNot : {0 ts : Vect n Type}
+  -> (p : Rel ts)
```
gallais, line 32: Same here.

libs/contrib/Decidable/Decidable/Extra.idr
```diff
@@ -47,3 +51,23 @@ doubleNegationExists {ts} {r} @{dec} nnxs =
...
+public export
+negateDec : (1 dec : Dec a) -> Dec (Not a)
+negateDec (Yes pf) = No ($ pf)
+negateDec (No npf) = Yes npf
+
+
+||| If a relation is decidable, then so is its complement
+public export
+[DecidableComplement] {ts : Vect n Type} -> {r : Rel ts} -> (posDec : Decidable n ts r) =>
+  Decidable n ts (complement {ts} r) where
```
gallais, line 66: Why not go via `curry/uncurry` and `notToComplement` & `complementToNot` lemmas?

libs/contrib/Decidable/Finite/Fin.idr
```diff
@@ -0,0 +1,91 @@
+module Decidable.Finite.Fin
+
```
gallais, line 2: This seems way too complicated. I expect that a simple:

* either `P FZ` holds
* or we use the fact that `P . FS` is decidable

ought to work (the size of `n` is structurally decreasing).

Cf. a crude attempt at deciding the universal quantifier in Agda
here: https://github.com/agda/agda-stdlib/issues/1181

gallais (changes requested):

libs/contrib/Decidable/Finite/Fin.idr
JoeyEremondi: I might be missing something obvious, but that seems fundamentally different than what I'm doing, because in your version, `P` is a predicate on Nats, where in this one, `P` is a predicate on `Fin n` for *a fixed n*, which is where all the complication comes from (since you get a value of a smaller type when pattern matching on  Fin).

Now that I think about it, I guess we can prove the more general version for `P : Nat -> Type`, and for any `P' : (Fin n) -> Type`, we could call the Nat version with `\x => (lte : LT x n ** P' (natToFinLTE x lte) )`. Which kind of makes me question the value of Fin overall vs `Fin' n = (x : Nat ** LT x n)`, since it would be 

Alright, I'll try revising it to the `Nat` version. Thanks for the feedback!

gallais: Here is what I meant

```idris
dec : (n : Nat) -> {0 p : Fin n -> Type} ->
      (pdec : (k : Fin n) -> Dec (p k)) ->
      Dec (k ** p k)
dec Z pdec = No (absurd . fst)
dec (S n) pdec =
  case pdec FZ of
    Yes pz => Yes (FZ ** pz)
    No npz => case dec n (\ k => pdec (FS k)) of
      Yes (k ** pk) => Yes (FS k ** pk)
      No npk => No $ \case
        (FZ ** pz) => absurd (npz pz)
        (FS k ** pk) => absurd (npk (k ** pk))
```

JoeyEremondi: ... oh. Yeah, I hadn't thought of doing `(\ k => pdec (FS k))` in the recursive call. That is much better. I'll put it all together and revise the PR.

JoeyEremondi:
Alright, I've simplified everything. The relation stuff is now generic in terms of `chain` (i.e. map), and I've used your massively simplified decision procedure for the Exists for finite. I also made the exists->forall building function generic. Thanks for the advice and help!

libs/contrib/Data/Nat/Order/Properties.idr
```diff
@@ -89,7 +89,24 @@ multLteMonotoneRight (S k) a b a_lte_b = CalcWith {leq = LTE} $
...
+  rewrite sym (plusZeroRightNeutral b) in ra_lte_rb
+multLteImpliesLeft 0 b (S (S k)) {  gt = gt} ra_lte_rb = LTEZero
+multLteImpliesLeft (S a) 0 (S (S k)) {  gt = gt} ra_lte_rb =
+  absurd $ replace {p = LTE _} (multZeroRightZero k) ra_lte_rb
+multLteImpliesLeft (S a) (S b) (S (S k)) {  gt = gt} ra_lte_rb
+  = let lte = fromLteSucc ra_lte_rb
+    in multLteImpliesLeft (S a) (S b) (S k) $
+       rewrite plusSuccRightSucc a (mult k (S a)) in
+       rewrite plusSuccRightSucc b (mult k (S b)) in
+       ?rhs
```
gallais: There's a hole left here.

libs/contrib/Decidable/Decidable/Extra.idr
```diff
@@ -53,21 +53,47 @@ doubleNegationExists {ts} {r} @{dec} nnxs =
...
+
+
+decideTransform :
+  {n : Nat}
+  -> {ts : Vect n Type}
+  -> {r : Rel ts}
+  -> {t : Type -> Type}
+  -> (tDec : {a : Type} -> Dec a -> Dec (t a))
+  -> (posDec : Decidable n ts r)
+  -> All ts (chain {ts} Dec (chain {ts} t r))
```
gallais, line 65: Is this return type `Decidable n ts (chain t r)`?

JoeyEremondi: Not quite, it's the type of the `decide` field of `Decidable`. My worry was that adding the generic one as an instance would clog up the interface search.

gallais: Oh I see. That's unfortunate: we should have a name we can use to
talk about a relation being decidable regardless of whether we want
to declare an implementation of the interface.

We could maybe introduce `IsDecidable` and have

```idris
interface Decidable n ts r where
  decide : IsDecidable n ts r
```

?

JoeyEremondi: Sounds good. Just pushed with that change

JoeyEremondi:
Okay, I did the IsDecidable suggestion, it looks much cleaner

libs/contrib/Decidable/Decidable/Extra.idr
```diff
@@ -47,3 +51,49 @@ doubleNegationExists {ts} {r} @{dec} nnxs =
...
+
+
+
+decideTransform :
+  {n : Nat}
+  -> {ts : Vect n Type}
+  -> {r : Rel ts}
+  -> {t : Type -> Type}
+  -> (tDec : {a : Type} -> Dec a -> Dec (t a))
+  -> (posDec : Decidable n ts r)
```
gallais, line 64: If this is not an `auto` argument then is it better to use the
interface or rather to use `IsDecidable` instead? I think building
dictionaries may be a bit more annoying than just passing the
function directly?

e.g. in `Data.List` we have:

```idris
lookupBy : (a -> a -> Bool) -> a -> List (a, b) -> Maybe b
```
and not

```idris
lookupBy : (eq : Eq a) -> a -> List (a, b) -> Maybe b
```

JoeyEremondi:
Alright, made the change you've suggested.
That said, given the rewrites with linearity that are currently happening and the bootstrap troubles, I'd be fine to put this on the backburner and put it in an external library until that's all sorted out.

gallais:
I don't think we use any of these in the compiler itself and none
of these changes use the `1` modality so I think it should be OK
to merge.

PR #867, opened 2020-12-19
Important `LazyList`'s operations were made to be truly lazy

buzden:
The current `LazyList`'s `Applicative` and `Monad` were implemented internally through usual, eager `foldr` (through `concatMap`). This led to lost laziness when using operations from these interfaces.

I propose a special lazy variant of `foldr` for `LazyList`s and reimplementing those operations using it. A function that is used to 
implement through is made `public export` because original implementations or `Applciative` and `Monad` operations were `public export` and are left so.

Lazy `concatMap` could be generalized if we had `LazyFoldable` and `LazyMonoid` interfaces, but it seems that there are not so many potential usages of such interfaces, thus this function is implemented in its specialized form for `LazyList`s.

Also, I left implementation of eager `Semigroup`'s `<+>` untouched despite the implementation of `++` is the same, because I think that lots of delaying and forcing may make computation slower when using this operation.

### DIFF
--- a/libs/base/Control/Monad/Error/Either.idr
+++ b/libs/base/Control/Monad/Error/Either.idr
@@ -107,6 +107,9 @@ Foldable m => Foldable (EitherT e m) where
   foldr f acc (MkEitherT e)
     = foldr (\x,xs => either (const xs) (`f` xs) x) acc e
 
+  foldrLazy f acc (MkEitherT e)
+    = foldrLazy (\x,xs => either (const xs) (`f` xs) x) acc e
+
   null (MkEitherT e) = null e
 
 public export
--- a/libs/base/Data/List1.idr
+++ b/libs/base/Data/List1.idr
@@ -119,6 +119,7 @@ Monad List1 where
 export
 Foldable List1 where
   foldr c n (x ::: xs) = c x (foldr c n xs)
+  foldrLazy c n (x ::: xs) = c x (foldrLazy c n xs)
   null _ = False
 
 export
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -418,6 +418,9 @@ public export
 implementation Foldable (Vect n) where
   foldr f e xs = foldrImpl f e id xs
 
+  foldrLazy op init [] = init
+  foldrLazy op init (x::xs) = x `op` foldrLazy op init xs
+
   null [] = True
   null _ = False
 
--- a/libs/contrib/Data/List/Lazy.idr
+++ b/libs/contrib/Data/List/Lazy.idr
@@ -10,6 +10,23 @@ data LazyList : Type -> Type where
   Nil : LazyList a
   (::) : (x : a) -> (xs : Lazy (LazyList a)) -> LazyList a
 
+--- Truly lazy functions ---
+
+public export
+(++) : LazyList a -> Lazy (LazyList a) -> LazyList a
+[]      ++ ys = ys
+(x::xs) ++ ys = x :: (xs ++ ys)
+
+public export
+foldrLazy : (func : elem -> Lazy acc -> acc) -> (init : acc) -> (input : LazyList elem) -> acc
+foldrLazy _  init [] = init
+foldrLazy op init (x::xs) = x `op` foldrLazy op init xs
+
+-- Specialized variant of `concatMap` with both `t` and `m` being `LazyList`.
+public export
+bindLazy : (a -> LazyList b) -> LazyList a -> LazyList b
+bindLazy f = foldrLazy ((++) . f) []
+
 --- Interface implementations ---
 
 public export
@@ -68,7 +85,7 @@ Functor LazyList where
 public export
 Applicative LazyList where
   pure x = [x]
-  fs <*> vs = concatMap (\f => map f vs) fs
+  fs <*> vs = bindLazy (\f => map f vs) fs
 
 public export
 Alternative LazyList where
@@ -77,7 +94,7 @@ Alternative LazyList where
 
 public export
 Monad LazyList where
-  m >>= f = concatMap f m
+  m >>= f = bindLazy f m
 
 -- There is no Traversable instance for lazy lists.
 -- The result of a traversal will be a non-lazy list in general
--- a/libs/contrib/Data/List/Lazy.idr
+++ b/libs/contrib/Data/List/Lazy.idr
@@ -10,22 +10,16 @@ data LazyList : Type -> Type where
   Nil : LazyList a
   (::) : (x : a) -> (xs : Lazy (LazyList a)) -> LazyList a
 
+||| Convert any foldable structure to a lazy list.
+toLazyList : Foldable t => t a -> LazyList a
+toLazyList = foldrLazy (::) []
+
 --- Truly lazy functions ---
 
 public export
 (++) : LazyList a -> Lazy (LazyList a) -> LazyList a
 []      ++ ys = ys
 (x::xs) ++ ys = x :: (xs ++ ys)
-
-public export
-foldrLazy : (func : elem -> Lazy acc -> acc) -> (init : acc) -> (input : LazyList elem) -> acc
-foldrLazy _  init [] = init
-foldrLazy op init (x::xs) = x `op` foldrLazy op init xs
-
--- Specialized variant of `concatMap` with both `t` and `m` being `LazyList`.
-public export
-bindLazy : (a -> LazyList b) -> LazyList a -> LazyList b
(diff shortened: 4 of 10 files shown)


buzden:
I thought that lazy variant of `foldr` can be actually useful to be added to the `Foldable` interface. For now, `and` , `or`, `any` and `all` operations in `Prelude.Interfaces` are implemented through `foldl` and shortcutting of `and` and `or` is claimed by having `Foldable t => t (Lazy Bool)` as an argument.

But since `&&` and `||` operations are lazy on their second argument, as soon as we have lazy variant of `foldr` in the `Foldable` interface, we can implement a short-cutting `and` and `or` for any `Foldable t => t Bool`, as I would expect, actually.

buzden:
I've added lazy variant of `foldr` to this PR, but put it in a separate commit to be able to abandon it because it, actually, differs from the original PR intension.

buzden:
I can say that I personally think even more radically. I think that `foldr` itself should require lazy folding function and eager version of `foldr` should be named specially to show that all people who use `foldr` with an eager folding function is aware of what they are doing.

gallais:
Isn't the lazy variant just the usual one with a different `acc`?
So we don't need the duplication, we can write `foldr {acc = Lazy acc}`?

buzden:
> Isn't the lazy variant just the usual one with a different `acc`?
> So we don't need the duplication, we can write `foldr {acc = Lazy acc}`?

Well, `foldr {acc = Lazy acc}` has the signature
```idris
(func : elem -> Lazy acc -> Lazy acc) -> (init : Lazy acc) -> (input : t elem) -> Lazy acc
```
which is not the same as proposed
```idris
(func : elem -> Lazy acc -> acc) -> (init : acc) -> (input : t elem) -> acc
```
especially, in the `func` argument.

For instance, you cannot easily do
```idris
and : Foldable t => t Bool -> Bool
and = foldr {acc=Lazy Bool} (&&) True
```

You would need to write something like

```idris
and : Foldable t => t Bool -> Bool
and = force . foldr {acc=Lazy Bool} (delay ... (&&)) True
  where
  (...) : (c -> d) -> (a -> b -> c) -> (a -> b -> d)
  (...) = (.) . (.)
```

Neither you can do
```idris
toLazyList : Foldable t => t a -> LazyList a
toLazyList = foldr {acc=Lazy (LazyList a)} (::) []
```

, you'll need something like

```idris
toLazyList : Foldable t => t a -> LazyList a
toLazyList = force . foldr (delay ... (::)) []
```

All normal "lazy" operations (like `&&` and lazy list's `::` above) have general form of `a -> Lazy b -> c`, not `a -> Lazy b -> Lazy c`.

So, it is unclear why do we need extra delaying and forcing -- this clutters the code and perhaps influences the runtime.

buzden:
> Isn't the lazy variant just the usual one with a different `acc`?
> So we don't need the duplication, we can write `foldr {acc = Lazy acc}`?

To estimate the difference, I experimented on the original commit if this PR, i.e. on 7ba761f47b73b6808e8793830e347fae4164796b.

In the original commit I proposed `foldrLazy` with the first argument of type `elem -> Lazy acc -> acc` only specialised for `LazyList` (i.e. not to the `Foldable` interface). In that commit I also reimplemented composition functions of `Applicative` and `Monad` interfaces to not to use general (and eager) `concatMap`, but to use specialised version for `LazyList`s that use specialised `foldrLazy`. I called this specialised version `bindLazy`:

```idris
foldrLazy : (func : elem -> Lazy acc -> acc) -> (init : acc) -> (input : LazyList elem) -> acc
foldrLazy _  init [] = init
foldrLazy op init (x::xs) = x `op` foldrLazy op init xs

-- Specialized variant of `concatMap` with both `t` and `m` being `LazyList`.
bindLazy : (a -> LazyList b) -> LazyList a -> LazyList b
bindLazy f = foldrLazy ((++) . f) []
```

For the experiment, I reimplemented the function `bindLazy` to use the original `foldr` with `acc` being `Lazy (LazyList b)` as you suppose:

```idris
bindLazy : (a -> LazyList b) -> LazyList a -> LazyList b
bindLazy f = force . foldr {acc = Lazy (LazyList b)} (\e, acc => delay $ f e ++ acc) []
```

by the way, you can actually write it _hiding_ all the delays and forcings:

```idris
bindLazy f xs = foldr {acc = Lazy (LazyList b)} (\e, acc => f e ++ acc) [] xs
```

Then I run one my program that uses applicative and monadic combinations of lazy lists a lot on both variants (well, this program  is not yet opensource, but if needed, I'm sure I can do a special example for such an experiment). My original proposed variant of `bindLazy` through `foldrLazy` gave 3 seconds of run. Then I tried `foldr {acc = Lazy (LazyList b)}` variant and I had to stop it after 10 minutes of run.

So, the gain seems obvious to me.

andrevidela:
Can we see the program you're testing this on? It's a bit hard to judge what's going on if we can't reproduce the issue. No need
to share the entire program, a minimal example would be enough, thanks! :)

buzden:
Sure. You can take

```idris
module LazyLazy

import Data.List as L
import Data.List.Lazy

%default total

coef : Nat
coef = 10000

longId : Nat -> Nat
longId n = foldr max n $ L.replicate n n

level3 : LazyList Nat
level3 = iterateN coef (+1) 0 >>= replicate coef . longId >>= replicate coef . longId

main : IO ()
main = putStrLn $ show $ head' level3
```

To measure time of run, I needed to build an executable, thus I used an ipkg-file:
<details>
  <summary>lazylazy.ipkg</summary>
  
```
package lazylazy

modules = LazyLazy

depends = contrib

main = LazyLazy
executable = lazylazy
```
</details>

Then I run with the following command:
```
rm build/exec/ -r && idris2 --version && idris2 --build lazylazy.ipkg && time build/exec/lazylazy
```

The printed result on the original commit, when `bindLazy` is implemented through `foldrLazy`:

```
Idris 2, version 0.2.1-7ba761f47
1/1: Building LazyLazy (LazyLazy.idr)
Just 0

real    0m0.064s
user    0m0.044s
sys     0m0.021s
```

Result when `bindLazy` is implemented through `foldr {acc = Lazy ...}`:

```
Idris 2, version 0.2.1-d9126f9b5
1/1: Building LazyLazy (LazyLazy.idr)
Just 0

real    7m47.746s
user    7m45.270s
sys     0m2.428s
```

To ensure what exactly I did, there is the diff:
<details>
  <summary>git diff 7ba761f47 d9126f9b5</summary>
  
```
diff --git a/libs/contrib/Data/List/Lazy.idr b/libs/contrib/Data/List/Lazy.idr
index 894b8a6f..cc6d87d8 100644
--- a/libs/contrib/Data/List/Lazy.idr
+++ b/libs/contrib/Data/List/Lazy.idr
@@ -17,16 +17,6 @@ public export
 []      ++ ys = ys
 (x::xs) ++ ys = x :: (xs ++ ys) 
 
-public export
-foldrLazy : (func : elem -> Lazy acc -> acc) -> (init : acc) -> (input : LazyList elem) -> acc 
-foldrLazy _  init [] = init
-foldrLazy op init (x::xs) = x `op` foldrLazy op init xs
-   
--- Specialized variant of `concatMap` with both `t` and `m` being `LazyList`.
-public export
-bindLazy : (a -> LazyList b) -> LazyList a -> LazyList b
-bindLazy f = foldrLazy ((++) . f) []
-   
 --- Interface implementations --- 
 
 -- Semigroup's operation is an eager variant of concatenation (++).
@@ -50,6 +40,11 @@ Foldable LazyList where
   null []     = True
   null (_::_) = False
 
+-- Specialized variant of `concatMap` with both `t` and `m` being `LazyList`.
+public export
+bindLazy : (a -> LazyList b) -> LazyList a -> LazyList b
+bindLazy f = force . foldr {acc = Lazy (LazyList b)} (\e, acc => delay $ f e ++ acc) []
+   
 public export
 Functor LazyList where
   map f [] = []
```
</details>

buzden:
I rebased to the current `master` and took the chance to restore the original signature of `null`, i.e. returning `Bool`, not `Lazy Bool`.

gallais:
I'd rather investigate why the `foldr` using a `Lazy acc` is slow than
modify the whole `Foldable` interface to add essentially duplicated code
with a different type.

edwinb:
This seems to have stalled, and in the end I agree with @gallais that it's not clear that modifying the `Foldable` interface is the right solution when this doesn't affect most things. Unfortunately laziness is never going to be perfect in this type-driven style and it might be that `LazyList` needs rethinking. So I'm going to close this for now.

PR #869, opened 2020-12-21
Add Traversable to List1

andrevidela:
I've been needing this a lot lately so I thought I'd add it.

### DIFF
--- a/libs/base/Data/List1.idr
+++ b/libs/base/Data/List1.idr
@@ -120,6 +120,10 @@ Foldable List1 where
   null _ = False
 
 export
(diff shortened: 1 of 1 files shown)


libs/base/Data/List1.idr
```diff
@@ -117,7 +117,10 @@ Monad List1 where
 export
 Foldable List1 where
   foldr c n (x ::: xs) = c x (foldr c n xs)
-  null _ = False
```
gallais, line 120: Can you please add `null _ = False` back?

PR #870, opened 2020-12-23
If the width terminal is zero make it Unbounded

michaelmesser:
Fixes the extra newlines described in https://github.com/meraymond2/idris-vscode/issues/23

### DIFF
--- a/src/Idris/Pretty.idr
+++ b/src/Idris/Pretty.idr
@@ -290,7 +290,7 @@ render doc = do
   color <- getColor
   opts <- case consoleWidth of
                Nothing => do cols <- coreLift getTermCols
-                             pure $ MkLayoutOptions (AvailablePerLine cols 1)
+                             pure $ MkLayoutOptions (if cols == 0 then Unbounded else AvailablePerLine cols 1)
                Just 0 => pure $ MkLayoutOptions Unbounded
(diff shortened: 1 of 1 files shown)


src/Idris/Pretty.idr
```diff
@@ -302,7 +302,7 @@ renderWithoutColor doc = do
   consoleWidth <- getConsoleWidth
   opts <- case consoleWidth of
                Nothing => do cols <- coreLift getTermCols
-                             pure $ MkLayoutOptions (AvailablePerLine cols 1)
+                             pure $ MkLayoutOptions (if cols == 0 then Unbounded else AvailablePerLine cols 1)
                Just 0 => pure $ MkLayoutOptions Unbounded
                Just cw => pure $ MkLayoutOptions (AvailablePerLine (cast cw) 1)
```
gallais, line 307: Can you take this opportunity to lift this to the toplevel & then call that
new function in both `do` blocks rather than having duplicated code? Thanks!

michaelmesser: So do you want a function
```idris
getRealizedPageWidth : {auto o : Ref ROpts REPLOpts} -> Core PageWidth
```
that will be called from both render and renderWithoutColor. Does that have an acceptable name and type signature?

gallais: Looks good to me!

gallais:
Thanks for the PR!

Issue #849, opened 2020-12-15
Javascript: Allow node: lambdaRequire to use local javascript files.


Invisible-Rabbit-Hunter:
In the javascript backend, trying to require (via `node:lambdaRequire:./localFile`) a local javascript file results in incorrectly generated javascript, specifically the period is placed in the name of the `__require_._localFile` variable, which is a syntax error.
# Steps to Reproduce
Compile 
```idris
module Main

%foreign "node:lambdaRequire:./compat:compat"
compat : PrimIO ()

main : IO ()
main = primIO compat
```
via the node backend.

# Expected Behavior
The generated javascript is syntactically correct.
For example a name like this:
```javascript
const __require__compat = require('./compat');
```
# Observed Behavior
The following line is generated:
```javascript
const __require_._compat = require('./compat');
```
Note the period in the middle of the name.

michaelmesser:
I think this is easy to fix. I will make a PR.

michaelmesser:
The problem is [requireSafe](https://github.com/idris-lang/Idris2/blob/910711afe994166c41fa329086872713ad6f0f38/src/Compiler/ES/ES.idr#L59) is using an incomplete blacklist. Since node caches requires, is there even a need to store it in a variable? I believe we can just require it every time. If we still want to use variables, I will replace the blacklist with a whitelist and add a hash at the end to ensure there won't be duplicate names for the same string.

michaelmesser:
By my measurement, it takes 0.000109444ms (109.444 ms for 1,000,000) to require a module after it has already been required. I am fine to implement it either way.

michaelmesser:
My previous idea would require the user to type out the escape sequences which would be ideally avoided.

Looking at this more, I am not sure why `lambdaRequire` exists at all.

Instead of:
```idris
%foreign "node:lambdaRequire:fs:fp=>__require_fs.fstatSync(fp.fd, {bigint: true}).size"
prim__fileSize : FilePtr -> PrimIO Int
```
can't one just use lambda and call require in the lambda like this:

```idris
%foreign "node:lambda:fp=>require('fs').fstatSync(fp.fd, {bigint: true}).size"
prim__fileSize : FilePtr -> PrimIO Int
```

Node caches requires. I measured that requiring an already loaded module only has a 0.000172294ms penalty (4,000,000 requires in 689.177 ms).

[...]

PR #873, opened 2020-12-26
Remove lamdaRequire

michaelmesser:
Currently `lambdaRequire`  is [buggy](https://github.com/idris-lang/Idris2/issues/849) and it requires users type out the mangled name. Since Node caches `require`s I do not see any reason to keep it. If the tiny performance hit (0.000172294ms per require) is really an issue, I can adapt this to search for `require`s with a static string as an argument and replace them with a variable.

### DIFF
--- a/docs/source/backends/javascript.rst
+++ b/docs/source/backends/javascript.rst
@@ -30,11 +30,10 @@ expression.
 
 .. code-block:: idris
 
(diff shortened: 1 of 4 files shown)


michaelmesser:
Since @rbarreiro (the author of the javascript backend) gave this a thumbs-up, can this be merged?

PR #876, opened 2020-12-27
Add :! to REPL

michaelmesser:
Add `:!` from Idris 1 to the REPL. Only tested on macOS.

### DIFF
--- a/src/Idris/REPL.idr
+++ b/src/Idris/REPL.idr
@@ -800,6 +800,9 @@ process (Editing cmd)
 process (CGDirective str)
     = do setSession (record { directives $= (str::) } !getSession)
          pure Done
+process (RunShellCommand cmd)
+    = do coreLift (system cmd)
+         pure Done
 process Quit
     = pure Exited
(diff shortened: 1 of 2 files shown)


src/Idris/REPL.idr
```diff
@@ -800,6 +800,9 @@ process (Editing cmd)
 process (CGDirective str)
     = do setSession (record { directives $= (str::) } !getSession)
          pure Done
+process (RunShellCommand cmd)
+    = do coreLift (system cmd)
```
gallais, line 804: Should we check we are not getting back an error code?
At least to warn the user?

michaelmesser: @gallais I don't think so. Bash does not by default. Neither does Idris 1. Most commands print their error messages when they error. Is there a specific use case you had in mind?

gallais: > Is there a specific use case you had in mind?

Nope. I just wanted to make sure it was not an oversight.

PR #879, opened 2020-12-27
Remove linearity subtyping

### LINKED ISSUE #73: Multiplicity subtyping with dependent types can break linearity
<a href="https://github.com/KyleDavidE"></img></a> **Issue by [KyleDavidE](https://github.com/KyleDavidE)**
_Friday May 08, 2020 at 15:07 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/357_

----

# Steps to Reproduce

```
badEq : ((x: Nat) -> Nat) = ((1 x:Nat) -> Nat)
badEq = Refl
```

# Expected Behavior

The term should not typecheck because this can be used to cast `(x: Nat) -> Nat` to `(1 x:Nat) -> Nat`, this sort of trick could easily be used to make "safe" terms of type `IO ((0 x: Type) -> IO x -> x)` which could grotesquely break things.

# Observed Behavior

It type checks, though it is of note that 

1. If you reverse the order in the type definition, (i.e. `badEq : ( (1 x:Nat) -> Nat ) =  ( (x: Nat) -> Nat)`) it does not type check
2. This doesn't work with erased arguments

edwinb:
It's disappointing to have to do this, but I think necessary because various issue reports have shown it to be unsound (at least as far as inference goes) and, at the very least, confusing. This patch brings us back to the basic conversion rules of QTT.

On the one hand, this makes the 1 multiplicity less useful, because it means we can't flag arguments as being used exactly once which would be useful for optimisation purposes as well as precision in the type. On the other hand, it removes some complexity (and a hack) from unification, and has the advantage of being correct! Also, I still consider the 1 multiplicity an experiment.

We can still do interesting things like protocol state tracking, which is my primary motivation at least.

Ideally, if the 1 multiplicity is going to be more generall useful, we'll need some kind of way of doing multiplicity polymorphism in the future. I don't think subtyping is the way (I've pretty much always come to regret adding some form of subtyping).

I won't merge this immediately, because no doubt people will have opinions and (hopefully!) ideas on other ways to resolve the problem. Also, I need to check over the documentation. But whatever we do, I'd rather this was correct than overly flexible.

Fixes #73 (and maybe some others).


edwinb:
Oh, wonderful, of course this breaks the bootstrap build because of the change in rules in the libraries. Well there's an exciting puzzle to solve...

JoeyEremondi:
Can you elaborate on what you mean by "we can't flag arguments as being used exactly once"? I'm confused how this is different from the 1 multiplicity from QTT...

edwinb:
I mean if you have

    foo : (1 _ : List a) -> List a

...then that's still fine, but you can't any longer use `foo` where a function with unrestricted multiplicity would be expected (say, as an argument to `map`). With the subtyping rule, it was harmless to add those annotations, and potentially useful because you say more about the function. Now, it has more impact on how the function can be used.

JoeyEremondi:
Ahh, I see, so linearity is useful for the internals of libraries and such, but you don't want to just pepper your functions with it here and there for the sake of (potential) performance. Makes sense.

libs/base/Data/Vect.idr
```diff
@@ -267,16 +267,13 @@ zipWith f (x::xs) (y::ys) = f x y :: zipWith f xs ys
 
 ||| Linear version
 public export
-lzipWith : (f : (1 _ : a) -> (1 _ : b) -> c)
-      -> (1 _ : Vect n a)
-      -> (1 _ : Vect n b)
-      -> Vect n c
+lzipWith : (f : a -> b -> c) -> Vect n a -> Vect n b -> Vect n c
```
gallais, line 270: This is just `zipWith` now. So we should either give it back its type
or remove it entirely IMO.

gallais (commented):
Should we add a `relax : ((1 _ : a) -> b) -> a -> b` function to `base`?
I see that a lot of linearity annotations have been removed on the first
argument. If this is a common pattern then this could become useful (even
if we don't ship functions with `1` annotations as part of `base` anymore)
as a crutch to replace implicit subtyping.

libs/base/Data/Vect.idr
edwinb: Ah yes, good spot. `relax` seems like it could be useful too, and if it is, it probably belongs in the prelude.

edwinb:
I still haven't merged this because of an annoying issue in a session type example I'm tinkering with. Essentially, I need to be able to have a `writeIORef` which is linear in the value of the reference. With this patch, `liftIO` is no longer linear in the action, which means I need a whole new set of IO primitives. This is to be expected, given the nature of the patch, but I'd still quite like to find a less extreme solution to the problem!

I wonder if extending unification to allow delaying resolution of multiplicity variables would help. Anyway, I'm just going to leave this here and try some things, and defer any kind of decision on it for a little bit longer.

Issue #810, opened 2020-12-02
Both idris2 and contrib define a module called Text.Parser


shmish111:
https://github.com/idris-lang/Idris2/blob/master/src/Text/Parser.idr and https://github.com/idris-lang/Idris2/blob/master/libs/contrib/Text/Parser.idr both exist which can cause a problem if you wish to use both contrib and idris2api. In fact it seems to cause another issue as well: https://github.com/idris-lang/Idris2/issues/809

andrevidela:
Should be fixed with #895

PR #881, opened 2020-12-28
remove modules that are also defined in contrib

shmish111:
To fix https://github.com/idris-lang/Idris2/issues/810

### DIFF
--- a/idris2api.ipkg
+++ b/idris2api.ipkg
@@ -121,14 +121,6 @@ modules =
 
     Text.Bounded,
     Text.Literate,
-    Text.Quantity,
-    Text.Token,
-
-    Text.Lexer,
-    Text.Lexer.Core,
-
-    Text.Parser,
-    Text.Parser.Core,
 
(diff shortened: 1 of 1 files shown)


ShinKage:
Unfortunately it does not seem to really fix the issue, even with this PR merged I'm still getting errors:
```Exception: invalid memory reference.  Some debugging context lost```
I guess because the two version of the related modules are actually incompatible and during the compilation is somehow using both in an unexpected manner. Probably renaming the module in the compiler would be enough, but it would be interesting to know exactly what happens when two modules have the same name and how this is translated to installed libraries.

shmish111:
@ShinKage I noticed that for my use case this doesn't actually do anything, however I am not using `idris2 --install` but rather I am copying ttc files directly. I need to find out exactly what `--install` does but I _do_ know now that you can't tell it to install the files to an alternative location (which is what I want to do).

I am closing this as it doesn't seem to be useful at the moment

PR #882, opened 2020-12-28
Progress towards building with Idris 0.2.1

edwinb:
We've lost the ability to build with the previous version, largely due to some library changes. This is a start at making things compatible again, mostly by adding a compatibility layer in the source. At least some of these should probably be removed again after 0.2.2 is released. We should also add a step to CI that builds with https://www.idris-lang.org/idris2-src/idris2-latest.tgz so we don't get into this hole again.

The biggest remaining challenge is how to deal with the change of constructor in `Data.List1`. It might be that the easiest solution is to release a 0.2.1.1 which does little other than update the List1 API.

### DIFF
--- a/idris2api.ipkg
+++ b/idris2api.ipkg
@@ -6,6 +6,12 @@ modules =
     Algebra.Semiring,
     Algebra.ZeroOneOmega,
 
+    -- Compatibility libraries; things in the Idris 2 libraries we use, but
+    -- not necessarily available in the version being used for building
+    -- Idris 2 itself
+    Compat.Data.Bifunctor,
+    Compat.Data.These,
+
     Compiler.ANF,
     Compiler.Common,
     Compiler.CompileExpr,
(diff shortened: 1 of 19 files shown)


edwinb:
Actually, by means of a horrible (temporary :)) hack I believe I can do this without an intermediate release, which will be nice.

buzden:
Does it mean that similar hacks will be required every time when libs that are used by the current released compiler version change?

edwinb:
The hack woudn't have been necessary if we'd been more aware of the potential problem. If CI tells us when we're using a thing the previous version doesn't have, we'll be aware of it.

edwinb:
Actually I think what I'm going to try (as well as the present contents of this PR) to see if it works is:

1. Copy the necessary modules from base (`Data.List1` and the things that import it) into `Compat.Data.*`
2. Import those instead of the base versions
3. Add `Compat.Idris021` (or something like that) for the other few library functions that are missing
4. Release 0.2.2
5. Undo all of the above so that it's all pretty again
6. Find a way to teach CI to check things continue to build with 0.2.2

This will be a bit boring but I suppose it'll keep me busy over the rest of the holidays.

edwinb:
There's probably a better scheme. I'll close this and come back to it, rather than think out loud in the comments!

PR #883, opened 2020-12-29
Build from stable as a CI step

mattpolzin:
Following up with a request by @edwinb (via Slack #idris channel) that new PRs are built by the latest stable compiler as a CI step to ensure that each version can always be built by the previous version.

Unsurprisingly, the new CI workflow will fail until we get `master` back to a compatible state with v0.2.1.

I took the liberty of also removing the uses of `ACTIONS_ALLOW_UNSECURE_COMMANDS` where it was trivial to do so (i.e. everywhere except for the Windows workflow where insecure commands are used more liberally, although it would still be quite reasonable to update that Workflow as well).

I went with Edwin's idea of pulling the latest stable source tar file because that seemed like a pretty reasonable way to ensure we are talking about the stable compiler any user might get if they downloaded it as a standalone package.

### DIFF
--- a/.github/workflows/ci-build-from-stable.yml
+++ b/.github/workflows/ci-build-from-stable.yml
@@ -1,0 +1,40 @@
+name: Ubuntu
+on:
+  push:
+    branches:
+      - '*'
+    tags:
+      - '*'
+  pull_request:
+    branches:
+      - master
+
+env:
+  SCHEME: scheme
+  IDRIS2_TESTS_CG: chez
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install build dependencies
+        run: |
+          sudo apt-get install -y chezscheme
+          echo "$HOME/.idris2/bin" >> $GITHUB_PATH
+      - name: Install stable version
+        run: |
+          mkdir ../stable
+          cd ../stable
+          wget https://www.idris-lang.org/idris2-src/idris2-latest.tgz
+          tar -xf idris2-latest.tgz
+          cd Idris2*
+          make bootstrap SCHEME=chez
+          make install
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-build-from-stable.yml
```diff
@@ -0,0 +1,40 @@
...
+          tar -xf idris2-latest.tgz
+          cd Idris2*
+          make bootstrap SCHEME=scheme
+          make install
+          cd ..
+          rm -rf ./Idris2*
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build current from stable
+        run: make all && make install
```
gallais, line 36: Can we try the following? Basically: how much of the problem can be solved by "use the updated `base`"?

```suggestion
        run: make libs && make install-libs && make all && make install
```

buzden: At the moment `make libs` sets `IDRIS2` variable to `../../${TARGET}` (relatively to each lib's dir) where `TARGET` is `build/exec/idris2`.

So to do what you suggest, we need change it, i.e. to make `IDRIS2` variable to be tunable for lib targets so that we can set it to point to idris2 which is installed at previous steps.

mattpolzin: @gallais I think your suggestion changes the purpose of the CI step (which I am not saying is necessarily a bad thing).

What I have PRed here is a check that the current install instructions (https://github.com/idris-lang/Idris2/blob/master/INSTALL.md#3-installing-with-an-existing-idris-2) are not broken by the incoming PR.

Your suggested change is more inline with a check that there is some possible path forward, but potentially only given a deviation from the current build instructions.

mattpolzin:
Closing, https://github.com/idris-lang/Idris2/pull/888 serves roughly the same purpose and is the stated preference.

PR #884, opened 2020-12-29
Add proof that list reverse is involutive.

mattpolzin:
There's probably an argument for not adding this directly to `base`, but when I first needed it I was a bit surprised it wasn't already in `base`, so here I am suggesting we add it.

This does overlap with `reverseReverseId` currently living in `contrib`, but that version is built up from other properties defined inside `contrib`'s `Data.List.Reverse` whereas this version only requires definitions currently in `Data.List`; an additional argument for inclusion in `base` is that this definition is not as clean (best I have figured) without access to `revOnto` which is not exported from `Data.List` so I ended up copying `revOnto` out of `base` just to create a proof that `reverse` is involutory in my own project.

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -670,6 +670,14 @@ revAppend (v :: vs) ns
             rewrite appendAssociative (reverse ns) (reverse vs) [v] in
               Refl
 
+||| List reverse applied twice yields the identity function.
+export
+reverseInvolutory : (xs : List a) -> reverse (reverse xs) = xs
+reverseInvolutory [] = Refl
+reverseInvolutory (x :: xs) = rewrite revOnto [x] xs in
+                                rewrite sym (revAppend (reverse xs) [x]) in
+                                  cong (x ::) $ reverseInvolutory xs
+
 export
 dropFusion : (n, m : Nat) -> (l : List t) -> drop n (drop m l) = drop (n+m) l
 dropFusion  Z     m    l      = Refl
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -672,11 +672,11 @@ revAppend (v :: vs) ns
 
 ||| List reverse applied twice yields the identity function.
 export
-reverseInvolutory : (xs : List a) -> reverse (reverse xs) = xs
-reverseInvolutory [] = Refl
-reverseInvolutory (x :: xs) = rewrite revOnto [x] xs in
+reverseInvolutive : (xs : List a) -> reverse (reverse xs) = xs
+reverseInvolutive [] = Refl
+reverseInvolutive (x :: xs) = rewrite revOnto [x] xs in
                                 rewrite sym (revAppend (reverse xs) [x]) in
-                                  cong (x ::) $ reverseInvolutory xs
+                                  cong (x ::) $ reverseInvolutive xs
 
 export
 dropFusion : (n, m : Nat) -> (l : List t) -> drop n (drop m l) = drop (n+m) l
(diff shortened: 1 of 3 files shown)


andrevidela:
This looks good but I'm a bit surprised that `revOnto` and `revAppend` are here, specially since they are duplicated in `Data.List.Reverse` ! 

`reverse (reverse xs) = xs` seems fundamental enough that it should be in base, so my preference would be to clean up remove the duplicates in `Data.List.Reverse`. So that means, 

- make `revOnto` public, and remove `Data.List.Reverse.reverseOntoSpec`
- remove `Data.List.Reverse.reverseAppend` since it's the same as `Data.List.revAppend`
- remove `Data.List.Reverse.reverseReverseId` since it's the same a `reverseInvolutory`

This is also because we do not really mind breaking changes in `contrib` but breaking changes in `base` (such as removing `revAppend`) are to be avoided unless absolutely necessary.

What do you think?

mattpolzin:
I'm torn. 

On one hand, I agree with your assessment and your suggestions make sense to me because `contrib` feels like it should generally dovetail with `base` nicely. I definitely feel like this reverse involutory proof (in some capacity with some name) would be a good fit for `base` (as you say, it feels fundamental).

On the other hand, one nice thing about `Data.List.Reverse` that the author of that module _must_ have done on purpose is that it all builds elegantly on top of `reverseOntoAcc`. So, since I am not that author, and I am in fact _really_ new to this community, I hesitate to go and rip out small pieces of `Data.List.Reverse` that overlap with counterparts in `base` but are implemented in fundamentally different and self-contained ways.

I'd feel better if yourself and others who have been around for longer took a strong stance on this (though maybe that's what you were already doing in your previous comment).

ohad:
Tagging @rgrover (the person who committed `contrib/Data.List.Reverse`).

Another consideration: exposing and adding to base means that removing these later will break code. 

My opinion: 
In this case it's unlikely we'll stop supporting `revOnto` or since `reverseInvolutory` is only a few lines of code, I see no harm in including it, but I'm also happy for it to stay in `contrib`.

I'm supportive of removing the redundancy between what's in the `contrib` package and `base`.

A positive take on trimming down the `contrib` package is that as the system matures and definitions harmonise and prove useful, code/api can migrate from `contrib` to `base`.

gallais:
Naming-wise, shouldn't it be involutive rather than involutory?

mattpolzin:
I came by involutory first for some reason but they appear to be synonymous and involutive undeniably fits in with subjective/injective/etc better so sounds good to me. I’ll push that name change today.

mattpolzin:
Here's another idea to throw out there if we want to avoid duplication between `base` and `contrib`:
If we wall agree `reverseInvolutive` is fundamental enough for `base` but generally feel that we should hold off on moving the other properties in `contrib/Data/List/Reverse.idr` into `base` then:
1. Remove `reverseReverseId` from the contrib module (as a duplicate of the new base library property `reverseInvolutive`).
2. Remove `reverseAppend` from contrib module (as a duplicate of the existing base library property `revAppend`).
3. Leave the `base` version of `revOnto` private because it truly is only a helper -- even though the `contrib` module duplicates this function, the `contrib` module itself only uses it in `reverseCons` which in turn is only used to prove `reverseAppend` (same signature as `revAppend` in `base`) and `reverseReverseId` (which we are discussing moving into `base` right now as `reverseInvolutive`).

This still leaves a bit of duplication in `contrib` but it puts off decision making about the rest of that module, reducing duplication on the whole, moves the involutive property into `base`, and leaves `revOnto` private thus not increasing the API surface area of `base` by more than the new involutive property.

mattpolzin:
Although we may still hear from @rgrover, I went ahead and made the changes outlined in my most recent comment to keep this PR moving forward.

andrevidela:
Since we haven't heard back I'll merge this, thanks a lot! :)

PR #887, opened 2020-12-29
Add it to REPL

michaelmesser:
Implement `it` in the REPL. Since this is in the contributions wanted and is easy to implement, I figured I would add it. I would be willing to try improve `:let` and add `:unlet` if that is wanted, but I would need to know if and how the behavior should be different from Idris 1 before I start. I'm not entirely sure if I did this right, but it works.

### DIFF
--- a/src/Idris/REPL.idr
+++ b/src/Idris/REPL.idr
@@ -643,9 +643,10 @@ process (Eval itm)
                  ntm <- norm defs [] tm
                  logTermNF "repl.eval" 5 "Normalised" [] ntm
                  itm <- resugar [] ntm
+                 ty <- getTerm gty
+                 addDef (UN "it") (newDef emptyFC (UN "it") top [] ty Private (PMDef defaultPI [] (STerm 0 ntm) (STerm 0 ntm) []))
                  if showTypes opts
-                    then do ty <- getTerm gty
-                            ity <- resugar [] !(norm defs [] ty)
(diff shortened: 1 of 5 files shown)


gallais:
Can we add a test case? In particular what happens if you evaluate 2 separate
expressions and then call `it`? It looks to me like you may run into trouble
and get an ambiguity error?

michaelmesser:
@gallais What do you mean by two separate expressions? Can you give me an example?

gallais:
My understanding of the patch is that after:

```idris
Main> 2
2
Main> 4
4
Main> 
```

we would have 2 new constants in scope `it = 2` and `it = 4` and when we call
`it` at the REPL there is no way for Idris to pick one over the other.

michaelmesser:
@gallais No. [addDef](https://github.com/idris-lang/Idris2/blob/771e4034fad1c4fe82122bae4ddaded0e1a88704/src/Core/Context.idr#L1093) calls [addCtxt](https://github.com/idris-lang/Idris2/blob/771e4034fad1c4fe82122bae4ddaded0e1a88704/src/Core/Context.idr#L433). addCtxt "Add the name to the context, or update the existing entry if it's already there." I can add a test if you want. I still may have made mistakes. I have no idea what `STerm` is but it seems to work.

michaelmesser:
@gallias I added a simple test.

gallias:
Hi there!
You've mentioned wrong person;-)


Cheers,
Anna

PR #888, opened 2020-12-29
Add CI action to build from older version

edwinb:
Specifically, at the moment, to build from Idris 0.2.2 (which is not yet released, but is basically 0.2.1 plus a few bits that are needed to build the latest version). I suggest we keep this at the earliest version known to work, though there is also no need to be shy about increasing the minimum requirements when we need to.

This CI action skips the tests - it's just about knowing whether the APIs are still up to date.

I'll release 0.2.2 soon, then 0.3.0 not long after, after checking over the latest PRs.

### DIFF
--- a/.github/workflows/ci-buildprevious.yml
+++ b/.github/workflows/ci-buildprevious.yml
@@ -29,7 +29,7 @@ jobs:
           wget https://www.idris-lang.org/idris2-src/idris2-0.2.2.tgz
           tar zxvf idris2-0.2.2.tgz
           cd Idris2-0.2.2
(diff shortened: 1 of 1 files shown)


edwinb:
I see I missed #883 before doing this. I think I prefer having a specific lower version bound, though. Most of the incompatibilities I found were really small things that just required writing something a slightly less neat way, and that sort of thing shouldn't really necessitate changing the minimum requirement. But, more generally, I'd just like to keep track of what the minimum requirement for building actually is, and this way we have half a chance of knowing that!

PR #889, opened 2020-12-29
Add rlwrap information to docs

JonathanLorimer:
This PR adds some documentation about how one can get command history in the Idris2 repl by using the rlwrap utility.


andrevidela (approved):
Looks good to me

PR #892, opened 2020-12-30
modified definition of ltrim using strM

mapf0ld:
Modified the definition of ltrim function using strM, which uses primitive operations a lot, to make trim function work faster. This change is directly adopted from the Idris1 library source so there should be no issues. I tried building the code myself on my local PC and both bootstrapping by chezscheme and self-hosting build were successful.


mapf0ld:
Oops something is wrong...

mapf0ld:
It seems that the totality issues aren't yet resolved. I'll take care of it.

PR #894, opened 2020-12-31
[ refactor ] ltrim in terms of asList

mapf0ld:
slightly modified Data.Strings.ltrim using strM primitive to make it work much faster. 
As strM is a primitive, I needed to use assert_total on a recursive call of ltrim
to convince the totality checker of the totality.


gallais:
The purpose of `StrM` (what is this name supposed to mean? Would `Uncons`
be a better one?) seems to be to only inspect the first element of the
string.

My advice would be to implement another view first (for instance called
`Unpack`) that goes something like:

```idris
data Unpack : String -> Type where
  Nil  : Unpack ""
  (::) : (c : Char) -> {str : String} -> Lazy (Unpack str) -> Unpack (strCons c str)
```

You should be able to write `unpack : (str : String) -> String` by using
`strM` and some `assert_total`. Once that is done, you'd be able to use
the recursion style you want to without any annotations because the `Unpack`
family gives you something to do structural induction on.

ShinKage:
There is also `Data.String.Iterator` which has an uncons type similar to the one suggested by @gallais and also has lazy unpacking and folds.

mapf0ld:
> The purpose of `StrM` (what is this name supposed to mean? Would `Uncons`
> be a better one?) seems to be to only inspect the first element of the
> string.
> 
> My advice would be to implement another view first (for instance called
> `Unpack`) that goes something like:
> 
> ```idris
> data Unpack : String -> Type where
>   Nil  : Unpack ""
>   (::) : (c : Char) -> {str : String} -> Lazy (Unpack str) -> Unpack (strCons c str)
> ```
> 
> You should be able to write `unpack : (str : String) -> Unpack str` by using
> `strM` and some `assert_total`. Once that is done, you'd be able to use
> the recursion style you want to without any annotations because the `Unpack`
> family gives you something to do structural induction on.

strM itself is defined in libs/base/Data/Strings.idr as prelude/Prelude/Strings.idr in Idris1, 
and in Idris1 ltrim is implemented with the strM primitive and assert_total, like this
```Idris
trim : String -> String
ltrim xs with (strM xs)
    ltrim "" | StrNil = ""
    ltrim (strCons x xs) | StrCons _ _
        = if (isSpace x) then assert_total (ltrim xs) else (strCons x xs)
```
so I just copied this definition in Idris1. I thought it better not to modify too much code, but would it be a problem?

Russoul:
@ShinKage But there appears to be no way to get the String back from a StringIterator other than traversing the lazy list in full anyway

ShinKage:
@Russoul I see, we should add a function to get back the remaining part of the string from a StringIterator.

gallais:
Given this was getting stale, I have pushed a commit implementing the view I
was suggesting & using it to implement `ltrim`. A string iterator would have
worked, except for the fact they're in `contrib` and this is in `base`.

Additionally, this is not completely useless in the sense that the view actually
says something about the shape of the string it is indexed by.

PR #896, opened 2020-12-31
Add withIteratorString

Russoul:
The proposed function allows for more effective backend-dependent iterator-to-string conversions.

Pinging #894

### DIFF
--- a/libs/contrib/Data/String/Iterator.idr
+++ b/libs/contrib/Data/String/Iterator.idr
@@ -32,6 +32,17 @@ export
 withString : (str : String) -> ((1 it : StringIterator str) -> a) -> a
 withString str f = f (fromString str)
 
+||| Runs the action `f` on the slice `res` of the original string `str` represented by the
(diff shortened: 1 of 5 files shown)


edwinb:
Thanks for this. One thing I wonder, if we have functions like this which could be implemented in Idris, but are more efficient done directly using the backend's primitives - should there be a way to fall back to the Idris implementation on other back ends?

It could probably be done with an FFI string of "idris_default:<function_name>" without too much trouble.

Independently of that, I'll merge this anyway.

Russoul:
We don't have that many foreign functions yet. And it's in our interest to keep that number low, to be as portable as possible. So I am not sure we super need this.

Moreover, AFAICR backends currently take FFI strings as a whole. Meaning they'll have to adapt to handle search and insertion of appropriate default function calls. While Idris can't take responsibility of handling defaults, because it doesn't know whether to fall back until it passes the control of the compilation flow to the backend.

PR #897, opened 2021-01-01
test: Change function definitions in "Generic.idr".

VoidNoire:
Seemed like the doubling functions had incorrect function definitions, so I changed them such that they actually double their inputs instead of squaring them.


andrevidela (approved):
Looks good to me! I checked with the book too and indeed, those should have been additions

PR #898, opened 2021-01-02
Apply small refactors all around

GustavoMF31:
Use find, elemAt, zip, etc instead of direct recursion.

I noticed a couple of these patterns were repeated throughout codebase and refactored the ones that I could find. (Actually, I remember having seen another function that was just a specialized `find` somewhere, but can't quite remember what file that was in. Does anyone remember such thing?).

I timed how long it took to build before and after the changes just to make sure there was no performance impact, and it took pretty much the same: 4m:14s before and 4m:10s after.


GustavoMF31:
Looks like I got `getUsings` wrong. (I wonder why I didn't get a type error. I must have messed something up). Should be fixed now.

edwinb:
If the performance is the same and we can still build from the previous version, then a bit of tidying is always nice to see, thanks. I can merge this once the conflict is resolved.

GustavoMF31:
There we go!

I just haven't got around to timing it again. I'm skeptical of trusting the previous numbers, as there was clearly an uncaught type error in the code. (If I were testing the right builds, they would have caught the error, right?).

GustavoMF31:
I've now timed it again! I got 3m47s without the changes and 3m37s with the changes, so I'm confident there is no performance impact. I thinks it's ready to merge.

src/Idris/REPLOpts.idr
```diff
@@ -46,7 +46,7 @@ export
 defaultOpts : Maybe String -> OutputMode -> List (String, Codegen) -> REPLOpts
 defaultOpts fname outmode cgs
     = MkREPLOpts False NormaliseAll fname (litStyle fname) "" "vim"
```
gallais, line 48: Now that we have named applies, it may be worth refactoring this unreadable
`MkREPLOpts` application to:

```idris
MkREPLOpts
  { showTypes = ...
  , evalMode  = ...
  , ...
  }
```

? :)

GustavoMF31: Sure! This way is definitely more readable. By the way, is this the change that is mentioned in the changelog as "Added new syntax for named applications of explicit arguments"? I read that and had no idea what it was about 😅. `f {x [= t], x [= t], ...}` is pretty cryptic though.

GustavoMF31:
I took the opportunity to refactor the creation of most big records to use the named application syntax. I also realized that some of the manipulation of `EState`'s could be more straightforwardly expressed as record updates, so I refactored that as well. I was a bit afraid that it was done the other way due to performance, so I timed the `make clean && make all` process using the version with record updates, and got 3m45s, which is pretty much the same I was getting before the change.

gallais:
This is currently blocked on #925

fabianhjr:
Hi @gallais, I did add the change you suggested a bit before your comment. It is currently green and ready to merge.

gallais:
I know I was waiting for the build to finish. :)

PR #900, opened 2021-01-03
Add `snoc` to Data.Vect

andrevidela:
Snoc adds an element at the end of the vector. The main use case
is to get the expected type signature
`Vect n a -> a -> Vect (S n) a` instead of
`Vect n a -> a -> Vect (n + 1) a` which you get by using `++ [x]`

I've needed this a lot lately, `List` already has `snoc` so I thought
`Vect` should get it too

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -149,6 +149,22 @@ public export
 (++) []      ys = ys
 (++) (x::xs) ys = x :: xs ++ ys
 
+||| Add an element at the end of the vector.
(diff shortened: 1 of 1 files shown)


gallais:
Do you need the `1` quantities? Because of the upcoming removal of
linearity subtyping (#879), most of the linearity annotations are
going to be removed from the std libs.

andrevidela:
Ah indeed they might be superfluous, I'll remove them

andrevidela:
I removed them

PR #901, opened 2021-01-03
Add count to prelude

andrevidela:
I've needed this function a lot lately so I thought other people might
need it too.

I've placed it in `Types` because it cannot live in `Interfaces`. Indeed,
since the return type is `Nat` it would require importing `Types` making
the imports circular. Another solution would have been to add it to `base`
rather than `prelude` but I could not find a suitable place for it. Finally,
another solution would have been to place it in `Interfaces` and have it 
depend on the `Num` interface and return a member of `Num` rather than
`Nat` but that meant worse inference and inaccurate semantics since the
count can never be negative.


gallais:
I guess it could also be implemented in terms of the more fundamental

```idris
foldMap :: (Foldable t, Monoid m) => (a -> m) -> t a -> m
```

that we do not currently have.

andrevidela:
@gallais I've added `foldMap` as well, note that I cannot use `foldMap` to implement `count` here because the `Monoid` instance for `Nat` lives in `base`

PR #902, opened 2021-01-03
Interactive & REPL :search feature

mattpolzin:
I don't believe this is as robust of a search feature as Idris 1 had, but for what it is I have found it very useful personally. Perhaps something is better than nothing here in the short term? 

As someone without a strong FP (and perhaps particularly Haskell) background, having the ability to search by type signature is incredibly powerful because I don't just happen to know what certain functions tend to be called.


src/Idris/REPL.idr
```diff
@@ -715,14 +749,23 @@ process (Exec ctm)
...
-         pure $ ProofFound itm
+process (TypeSearch searchTerm@(PPi _ _ _ _ _ _))
+    = do defs <- branch
+         let ctxt = gamma defs
+         rawTy <- desugar AnyExpr [] searchTerm
+         let bound = piBindNames replFC [] rawTy
+         (ty, _) <- elabTerm 0 InType [] (MkNested []) [] bound Nothing
+         ty' <- toResolvedNames ty
+         filteredDefs <- do names   <- allNames ctxt
+                            defs    <- pure $ catMaybes !(traverse (flip lookupCtxtExact ctxt) names)
```
andrevidela, line 760: Small syntactic suggestion, but I think this works:

```suggestion
                            defs    <- catMaybes <$> (traverse (flip lookupCtxtExact ctxt) names)
```

src/Idris/REPL.idr
```diff
@@ -607,6 +609,38 @@ loadMainFile f
...
+                          ty <- normaliseHoles defs [] (type def)
+                          pure [(show n) ++ " : " ++ (show !(resugar [] ty))]
+
+equivTypes : {auto c : Ref Ctxt Defs} ->
+             (ty1 : ClosedTerm) ->
+             (ty2 : ClosedTerm) ->
+             Core Bool
+equivTypes ty1 ty2 = do defs <- get Ctxt
+                        True <- pure (!(getArity defs [] ty1) == !(getArity defs [] ty2))
+                          | False => pure False
```
andrevidela, line 637: I think this is `when`

```suggestion
                        when (!(getArity defs [] ty1) == !(getArity defs [] ty2))) (pure False)
```

src/Idris/REPL.idr
mattpolzin: Thats definitely better; thanks!

src/Idris/REPL.idr
mattpolzin: I don't think that captures the same meaning. I want the entire `do` block to short circuit with `pure False` when the arities differ as opposed to executing something conditionally (unless you're talking about the entire rest of the block as the thing being optionally executed).

That said, it would not surprise me if there _was_ something better suited to the task here. It feels awkward to take the `False` case and just pass it along to `pure False` but that happens to be how I am getting the short circuit behavior, too.

andrevidela: Ah yes you're right! This is the right thing to do then :)

edwinb:
Thanks! Yes, having something implemented that can be improved on is definitely better than not having it at all. I'm about to release a new version, and it's probably safer to merge after that just in case.

Issue #53, opened 2020-05-20
Automated Documentation generation from source code.


edwinb:
<a href="https://github.com/jeetu7"><img src="https://avatars1.githubusercontent.com/u/1040586?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jeetu7](https://github.com/jeetu7)**
_Friday Apr 17, 2020 at 10:07 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/295_

----

This issue pertains to the mechanism for genreating html documentation automatically from source idr files. It might be something like hackage/haddock or python-sphinx-autodoc. Here we can document the approaches and progress till we have something implemented. Afterwards, further sub-issuse can be created.

The documentation extracted can be linked with browsable docs of the source code. #294

edwinb:
<a href="https://github.com/clayrat"><img src="https://avatars0.githubusercontent.com/u/321557?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [clayrat](https://github.com/clayrat)**
_Friday Apr 17, 2020 at 11:37 GMT_

----

Note that Idris1 already supports this with `--mkdoc` cmdline option, but it is broken due to https://github.com/idris-lang/Idris-dev/issues/2161

edwinb:
This exists now, although there is more to do - browsable source code would be nice for example. Even so, I think we can close this one now.

PR #903, opened 2021-01-04
WIP: Generate documentation from source code (aka `--mkdoc` support)

cypheon:
This is an initial WIP version of automated documentation generation from source code (implements feature request #53).

When running `idris2 --mkdoc prelude.ipkg`, docs are created in the `build/docs/` directory.

Example output for the "base" package can be found here: https://idris2docs.sinyax.net/data/base/index.html

Known issues:
 * types are not shown where they should be (although most of the time, the type is in the first line of the documentation)
 * "documentation" is currently just a `<pre>` tag with the output equivalent to `:doc` in the REPL
 * namespace members are sorted alphabetically, this may not be the best choice
 * `--clean` does not remove the generated docs
 * code needs some cleanup (and probably moved into a separate file, as it pollutes `Package.idr` with lots of HTML output-specific code, which will get worse as more features are implemented)

It is a very early stage, basically just the output from `:browse` and `:doc` stitched together into an HTML file, but is (at least to me) surprisingly useful to get an overview of a package.

This is definitely not ready to be merged, I created this PR to let people know that I'm working on it and as a place for feedback and discussion of the next steps.

PS: The stylesheet is a 1:1 copy from Idris 1, so docs may look somewhat familiar


gallais (approved):

edwinb:
This is great, thanks! I'll see if I can get some docs up on idris-lang.org shortly.

PR #905, opened 2021-01-05
Add public export to types/functions in Data.Fin.Order

alebahn:
I was trying to make use of Data.Fin.Order and realized that the code was not exported. I've added "public export" to the types and functions in Data.Fin.Order.

Let me know if you think any of them should be export instead of public export.


andrevidela (approved):
Looks good to me!

PR #908, opened 2021-01-07
Fix REPL execExp and fix "it"

### LINKED ISSUE #890: :exec using variable defined in :let causes Exception: variable is not bound
# Steps to Reproduce

```idris
Main> :let x : String
Main> :let x = "Test"
Main> :exec putStrLn x
```

# Expected Behavior

```idris
Main> :exec putStrLn x
Test
Main>
```

# Observed Behavior
Chez:
```idris
Main> :exec putStrLn x
Exception: variable Main-x is not bound
Main>
```
Node:
```idris
Main> :exec putStrLn x
/Users/michaelmesser/build/exec/_tmp_node.js:51
try{PrimIO_unsafePerformIO(undefined, Prelude$5fIO_putStrLn(undefined, Prelude$5fIO_$5f$5fImpl$5fHasIO$5fIO(), Main_x));}catch(e){if(e instanceof IdrisError){console.log('ERROR: ' + e.message)}else{throw e} }
                                                                                                                                                                                                      ^

ReferenceError: Main_x is not defined
    at Object.<anonymous> (/Users/michaelmesser/build/exec/_tmp_node.js:51:112)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1025:10)
    at internal/main/run_main_module.js:17:11
```

michaelmesser:
- In my PR for `it` I copied what Idris 1 did and that breaks referential transparency so I am fixing it in this PR
- Store a name to the last result in `REPLOpts`
- Clear that name on load
- Use `ILocal` to define it for expressions 
- Make a generalized version of `checkLocal` so it can be used to on declarations too
- Factors out parts of `execExp` and `compileExp` in to `prepareExp`
- Invoke `compileAndInlineAll` in `prepareExp`
- Change `it` to `replFC`
- `addToSave` `it`
- Use `DN` to display the machine name for `it` as `it`
- Add a test
- Resolves #890 
- Not sure if `compileAndInlineAll` recompiles already compiled expressions so this could be inefficient
- I have no idea why this fails bootstrap
- I'm open to better ways if anyone knows of any


src/Idris/REPL.idr
```diff
@@ -644,7 +653,9 @@ process (Eval itm)
                  logTermNF "repl.eval" 5 "Normalised" [] ntm
                  itm <- resugar [] ntm
                  ty <- getTerm gty
-                 addDef (UN "it") (newDef emptyFC (UN "it") top [] ty Private (PMDef defaultPI [] (STerm 0 ntm) (STerm 0 ntm) []))
+                 let itName = UN "it"
+                 addDef itName (newDef replFC itName top [] ty Private (PMDef defaultPI [] (STerm 0 ntm) (STerm 0 ntm) []))
+                 addToSave itName
```
Russoul, line 658: We probably don't want to save the temporary definition to a `.ttc` file

michaelmesser: @Russoul If I do not `addToSave itName` then `:exec putStrLn it` fail (since `it` is not in `toIR`). Should I try just adding `it` to `toIR` and avoid adding `it` to `toSave`?

Russoul: Yeah, you can try that. I don’t know how Idris handles writing to ttc when inside the REPL. It probably won’t write anything, but still better to be sure.

michaelmesser: @Russoul The way I implemented `it` is fundamentally broken. I have a few ideas to fix this issue that I will try.
```idris
Main> 1
1
Main> :let f : Integer -> Integer
Main> :let f x = x + it
Main> f 1
2
Main> f 1
3
Main> :let p : f 1 = 4
Main> :let p = Refl
Main> p
Refl
Main> 4
4
Main> :let q : f 1 = 5
Main> :let q = Refl
Main> p
Refl
Main> :t p
Main.p : f 1 = 4
Main> :t q
Main.q : f 1 = 5
Main> :let g : (a = b) -> (a = c) -> (b = c)
Main> :let g Refl Refl = Refl
Main> g p q
Refl
Main> :t g p q
g p q : 4 = 5
Main>
```

Russoul: Ah, I see. The temporary definition gets replaced with a new one each time. But the previous one depends on the new, breaking referential transparency. You can generate a new unique name for every next `it`. Maybe that’s what you have in mind already.

michaelmesser: @Russoul My ideas are the following:
1. `let it = {result_of_last_exp} in {current_exp}` and avoid messing with the context at all
2. Somehow make the `it`s unique (different namespaces maybe)?
3. Somehow force `it` to be inlined immediately
4. `it0`, `it1`, `it2`, ...

Can you clarify exactly what your idea is?

Russoul: In the REPL we only have global expressions/definitions AFAIK. So your best bet is probably to globally store an Int - your next unique`it` name ID. And then have something like: `MN (“it”, nextId)`. But you’ll also need to substitute this machine generated name for actual (UN “it”) that user inputs.

michaelmesser: @Russoul Why not `let it = {result_of_last_exp} in {current_exp}`? It seems simplest.

Russoul: That could work. But let bindings don’t reduce in Idris 2. 
For example:
`let x = 2 in the (x === 2) Refl` doesn’t type check. That could be confusing for users (`it` won’t reduce in types)

michaelmesser: @Russoul Would `where` work?

Russoul: let/where local definitions would work. But you'd still have to rename each clause to a unique name.
So I think it would be easier to stick with the global definition approach.

P.S. Just in case you don't know, we have irreducible let bindings and reducible let/where local definitions. Here is the difference:
```
doesntWork : 2 === 2
doesntWork = let x = 2
              in the (x === 2) Refl -- let binding here, doesn't reduce in (x === 2)

works : 2 === 2
works = let x : ?
            x = 2
         in the (x === 2) Refl -- local definition introduced via `let`, reduces normally in (x === 2)

```

michaelmesser: @Russoul `ILocal` / `PLocal` seems to work well for expressions. Is there an equivalent for declarations? Maybe something with namespaces? I do not see a simple way to transform all occurrences of `UN "it"` into the machine name.

Russoul: To transform occurrences you can utilise:
https://github.com/idris-lang/Idris2/blob/347829755786b13028333b2584599845ce6eca5b/src/TTImp/Utils.idr#L244
^--- for expressions

https://github.com/idris-lang/Idris2/blob/347829755786b13028333b2584599845ce6eca5b/src/TTImp/Utils.idr#L254
^--- for pattern matching clauses

So my idea is:

1. Construct a global definition out of what the user enters with a name `MN "it" (uniqueId + 1)`
2. Rename all occurrences of `UN "it"` to `MN "it" uniqueId ` in its body (one `PatClause `)
3. Add the definition
4. Increment the `uniqueId` counter

michaelmesser: I used `genName` to get a unique name and am storing the name in `REPLOpts`. For expressions I am using `ILocal` and I believe it works correctly.  I believe I can modify the code from `checkLocal` to work for declarations. If that fails I will try to use `substNameClause`.

michaelmesser: @Russoul Adapting `checkLocal` from `RawImpl` to `List ImpDecl` worked. To reduce duplication, should I write helper function and rewrite `checkLocal` in terms of that helper function? `checkLocal` is about 100 lines long.

michaelmesser: @Russoul When `:load` is invoked the definitions are cleared. Should `it` be cleared too?

michaelmesser: The last problem is making:
```
Main> :t it
Main.{evalResult:571} : Integer
```
be:
```
Main> :t it
it : Integer
```

Russoul: Don't know about `checkLocal`. It is used internally to type check let/where clauses inside a global definition.
And I don't have a particular feeling whether it's okay to do the `it` this way or not. If it turns out to work without
making too many internal changes, then it's probably fine. 

You can mark the PR as WIP (just in case) and push your potential implementation here - for others to see and comment on.

michaelmesser: @Russoul I have something that I think is reasonable now. I never thought `it` would be this complicated. Thank you for your help.

gallais:
Looks really good! Planning to merge as soon as the CI is green.

michaelmesser:
@gallais Not sure why CI is red. It doesn't like a test I added. But I don't understand why it fails. Also not sure if using `DN` to hide the machine name for `it` is a good idea or not.

gallais:
> Can bootstrap be updated?

The bootstrap should not matter: the tests are only run with the dev
version of Idris2. Otherwise we couldn't possibly fix bugs because the
bootstrap & fixed version would return different results! :D

I'll investigate.

michaelmesser:
@gallais I think a bug in bootstrap makes the resulting compiler (my PR built from bootstrap) have a bug. But I don't understand it so I could be completely wrong.

gallais:
It works just fine on my machine. :thinking:

michaelmesser:
@gallais Is it something to do with Racket vs Chez?

gallais:
I was careful to use `make bootstrap-racket` precisely to try to reproduce
the failure. In the end, it may just have been a CI fluke because the test
just worked after merging master in the branch. :)

michaelmesser:
@gallais The CI is failing on master

stefan-hoeck:
I'm a Nix user. Building fails on my system too. I'll try and figure out what's going on.

stefan-hoeck:
So, in my case the failure to build was related to changes made in #967. There, file package.nix was adjusted. Since I'm using my own version of this file, those adjustments where not included automatically in my case. Could this also be the issue with the CI?

gallais:
This is a known, unrelated, issue: #957 

(Also can you please stop @-ing me in every message? I see notifications
on github every time issues are modified so I will not miss new messages.
But I get push notifications on my phone when I'm being explicitly mentioned
which is a bit noisy for normal discussions).

PR #909, opened 2021-01-08
added MaybeT monad transformer to base

stefan-hoeck:
This adds MaybeT to base. The implementation was based on the one of EitherT. Note: As far as I can see, this was not in Idris1 but neither was EitherT, so I thought it should be fine if I put this in base.

### DIFF
--- a/libs/base/Control/Monad/Maybe.idr
+++ b/libs/base/Control/Monad/Maybe.idr
@@ -108,7 +108,7 @@ Functor m => Functor (MaybeT m) where
 public export
 Foldable m => Foldable (MaybeT m) where
   foldr f acc (MkMaybeT e)
-    = foldr (\x,xs => maybe acc (`f` xs) x) acc e
+    = foldr (\x,xs => maybe xs (`f` xs) x) acc e
 
   null (MkMaybeT e) = null e
 
--- a/libs/base/Control/Monad/Maybe.idr
+++ b/libs/base/Control/Monad/Maybe.idr
@@ -73,9 +73,6 @@ nothing = MkMaybeT $ pure Nothing
 -- Interface Implementations
 -------------------------------------------------
 
-on : (b -> b -> c) -> (a -> b) -> a -> a -> c
-on f g x y = g x `f` g y
-
 public export
 Eq (m (Maybe a)) => Eq (MaybeT m a) where
  (==) = (==) `on` runMaybeT
@@ -152,4 +149,3 @@ public export
 MonadState s m => MonadState s (MaybeT m) where
   get = lift get
   put = lift . put
-
--- a/libs/base/Control/Monad/Maybe.idr
+++ b/libs/base/Control/Monad/Maybe.idr
@@ -1,0 +1,154 @@
+module Control.Monad.Maybe
+
+-------------------------------------------------
+-- The monad transformer `MaybeT m a` equips a monad with the ability to
+-- return no value at all.
+
+-- Sequenced actions of `MaybeT m a` produce a value `a` only if all of the
+-- actions in the sequence returned a value.
+
+-- This is isomorphic to `EitherT ()` and therefore less powerful than `EitherT e a`
+-- ability to not return a result.
+-------------------------------------------------
+
+import Control.Monad.Trans
+import Control.Monad.Reader
+import Control.Monad.State
+import Data.Maybe
+
+public export
+data MaybeT : (m : Type -> Type) -> (a : Type) -> Type where
+  MkMaybeT : (1 _ : m (Maybe a)) -> MaybeT m a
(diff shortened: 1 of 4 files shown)


libs/base/Control/Monad/Maybe.idr
```diff
@@ -0,0 +1,154 @@
...
+export
+%inline
+nothing : Applicative m => MaybeT m a
+nothing = MkMaybeT $ pure Nothing
+
+-------------------------------------------------
+-- Interface Implementations
+-------------------------------------------------
+
+on : (b -> b -> c) -> (a -> b) -> a -> a -> c
```
gallais, line 76: This should probably be moved somewhere more convenient.
E.g. next to `id` in `Prelude.Basics`

buzden: If it would, the same function should be removed from `Control.Monad.Either`.

stefan-hoeck:
OK, I change the following two things:

- The Semigroup instance of MaybeT still used the Applicative implementation although the comments in the code suggested otherwise. I changed that to the more efficient monadic version and added a doc string explaining this decision.
- Utility function `on` was moved to Prelude.Basics together with some documentation and an example use case. The private versions of `on` were removed from Control.Monad.MaybeT and Control.Monad.EitherT.

If someone could quickly review the documentation for `Prelud.Basics.on` that would be great (I'm no native English speaker).

buzden:
> If someone could quickly review the documentation for `Prelud.Basics.on` that would be great (I'm no native English speaker).

[Haskell's library](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Function.html#v:on) has good description: "`on b u x y` runs the binary function `b` _on_ the results of applying unary function `u` to two arguments `x` and `y`". I personally think (despite I'm neither native speaker) that existing "Binary function composition" is too general: it could describe a function with e.g. `(c -> d) -> (a -> b -> c) -> a -> b -> d` signature.

By the way, don't you think it's reasonable do add
```idris
infixl 0 `on`
```
nearby?

stefan-hoeck:
I changed the documentation to the one found in Haskell. I left out the part about algebraic properties, since most other functions in the prelude don't state these kinds of laws either. We probably could prove these laws anyway, but I'm not sure whether these proofs should be in Prelude.Basics.

As suggested by @buzden, I also added `infixl 0 `on``.

libs/base/Control/Monad/Maybe.idr
```diff
@@ -0,0 +1,151 @@
...
+  neutral = nothing
+
+public export
+Functor m => Functor (MaybeT m) where
+  map f m = MkMaybeT $ map f <$> runMaybeT m
+
+public export
+Foldable m => Foldable (MaybeT m) where
+  foldr f acc (MkMaybeT e)
+    = foldr (\x,xs => maybe acc (`f` xs) x) acc e
```
gallais, line 111: I think this is a bug: in case `x` is nothing we should return the
current accumulator (`xs`) rather than discarding it and returning
the default one (`acc`).

```suggestion
    = foldr (\x,xs => maybe xs (`f` xs) x) acc e
```

gallais (changes requested):

libs/base/Control/Monad/Maybe.idr
stefan-hoeck: Thanks for spotting this.

gallais: Oh, this is really bad.

Can you put the fix in a separate PR? I'd like to squash this one
but this fix should definitely be a separate commit.

libs/base/Control/Monad/Maybe.idr
```diff
@@ -0,0 +1,151 @@
...
+public export
+Show (m (Maybe a)) => Show (MaybeT m a) where
+  showPrec d (MkMaybeT x) = showCon d "MkMaybeT" $ showArg x
+
+||| Corresponds to the Semigroup instance of Maybe
+||| 
+||| Note: This could also be implemented with only an Applicative
+||| prerequisite: `MkMaybeT x <+> MkMaybeT y = MkMaybeT $ [| x <+> y |]`
+||| However, the monadic version is more efficient for long-running effects,
+||| only evaluating the second argument if the first returns `Nothing`.
```
clayrat, line 93: I guess [selective functors](https://github.com/clayrat/idris-selective) would be handy here.

stefan-hoeck: Thanks for the interesting read @clayrat . I didn't know about selective functors.

edwinb:
Looks like the bugs have been squashed, so I'll merge this now. Thanks for the fix and the new contribution.

PR #914, opened 2021-01-11
Eq, Ord, and Show instances for LazyList

stefan-hoeck:
Addes Eq, Ord, and Show implementations for LazyList based on the ones for List.


edwinb:
Thanks for the contribution!

PR #916, opened 2021-01-12
Name quotes are suggested to use single brace instead of double

buzden:
An extremist's solution of problems mentioned in #838 (actually, this is a further continuation of #838).

Well, I'm sure there was a reason to make name quotes to have double braces (when the other quotes have single-character syntax). But I don't see this reason since only names are allowed between those braces.

But this double-brace syntax clashes with implicit and named parameters syntax in case when actual parameter's expression ends with an implicit or named parameter or auto-parameter bind (i.e. `@{..}`). I showed in #838 an example like
```idris
h : Nat -> Nat
h c = g {b = f {a=c}}
```

So what are the arguments to not doing this?

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -398,7 +398,7 @@ mutual
            pure (PDotted (boundToFC fname b) b.val)
     <|> do b <- bounds (symbol "`(" *> expr pdef fname indents <* symbol ")")
            pure (PQuote (boundToFC fname b) b.val)
-    <|> do b <- bounds (symbol "`{{" *> name <* symbol "}}")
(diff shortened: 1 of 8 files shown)


buzden:
"Build from previous version" action is falling because of the broken 0.2.2 release tar.

edwinb:
I think the only reason for using double braces was because that's what Idris 1 did. Reflection is already incompatible with Idris 1, so a change is fine.

This PR is an unfortunate casualty of me not being able to pay attention enough to Idris lately. Sorry about that. I will have a quick go at fixing up the conflicts.

edwinb:
Reopened as #1715

PR #917, opened 2021-01-12
Benchmark Suite

jmd28:
This is the beginnings of benchmark suite I wrote to aid a research project undertaken by myself and @edwinb on optimistation of the refc backend.

We're merging it into the main repo so that anyone will be able to contribute their own benchmark tests (see README.md details!)


edwinb:
Thanks! The previous version CI failure is my fault, I'm on the case...

buzden:
> Thanks! The previous version CI failure is my fault, I'm on the case...

The previous PR is also affected.

edwinb:
Thanks for this, merged now, I think this'll turn out to be useful. Hopefully it'll grow over time...

What would be especially useful would be "real" programs (or, at least, components of real programs) which will naturally give us more realistic results than anything we contrive.

PR #918, opened 2021-01-12
Don't add implicits after non-given explicits.

### LINKED ISSUE #67: Implicits in Interface Functions cause Typecheck Error
<a href="https://github.com/fabianhjr"></img></a> **Issue by [fabianhjr](https://github.com/fabianhjr)**
_Wednesday Apr 29, 2020 at 03:34 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/336_

----

# Steps to Reproduce

```
module Test

interface Foo a where
  bar : a -> {auto ok: ()} -> a 
```

# Expected Behavior

Typecheck (Like Idris1)

# Observed Behavior

```
1/1: Building Test (Test.idr)
Test.idr:4:3--5:1:While processing left hand side of bar at Test.idr:4:3--5:1:
ok is not a valid implicit argument in bar
```

Edit: maybe related to #8, however I couldn't change it to a constraint as a workaround since I was restricting the domain of a division to nonzero elements.

mr-infty:
The function `TTImp.TTImp.implicitsAs` is changed so that implicit
arguments that come after explicit ones that are not given are no longer
added. This fixes #67.

The reasoning behind this is that the former behaviour---of adding all
missing implicits---makes `implicitsAs` return malformed terms in
certain cases where the underlying function type has implicit arguments
that come after explicit ones. For example, the interface declaration

    interface Foo a where
      bar : a -> {auto ok : ()} -> a

and the function declaration

    foo : Void -> {auto ok : ()} -> Void
    foo = ?foo_hole

result in the errors

    ok is not a valid argument in bar

and

    ok is not a valid argument in foo

respectively, because the auto-implicit is added during elaboration. For
instance, in the second example the left hand side of the pattern
matching declaration

    foo = ?foo_hole

gets replaced by `foo {ok=ok}` (in `implicitsAs`), which is an invalid
function call (according to `TTImp.Elab.App.checkAppWith`).

The changes made are as follows.

The functions `TTImp.TTImp.setAs` and `TTImp.TTImp.findImps` now both
take a new argument `es` of type `List (Maybe Name)`.

In `setAs`, this variable is used to record the explicit arguments which
are explicitly given in the left hand side of the pattern matching
declaration.

This variable is then passed onto `findImps`, which uses it to determine
when it encounters an explicit variable not given and stop searching for
implicits.

### DIFF
--- a/src/TTImp/TTImp.idr
+++ b/src/TTImp/TTImp.idr
@@ -496,26 +496,26 @@ findImplicits tm = []
 export
 implicitsAs : {auto c : Ref Ctxt Defs} ->
               Defs -> List Name -> RawImp -> Core RawImp
-implicitsAs defs ns tm = setAs (map Just (ns ++ map UN (findIBinds tm))) tm
+implicitsAs defs ns tm = setAs (map Just (ns ++ map UN (findIBinds tm))) [] tm
   where
-    setAs : List (Maybe Name) -> RawImp -> Core RawImp
-    setAs is (IApp loc f a)
-        = do f' <- setAs is f
+    setAs : List (Maybe Name) -> List (Maybe Name) -> RawImp -> Core RawImp
+    setAs is es (IApp loc f a)
+        = do f' <- setAs is (Nothing :: es) f
              pure $ IApp loc f' a
-    setAs is (IAutoApp loc f a)
-        = do f' <- setAs (Nothing :: is) f
+    setAs is es (IAutoApp loc f a)
+        = do f' <- setAs (Nothing :: is) es f
              pure $ IAutoApp loc f' a
-    setAs is (INamedApp loc f n a)
-        = do f' <- setAs (Just n :: is) f
+    setAs is es (INamedApp loc f n a)
+        = do f' <- setAs (Just n :: is) (Just n :: es) f
              pure $ INamedApp loc f' n a
-    setAs is (IWithApp loc f a)
-        = do f' <- setAs is f
+    setAs is es (IWithApp loc f a)
+        = do f' <- setAs is es f
              pure $ IWithApp loc f' a
-    setAs is (IVar loc n)
+    setAs is es (IVar loc n)
         = case !(lookupTyExact n (gamma defs)) of
                Nothing => pure $ IVar loc n
                Just ty => pure $ impAs loc
-                                    !(findImps is !(nf defs [] ty)) (IVar loc n)
+                                    !(findImps is es !(nf defs [] ty)) (IVar loc n)
       where
         -- If there's an @{c} in the list of given implicits, that's the next
         -- autoimplicit, so don't rewrite the LHS and update the list of given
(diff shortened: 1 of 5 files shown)


Russoul:
Just one note. It currently doesn't respect the "bind-all-explicits" pattern (it's new):
```
foo : a -> b -> c -> {auto x : a} -> a -- fails.
foo {} = x
```

`{}` binds however many explicits there is in the signature. So `x` should be implicitly bound on the LHS.

Just an FYI: `{}` are parsed as `UN "_"`. There is a brief description of it in the docs.

mr-infty:
> Just one note. It currently doesn't respect the "bind-all-explicits" pattern (it's new):
> 
> ```
> foo : a -> b -> c -> {auto x : a} -> a -- fails.
> foo {} = x
> ```
> 
> `{}` binds however many explicits there is in the signature. So `x` should be implicitly bound on the LHS.
> 
> Just an FYI: `{}` are parsed as `UN "_"`. There is a brief description of it in the docs.

Thanks for the heads up! I've read the documentation, but just to make sure I understand it correctly: is the underscore notation supposed to skip *all* explicit arguments (as suggested by the example `dontCare`) or only all *named* explicit arguments (as suggested by the last sentence before the examples)?

Russoul:
It always skips both named and unnamed explicits regardless the form: `{}` or `{x, y, z, ... , _}`

Russoul:
Yep, that last phrase in the docs before the examples is misleading.

mr-infty:
@Russoul 
I've updated the code to make it compatible with the new syntax. I've also added your counterexample as a test-case in the regression test.

gallais:
> Just an FYI: `{}` are parsed as `UN "_"`.

This seems brittle. Is there a deep reason for not adding a new constructor
to the name datatype for `{}`?

Russoul:
I don't think there is a reason at all. Maybe only that `Name` is overused throughout Idris 2. Even core TT (`Term`) is indexed over `List Name`. While in fact, only variables, i.e `UN` and `MN`, are valid in `Term`. 
And this is what I am working on right now: replacing `Name` to a more precise datatype when appropriate in Idris codebase.
I'll remember to throw away `UN _` and have a constructor in `Name` for the bind-all-explicits pattern.

src/TTImp/TTImp.idr
```diff
@@ -496,26 +496,26 @@ findImplicits tm = []
 export
 implicitsAs : {auto c : Ref Ctxt Defs} ->
               Defs -> List Name -> RawImp -> Core RawImp
-implicitsAs defs ns tm = setAs (map Just (ns ++ map UN (findIBinds tm))) tm
+implicitsAs defs ns tm = setAs (map Just (ns ++ map UN (findIBinds tm))) [] tm
   where
-    setAs : List (Maybe Name) -> RawImp -> Core RawImp
-    setAs is (IApp loc f a)
-        = do f' <- setAs is f
+    setAs : List (Maybe Name) -> List (Maybe Name) -> RawImp -> Core RawImp
```
gallais, line 501: Can we get a comment documenting the different roles of these two
`List (Maybe Name)` arguments?

src/TTImp/TTImp.idr
```diff
@@ -529,20 +529,27 @@ implicitsAs defs ns tm = setAs (map Just (ns ++ map UN (findIBinds tm))) tm
                          pure (x :: ns')
         updateNs n [] = Nothing
 
-        findImps : List (Maybe Name) -> NF [] -> Core (List (Name, PiInfo RawImp))
-        findImps ns (NBind fc x (Pi _ _ Explicit _) sc)
-            = findImps ns !(sc defs (toClosure defaultOpts [] (Erased fc False)))
+        findImps : List (Maybe Name) -> List (Maybe Name) -> NF [] -> Core (List (Name, PiInfo RawImp))
```
gallais, line 532: Same comment here: a bit of documentation will make it easier for
future readers to understand what is going on.

tests/idris2/reg036/Test.idr
```diff
@@ -0,0 +1,10 @@
+module Test
+
+interface Foo a where
+  bar : a -> {auto ok: ()} -> a
+
+foo : Void -> {auto ok: ()} -> Void
+foo = ?foo_hole
+
+baz : a -> b -> c -> {auto x : a} -> a
+baz {} = x
```
gallais, line 10: I think I have found a bug involving shadowing:

```idris
foo : {a : Type} -> Void -> {a : a} -> Void -> {auto ok : ()} -> Void
foo _ {a = c} = ?foo_hole
-- ^   ^--- This should bind the second implicit `a`,
-- |        the one that comes after the first explicit argument
-- `-- there should be an implicitly bound `a : Type` here
```

gives me a hole with `c : Type` rather than `c : a`.

gallais (commented):
I think I have found a bug involving shadowing.

gallais:
I guess the bug I have found precedes this PR because `master` currently fails
in a similar way with:

```idris
foo : {a : Type} -> Void -> {a : a} -> Void
foo _ {a = c} = ?foo_hole
```

So I'd be willing to merge this PR with the only change being the added documentation.

src/TTImp/TTImp.idr
mr-infty: Sure, I'll add a comment in order to make it clearer to the reader!

tests/idris2/reg036/Test.idr
mr-infty: I'm surprised to learn that this is legal Idris code. Does Idris have separate namespaces for term and type variables?

mr-infty:
@gallais I've added some documentation. I hope, it makes the meaning of the code clearer.

tests/idris2/reg036/Test.idr
gallais: Why shouldn't it be legal? It's not a namespace issue, it's classic
shadowing in a telescope: when checking the type `a` in the telescope
`pre -> (x : a) -> post`, the variables bound in `pre` are in scope
(but not `x` itself as no binding is self-referential) and when checking
`post`, the variables `pre, x` are in scope (thus any `x` in `pre` is
shadowed by the most-recently bound one).

gallais:
Thank you, this is very good. It has even clarified some of the things
I had not realised after carefully reading this PR. Looks to me like
the new application syntax is quite ambiguous. I'll open another issue.

tests/idris2/reg036/Test.idr
mr-infty: Ah, I see. Thanks for clarifying that for me!

Issue #893, opened 2020-12-30
`with`-rule misses the `proof` keyword 


clayrat:
# Steps to Reproduce

```
thenCompare : Ordering -> Lazy Ordering -> Ordering
thenCompare LT y = LT
thenCompare EQ y = y
thenCompare GT y = GT

foo : (a, b : Nat) -> thenCompare (a `compare` b) o = a `compare` b
foo a b with (a `compare` b)
  foo a b | LT = Refl
  foo a b | EQ = ?hole
  foo a b | GT = Refl
```

# Expected Behavior

We should be able to add `proof prf` to the `with` rule and finish the hole with `sym prf`, as in Idris1

# Observed Behavior

We're forced to fall back on matching directly on `a` and `b`.

buzden:
You could use a `with (@@ (a `compare` b))` from `Syntax.WithProof` in `contrib` as an ersatz.

edwinb:
Given `Syntax.WithProof` as @buzden suggests, I've so far not been enthusiastic about implementing more syntax, since it's a feature that isn't used much and can be done in code already.

gallais:
`Syntax.WithProof` is not as powerful though: `compare a b` will not be
abstracted over in the goal type. To recover the full power of the inspect
idiom you need to nest 2 withs. Or use a multiwith, something I have been
trying to implement for a while now but the well-scoped syntax keeps kicking
my butt.

clayrat:
I've actually realized that the initial example is bogus, let's say I have something like the following instead:

```
foo : (a, b : Nat) -> Bool
foo  Z    b = False
foo (S _) b = False

notFoo : (a, b : Nat) -> Not (foo a b = True)
notFoo  Z    _ = uninhabited
notFoo (S _) _ = uninhabited

bar : (a, b : Nat) -> (foo a b) && c = foo a b
```

(where instead of `Nat`s and `Bool`s I have inductive types with multiple constructors, making it unpleasant to repeatedly match on all possible combinations in proofs). Now with `@@` I can finish `bar` like this:

```
bar a b with (@@ (foo a b))
  bar a b | (True  ** ab) = absurd $ notFoo a b ab
  bar a b | (False ** ab) = rewrite ab in Refl
```

I'm not sure how to regain the power of `with proof` syntax by nested matching however, so that one could keep the first line but drop the `rewrite` in the last line. The problem seems to be that subsequent matching acts both on the goal and `ab`, whereas the original syntax kept `ab` intact.

[...]

PR #919, opened 2021-01-12
add Inspect idiom

clayrat:
As suggested in https://github.com/idris-lang/Idris2/issues/893


gallais:
I reused the names in agda's stdlib in the original snippet but
they're maybe not the best? Here are propositions:

`Reveal f` is the the relation corresponding to the graph of `f`
so we could maybe call the record `Graph` instead?

As for `inspect`, I could see either `remember` (because that is
what we are trying to do: remember the origin of `f x`) or some
sort of `$`-like infix syntax because it is just a fancy apply.

clayrat:
I do like `remember`! However `Graph` is IMO too overloaded. We need a more "technical" word, maybe `FnGraph` or `Section`, [as in topology](https://en.wikipedia.org/wiki/Section_(fiber_bundle))?

gallais:
`Image` would also work as `Image f x` is the predicate that only
accepts the `y`s that are such that `f x === y`.

clayrat:
`Image` is also a bit generic, could be confusing out of context. I'm thinking to go with `FnGraph`, wdyt?

gallais:
If we are worried about things being confusing out of context,
we may put it in `Data.Function` and that should be unambiguous.

clayrat:
Ok, let's go with `Data.Fun.Graph` then.

PR #920, opened 2021-01-13
fixed bug in Foldable of EitherT

stefan-hoeck:
This fixes an issue in the Foldable implementation of EitherT that came up during the writing of MaybeT (#909).


edwinb:
Thanks for the fix!

PR #922, opened 2021-01-13
Special variants of `DPair` as records

buzden:


gallais:
Not sure what the purpose of this is. We already have `fst` and `snd`
projections for these and here we are introducing other names for the
same functions?

buzden:
There are two reasons none of which is strong:
- the current versions of `fst`s and `snd`s of `Subset` and `Exists` are linear, `.value` and `.prf` functions from the record definition are not;
- with record we can have a nice `x.value` and `x.prf` syntax for getting the parts of these pairs (as we have now with `.fst` and `.snd` for usual `DPair`s).

Well, from rational point of view, maybe it's better to change these definitions with `fst` and `snd` fields and make them non-linear (to have `Subset` and `Exists` definitions be consistent with `DPair`).

gallais:
> UPD. For some unknown reason, I (when experimenting) didn't manage to have a field named fst (however, a field named snd typechecks easily).

This works for me:

```idris
record Exists {0 type : _} this where                                                                                           
    constructor Evidence
    0 fst : type
    snd : this fst
```

buzden:
> This works for me

Yes, that was my bad.

I pushed this change. So, what do you think about just making `Exists` and `Subset` be declared purely as records, like `DPair`?

gallais:
Given that #928 reintroduces the namespaces you are removing here,
could we just keep them all along? That would also guarantee that
we keep `Exists.Exists` as a valid name for `Exists` as it's currently
defined in that namespace but you have peeled that outer layer off
in this PR.

buzden:
Yes, it makes sense. I'll do this.

PR #924, opened 2021-01-13
Implement Racket Futures Support

fabianhjr:
Using Racket's Futures: https://docs.racket-lang.org/reference/futures.html


fabianhjr:
Hi @gallais, this adds support for futures on the Racket CG, it is needed for #827.

Issue #943, opened 2021-01-17
ElabReflection error: can't reify as Namespace


bamboo:
# Steps to Reproduce

Compile and run the following Idris program:

```idris
module Main

import Language.Reflection

%language ElabReflection

%runElab declare [
  INamespace EmptyFC ["Foo"] [
    IClaim EmptyFC MW Export [] (MkTy EmptyFC (UN "foo") (IPrimVal EmptyFC StringType)),
    IDef EmptyFC (UN "foo") [
      PatClause EmptyFC (IVar EmptyFC (UN "foo")) (IPrimVal EmptyFC (Str "42"))
    ]
  ]
]

main : IO ()
main = putStrLn Foo.foo
```

# Expected Behavior

Compilation should succeed and the program should print 42.

# Observed Behavior

Program fails to compile with:
```
Can't reify as Namespace

/code/src/Main.idr:8:22--8:29
   |
 8 |   INamespace EmptyFC ["Foo"] [
   |                      ^^^^^^^
```

Curiously, if the program is changed to forcefully inject a `Namespace` at the expected location, it works:

```idris
  INamespace EmptyFC (believe_me (MkNS ["Foo"])) [
 ```

PR #944, opened 2021-01-17
Fix `Language.Reflection` definition of `INamespace

bamboo:
See #943

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -44,6 +44,7 @@ Nicolas Biri
 Niklas Larsson
 Ohad Kammar
 Peter Hajdu
+Rodrigo Oliveira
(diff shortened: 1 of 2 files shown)


michaelmesser:
I wonder if there is a simple way to have the type checker catch these problems at compile time.  I also wonder if `RawImp` and `TTImp` and the other types can be merged into one.

gallais:
> I wonder if there is a simple way to have the type checker catch these problems at compile time

I did get type errors at the time but I chose on purpose to leave the user-facing
datatype as `List String` and fix the plumbing around it, thinking that this would
make life easier for client code by not breaking it.

Clearly I was wrong. :disappointed: 

Thanks for the report & the fix @bamboo

Issue #945, opened 2021-01-17
Confusing error message for "Can't solve constraint"


farnoy:
Hi, I have two problems with the code posted below. The first one is that I'm not sure where the constraint is coming from, the second one with the compiler not showing me how it arrived at this constraint.

The problem I'm solving is to create a data type that associates natural numbers with the power of two needed to represent them. You can think of this as "how many bits to store n".

# Steps to Reproduce

```idris
module Repro

import Data.Nat

%default total

pow2 : Nat -> Nat
pow2 n = power 2 n

plusOneLTEPlusSelf : (a : Nat) -> GT a 0 -> LTE (S a) (a + a)
plusOneLTEPlusSelf 0 _ impossible
plusOneLTEPlusSelf (S n) _ =
  let y = lteAddRight (S n) {m=n} in
    rewrite sym (plusSuccRightSucc n n) in
      LTESucc y

greaterPowerStillGreater : (a, b : Nat) -> GT a 0 -> a = pow2 b -> LTE (S a) (pow2 (S b))
greaterPowerStillGreater a b gt0 prf =
  rewrite sym prf in
    rewrite plusZeroRightNeutral a in
      plusOneLTEPlusSelf a gt0

data NextPowerOfTwo : (n, p : Nat) -> { auto proof : LTE n (pow2 p) } -> Type where
  NextPowZero : NextPowerOfTwo 0 0
  NextPowSucc : NextPowerOfTwo a b {proof=renamed}
             -> { auto p : LTE (S a) (pow2 b) }
             -> NextPowerOfTwo (S a) b
  NextPowSuccPow : NextPowerOfTwo a b {proof=renamed}
                -> { auto p : a = pow2 b }
                -> { auto gt0 : GT a 0 }
                -> NextPowerOfTwo (S a) (S b) {proof=greaterPowerStillGreater a b gt0 p}

test0 : NextPowerOfTwo 0 0
test0 = NextPowZero

test1 : NextPowerOfTwo 1 0
test1 = NextPowSucc test0

test2 : NextPowerOfTwo 2 1
test2 = NextPowSuccPow test1

test3 : { auto k : NextPowerOfTwo 2 1 } -> NextPowerOfTwo 3 2
test3 {k} = NextPowSuccPow k
```

# Expected Behavior

I expected this to compile & work and for the compiler to be able to proof search my data structure to provide the `auto` implicits.

# Observed Behavior

Errors whenever I try to use `NextPowSuccPow`:
```
[...]

PR #948, opened 2021-01-17
libs: mark Data.Nat.lteAddRight as public export

farnoy:
I encountered a lack of this export in idris-lang/idris2#945

We might want to export more of these helpers as `public export`


gallais:
`LTE` is proof irrelevant so it should not really matter which proof is generated.
Then again this is exactly the implementation one would expect so maybe it's not
that big of a deal to expose it.

PR #949, opened 2021-01-17
[ unify, fix #589 ] Postpone when instantiating if type is unknown

### LINKED ISSUE #589: Unification failure when using functions in dependent pairs
# Steps to Reproduce

Type check the following snippets:

```idris
ac : forall r. ((x : a) -> (y : b ** r x y)) -> (f : a -> b ** (x : a) -> r x (f x))
ac g = (\x => fst (g x) ** \x => snd (g x))
```

```idris
ac : forall r. ((x : a) -> (y : b ** r x y)) -> (f : a -> b ** (x : a) -> r x (f x))
ac g = (\x => fst (g x) **
        \x => let sgx = snd (g x)
              in sgx)
```

```idris
ac : forall r. ((x : a) -> (y : b ** r x y)) -> (f : a -> b ** (x : a) -> r x (f x))
ac g = (\x => fst (g x) **
        \x => let sgx : r x (fst (g x))
                  sgx = snd (g x)
              in sgx)
```

# Expected Behavior

All three snippets should type check successfully.

# Observed Behavior

The first two do not type check successfully and produce the following error:

```
While processing right hand side of ac:
When unifying r x (fst (g x)) and r x (?fst x)
Can't make solution for Main.{fst:464} at ?Main.{a:462}_[b[3], a[2], r[1], g[0]] at:
```

The third snippet does type check successfully.

Russoul:
There are cases when the type of a meta variable in process is yet to be discovered while the meta is about to be instantiated.

This PR adjusts the process of instantiating metas to be able to postpone if the type is not properly expanded.

Three example cases, previously failing, are shown in the test file.

There is one flaw that I suspect to be possible, when in some cases even postponing doesn't help to find a unifier.
I don't know a way to predict whether a delay is needed or the process is conclusively stuck. So extra steps may be carried out
 for nothing. But still better than having false negatives.


edwinb:
I'm making a few changes to this to bring it up to date. I'll need to close this then open a new PR with a couple of modifications. Thanks for the fix, and sorry to leave it lingering so long!

PR #953, opened 2021-01-19
Remove leaked result file

fabianhjr:
Introduced as:

```
7a7504c95 (Guilherme Silva 2021-01-15 14:20:52 -0300 1) /nix/store/nw0mm1z2afpxv5rgf6w3apxc8pxwgd6h-idris2-0.3.0
```


buzden:
Maybe, it should be added to the `.gitignore`?

PR #956, opened 2021-01-19
[ fix #954 ] Allow RF names for interface methods

### LINKED ISSUE #954: Dot syntax functions inside interface implementations do not work
# Steps to Reproduce
```idris
interface I i where
    (.dot) : i -> Int

data D = MkD

I D where
    (.dot) _ = 0
```

# Expected Behavior
Compiles. 
`(MkD).dot` is `0`.

# Observed Behavior
```
Error: .dot is not a method of I

test.idr:7:5--7:11
   |
 7 |     (.dot) _ = 0
   |     ^
```

---
# Additional Cases

<details>
<summary> Click to view </summary>

The following **does** compile and results with a hole, as expected:
```idris
interface I i where
    (.dot) : i -> Int

data D = MkD

I D where
```
```
Main> (MkD).dot
.dot MkD
```

---

Default implementation fails to compile as well:
```idris
interface I i where
    (.dot) : i -> Int
    (.dot) _ = 0

data D = MkD

I D where
```
```
Error: .dot is not a method of I

test.idr:7:1--7:10
   |
 7 | I D where
   | ^^^^^^^^^
```

But only when the interface is implemented:
```idris
interface I i where
    (.dot) : i -> Int
    (.dot) _ = 0

data D = MkD
```
```
Main> :t (.dot)
Main.(.dot) : I i => i -> Int
Main> (MkD).dot
Error: Can't find an implementation for I D.

(interactive):1:1--1:10
   |
 1 | (MkD).dot
   | ^^^^^^^^^
```

---

This **does** compile and produces the required behaviour, even though it looks wrong (found by @Russoul) :

```idris
interface I i where
    (.dot) : i -> Int

data D = MkD

I D where
    dot _ = 0
```
```
Main> (MkD).dot
0
```

But the same workaround does not work with default implementations:

```idris
interface I i where
    (.dot) : i -> Int
    dot _ = 0

data D = MkD

I D where
```
```
Error: No method named dot in interface I

test.idr:3:5--3:14
   |
 3 |     dot _ = 0
   |     ^^^^^^^^^
```
</details>

gallais:


tests/idris2/interface022/run
```diff
@@ -0,0 +1,3 @@
+$1 --no-color --console-width 0 DotMethod.idr < input
```
ska80, line 1: `--no-banner` should be added to hide the banner that changes with each commit.

PR #960, opened 2021-01-20
Transformer overhaul

stefan-hoeck:
This PR restructures the existing monad transformer modules in base for reasons given below. It also adds two new transformer interfaces and data types similar to the ones found in Haskell packages [transformers](https://hackage.haskell.org/package/transformers-0.5.6.2) and [mtl](https://hackage.haskell.org/package/mtl). The new interfaces (`MonadWriter` and `MonadRWS`) and data types (`WriterT` and `RWST`) where also in base in Idris1.

In addition, the module layout was redesigned in the following way:

- Each transformer interface like `MonadError` and `MondState` was moved to its own interface module to avoid cyclic dependencies. This corresponds to the design of transformers and mtl.

- Each transformer data type with a corresponding transformer interface was moved to a separate module, while the original module publicly exports both, the interface module and the data module. Again, this corresponds to the design of mtl and transformers.

- Every transformer interface like `MonadError` and `MonadState` comes with an implementation for every core transformer data type (`MaybeT`, `EitherT`, `ReaderT`, `RWST`, `StateT`, and `WriterT`).

- All functions where annotated with `public export` and `%inline`. In my opinion, these core functions should be available in proofs and the inlinig annotations where taken from the corresponding Haskell implementations. I am no expert when it comes to inlining stuff, so people might disagree on this. In that case, I'll undo the inlining annotations. I also annotated interface implementations with `%inline`, assuming that this would inline the corresponding function implementation.

- The `MonadReader` implementation for `WriterT` differs from the one in mtl. See notes in the source code for the reasoning behind this decision.

- Only CPS-versions of `WriterT` and `RWST` are included, since according to the mtl-documentation only these seem to be able to run in constant space. I did not verify, whether this holds in Idris.

- The Idris version of StateT has the types the other way round compared to the Haskell version:

     Idris2:  `s -> (s,a)`
     Haskell: `s -> (a,s)`
  
  I don't know the reason behind this design decision, but would prefer the Idris2 version to be the same as in Haskell to make things less confusing for Haskell developers coming to Idris2. However, for the time being I left this untouched in order not to break existing code. I used the same order in `MonadState` but am happy to change this, if people agree on going back to the version found in Haskell.


libs/base/Control/Monad/RWS/CPS.idr
```diff
@@ -0,0 +1,133 @@
+||| Note: The difference to a 'stricht' RWST implementation is
+||| that accumulation of values does not happen in the
+||| Applicative and Monad instances but when invoking `Writer`-specific
+||| functions like `writer` or `listen`.
```
gallais, line 4: typos: "stricht" -> "strict"

libs/base/Control/Monad/Writer/CPS.idr
```diff
@@ -0,0 +1,107 @@
+||| Note: The difference to a 'stricht' Writer implementation is
```
gallais, line 1: same typo

libs/base/Control/Monad/Writer/CPS.idr
```diff
@@ -0,0 +1,107 @@
...
+module Control.Monad.Writer.CPS
+
+import Control.Monad.Identity
+import Control.Monad.Trans
+
+||| A writer monad parameterized by:
+|||
+|||   @w the output to accumulate.
+|||
+|||   m@ The inner monad.
```
gallais, line 14: Did you mean `@m`?

gallais (approved):

libs/base/Control/Monad/RWS/CPS.idr
stefan-hoeck: Fixed, thanks. I'm Swiss. We are quite fond of our 'ch's. :-)

libs/base/Control/Monad/Writer/CPS.idr
stefan-hoeck: Fixed.

libs/base/Control/Monad/Writer/CPS.idr
stefan-hoeck: Fixed.

PR #967, opened 2021-01-20
Refactor bootstrap and CI action to speedup CI

fabianhjr:
There seemed to be some duplication of building and testing.


fabianhjr:
CI times reduced substantially, except for racket that takes a while.

.github/workflows/ci-macos.yml
```diff
@@ -24,9 +24,9 @@ jobs:
           brew install chezscheme
           brew install coreutils
           echo "::add-path::$HOME/.idris2/bin"
-      - name: Build and test Idris 2 from bootstrap
+      - name: Build Idris 2 from bootstrap
         run: make bootstrap && make install
         shell: bash
-      - name: Build and test self-hosted
-        run: make clean && make all && make test INTERACTIVE=''
+      - name: Test self-hosted
```
gallais, line 30: Self-hosted means compiled using itself as the compiler. By removing this step
you are testing the executable compiled using the bootstrap, not using itself.

gallais (commented):
I haven't looked further down the PR because I am assuming that all of these
changes stem from the same confusion about compiled vs. self-hosted?

.github/workflows/ci-macos.yml
fabianhjr: On the makefile it was already doing that.

stage2 does a clean and rebuild.

fabianhjr: Ah wait, NVM.

So:

Bootstrap -> HEAD -> HEAD so it is important to check that HEAD can build HEAD.

Makefile
```diff
@@ -12,14 +12,13 @@ MAJOR=0
...
-    # inside a git repo
-    ifneq ($(shell git describe --exact-match --tags >/dev/null 2>&1; echo $$?), 0)
-        # not tagged as a released version, so add sha1 of this build in between releases
-        GIT_SHA1 := $(shell git rev-parse --short=9 HEAD)
-    endif
+	# inside a git repo
+	ifneq ($(shell git describe --exact-match --tags >/dev/null 2>&1; echo $$?), 0)
+		# not tagged as a released version, so add sha1 of this build in between releases
+		GIT_SHA1 := $(shell git rev-parse --short=9 HEAD)
+	endif
```
fabianhjr, line 21: This were using space instead of tabs in the Makefile

.github/workflows/ci-macos.yml
fabianhjr: Ok @gallais, amended and added back as a separate stage to be able to compare time difference on self-host compile and tests.

bootstrap.sh
```diff
@@ -1,52 +0,0 @@
...
-mkdir -p ../build/exec
-mkdir -p ../build/exec/idris2_app
-install idris2-boot ../build/exec/idris2
-install idris2_app/* ../build/exec/idris2_app
-
-cd ..
-
-# TODO: Unify with bootstrap-rkt
-PREFIX=${PWD}/bootstrap
-
```
fabianhjr, line 32: Addressed this TODO by splitting into 2 stages, stage2 being shared.

fabianhjr:
Screenshot from 2021-01-21 19-30-02

Jobs got split and now produce artifacts.

gallais:
I did not know github actions had stages. Am I understanding correctly that
the full racket workflow (build-bootstrap-racket + self-host-racket) takes
50min? That seems longer than the current one (30m for the latest build on
master).

buzden:
> That seems longer than the current one

It seems that previously a CI action for Racket didn't try to build a self-hosted compiler but now this stage is added, that's why it takes more time.

bootstrap-stage2.sh
```diff
@@ -0,0 +1,40 @@
...
+    IDRIS_PREFIX=$PREFIX
+    SEP=":"
+fi
+
+# Default to chez CG
+if [ "$1" = "" ]; then
+    IDRIS2_CG="chez"
+else
+    IDRIS2_CG=$1
+fi
```
gallais, line 21: How about `IDRIS2_CG ?= "chez"`? i.e. only override the variable if it's not set rather than
relying on an argument being passed to the script.

fabianhjr:
Yes, racket wasn't being self-hosted. Additionally I tried a bit to add caches to the bootstrap phase but gave up since it wasn't being picked up for some reason that I couldn't figure out.

gallais:
Is this ready? It looks ready.

fabianhjr:
Yeah it is

PR #970, opened 2021-01-21
adds command for self-hosting when using racket

rjregenold:
When trying to perform a self-hosting build, it will fail if you're not using `scheme`. This PR updates the docs with a command that succeeds when using `racket`.


gallais:
Thanks!

PR #972, opened 2021-01-21
Port Data.List.unzip from Idris1

ska80:


libs/base/Data/List.idr
```diff
@@ -358,6 +358,20 @@ export
 zip3 : List a -> List b -> List c -> List (a, b, c)
 zip3 = zipWith3 \x, y, z => (x, y, z)
 
+||| Split a list of pairs into two lists
+export
+unzip : List (a, b) -> (List a, List b)
```
gallais, line 363: Can we have `unzipWith : (a -> (b, c)) -> List a -> (List b, List c)` too? :pleading_face: 
I am always annoyed that we have `zipWith`, `zip`, `unzip` but no `unzipWith`.

stefan-hoeck: Also, if we have `zip3`, we should probably also have `unzip3` and `unzipWith3`.

ska80: > Can we have `unzipWith : (a -> (b, c)) -> List a -> (List b, List c)` too? 🥺
> I am always annoyed that we have `zipWith`, `zip`, `unzip` but no `unzipWith`.

Good idea.  I will add them.

stefan-hoeck: Just one more note: If we add `unzipWith`, we should have this also for `Vect`, `Stream`, and `List1` (probably also for `Maybe` and `Either e`) to be consistent across different (un)zippable data types. I wonder, whether we actually should have an interface for this (like MonadZip in Haskell). Probably this should not all go in this PR, though?

ska80: Added `zipWith` and `zipWith3`

libs/base/Data/List.idr
```diff
@@ -358,19 +358,31 @@ export
 zip3 : List a -> List b -> List c -> List (a, b, c)
 zip3 = zipWith3 \x, y, z => (x, y, z)
 
+||| Split a list by applying a function into two lists
+export
+unzipWith : (a -> (b, c)) -> List a -> (List b, List c)
+unzipWith f [] = ([], [])
+unzipWith f (x :: xs) with (unzipWith f xs)
+  unzipWith f (x :: xs) | (lefts, rights) = (fst (f x) :: lefts, snd (f x) :: rights)
```
gallais, line 366: I don't think we need to use a `with` (that means generating an auxiliary definition)
and we should probably avoid duplicating the `f x` call. So I'd write something like
this instead (and something similar for `unzipWith3`):

```suggestion
unzipWith f (x :: xs)
  = let (b, c) = f x
        (bs, cs) = unzipWith f xs
    in (b :: bs, c :: cs)
```

PR #974, opened 2021-01-21
added System.Console.GetOpt to contrib

stefan-hoeck:
This adds an Idris2 port of the corresponding Haskell module [GetOpt](http://hackage.haskell.org/package/base-4.14.1.0/docs/System-Console-GetOpt.html) as discussed in #935.

Note: Since this uses its own implementation of `unzip3` internally, we probably should wait with merging this until after #972 is merged.


stefan-hoeck:
`unzip3` has been removed. (And sorry for hearting myself. That was unintentional.)

PR #979, opened 2021-01-22
Remove Control.Monad.Trans.Either from contrib

stefan-hoeck:
#554 introduced its own `EitherT` implementation to contrib. This PR removes `Control.Monad.Trans.Either` from contrib and makes `Control.Validation` use the `EitherT` implementation from base.


andrevidela:
Ah excellent! I didn't even realise the wires were getting crossed, thanks!

PR #982, opened 2021-01-22
[ new ] Order relations on Int

gallais:


libs/contrib/Data/Int/Order.idr
```diff
@@ -0,0 +1,354 @@
...
+-- Our solution in this file is to use unsafe primitives to manufacture such
+-- proofs. This means we are going to essentially postulate some *conditional*
+-- results. We do not want such conditional results to reduce to canonical
+-- forms too eagerly.
+
+-- Indeed the statement `GT 0 1 -> EQ 0 1` should be provable because 0 is not
+-- greater than 1. But its proof should not reduce to a constant function
+-- returning the value `Refl` because it is not true that `0` and `1` can be
+-- unified. If the proof were to behave this way, we could, in an absurd context,
+-- coerce values from any type to any other and cause segmentation faults.
```
stefan-hoeck, line 35: This is really cool, especially since it is so well documented!

It would be great to have a (hint of an) example, how exactly a reducing proof of `GT 0 1 -> EQ 0 1` could be abused.

Anyway, thank you very much for this!

gallais: This is not specifically about `GT 0 1 -> EQ 0 1` but the principle is the same.
If you look at my [original file](https://github.com/gallais/pearl-binary-search/blob/main/src/Data/Int/Order.idr#L26) you can see that I was re-exporting `MkLT`.
I had to tighten this down because of a similar issue.

If you are allowed to pattern-match on `MkLT`, you could run into the 
theoretically impossible `crash` used in `irrefl` e.g.

```idris
argh : (p : Void) -> (lie : LT a a) -> irrefl lie === p
argh p (MkLT eq) with (a < a)
  argh p (MkLT Refl) | True  = ?crash -- type: idris_crash (...) === p
  argh p (MkLT eq)   | False = ?sensiblepath
```

Now, in the current evaluator `idris_crash` is a blocked primitive so it's not
that big of a deal: you just see this crash in the goal. But if we start doing
advanced things with evaluation (like compiling to native code and running that)
we could very well trigger the crash and see it blow up in our faces.

The problem in that example was that I was able to make `lie : LT a a` reduce to
a canonical form `MkLT Refl` which allowed `irrefl` to reduce & crash.

The idea behind the `GT 0 1 -> EQ 0 1` example is similar: use
`reflect : EQ a b -> a === b` to get an equality proof that also reduces to `Refl`
and then you can cast an `()` to a `(Int, Int)` & try to access its second projection
and that should break evaluation pretty badly (and potentially use a segfault if
we are using low level primitives in the evaluator).

PR #986, opened 2021-01-25
Add zip*/unzip* for List1

ska80:
Unfortunately, I wasn't just able to reuse `Data.List.zip*/unzip*` on `Data.List1`'s tail due to the circular
module dependency error as `Data.List` already depends on `Data.List1`

### DIFF
--- a/libs/base/Data/List1.idr
+++ b/libs/base/Data/List1.idr
@@ -10,6 +10,8 @@ record List1 a where
   head : a
   tail : List a
 
+%name List1 xs, ys, zs
+
 ------------------------------------------------------------------------
 -- Conversion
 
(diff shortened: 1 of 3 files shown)


libs/base/Data/List1.idr
```diff
@@ -141,3 +143,65 @@ Ord a => Ord (List1 a) where
...
+  where
+    zipWith3' : List a -> List b -> List c -> List d
+    zipWith3' [] _ _ = []
+    zipWith3' _ [] _ = []
+    zipWith3' _ _ [] = []
+    zipWith3' (x :: xs) (y :: ys) (z :: zs) = f x y z :: zipWith3' xs ys zs
+
+export
+zip3 : List1 a -> List1 b -> List1 c -> List1 (a, b, c)
+zip3 = zipWith3 \x, y, z => (x, y, z)
```
gallais, line 175: Note that you can use sections i.e. `(,,)`.

ska80: Applied this suggestion in `Data.List`, `Data.List1` and `Data.Stream`

PR #987, opened 2021-01-25
[ refactor ] Rename Data.Strings to Data.String (1/2)

andylokandy:
The reason why I submit this PR is that the other modules in `Data` are singular and the module name `Data.Strings` is out of sync with `Data.String.Extra` in `contrib`. Therefore, I think this will make the naming more consistent.

But this is indeed a big breaking change, we may break many many existing Idris1 and Idris2 codes... So I think it has to be thought carefully before being accepted.


gallais:
This makes it impossible to build from the previous version so won't be
accepted as is. A move towards the end result of this PR is to first
move `Data.Strings` to `Data.String` and then reexport publicly the content
in `Data.Strings`.

After a minor release has happened, you'll be able to rename the `Data.Strings`
import to `Data.String` ones without breaking the build chain.

libs/base/Data/Strings.idr
```diff
@@ -1,3 +1,4 @@
+||| To be deprecated
 module Data.Strings
```
gallais, line 3: Can we not replace the content of this module with `import public Data.String`?

andylokandy: This was problematic on my machine. But I seems not able to reproduce it... I'll push it here and let the CI try it.

andylokandy:
```
Error: While processing right hand side of show. Undefined name Strings.singleton. 

System/Path.idr:110:22--110:44
     |
 110 |   show (Disk disk) = Strings.singleton disk ++ ":"
     |                      ^^^^^^^^^^^^^^^^^^^^^^
```

andylokandy:
```
Error: While processing right hand side of show. Undefined name Strings.singleton. 

System/Path.idr:110:22--110:44
     |
 110 |   show (Disk disk) = Strings.singleton disk ++ ":"
     |                      ^^^^^^^^^^^^^^^^^^^^^^
```

gallais: The trick was to import qualified! ;)

PR #988, opened 2021-01-26
Make `zip` usable infix 

mb64:
Since `(a, b, c)` is the same as `(a, (b, c))`, this allows for chaining `zip` in a nice-looking way:

```idris
zipFour : List a -> List b -> List c -> List d -> List (a,b,c,d)
zipFour x y z w = x `zip` y `zip` z `zip` w
```

I chose `infixr 6`, since that seemed reasonable enough to me, but I'm open to suggestions for a better precedence.

### DIFF
--- a/libs/base/Data/Zippable.idr
+++ b/libs/base/Data/Zippable.idr
@@ -21,6 +21,8 @@ interface Zippable z where
   zip : z a -> z b -> z (a, b)
   zip = zipWith (,)
 
+  infixr 6 `zip`
+
   ||| Combine three parameterised types by applying a function.
   ||| @ z the parameterised type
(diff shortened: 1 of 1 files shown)


buzden:
`Stream`s also have a `zip` function. Since you did it for `List`s and `Vect`s, why not to do it for `Stream`s?

ska80:
> `Stream`s also have a `zip` function. Since you did it for `List`s and `Vect`s, why not to do it for `Stream`s?

and for `Data.List1` once #986 is merged

mb64:
Rebased to master.  The new `Zippable` interface is nice, it makes this change a lot simpler.

PR #989, opened 2021-01-26
Fix of laziness of applicative and monad operations of `LazyList`

buzden:
This is the weaker (and actually, the original) version of #867. Suggestion of addition of lazy `foldr` to the `Foldable` interface seems to be accepted not as enthusiastic as I anticipated, but I think we need at least to fix existing `LazyList` implementation.

The current implementation of `LazyList`'s functions `<*>` from `Applicative` and `>>=` from `Monad` are not lazy, thus resulting to too eager and costly computation when there are potentially big lazy lists being combined with `<*>` or `>>=`.

This PR fixes this and seems to not to add anything controversial, unlike the #867.

### DIFF
--- a/libs/contrib/Data/List/Lazy.idr
+++ b/libs/contrib/Data/List/Lazy.idr
@@ -10,6 +10,22 @@ data LazyList : Type -> Type where
   Nil : LazyList a
   (::) : (x : a) -> (xs : Lazy (LazyList a)) -> LazyList a
 
+--- Truly lazy functions ---
+
+public export
+foldrLazy : (func : elem -> Lazy acc -> acc) -> (init : Lazy acc) -> (input : LazyList elem) -> acc
+foldrLazy _  init [] = init
+foldrLazy op init (x::xs) = x `op` foldrLazy op init xs
+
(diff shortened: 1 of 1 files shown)


libs/contrib/Data/List/Lazy.idr
```diff
@@ -10,6 +10,23 @@ data LazyList : Type -> Type where
...
 
+--- Truly lazy functions ---
+
+public export
+(++) : LazyList a -> Lazy (LazyList a) -> LazyList a
+[]      ++ ys = ys
+(x::xs) ++ ys = x :: (xs ++ ys)
+
+public export
+foldrLazy : (func : elem -> Lazy acc -> acc) -> (init : acc) -> (input : LazyList elem) -> acc
```
gallais, line 21: Shouldn't `init` be lazy too? If `func` throws away its second argument and `input`
is non-empty, there is no reason to evaluate `init`!

buzden: I agree with you. I was solving a particular problem and wasn't thinking a little further.

PR #991, opened 2021-01-26
Remove unused modules

ska80:
Should these modules be removed?  They are not used and mentioned anywhere and probably were left after experimentation.


edwinb:
I think they should be removed, yes. They should probably never have been there in the first place.

buzden:
I'm wondering why `Control.App.Network` is not affected: it seems even to not to typecheck.

ska80:
> I'm wondering why `Control.App.Network` is not affected: it seems even not to typecheck.

I guess it is still WIP

buzden:
> > I'm wondering why `Control.App.Network` is not affected: it seems even not to typecheck.
> 
> I guess it is still WIP

Let's see. It looks like an early experiment before the `network` library established.

PR #993, opened 2021-01-27
Add missing Bits64->Int cast

peterohanley:
Looks like it was just missing. The Int->Bits64 and Integer->Bits64
cases are identical, so casting back can also be identical.

I ran into this in some binary parsing code.


gallais:
Thanks. We keep hitting these missing cases.

I wonder whether we could refactor the code to ensure that all of the casts are
covered in all of the backends.

PR #997, opened 2021-01-28
[new] Add Colist and Colist1 to base

stefan-hoeck:
This adds `Colist` and `Colist1` codata types to base as requested in a comment by gallais [here](https://groups.google.com/g/idris-lang/c/QWhCihTtI_w). Just tell me if this should rather go to contrib instead.

If this gets accepted and stays in base, we might consider adjusting some of the existing partial functions like `takeBefore` and `takeUntil` in Data.Stream to return `Colist`s instead.


libs/base/Data/Colist.idr
```diff
@@ -0,0 +1,228 @@
...
+||| This stops with `Nil` once the function returns `Nothing`.
+public export
+iterateMaybe : (f : a -> Maybe a) -> Maybe a -> Colist a
+iterateMaybe _ Nothing  = Nil
+iterateMaybe f (Just x) = x :: iterateMaybe f (f x)
+
+||| Generate an infinite `Colist` by repeatedly applying a function
+||| to a seed value.
+public export
+unfoldr : (f : s -> (s,a)) -> s -> Colist a
```
gallais, line 70: This is the unfold for `Stream` so I'm not sure it's worth including it here too.
There is also no meaningful notion of unfoldl so there is no need for the trailing 'r'.

libs/base/Data/Colist.idr
```diff
@@ -0,0 +1,228 @@
...
+unfoldr f s = let (s2,a) = f s
+               in a :: unfoldr f s2
+
+||| Generate an `Colist` by repeatedly applying a function
+||| to a seed value.
+||| This stops with `Nil` once the function returns `Nothing`.
+public export
+unfoldrMaybe : (f : s -> Maybe (s,a)) -> s -> Colist a
```
gallais, line 78: This is the proper `unfold` for Colist.

libs/base/Data/Colist.idr
```diff
@@ -0,0 +1,228 @@
...
+isNil [] = True
+isNil _  = False
+
+||| True, if the given `Colist` is non-empty.
+public export
+isCons : Colist a -> Bool
+isCons [] = False
+isCons _  = True
+
+||| Append two `Colist`s.
```
gallais, line 99: It's also pretty convenient to have functions such as
```idris
lappend : List a -> Colist a -> Colist a
rappend : Colist a -> List a -> Colist a
```

libs/base/Data/Colist1.idr
```diff
@@ -0,0 +1,205 @@
...
+replicate (S k) x = x ::: replicate k x
+
+||| Produce a `Colist1` by repeating a sequence
+public export
+cycle : List1 a -> Colist1 a
+cycle (h ::: [])          = repeat h
+cycle (h ::: t@(y :: ys)) = h ::: run y ys
+  where run : a -> List a -> Colist a
+        run v []        = v :: run h t
+        run v (y :: ys) = v :: run y ys
```
gallais, line 65:
```suggestion
cycle (x ::: xs) = run x xs
  where run : a -> List a -> Colist a
        run v []        = v :: run x xs
        run v (y :: ys) = v :: run y ys
```

gallais (commented):
A couple of comments on lots of good stuff.

libs/base/Data/Colist1.idr
stefan-hoeck: I don't think this works, does it? run should now have return type `Colist1 a` instead of `Colist a`. A shorter version
that would work would be

```
cycle (h ::: t) = h ::: cycle (t <+> [h])
```

That would be short and clean.

libs/base/Data/Colist.idr
stefan-hoeck: Removed. I also had to adjust (and rename) `unfoldr` for `Colist1`.

libs/base/Data/Colist.idr
stefan-hoeck: Renamed this one and wrote a similar `unfold` function for `Colist1`.

libs/base/Data/Colist.idr
stefan-hoeck: I added `appendl` and `lappend` to `Colist` and `Colist1`. I chose the names to be consistent with the ones in `Data.List1`.

libs/base/Data/Colist1.idr
stefan-hoeck: Changed the implementation to `cycle (h ::: t) = h ::: cycle (t ++ [h])`.

stefan-hoeck:
@andorp I don't have a strong opinion on this and will be happy to move stuff around if people agree on a new `CoData` module hierarchy.

libs/base/Data/Colist1.idr
gallais: Good point. It's pretty annoying we don't have a type level notion of
"being guarded" because we really ought to be able to write this like
so:

```idris
cycle1 : List1 a -> Colist1 a
cycle1 xxs = xxs `lappend1` cycle1 xxs
```

where `lappend1 : List1 a -> GUARDED (Colist1 a) -> Colist1 a`

gallais:
I don't know about `Data` vs. `CoData`: there are mixed inductive / coinductive
structures that would not neatly fit into one or the other.

E.g.

```idris
||| Infinitely often `a` separated by finite segments of `b`
data IOB : (a, b : Type) -> Type where
  A : a -> Inf (IOB a b) -> IOB a b
  B : b -> IOB a b -> IOB a b

||| total because terminating (i.e. inductive-like)
untilA : IOB a b -> IOB a b
untilA iob@(A _ _) = iob
untilA (B _ iob) = untilA iob

||| total because productive (i.e. coinductive-like)
brrrr : a -> IOB a b
brrrr a = A a (brrrr a)
```

libs/base/Data/Colist1.idr
stefan-hoeck: Something like this "GUARDED" keyword would indeed be quite useful. I'm at the moment experimenting with Cotrees (possibly infinite rose trees) and writing total functions often means repeating quite a bit of code, since it's for instance not possible to use a recursive call in Functor `map`.

Anyway, if my experiments lead to some useful results and if there is some interest in having Tree and Cotree in base or contrib, I'll be happy to add those as well.

libs/base/Data/Colist.idr
```diff
@@ -0,0 +1,231 @@
...
+public export
+Monoid (Colist a) where
+  neutral = Nil
+
+public export
+Functor Colist where
+  map f []        = []
+  map f (x :: xs) = f x :: map f xs
+
+public export
```
mattpolzin, line 197: Is it correct (or is it important) that [`Applicative` on `List`](https://github.com/idris-lang/Idris2/blob/master/libs/prelude/Prelude/Types.idr#L357) operates differently than this implementation on `Colist`?

on `List` it currently applies each function to each value and concats the results together.

buzden: It looks like the current implementation mimics the `Data.Stream`'s implementation of `Applicative` and `Monad`. So, `Colist` is rather a `Stream` that can be finite, than a `List` that can be infinite.

stefan-hoeck: @mattpolzin: I think this is correct and important. Let's try and write a total implementation of `concatColist` in the spirit of `Data.List`

```
concatColist : Colist (Colist a) -> Colist a
concatColist Nil = Nil
concatColist ((x :: xs) :: xss) = x :: concatColist (xs :: xss)
concatColist (Nil :: xss) = concatColist xss
```

The last pattern match is problematic, since the recursive call is not guarded by `Inf`. In order to decide what the next element is, it might need to traverse the whole (possibly infinite) `Colist`. Indeed: `isNil . concatColist $ repeat Nil` will loop forever. Therefore, the module provides not `concatColist`, since for the time being I decided to only provide total functions.

But in order to implement `Applicative` and `Monad` in the `List` sense, we need something along the lines of `concatColist`.

The `Monad` instance for `Stream` behaves similarily: It treats a `Stream a` as a function `Nat -> a` (indeed: the two are isomorphic!) and the `Monad` instance of `Stream` behaves the same as the `Monad` instance for `Reader Nat`. This is, by the way, the same behavior like for `Vect n a` which is isomorphic to `Reader (Fin n) a` and whose `Monad` behaves the same as the one for `Reader (Fin n)`. Such data types are sometimes called _Naperian_ (see for instance [The Essense of the Iterator Pattern](https://www.cs.ox.ac.uk/jeremy.gibbons/publications/iterator.pdf)).

Now a `Colist` is not of a fixed size or shape, so it is not a Naperian data type. Still, it can be viewed as a function `index : Nat -> Maybe a` with the additional invariant that if `index n = Nothing` then `index (S n) = Nothing` and therefore its `Monad` and `Applicative` implementations behave like the ones for `ReaderT Nat Maybe`.

`Colist1` behaves the same with the only additional invariant that it always holds at least one element and therefore `index 0` always returns a `Just`.

However, all this writing made me realize I never verified the Applicative and Monad laws for these implementations. I'll give that a try in Idris and will report back here if I got stuck (or the laws are violated!).

mattpolzin: Thanks for the great reply! It makes intuitive sense to me now to implement it this way.

stefan-hoeck: I already got stuck proving the Functor laws, haha. I tried something along the following lines

```
I : a -> a
I a = a

functorId : (v : Colist a) -> (map I v = v)
functorId []       = Refl
functorId (x :: y) = cong2 (::) Refl ?foo
```

The problem her is, that `?foo` is cluttered with `Delay`s and `Force`s, and I have no idea how to get rid of those (its type is actually `Delay (map I Force y) = y`). Could someone please show me, if (and how) it is possible to properly deal with this? It would be great to have actual proofs of these laws in the module.

stefan-hoeck: Ok, I found the following [blog post](https://www.howtobuildsoftware.com/index.php/how-do/rS0/verification-idris-coinduction-proof-of-streams-functor-laws) describing a technique for proving equality of codata. In my case, this didn't work out in Idris2 (still problems with Delay and Force), but from this (and the isomorphisms mentioned above), I got the idea of using an element-wise equivalence: Two Colists can be considered equal, if `index` returns the same result for both of the for every natural number. Once again, I'll report back if this works out as expected.

buzden: > The problem her is, that `?foo` is cluttered with `Delay`s and `Force`s, and I have no idea how to get rid of those (its type is actually `Delay (map I Force y) = y`). Could someone please show me, if (and how) it is possible to properly deal with this? It would be great to have actual proofs of these laws in the module.

You can match on a `Delay`, actually, to force it:
```idris
functorId : (xs : Colist a) -> map Prelude.id xs = xs
functorId [] = Refl
functorId (x::Delay xs) = cong (x::) $ functorId xs
```

buzden: > I think this is correct and important

Sorry for repeating myself again, but I really think that the documentation comment on the `Colist` definition (that is currently "A possibly infinite list") is misleading and should be replaced with "A possibly finite Stream".

stefan-hoeck: @buzden Sorry, I didn't realize this was a comment regarding the data type's short description. I adjusted the descriptions accordingly.

stefan-hoeck: So, I found that the Monad laws for this implementation actually do not hold. Namely the law of associativity, which, expressed in terms of `join`, states that

```
join . join = join . map join
```

Minimal counter example:

```
Test : Colist (Colist (Colist Nat))
Test = [ [[1]], [[],[2,3]] ]

join Test = [[1],[2,3]]

join (join Test) = [1,3]

map join Test = [[1],[]]

join (map join Test) = [1]
```

I therefore removed the Monad implementations for Colist and Colist1.

gallais: Yeah the right notion of equality on coinductive types is bisimilarity
(aka a coinductive notion of pointwise equality). Ohad has a student
incoming who will work on a prototype changing the notion of equality
in Idris to correspond to this (and change equality of functions to
pointwise equality too).

In the meantime, `(=#=)` in the linked post is the right thing to do.

andorp:
> I don't know about `Data` vs. `CoData`: there are mixed inductive / coinductive
> structures that would not neatly fit into one or the other.
> 
> E.g.
> 
> ```idris
> ||| Infinitely often `a` separated by finite segments of `b`
> data IOB : (a, b : Type) -> Type where
>   A : a -> Inf (IOB a b) -> IOB a b
>   B : b -> IOB a b -> IOB a b
> 
> ||| total because terminating (i.e. inductive-like)
> untilA : IOB a b -> IOB a b
> untilA iob@(A _ _) = iob
> untilA (B _ iob) = untilA iob
> 
> ||| total because productive (i.e. coinductive-like)
> brrrr : a -> IOB a b
> brrrr a = A a (brrrr a)
> ```

Good point. I wonder if there any famous structure which has this property. I can't think of any from top of my head.

gallais:
> Good point. I wonder if there any famous structure which has this property.
  I can't think of any from top of my head.

[Stream transducers](https://arxiv.org/abs/0905.4813) are the first thing that come to mind.
They're a cool little formalisation project that could go into `contrib` if
someone is interested in having a look.

PR #998, opened 2021-01-28
Make :typeat a useful command

GustavoMF31:
## Overview
The `:typeat` command is very simple: It takes the cursor position in the file and then shows the type of the name under it. It works by using the `names` field of the metadata: It is a mapping from file contexts, a type that holds a file name and start and end positions in the file, to the name and type of the variable.

It is different from the regular `:t` in that `:t` gets the information from the global definitions and thus doesn't work for local variables.

The problem was that the `names` field of the metadata was really messy. It was missing some stuff and also including some nodes that don't actually exist in the source code, making `:typeat` work pretty poorly. This PR is about cleaning up some file context stuff to populate the metadata accurately and then be able to get types of local variables with `:typeat`.

Note: I'm currently using the command through my fork of idris2-vim. I've submitted a [pull request](https://github.com/edwinb/idris2-vim/pull/24) to include the changes in the main repo.

## Improvements
- Include names defined in `where` and `let`
- Include the name in the lhs of `let` bindings
- Include the name in the lhs of `do` bindings
- Include variables bound by lambdas
- Include variables bound by as-patterns
- Create a type representing file contexts that cannot be empty
- Track file contexts for `App` nodes when evaluating to normal form

## Fixes
- Give precedence to the local name when there is shadowing
- Make `:typeat` work within type signatures
- Make `:typeat` work within !-notation
- Make `:typeat` work around record update syntax
- Give empty file contexts to some AST nodes generated by desugaring lambdas
- Give the correct file context to function applications
- Track file contexts better around interfaces
- Make `:typeat` work within record definitions

I've also added tests for most of the fixes and improvements.

## Breaking changes

Some of the modifications require storing more file context information in AST nodes. This means it's a breaking change for Elaborator Reflection, but fixing existing code is just a matter of adding more `EmptyFC`.

Since the nodes are saved to TTC, I'm pretty sure I should increase the TTC version. However, when I push the commit that does it, the "build-api" ci job fails (I've left it there so you can see the problem). Maybe it's not set up to handle a TTC version increase. Can someone help me figure out how to get it working?

## Performance benchmarks
To make sure things were not made slower, I timed how long it took for each version of Idris2 (master and my branch typeat) to build master. This time I did an average of 5 builds to be a bit more precise. In the end, building using the branch with the changes took on average 1.6% longer, which I'd say is acceptable.

## Things that are still broken

I also ran into some other problems, but I decided to leave them as is in order to not make this PR any longer than it already is. They are:

- Global definitions are not included in the metadata
- Variables in interface constraints get wrong file contexts
- The name of a function in the lhs is not included in the metadata
- The name of a function in the lhs in an interface implementation gets a machine generated name
- Record updates with the new syntax generate nodes with wrong file contexts
- Variables in the lhs of a function definition with types that correspond to type variables show `?_ [no locals in scope]` as the type


GustavoMF31:
I've taken a look at the "build-api" job, and it looks like it's using the previous version to build and test the api of the current version, which seems a bit odd. Shouldn't the current version be the one required to build and work with the api? I've tested by pushing to a separate branch and if "build-api" is changed to use the artifact "idris2-nightly" (the current version) the test passes.
There might be some compatibility reason for the way things are, though. Does anyone know of such thing?

However, I still couldn't understand why building the api with the previous version fails when the TTC version is increased. I can't imagine how the different versions get messed up if the current version of idris2 is not even installed at that point. (But there is always the possibility I understood the job wrong).

src/Idris/Elab/Interface.idr
```diff
@@ -284,12 +287,14 @@ elabInterface {vars} fc vis env nest constraints iname params dets mcon body
          let conName_in = maybe (mkCon fc fullIName) id mcon
          -- Machine generated names need to be qualified when looking them up
          conName <- inCurrentNS conName_in
-         let meth_sigs = mapMaybe getSig body -- (FC, RigCount, List FnOpt, Name, (Bool, RawImp))
-         let meth_decls = map (\ (f, c, o, n, b, ty) => (n, c, o, b, ty)) meth_sigs
+         let meth_sigs = mapMaybe getSig body -- (FC, FC, RigCount, List FnOpt, Name, (Bool, RawImp))
+         let meth_decls = map (\ (f, f', c, o, n, b, ty) => (n, c, o, b, ty)) meth_sigs
```
gallais, line 291: This is getting out of hand: we should probably introduce record types for these
different structures we are passing around during elaboration. That would have the
added benefit of documenting what these are meant to be (e.g. the type of `elabMethods`
is modified below to return 2 `FC` instead of one and it's impossible to know what
they correspond to without looking at the implementation).

src/Idris/Pretty.idr
```diff
@@ -197,9 +198,43 @@ mutual
...
+          <|> do
+            nName <- getPRefName n
+            guard (isUnderscoreName nName)
+            pure continuation
+
+        where
+          isUnderscoreName : Name -> Bool
+          isUnderscoreName (UN "_") = True
+          isUnderscoreName (MN "_" _) = True
+          isUnderscoreName _ = False
```
gallais, line 224: Please move this function to `Core.Name`

src/Idris/REPL.idr
```diff
@@ -440,7 +448,11 @@ processEdit (GenerateDef upd line name rej)
                     put ROpts (record { gdResult = Just (line, searchdef) } ropts)
                     Just (_, (fc, cs)) <- nextGenDef rej
                          | Nothing => pure (EditError "No search results")
-                    let l : Nat =  integerToNat (cast (snd (startPos fc)))
+
+                    let l : Nat = integerToNat $ cast $ case fc of
+                         (MkFC _ (_, col) _) => col
+                         _ => the Int 0
+
```
gallais, line 455: The same pattern is use twice, can we introduce `startCol : FC -> Int` in `Core.FC`?

src/Idris/REPLCommon.idr
```diff
@@ -51,32 +51,39 @@ printError : {auto o : Ref ROpts REPLOpts} ->
...
                                               toSExp !(renderWithoutColor msg),
                                               -- highlighting; currently blank
                                               SExpList []],
                                     toSExp i])
   where
+    -- TODO: Display a better message when the error doesn't contain a location
+    defaultFC : FC -> NonEmptyFC
+    defaultFC EmptyFC = ("", (0, 0), (0, 0))
+    defaultFC (MkFC fname start end) = (fname, start, end)
+
```
gallais, line 86: This should be moved to `Core.FC`. It could be reused in other places where you're doing
essentially this (e.g. the definition of `constructSExp`).

gallais (commented):
This is exciting!

Russoul:
#1030 should do it. Here is a merged run I did: https://github.com/Russoul/Idris2/runs/1833442179
"Install libs" shouldn't be needed anymore, though I forgot to take your commits, gallais, with CI changes out before running the merged CI test...

src/Core/Value.idr
```diff
@@ -75,15 +75,18 @@ mutual
   data NF : List Name -> Type where
        NBind    : FC -> (x : Name) -> Binder (NF vars) ->
                   (Defs -> Closure vars -> Core (NF vars)) -> NF vars
-       NApp     : FC -> NHead vars -> List (Closure vars) -> NF vars
+       -- Each closure is associated with the file context of the App node that
+       -- had it as an argument. It's necessary so as to not lose file context
+       -- information when creating the normal form.
```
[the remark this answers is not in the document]
gallais: Closures in `NDelay` are not tagged with an `FC` so I think we can shelve this one for now.

GustavoMF31:
Oh, sorry for taking so long here, looks like you took over. I had the start of the refactoring already done locally, I should have communicated better. Regardless, thanks for taking a look at the "build-api" problems, gallais and Russoul!

gallais:
We were starting to get merge conflicts so I figured I'd better get on
with it because it's a big one and it was going to deteriorate rapidly.

gallais:
Planning to merge this tomorrow afternoon.

Issue #555, opened 2020-08-13
String interpolation for Idris2


andrevidela:
TLDR: String interpolation, If you want to comment, read the 3rd section first.

[String interpolation](https://en.wikipedia.org/wiki/String_interpolation) (Also known as variable interpolation, variable expansion, or variable substitution) is a common feature of modern programming languages that allows the programmer to write a string literal and insert executable code inline for it to be automatically inserted within the string by the compiler.

Since this proposal is mostly syntactic, in order to avoid the [_Wadler effect_ ](https://wiki.haskell.org/Wadler's_Law), we are going to split it in 3 parts:

- The proposal itself
- The technical details
- The research conducted to reach the proposal

To avoid endless bike-shedding I invite you to comment on the first two, the last one aims to answer all other questions or remarks that might slow down or pollute the discussion thread with less pressing concerns.

# 1. The proposal

We propose introducing this new syntax for string literals

```scala
"this is a normal string"
s"This is an {interpolated} string"
"normal strings are {raw} strings"
s"interpolated string \{need\} to escape curly braces if you want to show them."
```

Expressions within brackets can be arbitrary complex as long as they return a type that implements `Show` :

```scala
s"This is a string with a complex value: {replace (plusCommutative n m) $ map isEven record.field}"
```

# 2. Technical details

String interpolation has 2 goals:
- Improve readability
- Inform the compiler of the user’s intention about string manipulation

While the first one can only be subjectively evaluated, the second one has a more concrete consequence for the compiler. Take this example

```haskell
let name = "Susan"
    greeting = s"hello {name}" in 
    putStrLn greeting
```

In it, the compiler can _see_ that the variable `name` comes from a string literal. From which the compiler can perform the substitution _at compile time_ rather than at runtime, outputting the code (given some inlining)

```haskell
putStrLn "hello Susan"
```

instead of 

```haskell
putStrLn ("hello " <+> "Susan")
```

In other words, the compiler can avoid runtime concatenation operation and perform those text replacement at compile-time.

## A taste of linearity

Interestingly enough, since this substitution is performed at _compile time_ it is semantically equivalent to a function call with linearity `0`, that is, the following should compile:

```haskell
let 0 name = "Susan"
    1 greeting = s"hello {name}" in
    putStrLn greeting
```

[...]

PR #999, opened 2021-01-29
Support raw string

andylokandy:
This PR proposed a new simple syntax for raw string, aka, the string literal with no escape.

For example, a template string that contains lots of quotes can be written in:
```idris
template : String
template = #"[project]
name: "project"
version: "0.1.0"

[deps]
"semver" = 0.2
"#
```

before:

```idris
template : String
template =  "[project]\n"
         ++ "name: \"project\"\n"
         ++ "version: \"0.1.0\"\n"
         ++ "\n"
         ++ "[deps]\n"
         ++ "\"semver\" = 0.2\n"
```

### DIFF
--- a/src/Parser/Support.idr
+++ b/src/Parser/Support.idr
@@ -138,64 +138,66 @@ getEsc "SP" = Just '\SP'
 getEsc "DEL" = Just '\DEL'
 getEsc str = Nothing
 
-escape' : List Char -> Maybe (List Char)
-escape' [] = pure []
-escape' ('\\' :: '\\' :: xs) = pure $ '\\' :: !(escape' xs)
-escape' ('\\' :: '&' :: xs) = pure !(escape' xs)
-escape' ('\\' :: 'a' :: xs) = pure $ '\a' :: !(escape' xs)
-escape' ('\\' :: 'b' :: xs) = pure $ '\b' :: !(escape' xs)
-escape' ('\\' :: 'f' :: xs) = pure $ '\f' :: !(escape' xs)
-escape' ('\\' :: 'n' :: xs) = pure $ '\n' :: !(escape' xs)
-escape' ('\\' :: 'r' :: xs) = pure $ '\r' :: !(escape' xs)
-escape' ('\\' :: 't' :: xs) = pure $ '\t' :: !(escape' xs)
-escape' ('\\' :: 'v' :: xs) = pure $ '\v' :: !(escape' xs)
-escape' ('\\' :: '\'' :: xs) = pure $ '\'' :: !(escape' xs)
-escape' ('\\' :: '\"' :: xs) = pure $ '\"' :: !(escape' xs)
-escape' ('\\' :: 'x' :: xs)
-    = case span isHexDigit xs of
-           ([], rest) => assert_total (escape' rest)
-           (ds, rest) => pure $ cast !(toHex 1 (reverse ds)) ::
-                                 !(assert_total (escape' rest))
+escape' : List Char -> List Char -> Maybe (List Char)
+escape' _ [] = pure []
+escape' escapeChars (x::xs)
+    = assert_total $ if escapeChars `isPrefixOf` (x::xs)
+         then case drop (length escapeChars) (x::xs) of
+                   ('\\' :: xs) => pure $ '\\' :: !(escape' escapeChars xs)
+                   ('\n' :: xs) => pure !(escape' escapeChars xs)
+                   ('&' :: xs) => pure !(escape' escapeChars xs)
+                   ('a' :: xs) => pure $ '\a' :: !(escape' escapeChars xs)
+                   ('b' :: xs) => pure $ '\b' :: !(escape' escapeChars xs)
+                   ('f' :: xs) => pure $ '\f' :: !(escape' escapeChars xs)
+                   ('n' :: xs) => pure $ '\n' :: !(escape' escapeChars xs)
+                   ('r' :: xs) => pure $ '\r' :: !(escape' escapeChars xs)
+                   ('t' :: xs) => pure $ '\t' :: !(escape' escapeChars xs)
+                   ('v' :: xs) => pure $ '\v' :: !(escape' escapeChars xs)
+                   ('\'' :: xs) => pure $ '\'' :: !(escape' escapeChars xs)
+                   ('"' :: xs) => pure $ '"' :: !(escape' escapeChars xs)
+                   ('x' :: xs) => case span isHexDigit xs of
+                                       ([], rest) => escape' escapeChars rest
+                                       (ds, rest) => pure $ cast !(toHex 1 (reverse ds)) ::
+                                                             !(escape' escapeChars rest)
+                   ('o' :: xs) => case span isOctDigit xs of
+                                       ([], rest) => escape' escapeChars rest
+                                       (ds, rest) => pure $ cast !(toOct 1 (reverse ds)) ::
+                                                             !(escape' escapeChars rest)
+                   xs => case span isDigit xs of
+                              ([], (a :: b :: c :: rest)) =>
+                                case getEsc (fastPack (the (List _) [a, b, c])) of
+                                     Just v => Just (v :: !(escape' escapeChars rest))
+                                     Nothing => case getEsc (fastPack (the (List _) [a, b])) of
+                                                     Just v => Just (v :: !(escape' escapeChars (c :: rest)))
+                                                     Nothing => escape' escapeChars xs
+                              ([], (a :: b :: [])) =>
+                                case getEsc (fastPack (the (List _) [a, b])) of
+                                     Just v => Just (v :: [])
+                                     Nothing => escape' escapeChars xs
+                              ([], rest) => escape' escapeChars rest
+                              (ds, rest) => Just $ cast (cast {to=Int} (fastPack ds)) ::
+                                              !(escape' escapeChars rest)
+         else Just $ x :: !(escape' escapeChars xs)
   where
     toHex : Int -> List Char -> Maybe Int
     toHex _ [] = Just 0
     toHex m (d :: ds)
         = pure $ !(hex (toLower d)) * m + !(toHex (m*16) ds)
-escape' ('\\' :: 'o' :: xs)
-    = case span isOctDigit xs of
-           ([], rest) => assert_total (escape' rest)
-           (ds, rest) => pure $ cast !(toOct 1 (reverse ds)) ::
-                                 !(assert_total (escape' rest))
-  where
+
     toOct : Int -> List Char -> Maybe Int
     toOct _ [] = Just 0
     toOct m (d :: ds)
         = pure $ !(oct (toLower d)) * m + !(toOct (m*8) ds)
-escape' ('\\' :: xs)
-    = case span isDigit xs of
-           ([], (a :: b :: c :: rest)) =>
-               case getEsc (fastPack (the (List _) [a, b, c])) of
-                   Just v => Just (v :: !(assert_total (escape' rest)))
-                   Nothing => case getEsc (fastPack (the (List _) [a, b])) of
-                                   Just v => Just (v :: !(assert_total (escape' (c :: rest))))
-                                   Nothing => escape' xs
-           ([], (a :: b :: [])) =>
-               case getEsc (fastPack (the (List _) [a, b])) of
-                   Just v => Just (v :: [])
-                   Nothing => escape' xs
-           ([], rest) => assert_total (escape' rest)
-           (ds, rest) => Just $ cast (cast {to=Int} (fastPack ds)) ::
-                                 !(assert_total (escape' rest))
-escape' (x :: xs) = Just $ x :: !(escape' xs)
 
 export
-escape : String -> Maybe String
-escape x = pure $ fastPack !(escape' (unpack x))
+escape : Nat -> String -> Maybe String
+escape hashtag x = let escapeChars = '\\' :: replicate hashtag '#' in
+                       fastPack <$> (escape' escapeChars (unpack x))
 
 export
 getCharLit : String -> Maybe Char
 getCharLit str
-   = do e <- escape str
+   = do e <- escape 0 str
         if length e == 1
            then Just (assert_total (prim__strHead e))
            else if length e == 0 -- parsed the NULL character that terminated the string!
--- a/src/Idris/IDEMode/Parser.idr
+++ b/src/Idris/IDEMode/Parser.idr
@@ -28,7 +28,7 @@ ideTokens : TokenMap Token
 ideTokens =
     map (\x => (exact x, Symbol)) symbols ++
     [(digits, \x => IntegerLit (cast x)),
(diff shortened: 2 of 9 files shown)


buzden:
Maybe, then, properly aligned raw string literals should remove appropriate count of spaces? I.e., for example

```idris
test : IO ()
test =
  do
    putStr configTemplate
    putStr """foo
               bar
              baz
           """
```

should be printed as
```
foo
 bar
baz
```

Because, I think, this example from your test

```idris
test =
￼  do
￼    putStr configTemplate
￼    putStr """foo
￼bar
￼baz
￼"""
```

looks ugly, especially in the do-notation (however, it looks so anyway).

andylokandy:
@buzden That a neat idea. But this ugly(that's true) syntax is adopted by Rust, Scala, Js, Python. Anyway, I'll try out your idea.

buzden:
> @buzden That's a neat idea. But this ugly(that's true) syntax has adopted by Rust, Scala, Js, Python. Anyway, I'll try your idea.

I know only about Scala, to bite ugliness, they have to use
```scala
val s = """foo
          | bar
          |baz
        """.stripMargin
```

So, I mean, it's solvable afterwards, but why not to foresee it at the initial implementation.

gallais:
Can't you write

```idris
test =
￼  do
￼    putStr configTemplate
￼    putStr $ unlines 
       [ "foo"
       , " bar"
       , "baz"
       ]
```

? This complexifies the lexer for very little gain IMO.

andrevidela:
@gallais You _can_ but the gain is when you are dealing with large templated strings that are specially useful for templating configuration files, testing and outputing large strings.

gallais:
Wouldn't an `unsafePerformIO (readFile "my_template.txt")` be more ergonomic?

andrevidela:
Yes and no, it would work when your template file is static but not when it's dynamically generated. Additionally it doesn't allow you to use the string at compile time or to see what the string even is when looking at code. And it doesn't address the testing or large string output use-cases.

Typically the test for the Idris compiler make use of 4 files, the expected output, the run instructions, the input and the idris2 program.

All those could be raw strings instead of different files, allowing you to edit a given test by looking at a single file rather than 4 files. Abstracting those tests into strings rather than files also allow you to abstract over them in order to keep them in sync. Reducing the size of diffs and ensuring the tests themselves aren't buggy. I'm not suggesting we do that _here_ but this is an example of where someone writing their own tests might expect multi-line raw strings to exist.

I've personally needed them to write down network requests by hand

andrevidela:
I strongly encourage reading Rust's discussion for their raw string https://github.com/rust-lang/rust/issues/9411 and swift's accepted proposal (the discussion is a bit too long) https://github.com/apple/swift-evolution/blob/b2c3fc5320a215fd7e0587e0069cb403a91218b8/proposals/0168-multi-line-string-literals.md

andylokandy:
I've updated the syntax as suggested by @buzden

andrevidela:
I don't think this quite addresses the problems @buzden mentionned. Namely

## Leading line breaks aren't trimmed away

You seem to trim the last linebreak but not the first one

```
let testValue = """
    <?xml version="1.0"?>
    <catalog>
        <book id="bk101" empty="">
            <author>\(author)</author>
            <title>XML Developer's Guide</title>
            <genre>Computer</genre>
            <price>44.95</price>
            <publish_date>2000-10-01</publish_date>
            <description>An in-depth look at creating applications with XML.</description>
        </book>
    </catalog>
    """
```
Would insert an additional line return at the begining. With the current design, you force everyone to have at least 3 space of indentation to the left which isn't necessary nor useful. Trimming the leading line break would allow to stick
the content of a string to the left, which is particularly welcome when pasting from a reference string for example:

```
let testHTML = """
<html>
  <head> … </head>
  <body> … </body>
</html>
"""
```

## The trimming is done using the leading `"""` instead of the closing one

This is a problem because the leading `"""` won't be at the same indentation as the rest of the block, unless you want to force it this way, but it would be nice if we had a way to _chose_ the indentation of the following block. Using the closing `"""` and trimming the line returns around them allows for this flexibility, aka

```
let testValue = """
----<?xml version="1.0"?>
----<catalog>
----    <book id="bk101" empty="">
----        <author>\(author)</author>
----        <title>XML Developer's Guide</title>
----        <genre>Computer</genre>
----        <price>44.95</price>
----        <publish_date>2000-10-01</publish_date>
----        <description>An in-depth look at creating applications with XML.</description>
----    </book>
----</catalog>
----"""
```

All the `----` are trimmed out, and similarly if the closing `"""` was further to the left

Similarly, 
```
let trimmed = """
    { "username" : "john",
      "age" : 23 }
        """
```
Would be an error since the closing brace is going to trim non-whitespace. This is already handled, the error message is unsatisfactory but we can fix it later.

## No words about edge cases 

What do those do / what should these do?

```
let emptyString = """"""
    quoteString = """""""
    doubleQuote = """"""""
    string = """"hello""""
    nested1 = """"""""""""
    nested2 = """this is a multiline string in Idris:
"""hello I am
broken"""
"""
    nested3 = """this is a multiline string in Idris:
\"""hello I am
not a raw string anymore\"""
"""
```

## Discussion
```
emptyString = """""" 
```

This works if you think of `"""` as "raw strings" unfortunately, they as raw as long as you don't use `"""` again, which we might decide is the indended behaviour. However, I do not see why someone would write `""""""` instead of `""` and so I can see how someone might make the argument that `""""""` should be an error.

```
quoteString = """""""
doubleQuote = """"""""
string = """"hello""""
```

Similarly here, if `"""` means "raw strings" this should parse, but it doesn't, if this were to mean "multi-line string" I would expect this to be an error.

```
    nested1 = """"""""""""
```

Same for this one but it's a bit harder to see, IMO this should never parse, but if we were strict about `"""` meaning "raw strings" this should be ok.

```
    nested2 = """this is a multiline string in Idris:
"""hello I am
broken"""
"""
```

One could make the argument that this should parse if `"""` were really raw strings, but it seem to be very confusing now

```
    nested3 = """this is a multiline string in Idris:
\"""hello I am
not a raw string anymore\"""
"""
```

I would argue this should be possible but the semantics of raw strings is completely lost now since we've admitted escape characters.

## Proposed solution

1. Change the semantics of `"""` from "raw strings" to "multi-line strings, enforcing the fact that the start and end delmiters have to be on different lines, this allows to get rid of the first 5 edge cases and allows for `nested3`.
2. adopt Rust's solution for _raw strings_ which is to declare delimiters at the begining and end of the string such that the programmer decides which sequence of character cannot appear within the string
3. Disallow raw strings in multi-line strings
4. Allow multi-line strings to be interpolated with #555 

Number `3` means the following:
```
r#"""
   hello there
   """#
```
would be rejected, the reason for this is that the semantics of "raw string" fundamentally clashes with the trimming behaviour of multi-line strings. One can do the following:

```
r#"
hello there
"#
```

This would be slightly confusing since `""` cannot be used for multi-line strings but `r#""#` would. This is where we have to make an opiniated decision. I believe this is alright since "raw strings" aren't a feature that beginners will ask for or even interact with. It will be reserved to people writing templates, tests and regexes. IMO, If you can understand regexes, you can understand why `r#""#` is allowed to span multiple lines :)

Finally, the `"""` syntax isn't redundant with `r#""#` since raw string are fundamentally incompatible with interpolated strings, this would allow both strings and multi-line strings to be interpolated _and_ have single-line and multi-line raw strings.

## Conclusion

All of this should have been written in an issue but since the bulk of the work has been done for multi-line strings I would tend to merge it if everyone agrees with the point raised about trimming. If everyone is okay with this post I can write a proposal for supporting `raw strings` and how it would interact with strings, multi-line strings and string interpolation.

PS: you will notice those proposals don't stem from nowhere and are basically the swift and rust proposal smushed together.

ohad:
I don't quite follow. 
You might get strings like `""""""` if you auto-generate idris code and the string happens to be empty, so it should allowed.

We could add the convention that adjacent raw strings are concatenated, and then we can include `"""` inside a raw string by something like:
```idris
"""preamble
"""++Data.String.TripleQuotes++"""
rest"""
```

That's probably going to be easier than allowing arbitrary delimiters for raw strings.

andrevidela:
Unfortunately it's not as easy, this bit does not typecheck:

```
("""++Data.String.TripleQuotes++)"""
rest"""
```

because

```
                                """
rest"""
```
is an error. And this because of the trimming behaviour of `"""` requiring at least as many columns as the first set of `"""`. See https://github.com/idris-lang/Idris2/pull/999/files#diff-9d7142eda56885e2568717315a20b0cd73a81dca65f17e6a058807bd90196e94R195

It would also be an error in the version I propose since the last set of `"""` need to be on the next line in order to indicate how much whitespace to trim.

ohad:
Ah, I missed that. But that's orthogonal, and would be ironed out quickly.

andrevidela:
Please tell us ! I would love to know how :)

ohad:
There should be a way to have a raw string that doesn't gobble up the indentation, probably by adding a qualifier to the initial """ as in one of your suggestions, e.g.:
```idris
"""foo
   bar""" == "foo\n   bar"
t"""foo
    bar""" == "foo\nbar"
```

andylokandy:
### Replying to @andrevidela's detailed discussion

> Leading line breaks aren't trimmed away

This looks like a strange request to me. I don't know if any popular language's raw string behaves like that. It involves a special rule that the line wrap very after the `"""` will be ignored. 

I wouldn't say these two below have the same output if I didn't read into the language spec.

```
"""first line
second line"""
```

```
"""
first line
second line"""
```

> The trimming is done using the leading """ instead of the closing one

It has trade-offs: 

1. We can't have indent spaces in the last line.
2. People are hard to tell whether the start quote or the end quote determines the indentation.

> No words about edge cases

This implementation adopted a dead simple rule: if the first `"""` is met, consume the string until the next `"""` is found. So many of your examples should be invalid.

### Conclusion

From the discussions above a few syntax options are revealed:

- Q1. Whether to strip margin or not?
    1. Strip margin.
        1.1 Use the indent of the first quote. (current commit)
        1.2 Use the indent of the last quote. (@andrevidela)
    2. Don't strip margin. (https://github.com/idris-lang/Idris2/pull/999/commits/91caf09bd81ac4735d7717551cfe5393c6fbf452)
   
- Q2. Whether to allow dynamic escape?
    1. Don't allow. (current commit)
    2. Rust-style escape. (@andrevidela)
   
- Q3. Whether to strip the first line wrap or not?

andrevidela:
Thanks for you answer @andylokandy and apologies for the wall of text. I think your conclusion sumarises perfectly the situation but I see that I haven't expressed what motivated this entire discussion in the first place well enough. I think we can summarise it with the following statement:

> Raw and multi-line string are fundamentally incompatible

Raw-strings aim to render the content of the string _as  is_ without any change to its representation, disregarding escape characters or double quotes `"`. Multi-line strings however allow some affordances in trimming whitespace, allowing escape characters, and eventually allowing interpolated strings with #555. And as I understand your motivation of rendering file contents, interpolated strings is definitely something that should work in multi-line strings. And interpolated strings can _never_ be allowed in raw strings. With this information we can add to your conclusion how each conclusion fits with the hypothesis that we should separate raw and multiline strings:

- Q1. Whether to strip margin or not?
  1. strip margins (multi-lines trings, not raw strings)
  2. don't strip (raw strings, which means no interpolation)
- Q2. Whether to allow dynamic escape?
  1. Don't allow (raw strings, but we can't document any syntax which has `"""` as tokens)
  2. Rust-style escape (raw strings)
- Q3. Whether to strip the first line wrap or not?
  1. trim it (multi-line strings, forbit any line that starts with `"""\w`)
  2. Don't trim (raw-strings, but now the text is jagged between the first line and the remaining ones)

ohad:
Looking at #1004 , I think the answers are:

    Q1. Whether to strip margin or not?
        A: don't strip
    Q3. Whether to strip the first line wrap or not?
        A: Don't trim 

Re. Q2: since implementing the Rust style variable delimiters might need a more sophisticated implementation, I propose we use a future compatible version that fixes the delimiter. For example:
```idris
let hello = r"""This is a raw string"""
```
When we get to implement the Rust-style variable delimiters, this will be a special case where the special delimiter `xy"` is given by `x := y := "`.

andylokandy:
@ohad I think the dynamic delimiter shouldn't be the quote itself. Or else we can't tell where is the start of the content.

ohad:
Depends how you set it up. In any case, I have no personal preference, and I now see you've implemented the more general mechanism. Great, and thanks!

andylokandy:
Sorry for changing the PR back and forth. I finally convinced myself that raw string and interpolation string are orthogonal in the Swift way. And I've found it easy to implement.

andrevidela:
Yes they are! This is what I mean by "discussing the design goals of raw strings" in #1004 Raw strings in swift are a modifier over string literals, which also applies to multi-line string literals. But that is not the case with the Rust design of raw-strings. I'll review this tomorrow

src/Parser/Support.idr
```diff
@@ -138,64 +138,65 @@ getEsc "SP" = Just '\SP'
...
+                                     Nothing => case getEsc (fastPack (the (List _) [a, b])) of
+                                                     Just v => Just (v :: !(escape' escapeChars (c :: rest)))
+                                                     Nothing => escape' escapeChars xs
+                              ([], (a :: b :: [])) =>
+                                case getEsc (fastPack (the (List _) [a, b])) of
+                                     Just v => Just (v :: [])
+                                     Nothing => escape' escapeChars xs
+                              ([], rest) => escape' escapeChars rest
+                              (ds, rest) => Just $ cast (cast {to=Int} (fastPack ds)) ::
+                                              !(escape' escapeChars rest)
```
andrevidela, line 178: This is big enough that it should probably be in its own function

src/Parser/Support.idr
```diff
@@ -138,64 +138,65 @@ getEsc "SP" = Just '\SP'
...
-           (ds, rest) => Just $ cast (cast {to=Int} (fastPack ds)) ::
-                                 !(assert_total (escape' rest))
-escape' (x :: xs) = Just $ x :: !(escape' xs)
 
 export
-escape : String -> Maybe String
-escape x = pure $ fastPack !(escape' (unpack x))
+escape : Nat -> String -> Maybe String
+escape hashtag x = let escapeChars = '\\' :: replicate hashtag '#' in
+                       pure $ fastPack !(escape' escapeChars (unpack x))
```
andrevidela, line 194: This PR is a good opportunity to tidy up this kind of thing!
```suggestion
                       fastPack <$> (escape' escapeChars (unpack x))
```

andrevidela (commented):
Looks good! I only have some cosmetic suggestions

src/Parser/Support.idr
andylokandy: It's going to be fully refactored in multi-line PR, therefore we can leave it for now.

andrevidela: sounds good

andrevidela:
Looks good!

buzden:
Maybe, some documentation should be added and changelog updated?

andrevidela:
Yes, I'll update the issues and add the tracking for documentation and changelog tomorrow

PR #1001, opened 2021-01-29
Update Nix version in CI

### LINKED ISSUE #957: Flaky Nix CI Runs

Infinisil:
Fixes #957

### DIFF
--- a/.github/workflows/ci-nix.yml
+++ b/.github/workflows/ci-nix.yml
@@ -17,7 +17,7 @@ jobs:
           fetch-depth: 0
     - uses: cachix/install-nix-action@v12
       with:
-        install_url: https://github.com/numtide/nix-flakes-installer/releases/download/nix-3.0pre20200820_4d77513/install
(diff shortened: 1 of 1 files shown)


edwinb:
I don't know enough about Nix (which is why I worry a little bit that we have Nix support in the repo at all) but on the basis that there was a :heart: from the initial issue reporter, and it doesn't affect anything else, I'll merge this now. Thanks for the fix.

Issue #1004, opened 2021-01-30
Raw strings + multi-line string discussion


andrevidela:
This proposal aims to motivate and explain the design behind two new string features in Idris: Raw and multi-line strings. Each proposal has a motivation, from which follow design goals. The proposal answers the design goals at the end, a discussion section talks about the different alternatives and how they do, or do not answer the design goals stated. This 3000 words long expose ends with a note on implementation and an invitation to discuss the design goals.

# Multi-line strings

Mutli-line strings are a feature that allow to write string literals across multiple lines.

## Motivation

Strings are ubiquitous in programming and allow programmers to communicate through text. Another use for strings is to carry data that will be used by other systems that read strings. Typical examples of this are:

- Generate some HTML headers
- Print out some ascii-art.
- Write out manually an HTTP request.
- Output files given a known template.

What every example has in common is to write down a string literal across multiple lines. Additionally, some of them would benefit from string interpolation #555 while others would benefit from avoiding character escapes. The first two examples would benefit from avoid escaping characters, while the last two would benefit from string interpolation.

Here are each example with existing string syntax

### HTML header

```idris
let header = "<head>\n    <script type=\"text/javascript\" src=\"./tracking.js\"></script>\n</head>"
```
or
```idris
let header = "<head>\n"
          ++ "    <script type=\"text/javascript\" src=\"./tracking.js\"></script>\n" 
          ++ "</head>"
```

in order to show the layout of the file using vertical space.

### ascii-art

This example is taken from the idris compiler itself

```idris

banner : String
banner = "     ____    __     _         ___                                           \n" ++
         "    /  _/___/ /____(_)____   |__ \\                                          \n" ++
         "    / // __  / ___/ / ___/   __/ /     Version " ++ showVersion True version ++ "\n" ++
         "  _/ // /_/ / /  / (__  )   / __/      https://www.idris-lang.org           \n" ++
         " /___/\\__,_/_/  /_/____/   /____/      Type :? for help                     \n" ++
         "\n" ++
         "Welcome to Idris 2.  Enjoy yourself!"
```

As you can see this suffers from two problems: the `\` needs to be escaped, and the version number needs to be inserted. In this proposal I will not suggest to fix both of them at the same time but will provide a solution for each of them individually, leading to two different compromises that have different properties.

### Write out manually an HTTP request.

```idris
testRequest : String
testRequest = 
"POST /cgi-bin/process.cgi HTTP/1.1\n" ++ 
[...]

PR #1010, opened 2021-01-31
Support multi-line string

andylokandy:
Based on #999 and blocked on https://github.com/idris-lang/Idris2/pull/1000.

This PR implements the multi-line string as described in https://github.com/idris-lang/Idris2/issues/1004.

In brief, a multi-line string will:

1. trim the first and the last line.
2. trim the margin regarding the indentation of the tail quote. 
3. escape characters as the normal string.

With multi-line string, the example in #999 can be written in:
```idris
template : String
template = """
    [project]
    name: "project"
    version: "0.1.0"

    [deps]
    "semver" = 0.2
    """
```


andylokandy:
I can't see why the test framework outputs an extra line...

andylokandy:
Closed due to https://github.com/idris-lang/Idris2/pull/1097

PR #1017, opened 2021-02-02
[ fix #835 ] Keep names of implicit variables in with clauses

### LINKED ISSUE #835: With makes where-clause parameters turn into `_`
This might be related to #834, but I'm not sure.

# Steps to Reproduce
```idris
import Data.Nat

foo : {P : Nat -> Type}  -> Dec (P 0)
foo = ?rhs1
  where
      bar : {P2 : Nat -> Type}
        -> (x : Nat)
        -> P2 x
      bar x with (S x)
      bar x | x1 = ?rhs2
```

# Expected Behavior
The parameter types of the where-bound function should show up like normal
```
Main> :t rhs2
   P : Nat -> Type
   P2 : Nat -> Type
   x1 : Nat
   x : Nat
rhs2 : P2 x
```

# Observed Behavior
`P2` in the goal is printed as `_`.
Also, P2 is missing in the parameter list (same as #834)
```
Main> :t rhs2
   P : Nat -> Type
   x1 : Nat
   x : Nat
rhs2 : _ x
```

gallais:
I'll push a test case tomorrow.


gallais:
Phew! That was quite the hunt!

Issue #1023, opened 2021-02-04
Strange behavior of large hexadecimal literals


stefan-hoeck:
# Steps to Reproduce
Start up a REPL and enter the following expressions:

`the Bits64 0xffffffffffffffff`
`the Bits64 0x8000000000000000`

# Expected Behavior
`the Bits64 0xffffffffffffffff` prints 18446744073709551615 (2^64 - 1: max unsigned 64bit number)
`the Bits64 0x8000000000000000` prints 9223372036854775808 (= 2^63)

# Observed Behavior
`the Bits64 0xffffffffffffffff` prints 9223372036854775807 (2^63 - 1: max signed 64bit number)
`the Bits64 0x8000000000000000` prints 0

Replacing Bits64 by Integer behaves the same. Entering the numbers in decimal form behaves correctly.

gallais:
At least this bug is easy to track down: the code is decoding these literals
into an `Int`. :sweat_smile:

Issue #983, opened 2021-01-22
Missing bitwise operations


stefan-hoeck:
Idris2 provides bitwise operations like `prim__and_Int` or `prim__or_Bits64` for all primitive integer types with the exception of `prim__xor` which seems to be missing for `Bits64` and `Integer`. I found that the corresponding entries are missing from `src/Core/Primitives.idr`. Would it be sufficient to add those missing entries to `src/Core/Primitives.idr` to make operations `prim__xor_Bits64` and `prim__xor_Integer` available or are there other steps necessary (or is there another reason for them being missing)?

Also, would it make sense to make all these binary operations available (and visible) via some interface similar to Haskell's `Data.Bits`? Right now, they seem not to be documented anywhere.

I'll be happy to try and draft some PRs if this is of interest.

gallais:
Sounds very reasonable.

Note that you won't be able to expose the new primitives in the `libs/`
until after the modified compiler has been released.

stefan-hoeck:
I'm afraid I'll need some guidance here, since I'm unfamiliar with bootstrapping issues. If I understand correctly, I have to add the following line to `bootstrap/idris2_app/idris2.rkt` and `bootstrap/idris2_app/idris2.ss`:
```
 (define prim__xor_Integer (lambda (arg-0 arg-1) (blodwen-xor arg-0 arg-1)))
```

Also, I tried modifying `src/Core/Primitives.idr`:

```
bxor : Constant -> Constant -> Maybe Constant
bxor (I x) (I y) = pure $ I (prim__xor_Int x y)
bxor (BI x) (BI y) = pure $ BI (prim__xor_Integer x y)
bxor (B8 x) (B8 y) = pure $ B8 (prim__xor_Int x y)
bxor (B16 x) (B16 y) = pure $ B16 (prim__xor_Int x y)
bxor (B32 x) (B32 y) = pure $ B32 (prim__xor_Int x y)
bxor (B64 x) (B64 y) = pure $ B64 (prim__xor_Integer x y)
bxor _ _ = Nothing
```

I tried to build this, but got an error about `prim__xor_Integer` not being found. Should these two changes be done separately then? Or is the addition to the Scheme and Racket files even necessary, since those functions are exactly the same as prim__xor_Int anyway (it's done this way for other bitwise operations as well, so I just copied that style)?

stefan-hoeck:
Ok, I of course approached this the wrong way. I just found the discussion for #548, so I'll try and go along those lines.

[...]

PR #1026, opened 2021-02-04
[ new ] Bitwise XOR for Bits64 and Integer

stefan-hoeck:
This adds primitives `prim__xor_Bits64` and `prim__xor_Integer` to the compiler as discussed in #983. In addition, tests for all primitive bit ops have been added (which led to the detection of  #1023   :-) ).

I'll submit a new PR for including these in `bxor` in `src/Core/Primitives.idr` once the new compiler is released.


gallais:
You probably need to move your `basic054` test case because #1023 snatched that name.

stefan-hoeck:
> You probably need to move your basic054 test case because #1023 snatched that name.

Hmm, is this still an issue? I merge master into this branch and resolved conflicts (at least I thought so) after #1023 was committed. (See d84a5e8)

gallais:
Oh, I see. It thought you had overwritten `basic054` during your merge.

tests/idris2/basic054/input
```diff
@@ -0,0 +1,2 @@
+:exec main
+:q
```
gallais, line 2: Note that this file is useless: `run` does not use an `input`.

stefan-hoeck: OK. Removed the file.

stefan-hoeck:
> Edit: there are indeed changes though, coming from a broken merge rather
> than pre-emptively moving the test before merging AFAICT.

There is now only `tests/idris2/basic054/run`, which seems to be identical to your version. Why it still appears in the list of changed files, I don't know. Should I try and do something about that?

Also, some tests are failing. Are these failures to be expected here?

stefan-hoeck:
I just saw that I forget the '--cg node' option in the node tests, so please do not merge this yet. Will look into it tomorrow.

stefan-hoeck:
The node tests fail in several ways (for instance #1037 and #1039). I'll therefore do the following: I'll try and fix those issues first, adding most of the bitwise tests in this PR to the PRs with the fixes. Once the fixes are merged, I'll rebase and continue here.

stefan-hoeck:
I rebased this and activated the additional xor-tests. This should be ready for reviewing now.

Issue #801, opened 2020-11-28
JavaScript Backend: Ints are considered to be BigInts


DoctorRyner:
# Steps to Reproduce

1. Create a file named "Main.idr"

```idris
module Main

%foreign "javascript:lambda:x => console.log(x)"
prim__log : Int -> PrimIO ()

log : Int -> IO ()
log = primIO . prim__log

main : IO ()
main = log 1
```

2. Compile it with `idris2 Main.idr --codegen javascript -o index.js`
3. Run it with `node build/exec/index.js`

# Expected Behavior

Seeing 1

# Observed Behavior

Seeing 1n

Int in considered as BigInt when it really should correspond to a rounded number, while Integer should stay to be BigInt

rbarreiro:
why?

DoctorRyner:
Because when we use restricted Int we don't really expect to get an unrestricted BigInt. BigInts are in fact rarely used and in most cases we don't want to use them

rbarreiro:
The Number in JavaScript is really a float, hence I am not sure they are a better representation of Int.

DoctorRyner:
Well, PureScript and Elm handle it this way

DoctorRyner:
Oh, and Nat is considered to be a `BigInt` as well, so it means that I can't use Nat for indexing in JS

rbarreiro:
I think that both Elm and PureScript were made before BigInt was widely supported. 
I don't think Nat is supposed to be bounded.

DoctorRyner:
And I think they don't use BigInt so this simple code wouldn't cause any errors

<img width="415" alt="Screenshot 2020-11-29 at 1 40 37 PM" src="https://user-images.githubusercontent.com/24430919/100539576-8aa47580-3248-11eb-80da-a8f9226c4f04.png">
<img width="499" alt="Screenshot 2020-11-29 at 1 40 47 PM" src="https://user-images.githubusercontent.com/24430919/100539578-8d06cf80-3248-11eb-8e7f-51db2c67a5ed.png">

DoctorRyner:
But I'm not sure at all why I get this error

stefan-hoeck:
Sorry for chiming in on this discussion, but I have had to deal with this issue just recently.

[...]

Issue #623, opened 2020-08-31
Fun with Floats (Idris2 edition)


wenkokke:
As mentioned in #601, the results of floating-point arithmetic are non-deterministic, based on architecture and optimisation settings. The solution is to truncate floating-point numbers to 64-bits after every operation which may result in an overflow. However, probably comes with a significant runtime overhead, so I would encourage you to provide a flag akin to `-fno-excess-precision` option, which *optionally* inserts these truncations.

# Steps to Reproduce

Compile and run the following program, using `idris2 crash.idr -o crash && ./crash`:

```idris
module Main

infinity : Double
infinity = 1.0 / 0.0

maxValue : Double
maxValue = 1.7976931348623157e308

isInfinite : Double -> Bool
isInfinite = (== Main.infinity)

el : Bool -> Type
el False = Bool
el True  = Nat

defaultEl : (b : Bool) -> Main.el b
defaultEl False = False
defaultEl True  = 14

inc : Main.el (Main.isInfinite Main.infinity) -> Nat
inc = (+ 1)

cast : Main.el (Main.isInfinite ((Main.maxValue * Main.maxValue) / Main.maxValue))
     -> Main.el (Main.isInfinite Main.infinity)
cast x = x

boom : Nat
boom = inc (cast (defaultEl (isInfinite ((maxValue * maxValue) / maxValue))))

main : IO Unit
main = print boom
```

# Expected Behavior

I'd expect the program to print `15`.

# Observed Behavior

The program prints some large number, which is different every time.

wenkokke:
(See also <https://github.com/agda/agda/issues/4892>.)

Atomotron:
In light of what I'm learning about transcendental functions while thinking about PR #1203, I don't think it will be possible to iron the inconsistencies out of machine floats. To make Idris sound again, we could do the following two things: (both could and perhaps should coexist):

- Introduce a rational or a emulated FP type that does math in a consistent way, based on integers, or perhaps based on a tightly controlled machine float backend as outlined in [this paper](https://drops.dagstuhl.de/opus/volltexte/2019/11062/pdf/LIPIcs-ITP-2019-7.pdf). 

[...]

PR #1033, opened 2021-02-05
Add Data.Bits to base

stefan-hoeck:
This PR adds a new interface for bitwise operations to base. This is part of #983. Implementations for `Bits64` and `Integer` will be added as soon as a new compiler including the changes from  #1026 is released.

I decided against adding an implementation for `Int`. The `Int` type seems not to be specified very well. It is, for instance, unclear, whether it is supposed to be a signed 64bit or 32bit-precision integer, or whether this depends on the backend and operating system. Also, some bitwise operations on negative numbers seem to behave rather strangely. For instance, on Chez Scheme (64bit Linux), `prim__shl_Int 1 63 = 0`, while on GHC Haskell, the same expression yields `-9223372036854775808`. I'm nut sure whether I should file a bug report about this, or whether we should just agree on not officially supporting `Int` via `Data.Bits` for now.

Finally: For the time being we only support `shiftL`, `shfitR`, `and`, `or`, and `xor` as primitive operations. All other bitwise operations are implemented in terms of these core functions. This is probably not ideal, since typically backends provide optimized versions of these functions already (especially `popCount` is most certainly far too slow). I'm not sure what the general opinion about adding additional primitives (which have to be supported across all backends) is. I'd suggest, we release this with unoptimized implementations for the time being, adding additional primitives on demand.


gallais (commented):
Why is it that `shift`, `bit`, and all other operations talking about
positions take an `a` rather than a `Nat`?

stefan-hoeck:
> Why is it that shift, bit, and all other operations talking about
> positions take an a rather than a Nat?

That's the types of the primitive functions, which I didn't touch. If you consider casting a Nat to a primitive to be low enough, we could opt for more safety here and use Nats for positioning. People requiring raw performance could still call the prim__ versions directly.

peterohanley:
Can shiftL/R be removed from the prelude, then? They're not used anywhere but libs/contrib/System/Random.idr and it's a bit odd to have duplicated members of an interface but for only one type just floating around. https://github.com/idris-lang/Idris2/blob/8102e9e49510d783b1e9d587e97676f20802995e/libs/prelude/Prelude/Interfaces.idr#L65

stefan-hoeck:
> Can shiftL/R be removed from the prelude, then? They're not used anywhere but libs/contrib/System/Random.idr and it's a bit odd to have duplicated members of an interface but for only one type just floating around.

Oh dear, I forgot about that one. As stated in the initial comment, I didn't plan to add an instance of `Data.Bits` for `Int` since `Int` is not well specified. How should we go about this? I could add an instance of `Bits` but not `FiniteBits`, since the bit size of `Int` can vary across backends. There is, for instance, a discussion going on, whether `Int` on JS backends should be represented by `Number` (see  #801), in which case it could not be considered a 64bit number on that backend.

stefan-hoeck:
Seems like `shiftL` and `shiftR` are also used in the compiler in `src/Core/Primitives.idr`. gallais, could you quickly comment on this: Will it be safe to remove `shiftL` and `shiftR` from the Prelude and replace them with calls to the `prim__` versions in `src/Core/Primitives.idr`?

Also, should we consider removing bit ops (the primitives as well) from `Int` altogether? I think the main issue here is, that we use modular arithmetic to deal with `Int` overflows, which leads to incorrect behavior when left shifting positive values in such a way that they should actually become negative. We could take the position that users should stick to Bits types (or Integer, where the bit ops are properly supported by the backends) if they need to do bit operations.

gallais:
> Seems like shiftL and shiftR are also used in the compiler in src/Core/Primitives.idr. gallais, could you quickly comment on this: Will it be safe to remove shiftL and shiftR from the Prelude and replace them with calls to the prim__ versions in src/Core/Primitives.idr?

Yes that would be fine given it's what we are doing for all of the other primitives.

> Also, should we consider removing bit ops (the primitives as well) from Int altogether?

I think we should fix `Int` to a specific width and provide `Int32` and `Int64` versions
of the primitives so that user may pick the precision they want. As we have already seen
with floats (#623) letting different backends use different precisions is a *very* bad
idea for consistency.

stefan-hoeck:
> We could maybe have an associated index type in the Bits interface?

Note, that this would not prevent us from shifting an 8 bit number by - for instance - 64 bit, since we could just call `shiftL` repeatedly. It would, however, make sure we stayed within bounds if one of the backends chose to throw an exception in case of a too large shift. I'm also a bit concerned about the additional verbosity for using such basic functions (a weak argument, I know). As we often shift by a fixed number of bits (known at compile time), we should at least have an implementation of `fromInteger` with some auto implicits for `Subset`, so users are still able to conveniently write something like

```idris
the Bits8 13 `shiftL` 2
```

> I think we should fix Int to a specific width and provide Int32 and Int64 versions of the primitives so that user may pick the precision they want.

This sounds very reasonable. We'll still have to decide what the precision of `Int` should be (probably 64 bit, as this is how it is implemented right now in JS and Scheme), but that discussion should probably go into #1048.

gallais:
> since we could just call shiftL repeatedly.

Right, it's not as important for `shift(L/R)` but it is a bit more
for `complementBit` and `bit` and `clearBit` and `setBit`.

> we should at least have an implementation of fromInteger with some auto implicits for Subset

We know that `lt` on `Nat` reflects `LT`. So we can indeed have a

```idris
fromInteger : (k : Nat) -> {n : Nat} -> {auto 0 _ : lt k n === True} -> Subset Nat (`LT` n)
```

that will be guaranteed to work.

stefan-hoeck:
Suggestions from reviewers have been implemented:

- use an `Index` type in Data.Bits
- added some proofs to Data.Nat (such as `ltReflectsLT`)
- remove `shiftL` and `shiftR` from Prelude and adjusted usage of these functions in the compiler and libraries

Thanks for all the valuable suggestions.

stefan-hoeck:
Argh, some typedd-book tests use `shiftR` from the Prelude. In order not to diverge too much from the book, I'll try to do the following over the weekend: Fix bitops implementations for `Int` and add a (hopefully correct) implementation of Data.Bits for `Int`, fixing it at 64bit as this is how it is implemented on all backends right now. I'll then fix the typedd tests.

stefan-hoeck:
This should once again be ready to review. The gambit test was failing, though I don't know why. I changed it to actually build an executable, which works fine on my system but takes ages on github, so I'm not sure what to do about this. In addition, I fixed the typedd tests to use Data.Bits and updated the docs accordingly.

stefan-hoeck:
Rebased and cleaned up commit history.

stefan-hoeck:
gallais, I think I need some help here. With a fresh make and install, tests like `chez/chez032` fail on my system. The culprit seems to be the following line

```idris
intmin : Int
intmin = prim__shl_Int 1 63
```

which is evaluated to 0 at compile time instead of -9223372036854775808. However, if I build and install the latest commit twice, everything works fine as it does in the CI on github. Do you have an idea what's going on here? If I change this to the following version, everything works OK:

```idris
shl63 : Int -> Int
shl63 = (`prim__shl_Int` 63)

intmin : Int
intmin = shl63 1
```

This seems also to be the reason why the gambit test fails.

stefan-hoeck:
Fixed and rebased (gambit problem was introduced by my own carelessness). Please review.

gallais (approved):

libs/base/Data/Bits.idr
```diff
@@ -0,0 +1,192 @@
...
+--------------------------------------------------------------------------------
+
+public export
+interface Bits a => FiniteBits a where
+  ||| Return the number of bits in values of type `t`.
+  bitSize : Nat
+
+  ||| Returns the bitwise complement of a value.
+  complement : a -> a
+
```
buzden, line 173: A tempting thing to ask is why not to have a function connecting a `bitSize` or `FiniteBits` with `Index` type in `Bits` like
```idris
0 bitToIndex : Fin bitSize -> Index
```
I can imagine why they might not be isomorphic, but the function working at least one way should exist.

stefan-hoeck: Could you please elaborate on this? Do you see a use case for `bitToIndex` other than as a proof that `bitSize` and `Index` are properly correlated? Also, why might they not be isomorphic?

buzden: > Do you see a use case for bitToIndex other than as a proof that bitSize and Index are properly correlated?

Well, no, what I originally suggested is purely a proof of proper correlation and that's why it has `0` quantity.

However, it can be useful for generic programming, i.e. for code working with any type satisfying `FiniteBits`. Now you cannot use any shifting and bit setting functions unless you know exactly the type. Having a `bitToIndex` function you can have generic functions (even if you'll need some requirement on `bitSize` value in its signature). But then, `bitsToIndex` shouldn't have zero quantity.

> Also, why might they not be isomorphic?

I can imagine a proper instance for finite-bits-type (like, e.g. `Bits32`) with `Index` being `Integer`. In this case there is no proper function of type `Index -> Fin bitSize`. This is very artificial example and maybe it's better to require an isomorphism between `Fin bitSize` and `Index` -- I'm just not sure about this.

When requiring an isomorphism, probably it's better to have a special data structure coding it (i.e. having functions in both directions and showing extensional equality of applications of these functions to each other with the identity function).

stefan-hoeck: Thanks, buzden, for your kind explanations.

Since this PR is already quite big enough, I'd rather not blow it up any more by adding a new data type/interface for `Isomorphism`. But I'll give `bitsToIndex` a go, as it makes sense to correlate the two types.

libs/base/Data/Bits.idr
```diff
@@ -23,6 +24,13 @@ fromNat :  (k : Nat)
         -> Subset Nat (`LT` n)
 fromNat k = Element k (ltReflectsLT k n prf)
 
+||| Utility for implementing `bitsToIndex` in `FiniteBits`.
+export
+fromFin : Fin n -> Subset Nat (`LT` n)
+fromFin FZ     = Element 0 (LTESucc LTEZero)
```
buzden, line 30: I believe the right place for this function is in `Data.Fin`, somewhere near for `finToNat`, named like `finToNat'` or `finToBoundedNat`.

libs/base/Data/Bits.idr
```diff
@@ -0,0 +1,200 @@
...
+public export
+interface Bits a => FiniteBits a where
+  ||| Return the number of bits in values of type `t`.
+  bitSize : Nat
+
+  ||| Properly correlates `bitSize` and `Index`.
+  bitsToIndex : Fin bitSize -> Index {a}
+
+  ||| Returns the bitwise complement of a value.
+  complement : a -> a
```
peterohanley, line 176: Does `complement` belong in `FiniteBits`? `Integer` isn't `FiniteBits` yet

```
complement : Integer -> Integer
complement x = xor x -1
```

just like all the other implementations. Chez scheme `logxor` supports bignums at least. The missing piece is a `oneBits` in `Bits`

stefan-hoeck: I assume with `oneBits` you mean the value with the first bit set to 1?

peterohanley: No, all bits set to one. In twos complement that's -1, as in all the current implementations

libs/base/Data/Bits.idr
gallais: `Fin` is not efficient at the moment so you may want to have

```idris
bitToIndex : Subset Nat (`LT` bitSize) -> Index
```

instead.

libs/base/Data/Bits.idr
peterohanley: oneBits = complement zeroBits

libs/base/Data/Fin.idr
```diff
@@ -54,6 +55,13 @@ finToNatInjective FZ     (FS _) Refl impossible
 finToNatInjective (FS m) (FS n) prf  =
   cong FS $ finToNatInjective m n $ succInjective (finToNat m) (finToNat n) prf
 
+||| Convert a Fin to a bounded Nat
+public export
+finToBoundedNat : Fin n -> Subset Nat (`LT` n)
+finToBoundedNat FZ     = Element 0 (LTESucc LTEZero)
+finToBoundedNat (FS x) = let Element k prf = finToBoundedNat x
+                          in Element (S k) (LTESucc prf)
+
```
buzden, line 64: Being rebased to the current `master`, the second case could be simplified to
```idris
finToBoundedNat (FS x) = bimap S LTESucc $ finToBoundedNat x
```

stefan-hoeck: Neat! However, I'll rebase this and undo the addition of `finToBoundedNat`. After changing `bitsToIndex` to the more efficient type suggested by gallais, `finToBoundedNat` is unrelated to this PR.

libs/base/Data/Bits.idr
stefan-hoeck: I moved `complement` to Data.Bits and added `oneBits`. In addition, I moved `popCount` to `FiniteBits` as it makes more sense there. I just quickly checked with Haskell, where `popCount` is defined on `Bits`: There, `popCount ((-1) :: Integer)` returns -1, probably not what we want.

libs/base/Data/Bits.idr
stefan-hoeck: Changed `bitToIndex` as suggested by gallais.

libs/base/Data/Bits.idr
peterohanley: 👍

Just for completeness, complement in Bits prevents an instance for "unsigned bignums" e.g. Nat, but I don't think that instance is a good idea anyway, and adding a third class is not warranted here imo.

gallais:
Ugh. Didn't realise this was taking the gambit test from ~15 minutes to 45+ minutes.
We should maybe consider turning that test off if it cannot be made to be faster.

stefan-hoeck:
> Ugh. Didn't realise this was taking the gambit test from ~15 minutes to 45+ minutes.
> We should maybe consider turning that test off if it cannot be made to be faster.

I find writing tests for the gambit and racket backends to be quite a pain, to be honest. And yet, these tests are important if we continue supporting these backends as implementations can vary slightly across different Scheme flavors. So yeah, feel free to turn off the gambit test. We probably still need to think about how to make sure, the gambit (and racket) backends behave correctly.

Issue #1027, opened 2021-02-04
Raw strings proposal


andrevidela:
# Current status

This issue keeps track of the progress regarding the raw strings proposal. It comes from #1004.  Here is the todo list:

- [x] Implement raw strings with `#` as a delimiter for raw strings (implemented in #999)
- [x] Implement dynamic delimiters (#1034)
- [x] update changelog
- [x] update documentation

# Summary

While writing string literals one might be inconvenienced by the heavy amount of _escaping_ needed when displaying characters such as `\` or `"`. Raw strings provide syntax to dynamically change the escaping sequence of a string literal such that those character can be written plainly.

# Motivation 

Raw strings is a feature that allow programmers to write strings between special delimiters such that they do not have to escape anything and the string is interpreted as-is without any transformation from the compiler. This allows for the following use cases:

• Writing regex expression while avoiding escape characters
• Writing strings which are heavy in `"` which need escape characters
• Writing strings which contains special character that need to be rendered as is, specially characters like`\n`

Without raw strings, this is how the given example would be expressed in Idris:

```Idris
let regex = "(\\w\\d)+"
```

```idris
let documentation = "the token for multiline string is \"\"\""
```

```idris
let multilines = "in order to go to the next line, use \\n \n like so"
```

Those example get particularly bad when mixed together:

```idris
let regex = "\"\"\"\\n(\\s*\\w+\\n)*\\s*\"\"\""
```

Additionally, raw strings are useful when pasting code from other sources, typically one might want to insert a bash script as a literal string as part of a CI jobs:

```idris
ci_job : String 
ci_job = "rm -rf build/ttc # clean the ttc files to avoid reusing old build\n" ++
"make all # Compile everything \n"++
"make test | tee results.back # save the results as a backup file `results.back` is always the current run\n"++
"cp results.back ../previous\\ runs/test\\ results:\\ $(DATE).back # Save the current run for reference later\n"++
"                                                          # useful to compare a previous run with results.back\n"++
"make clean"
```

The design goal for this proposal is thefore as follow:

• Allow any string to be represented within raw string, including line breaks
• Allow dynamic escape sequences with `\#`

# The proposal

Reuse the syntax and semantics of Swift's raw strings `#"…"#`. It comes with the following benefits:

• Already familiar for Swift programmers
• Handles to escape characters using dynamic delimiters for the string
• The changes from normal strings are minimal

This means the previous examples can be rewritten like this:

```Idris
let regex = #"(\w\d)+"#
[...]

PR #1034, opened 2021-02-05
lexer: add a language-composable tokenizer

andylokandy:
Prerequisite of #555 #1027 

This PR adds a new lexer specification `Tokenizer` to replace `TokenMap`. Featured in supporting language composition, which means a language inside another language. It's incredibly common that language composition is needed. 

## Language composition

For example, a HTML document can embed CSS spec, which will look like:

```html
<!DOCTYPE html>
<html>
<head>
    <style>
        body {background-color: powderblue;}
        h1   {color: blue;}
        p    {color: red;}
    </style>
</head>
<body>
</body>
</html>
```

Yet another example is interpolate string described In #555.

```
let respond = "Your proposal is \{ if accepted then "accepted" else "rejected" }."
```

## Other features

The `Tokenizer` also supports dynamic delimiters for composition, solving the raw string [problem](https://github.com/idris-lang/Idris2/issues/1027#issue-801309426). And as a side-effect, the non-pairing bracket will be precisely identified in the lexer stage, which could be used to improve error messages.

## Example

The old one is out-of-date. For the lastest example, see https://github.com/idris-lang/Idris2/pull/1056.


andylokandy:
cc @andrevidela @ohad

andrevidela:
Does this work with nested comments? Can we already swap out the old tokenizer for this one?

andylokandy:
Should have no problem AFAICS

gallais:
It would be good to benchmark the processing of nested comments with this approach
vs. the old approach. Parsing already takes quite a long time so I think we should
make sure this is not worse before moving onwards. This is a perfect case study.

andylokandy:
According to my benchmark, the result is pretty satisfiable 🤔 

### Many comments
```
{-
"-}"
--}
{-
"-}"
--}
{-
"-}"
--}
......
```

| n | old | new |
| ------------- | ------------- |--- |
| 300k | 1.44s | 1.60s |
| 900k  | 3.10s | 3.60s |

### Many nested comments
```
{-
"-}"
{-
"-}"
{-
"-}"
......
-}
-}
-}
```
| n | old | new |
| ------------- | ------------- |--- |
| 300k | 7.32s | 4.25s |
| 900k  | 24.66s | 12.97s |

src/Idris/Parser.idr
```diff
@@ -390,7 +390,7 @@ mutual
            pure (PDotted (boundToFC fname b) b.val)
     <|> do b <- bounds (symbol "`(" *> expr pdef fname indents <* symbol ")")
            pure (PQuote (boundToFC fname b) b.val)
-    <|> do b <- bounds (symbol "`{{" *> name <* symbol "}" <* symbol "}")
+    <|> do b <- bounds (symbol "`{{" *> name <* symbol "}}")
```
gallais, line 393: This was a bug fix cf. #838. Is this reintroducing the bug?

andylokandy: This PR has a better solution to #838 and without [this problem](https://github.com/idris-lang/Idris2/pull/838#issuecomment-743215738),

gallais: Nice! <3

andylokandy:
I improved the error output for improper bracket paring from this:
```
Error: Parse error at line 5:9:
Expected 'case', 'if', 'do', application or operator expression (next tokens: [let, identifier y, symbol :=, symbol (, literal 42, in, identifier y, symbol +, identifier x, identifier baz])
```
to this:
```
Error: Bracket is not properly closed.

PError2.idr:5:18--5:19
   |
 5 | bar x = let y := (42 in y + x
   |      
```

andylokandy:
@andrevidela @gallais I think I'm generally finished on this PR, PTAL >_o

src/Libraries/Text/Lexer/Tokenizer.idr
```diff
@@ -0,0 +1,148 @@
...
+             let beginTok' = fastPack $ reverse beginTok
+             let tag = tagger beginTok'
+             let beginTok'' = MkBounded (mapBegin beginTok') False line col line' col'
+             let middle = middleFn tag
+             let end = endFn tag
+             let (midToks, (reason, line'', col'', rest'')) =
+                    tokenise end middle line' col' [] rest
+             case reason of
+                  reason@(ComposeNotClosing _ _) => Left reason
+                  _ => Right ()
```
andrevidela, line 119: What was the goal of this line? If i'm not mistaken, this desugars to
```
_<- case reason of
         reason@(ComposeNotClosing _ _) => Left reason
         _ => Right ()
```

which doesn't do anything

src/Libraries/Text/Lexer/Tokenizer.idr
```diff
@@ -0,0 +1,148 @@
...
+                    Either StopReason (List (WithBounds a), Int, Int, List Char)
+    getFirstMatch (Match lex fn) str
+        = let Just (tok, rest) = scan lex [] str
+                | _ => Left NoRuleApply
+              line' = line + cast (countNLs tok)
+              col' = getCols tok col
+              tok' = MkBounded (fn (fastPack (reverse tok))) False line col line' col' in
+              Right ([tok'], line', col', rest)
+    getFirstMatch (Compose begin mapBegin tagger middleFn endFn mapEnd) str
+        = do let Just (beginTok, rest) = scan begin [] str
```
andrevidela, line 106: Since none of this code is monadic it can be implemented with a single `let` :)

andrevidela (commented):
Looks good to me, the only thing bugging me is the repettition in `tokenise` which uses `reject` to stop the lexer. We can achieve this by adding a special `Stop` lexeme but I don't htink it's necessarily in scope for this PR. I can do this change later

src/Libraries/Text/Lexer/Tokenizer.idr
andylokandy: I want to shortcut and return ComposeNotClosing while to do nothing on the other variances.

andylokandy:
> We can achieve this by adding a special Stop lexeme but I don't htink it's necessarily in scope for this PR. I can do this change later

This brings a problem that whether the inner tokeniser knows the closing lexer. IMO, the closing lexer should depend on the stacking state, which means that it can't be a static rule in the inner tokeniser.

src/Libraries/Text/Lexer/Tokenizer.idr
andrevidela: I think you're looking for 

```
case reason of 
  (ComposeNotClosing _ _) ⇒ Left reason
  _ ⇒ ?restOfProgram
  ```

Comment (andylokandy) [reply src/Libraries/Text/Lexer/Tokenizer.idr:119]: I rather the way below, but it's not implemented internally:
```
let (midToks, (EndInput | NoRuleApply, line'', col'', rest'')) =
                    tokenise end middle line' col' [] rest
    | reason => Left reason
```

Comment (andrevidela) [reply src/Libraries/Text/Lexer/Tokenizer.idr:119]: Ah yes that would be nice, maybe it could be recorded in an issue and we could open a PR to implement it

File: src/Libraries/Text/Lexer/Tokenizer.idr (hunk 3 above)
Comment (andylokandy) [reply src/Libraries/Text/Lexer/Tokenizer.idr:106]: Do you like to keep it as it is or to break the rest following `case` as you said above?

Comment (andrevidela) [reply src/Libraries/Text/Lexer/Tokenizer.idr:106]: I think the `case` should be there for sure since the short-circuiting isn't happening right now. It's probably best to clean this up with two `lets`

```
let Just (beginTok, rest) = scan begin [] str
    …
    … in
    case reason of
         (ComposeNotClosing _ _) ⇒ Left reason
         _ ⇒ let …
```

it's not perfect but it removes the repetition, remove the misleading `do-notation` and enable the short-circuiting semantics for `ComposeNotClosing`

Comment (andylokandy) [reply src/Libraries/Text/Lexer/Tokenizer.idr:106]: I personally prefer not to split them even by introducing the repeating `let`, because they are logically in the same layer.

Comment (andrevidela) [reply src/Libraries/Text/Lexer/Tokenizer.idr:106]: Yes that would be ideal but even if you keep the do-notation, you'll have to split the statment in two in order to short-circuit the control flow on `ComposeNotClosing` errors:

```
do let …
   …
   case reason of
        (ComposeNotClosing _ _ _) =>Left reason
        _ ⇒ do let …
```

Comment (andylokandy) [reply src/Libraries/Text/Lexer/Tokenizer.idr:106]: > you'll have to split the statment in two in order to short-circuit the control flow

Why? Seems it's working. What did I miss?

Comment (andrevidela) [reply src/Libraries/Text/Lexer/Tokenizer.idr:106]: It's working but it's not doing what's advertised :) But this is the wrong thread! le me close this and reply on the other

File: src/Libraries/Text/Lexer/Tokenizer.idr (hunk 2 above)
Comment (andrevidela) [reply src/Libraries/Text/Lexer/Tokenizer.idr:119]: This is linked to this https://github.com/idris-lang/Idris2/pull/1034#discussion_r575457543

whether you keep `let` or `do` doesn't matter in terms of functionality. But here we want to short circuit the control flow. 


This is what we have:
```

do let …
   _ <- case reason of -- this line does not do anything
             reason@(ComposeNotClosing _ _) => Left reason
             _ => Right ()
   let …
```

This is what we would like

```
let Just (beginTok, rest) = scan begin [] str
      | _ => Left NoRuleApply
    …
    (midToks, (EndInput | NoRuleApply, line'', col'', rest'')) =
                    tokenise end middle line' col' [] rest
      | (_, (reason, _, _, _) => Left reason
    …
```

but it doesn't exist so here we what we need to do, either:

```
let Just (beginTok, rest) = scan begin [] str
      | _ => Left NoRuleApply
    … in
    case reason of
         Left (ComposeNotClosing _ _ _ ) = > Left reason
         Right _ = > let …
```

Or

```
do let (Just (beginTok, rest)) = scan begin [] str
     | _ => Left NoRuleApply
   let …
   case reason of
        Left (ComposeNotClosing _ _ _ ) = > Left reason
        Right _ = > do let …
```

Comment (andylokandy) [reply src/Libraries/Text/Lexer/Tokenizer.idr:119]: Thank you, that's clear enough. `ComposeNotClosing` is akin to `commit`, which stops trying other branches.

Comment (andylokandy) [reply src/Libraries/Text/Lexer/Tokenizer.idr:119]: I think it became worse....
```idris
    getFirstMatch (Compose begin mapBegin tagger middleFn endFn mapEnd) str
        = let Just (beginTok, rest) = scan begin [] str
                 | _ => Left NoRuleApply
              line' = line + cast (countNLs beginTok)
              col' = getCols beginTok col
              beginTok' = fastPack $ reverse beginTok
              tag = tagger beginTok'
              beginTok'' = MkBounded (mapBegin beginTok') False line col line' col'
              middle = middleFn tag
              end = endFn tag
              (midToks, (reason, line'', col'', rest'')) =
                    tokenise end middle line' col' [] rest in
              case reason of
                  reason@(ComposeNotClosing _ _) => Left reason
                  _ => let Just (endTok, rest''') = scan end [] rest''
                         | _ => Left $ ComposeNotClosing (line, col) (line', col')
                       line''' = line'' + cast (countNLs endTok)
                       col''' = getCols endTok col''
                       endTok' = fastPack $ reverse endTok
                       endTok'' = MkBounded (mapEnd endTok') False line'' col'' line''' col'''
                       Right ([endTok''] ++ reverse midToks ++ [beginTok''], line''', col''', rest''') in
```

andylokandy: CALLBACK HELL!! isn't it?

andrevidela: unfortunately I have to leave right now because of an unrelated obligation but here is what I came up with:

```
    getNext : (lexer : Lexer) ->  (line, col : Int) -> (input : List Char) -> Maybe (WithBounds String, List Char)
    getNext lexer line col str =
      let Just (token, rest) = scan lexer [] str
            | _ => Nothing
          line' = line + cast (countNLs token)
          col' = getCols token col
          tokenStr = fastPack $ reverse token
          parsedToken = MkBounded tokenStr False line' col' line' col'
       in pure (parsedToken, rest)

    getFirstMatch : Tokenizer a -> List Char ->
                    Either StopReason (List (WithBounds a), Int, Int, List Char)
    getFirstMatch (Match lex fn) str
        = let Just (bounded@(MkBounded tok' _ line' col' _ _), rest) = getNext lex line col str
                | _ => Left NoRuleApply
           in Right ([map fn bounded], line', col', rest)
    getFirstMatch (Compose begin mapBegin tagger middleFn endFn mapEnd) str
        = let Just (open@(MkBounded beginTok' _ line' col' _ _),  rest) = getNext begin line col str
                | _ => Left NoRuleApply
              tag = tagger beginTok'
              middle = middleFn tag
              end = endFn tag
              (midToks, (reason, line'', col'', rest'')) =
                    tokenise end middle line' col' [] rest
              in case reason of
                      (ComposeNotClosing _ _) => Left reason
                      _ => let Just (close@(MkBounded entTok'' _ line''' col''' _ _), rest''') = getNext end line'' col'' rest''
                                 | _ => Left $ ComposeNotClosing (line, col) (line', col')
                            in Right ([close] ++ reverse midToks ++ [open], line''', col''', rest''')

    getFirstMatch (Alt t1 t2) str
        = case getFirstMatch t1 str of
               Right result => Right result
               Left reason@(ComposeNotClosing _ _) => Left reason
               Left _ => getFirstMatch t2 str
```

it's not finished but you get the idea, does that seem compelling to you?

andrevidela:
@andylokandy I've tried adding my suggestion in a commit using the web editor but accidentally added whitespace! If you're happy with this suggestion, could you rebase and meld my commit into yours and remove the whitespace?

andrevidela:
looks like I've messed up something with bootstrap, I'll check this tomorrow

buzden:
There was a strange rebase copying commits from the `master` O_O

andrevidela:
I think I messed up something while rebasing a deleted commit…

PR #1036, opened 2021-02-06
[ fix ] All the REPL commands should be in CommandTable

ohad:
For historical reasons, these commands were treated specially, probably since they predate the command table. Move them into the command table so that they appear in the help menu.

I've renamed the execute shell command instruction from `:!` to `:sh` since the parser gets confused and things `:!` is an operator, rather than a shell command. Rightly so.


tests/chez/chez031/input
```diff
@@ -1,4 +1,4 @@
-:load Specifiers.idr
+:load "Specifiers.idr"
```
gallais, line 1: I was somewhat annoyed by this change but I am convinced that this is the right thing
to do given that file names may have e.g. spaces in them.

It probably is the right thing to do for `log` too: at the moment we run into issues
when parts of the chosen "topic" are reserved keywords.

gallais (approved):
Let's go!

tests/chez/chez031/input
clayrat: Ouch, this tripped me up for a few minutes until I decided to scan for latest related commits. We should definitely mention this in the changelog.

jfdm: I think this also breaks the IDE protocol! Idris-mode has stopped working, as I think it is expecting an unquoted string here.

ska80: > I think this also breaks the IDE protocol! Idris-mode has stopped working, as I think it is expecting an unquoted string here.

This problem is fixed in this repo: https://github.com/redfish64/idris2-mode

jfdm: The problem is *not* fixed. The IDE Protocol is now broken and no longer backwards compatible with Idris1 and the idris-mode.

gallais: I don't think backwards compatibility should trump bug fixing (like the fact
`:cd`, `:load`, etc. were implemented in a hacky way bypassing the command
table and were not documented)

jfdm: Right, bug fixing is important and another way to address this with backwards compatibility is to *also* update idris-mode and idris1 with the required changes.

What's the point in having an IDE Protocol if you are not going to make it coherent across the projects! There is no need to have separate modes for Idris.

andrevidela: It seems like the fix is to have the IDE protocol advertise a version number and implementations should match it, `idris-mode` and `idris2-mode` by be a bit too coarse as a versioning strategy.

chrrasmussen: I was also affected by this change, but I still think it make sense to make the code uniform, as in this pull request.

What really seems to be the issue is that some IDE plugins rely on an IDE Mode command called [`:interpret`](https://github.com/idris-lang/Idris2/blob/7ccc47712e4c61991b53b7fcddeaf8187e374007/src/Idris/IDEMode/Commands.idr#L65-L66), which allow arbitrary REPL commands to be run. Note that the IDE Mode command `:interpret` can still be useful if it just delegates input from the user, but it should not be expected to have a stable API (in contrast to the IDE Mode).

One way to fix the problem (at least until IDE plugins switch to solely using IDE Mode commands) is to allow both `:load "Specifiers.idr"` and `:load Specifiers.idr` (with and without quotes) to be valid REPL commands. This mimics the behaviour of the shell, where one needs to use quotes or apostrophes to allow spaces and some other special characters. As a bonus, it would make the REPL more ergonomic for users, as it would not require quotes in simple cases.

jfdm: I agree with @chrrasmussen  we should fix bugs and think maybe about UX. 

Reexamining the idris-mode code, [there is support for a protocol version number](https://github.com/jfdm/idris-mode/blob/master/inferior-idris.el#L69). So in another step we should update the protocol to report the version of the protocol the compiler talks. Thus, if my understanding of emacs is correct we can use this variable to assume a lower bound for when a command can work i.e. when protocol version is good enough for idris 2 things do idris2 two things otherwise fail graceful.

Not to mention, Idris2 *should* have been working in the first place out-of-the-box with the IDE-mode, new commands aside. It *does* implement the IDE protocol after all...

zenntenn:
Technically Idris2 isn't out of the box yet though

PR #1038, opened 2021-02-06
[fix #1037 and #1039] Fix truncation error and Int bit shifts in JS backends

### LINKED ISSUE #1037: Truncation error on Node backend
# Steps to Reproduce
Run

```idris
main : IO ()
main = do printLn (map (*0) [the Bits8 0])
          printLn [the Bits8 0]
          printLn (map (*0) [the Bits8 0] == [0])
```

# Expected Behavior
Prints
```
[0]
[0]
True
```

# Observed Behavior
Prints
```
[256]
[0]
True
```

Node shows similar behavior with other bounded types, printing the theoretically unreachable upper bound in each case.
I'm trying to untangle the generated JS code but am having a hard time figuring out what's going on here.

stefan-hoeck:


stefan-hoeck:
This fixes #1037 and #1039, both of which were found when writing tests for #1026. I fixed both issues together, so I could move the tests from #1026 here.

PR #1040, opened 2021-02-06
Update IOArray.fromList to use HasIO

SmiVan:
There seems to be no apparent reason why this function hasn't been updated to MonadIO and later to HasIO with everything else, I'm guessing it was missed by accident.


gallais:
Thanks!

PR #1047, opened 2021-02-09
[ re #1043 ] Fix typo in support.scm

### LINKED ISSUE #1043: Error in gambit backend: Unbound variable: exact-integer?
Trying to run `:exec main` in #1039 using the gambit backend, I get:

```
*** ERROR IN #<procedure #2> -- Unbound variable: exact-integer?
```

We should fix this & check whether any other primitives are missing!

stefan-hoeck:
This fixes #1043. I added a gambit test for this. It takes quite some time to build, though. I also wanted to add a racket test, but couldn't get racket to run properly on my system (seems to be a known issue on NixOS), so I let that one be for the time being.


gallais:
The CI fails once gambit (4.8.8-3, same version as me) is installed: the test's
output is missing all of the content that was meant to be printed?

Does the ci filter out stderr? Because we are not seeing the error I get.

stefan-hoeck:
> The CI fails once gambit (4.8.8-3, same version as me) is installed: the test's
> output is missing all of the content that was meant to be printed?
> 
> Does the ci filter out stderr? Because we are not seeing the error I get.

I tested this by running the Ubuntu CI from within a debian container, where we have access to gambit 4.9.3 ([see here](https://github.com/stefan-hoeck/Idris2/runs/1863599688)). The test fails there as well (same output: nothing). Could it be that the test takes too long to build (some upper limit for how much time a test may take)?

gallais:
With that latest change, I am now getting:

```
Main> Error: INTERNAL ERROR: compileExpr returned Nothing
```

!

gallais:
I've removed the CI which is unrelated to the typo. I'll open a separate PR for that
and this one can be merged even if we don't have a working CI for gambit yet.

PR #1050, opened 2021-02-09
[ new ] Data.OpenUnion

gallais:
Had to introduce this while looking at Freer and figured it may see good use in `contrib` too.
Note that this is the open union of *families*. But we could equally well define the open
union of types.


libs/prelude/Prelude/Interfaces.idr
```diff
@@ -176,10 +176,16 @@ interface Applicative m => Monad m where
...
 %allow_overloads (>>=)
 
+||| Sequencing of effectful composition
 public export
 (>>) : (Monad m) => m a -> m b -> m b
 a >> b = a >>= \_ => b
 
+||| Kleisli composition of effectful computations
+public export
+(>>>) : Monad m => (a -> m b) -> (b -> m c) -> (a -> m c)
```
buzden, line 186: Why not `>=>`?

gallais: I was reading the Freer paper, that's what they used & I assumed it was standard.
Looking at hoogle, `>=>` seems to have the more precise type (and `>>>` to
be used for arrows in a category which Kleisli arrows are).

gallais: Ha! As it turns out we had some of these in `Control.Monad.Syntax`.
I've moved them all to `Prelude.Interfaces`.

PR #1053, opened 2021-02-10
Generalize Category

michaelmesser:
Thanks to Russoul for telling me to add `| cat`.

### DIFF
--- a/libs/contrib/Control/Category.idr
+++ b/libs/contrib/Control/Category.idr
@@ -4,7 +4,7 @@ import Data.Morphisms
 
 
 public export
-interface Category (0 cat : Type -> Type -> Type) where
+interface Category (0 cat : obj -> obj -> Type) | cat where
   id  : cat a a
   (.) : cat b c -> cat a b -> cat a c
 


gallais:
The generalisation makes sense. But I can't think of a case where you'd know
`cat` but not `obj` so I'm not sure what the point of `| cat` is.

Russoul:
In cases like this:
```
public export
(>>>) : Category cat => cat a b -> cat b c -> cat a c
f >>> g = g . f
```
Idris won't use the provided instance, unless `cat` is made a determining argument.
I can't see myself that this is crucial or even useful at all. The above example was the one found
broken when generalising `Category`, so the `| cat` thing was sort of a fix.

gallais:
What if we're not using `id` and `(.)` for the `Category`'s fields?
It seems dodgy to overload names in such common use.

The two different types of overloading may be interacting in weird ways.

Russoul:
On the one hand, things can potentially become messy with overloading, but we can always disambiguate if needed.
On the other, there is so many different mathematical concepts or structures, if you will, that admit composition and it's natural for us humans to overload one acquainted symbol with it. We also have `%hide` for the rescue :)

Ah, that's one of the examples where simple name synonyms would be useful.

gallais:
`(.)` and `id` are part of the prelude & as a consequence users don't really
have the opportunity to *not* import them. That's why I think that adding
new, conflicting, definitions under the same name is a bad idea.

Can we at least try to use `<<<` instead of `(.)` and see whether that makes
the need for the (IMO hack) `| cat` disappear?

michaelmesser:
I tried renaming `(.)` to `test` and removing `| cat` and I get the same error.
```
While processing right hand side of >>>. Can't find an implementation for Category ?cat.

/Users/michaelmesser/Code/Idris2FRP/src/AutoTest.idr:25:11--25:19
    |
 25 | f >>> g = test g f
    |           ^^^^^^^^
```

PR #1054, opened 2021-02-10
[ new ] support record projections in refc backend

Z-snails:
Fixes records (specifically '.' syntax) being unsupported with refc backend.


gallais:
Can we please add a test case for this in `tests/refc`?

gallais:
Thanks!

PR #1056, opened 2021-02-10
Implement interpolated string

andylokandy:
This PR implements both #1027 and #555, basing on #1034.

For example, the listing below equals to `"Hello Idris2!"`:
```idris 
let idris = "Idris" in 
  "Hello \{idris ++ show 2}!"
```

For more details see #1027.


andrevidela:
we can probably close #1034 and focus on this PR

andylokandy:
@andrevidela Isn't it too large as a whole?

buzden:
> we can probably close #1034 and focus on this PR

It seems that #1034 has its own separate value disregarding raw strings.

andylokandy:
@gallais @andrevidela I've implemented the interpolated string and I reused this PR (originally for raw strings) since they are quite related IMO. 😄

src/Idris/Desugar.idr
```diff
@@ -277,6 +272,21 @@ mutual
       = pure $ IMustUnify fc UserDotted !(desugarB side ps x)
   desugarB side ps (PImplicit fc) = pure $ Implicit fc True
   desugarB side ps (PInfer fc) = pure $ Implicit fc False
+  desugarB side ps (PString fc strs)
+      = case !fromStringName of
+             Nothing => pure $ !(desugurStr strs)
```
gallais, line 277: `desugur` -> `desugar`?

src/Idris/Desugar.idr
```diff
@@ -277,6 +272,21 @@ mutual
...
+             Just f => pure $ IApp fc (IVar fc f)
+                                      !(desugurStr strs)
+    where
+      toRawImp : PStr -> Core RawImp
+      toRawImp (StrLiteral fc s) = pure $ IPrimVal fc (Str s)
+      toRawImp (StrInterp fc tm) = desugarB side ps tm
+      concat : RawImp -> RawImp -> RawImp
+      concat a b = IApp (getFC a) (IApp (getFC b) (IVar (getFC b) (UN "++")) a) b
```
gallais, line 285: Should we instead call `fastConcat`?

andylokandy: I don't know. Is there any case where `++` is needed? If no, should `++` be an alias for `fastConcat`?

src/Idris/Desugar.idr
andylokandy: 👍

src/Idris/Desugar.idr
gallais: `fastConcat` takes a whole `List` of strings and only allocates one big
string that contains the concatenation of them all. `foldr (++)` will
leads to the allocation of many intermerdiate strings.

`(++)` still has its uses when you're only putting 2 strings together.

andylokandy: I see. Another question is that `++` isn't in prelude. So is it a good choice to include ffi in the prelude?

gallais: Ah, I see. I guess it would make sense to add this as a primitive and insist that
each backend should implement it. But we can leave that as future work, in a separate PR.

src/Idris/Pretty.idr
```diff
@@ -277,6 +281,7 @@ mutual
       go d (PSectionR _ x op) = parens (go startPrec x <++> pretty op)
       go d (PEq fc l r) = parenthesise (d > appPrec) $ go startPrec l <++> equals <++> go startPrec r
       go d (PBracketed _ tm) = parens (go startPrec tm)
+      go d (PString _ xs) = parenthesise (d > appPrec) $ hsep $ punctuate "++" (prettyString <$> xs)
```
gallais, line 284: Is there a reason we are we not displaying it as an interpolated string?

gallais (commented):
One last question

src/Idris/Pretty.idr
andylokandy: If I don't understand wrong, this is used to display error message from TTImp, and since resugar is not implemented for interpolated string, it's useless to show it in interploated form.

PR #1059, opened 2021-02-11
Patch CVs and `sleep` in Racket

CodingCellist:
The Racket runtime blocks all threads on a foreign function call, meaning `sleep` has to use `blodwen-sleep` and not `idris2_sleep` in order for it only to pause the calling thread and not the entire runtime. This fixes that and adds comments to clarify this. Also included are some tests of the Racket concurrency primitives (which rely on `sleep`).


tests/racket/conditions005/Main.idr
```diff
@@ -0,0 +1,22 @@
...
+  do cvMutex <- makeMutex
+     cv <- makeCondition
+     ts <- for [1..n] $ \_ => fork $ do mutexAcquire cvMutex
+                                        conditionWait cv cvMutex
+                                        putStrLn "Hello mother"
+                                        mutexRelease cvMutex
+     putStrLn "Hello children"
+     sleep 1
+     conditionBroadcast cv
+     for ts $ \t => threadWait t
```
gallais, line 20: `IO` actions are not allowed to silently discard their result
anymore. So you need an `ignore $` in front of this `for (...)`.

gallais (commented):
I haven't merged this because AFAIU it causes #1060 and we are
using the `Futures` library in the test suite so we can't really
break that.

tests/racket/conditions005/Main.idr
CodingCellist: I thought I'd fixed that, but I must have undone it accidentally. It should be fixed now.

I re-enabled and ran the `Futures` tests, and whatever the issue was, it seems to be gone now. Hopefully it is safe to merge now, but let me know what you think  : )

gallais: Great news! By the way, I checked and we do have

```idris
for_ : (Foldable t, Applicative f) => t a -> (a -> f b) -> f ()
```

but it does not matter whether we use one or the other as they're exactly the same AFAICR.

PR #1063, opened 2021-02-12
[ new ] Perfect binary trees

gallais:


libs/contrib/Data/Nat/Exponentiation.idr
```diff
@@ -0,0 +1,83 @@
...
+import Data.Nat.Order
+import Decidable.Order
+import Syntax.PreorderReasoning
+import Syntax.PreorderReasoning.Generic
+
+%default total
+
+public export
+pow : Nat -> Nat -> Nat
+pow = Mon.(^)
```
buzden, line 16: Now we have `pow` in `contrib` and `power` in `prelude` that are practically the same, right?

gallais: I did not know about `power`. This one is (runtime) efficient.

libs/contrib/Data/Vect/Binary.idr
```diff
@@ -0,0 +1,53 @@
...
+
+public export
+zero : {bs : Bin} -> {n : Nat} -> Path n (suc bs)
+zero {bs} = case bs of
+  [] => Here (fromNat 0 n (ltePow2 {m = n}))
+  (O :: bs) => Here (fromNat 0 n (ltePow2 {m = n}))
+  (I :: bs) => There zero
+
+public export
+weaken : {bs : _} -> Path n bs -> Path n (suc bs)
```
gallais, line 34: TODO: prove this correct. `toNat (weaken p) === toNat p`

libs/contrib/Data/Monoid/Exponentiation.idr
```diff
@@ -0,0 +1,83 @@
...
+modularRec : Monoid a => a -> HalfRec n -> a
+modularRec v HalfRecZ = neutral
+modularRec v (HalfRecEven _ acc) = let e = modularRec v acc in e <+> e
+modularRec v (HalfRecOdd _ acc) = let e = modularRec v acc in v <+> e <+> e
+
+public export
+modular : Monoid a => a -> Nat -> a
+modular v n = modularRec v (halfRec n)
+
+infixl 10 ^
```
mr-infty, line 28: Is there a reason you make it *left*-associative? The common convention seems that exponentation is regarded as a right-associative operation.

gallais: It was a purely type-directed choice. In general `2 ^ m` has type `a` which makes it an
ideal candidate for further `^ m`  action at the same type. If you make it right associative,
the nested call to `^`  uses `Num Nat` instead of `Num a`.

mr-infty: Would that be a problem, if the nested call used `Num Nat`? Right-associativity seems to be the universal convention, in mathematics as well as in programming languages.

buzden: > Would that be a problem, if the nested call used `Num Nat`? Right-associativity seems to be the (near-)universal convention, in mathematics as well as in programming languages.

My experience doesn't let me to agree with you. E.g. [wikipedia does not agree with you either](https://en.wikipedia.org/wiki/Exponentiation#Identities_and_properties)

gallais: This wikipedia article does suggest we should make it right associative:

> Without parentheses, the conventional order of operations for serial exponentiation in superscript notation is top-down (or right-associative), not bottom-up[19][20][21][22] (or left-associative). That is,

mr-infty: > > Would that be a problem, if the nested call used `Num Nat`? Right-associativity seems to be the (near-)universal convention, in mathematics as well as in programming languages.
> 
> My experience doesn't let me to agree with you. E.g. [wikipedia does not agree with you either](https://en.wikipedia.org/wiki/Exponentiation#Identities_and_properties)

This wikipedia page states: 

>  conventional order of operations for serial exponentiation in superscript notation is top-down (or right-associative

How does that not agree with me?

In any case, using right-associativity has practical benefits, since e.g.  `((x^y)^z)^w` and already be written as `x^(y*z*w)`.

gallais:
I have reverted the addition of `weaken : {bs : _} -> Path n bs -> Path n (suc bs)` as
it is not proven correct and I am not at all convinced it is. I'd rather merge this as
is and come back to it later / talk to Wouter to figure out how we can define this one.

libs/contrib/Data/Monoid/Exponentiation.idr
gallais: Fixed. Thanks for the review!

PR #1065, opened 2021-02-14
[ new ] Reflect %hide

Russoul:
Reflecting %hide will help in another project: making constructors/projections/methods to be elaborated inside a nested
namespace. 

That change requires rewriting of qualified names in the Idris 2 source and the libraries.
For example `Prelude.(::)` would need to be written instead as `List.(::)`. But we can't directly fix up those, because
  we'd break backwards compatibility (bootstrapping and self-hosting from previous release). Instead we can rely on elaborator reflection to disambiguate a name based on compile time information.


edwinb:
I see this has been here for a while, sorry about that. Seems reasonable though - if you can bring the conflicts up to date then I'll merge it.

Russoul:
It's a prerequisite for #1112.  So it can wait until I've rebased and finished the latter.

edwinb:
We probably want this, but I see the thing that depends on it still needs some thought and there's plenty of conflicting files. I'm going to close this for now, we can reopen when the other details are resolved.

PR #1070, opened 2021-02-15
Add a total way of reading files in.

mattpolzin:
It's useful to be able to read files in within total code. I figured if I was adding a concept of fuel, I may as well add an efficient way to offset the start of the read operation.

I added tests (and caught a bug in my initial pass at the JS helper implementation so I am glad I did). As far as I could tell, this was the first base library test added.

### DIFF
--- a/libs/base/System/File.idr
+++ b/libs/base/System/File.idr
@@ -320,7 +320,7 @@ withFile filename mode onError onOpen =
      closeFile h
      pure $ Right res
 
-try : (Monad m) => m (Either FileError a) -> (a -> m (Either FileError b)) -> m (Either FileError b)
+try : Monad m => m (Either FileError a) -> (a -> m (Either FileError b)) -> m (Either FileError b)
 try a f = a >>= either (pure . Left) f
 
 readOnto : HasIO io => (acc : List String) ->
@@ -364,7 +364,7 @@ readFilePage offset fuel file
 export
 partial
 readFile : HasIO io => String -> io (Either FileError String)
-readFile = (map $ map (fastAppend . snd)) . readFilePage 0 forever
+readFile = (map $ map (fastConcat . snd)) . readFilePage 0 forever
 
 ||| Write a string to a file
 export
--- a/libs/base/System/File.idr
+++ b/libs/base/System/File.idr
@@ -1,8 +1,9 @@
 module System.File
 
+import public Data.Fuel
+
 import Data.List
 import Data.Strings
-import Data.Fuel
 import System.Info
 
 %default total
--- a/libs/base/System/File.idr
+++ b/libs/base/System/File.idr
@@ -320,6 +320,22 @@ withFile filename mode onError onOpen =
      closeFile h
      pure $ Right res
 
+try : (Monad m) => m (Either FileError a) -> (a -> m (Either FileError b)) -> m (Either FileError b)
+try a f = a >>= either (pure . Left) f
+
+readOnto : HasIO io => (acc : List String) ->
+                       (offset : Nat) ->
+                       (fuel : Fuel) ->
+                       File ->
+                       io (Either FileError (Bool, List String))
+readOnto acc _ Dry h = pure (Right (False, reverse acc))
+readOnto acc offset (More fuel) h
+  = do False <- fEOF h
+         | True => pure $ Right (True, reverse acc)
+       case offset of
+            (S offset') => fSeekLine h >> readOnto acc offset' (More fuel) h
+            0           => try (fGetLine h) (\str => readOnto (str :: acc) 0 fuel h)
+
 ||| Read a chunk of a file in a line-delimited fashion.
 ||| You can use this function to read an entire file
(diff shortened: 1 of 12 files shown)


mattpolzin:
Apparently my changes to the JS helper file triggered GitHub's linter to go crazy on all of the file's historical non-standard choices. I can fix those linter complaints before merge.

mattpolzin:
Well, at least the camel-case errors are probably not worth fixing because I assume the name structure is how Idris is finding the functions via FFI... I may have to figure out how to disable that one.

mattpolzin:
Oi, I don't think there is a configuration for the JS linter being used here (on purpose).

Can we disable for JS? I don't really want to take on the task of redoing the way we name our javascript FFI functions as part of this PR...

libs/base/System/File.idr
```diff
@@ -288,26 +307,61 @@ fPoll (FHandle f)
     = do p <- primIO (prim__fPoll f)
          pure (p > 0)
 
+||| Produce enough @Fuel@ to read the given number
+||| of lines.
 export
-readFile : HasIO io => String -> io (Either FileError String)
-readFile file
+lineCount : Nat -> Fuel
```
gallais, line 313: I think we can reuse `Data.Fuel`'s `limit` instead of introducing this identical definition.

libs/base/System/File.idr
```diff
@@ -288,26 +307,61 @@ fPoll (FHandle f)
...
+|||
+||| On success, returns a tuple of whether the end of
+||| the file was reached or not and the String read in
+||| from the file.
+|||
+||| Note that because we are chunking by lines, this
+||| function's totality depends on the assumption that
+||| no single line in the input file is infinite.
+export
+readFilePage : HasIO io => (offset : Nat) -> (until : Fuel) -> String -> io (Either FileError (Bool, String))
```
gallais, line 334: Would it make sense to get a `List String` instead?

I see that you are using `fastAppend` but that's still a lot of
appending especially if we are primarily interested in the lines? 

I guess we should at least expose the auxiliary function corresponding
to `read offset fuel []` plus the `openFile` error handling.

support/js/support_system_file.js
```diff
@@ -1,46 +1,65 @@
-const support_system_file_fs = require("fs")
+const support_system_file_fs = require('fs')
 
-function support_system_file_readLine(file_ptr){
-  const LF = 0x0a;
-  const readBuf = Buffer.alloc(1);
-  let lineEnd = file_ptr.buffer.indexOf(LF);
+// like `readLine` without the overhaead of copying characters.
```
gallais, line 3: typo: overhead*

tests/base/system_file001/run
```diff
@@ -0,0 +1,15 @@
+# @readFilePage@ uses primitive functions with definitions for both
+# C (supported by most backends) and Node.
+$1 --cg chez --no-color --console-width 0 --no-banner ReadFilePage.idr < input
+$1 --cg node --no-color --console-width 0 --no-banner ReadFilePage.idr < input
+
+# The following backends failed for reasons unrealted to this test. They can be
```
gallais, line 6: typo: unrelated*

gallais (commented):
Looks good!

> Can we disable linting for JS?

As long as the only errors are linting related, I'm happy to merge.
You should be able to turn it off by editing `.github/linters/.ecrc`,
I believe? I hope it's just a matter of adding `"*.js"` next to `"expected"`.

libs/base/System/File.idr
mattpolzin: Doh! Yeah, that'd be the way to go.

libs/base/System/File.idr
mattpolzin: Originally I thought it made sense to leave the return type closer to `readFile`, but I do like your idea of `List String`. 

One awkward thing is that each line still has a newline at the end but this is actually not an unprecedented behavior; I happen to know that in Elixir the `File.stream!/3` function defaults to reading by line with the same behavior (you need to drop the newline from the end of each line if you want to be rid of it).

My intuition for this function would be to return the equivalent of the file after running `lines` on it, but given the precedent I mentioned for the behavior of leaving the newlines in and the fact that doing so means fewer overall changes in this one PR, I vote we leave it this way (that is, change the signature to returning `List String` but accept that newlines are not stripped).

gallais:
One last thing: `fastAppend` has been deprecated in favour of `fastConcat`
(`append` is the name for the function putting 2 strings together rather a list of them).

libs/base/System/File.idr
mattpolzin: Can you elaborate on the following?

> I guess we should at least expose the auxiliary function corresponding
> to read offset fuel [] plus the openFile error handling.

I thought you might mean roughly what I did in my latest changes where I pulled `withFile` and `readOnto` out but I also got a bit lost in the weeds trying to deduplicate code and might have made an ill-advised decision with my `try` function.

gallais (approved):

PR #1078, opened 2021-02-17
DESTDIR support for make install

Qeenon:
This pull request adds support for `${DESTDIR}` environment variable to Makefile files.
It is useful to have to be able to move files after installing, to automate system wide installs.
it is documented here https://www.gnu.org/prep/standards/html_node/DESTDIR.html
it's optional and should not break install if not used.

But ideally I want to have some kind of DESTDIR support on `idris2 --install` too
Here: https://github.com/idris-lang/Idris2/blob/master/Makefile#L140-L144

I expect that it should be used somewhere here: https://github.com/idris-lang/Idris2/blob/master/src/Idris/Package.idr#L366-L367
I think adding ability to pass it as `--install --destdir=` option or checking environment variable could be great.

Thank you.

### DIFF
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -36,6 +36,7 @@ import Idris.REPLOpts
 import Idris.SetOptions
 import Idris.Syntax
 import Idris.Version
+import Idris.Env
 import IdrisPaths
 
 %default covering
@@ -366,8 +367,12 @@ install pkg opts -- not used but might be in the future
          -- Make the package installation directory
          let installPrefix = prefix_dir (dirs (options defs)) </>
                              "idris2-" ++ showVersion False version
+         maybeDestdir <- coreLift $ idrisGetEnv "DESTDIR"
+         let installPath = case maybeDestdir of
+              Just destdir => destdir </> installPrefix
+              Nothing => installPrefix
          True <- coreLift $ changeDir installPrefix
-             | False => throw $ InternalError $ "Can't change directory to " ++ installPrefix
+             | False => throw $ InternalError $ "Can't change directory to " ++ installPath
          Right _ <- coreLift $ mkdirAll (name pkg)
              | Left err => throw $ InternalError $ unlines
                              [ "Can't make directory " ++ name pkg
@@ -379,7 +384,7 @@ install pkg opts -- not used but might be in the future
          -- srcdir/build into it
          traverse_ (installFrom (name pkg)
                                 (srcdir </> build)
-                                (installPrefix </> name pkg)) toInstall
+                                (installPath </> name pkg)) toInstall
          coreLift_ $ changeDir srcdir
          runScript (postinstall pkg)
 
--- a/support/c/Makefile
+++ b/support/c/Makefile
@@ -4,6 +4,7 @@ TARGET = libidris2_support
 
 LIBTARGET = $(TARGET).a
 DYLIBTARGET = $(TARGET)$(SHLIB_SUFFIX)
+IDRIS2_DESTDIR := ${DESTDIR}${PREFIX}
 
 CFLAGS += -O2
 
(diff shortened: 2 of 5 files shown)


support/c/Makefile
```diff
@@ -51,8 +51,8 @@ cleandep: clean
 .PHONY: install
 
 install: build
-	mkdir -p ${PREFIX}/idris2-${IDRIS2_VERSION}/lib
-	mkdir -p ${PREFIX}/idris2-${IDRIS2_VERSION}/include
-	install -m 755 $(DYLIBTARGET) ${PREFIX}/idris2-${IDRIS2_VERSION}/lib
-	install -m 644 $(LIBTARGET)   ${PREFIX}/idris2-${IDRIS2_VERSION}/lib
-	install -m 644 *.h            ${PREFIX}/idris2-${IDRIS2_VERSION}/include
+	mkdir -p ${DESTDIR}${PREFIX}/idris2-${IDRIS2_VERSION}/lib
```
gallais, line 54: Can you please introduce `IDRIS2_DESTDIR` like you did in the main makefile?
This way we'll have the same structure everywhere. Same comment for the `refc`
Makefile.

gallais:
This sounds like a good idea.

> But ideally I want to have some kind of DESTDIR support on idris2 --install too

If you want to have a crack at it, you probably want to have a look at
`idrisGetEnv` in `src/Idris/Env.idr`

Qeenon:
> This sounds like a good idea.
> 
> > But ideally I want to have some kind of `DESTDIR` support on `idris2 --install` too
> 
> If you want to have a crack at it, you probably want to have a look at
> `idrisGetEnv` in `src/Idris/Env.idr`

Yes, I will be happy to try making it, Thank you!

Qeenon:
@gallais I think I need help on this now

ejconlon:
@Qeenon Just a shot in the dark, between the makefile steps you may have to copy each library from the calculated DESTDIR to the libdir in order for Idris to resolve them. For example, contrib probably depends on prelude, but has no way to checking the calculated DESTDIR for those compilation results. (Needless to say I would really love to see this fixed!)

Qeenon:
@ejconlon hello!

if I understand you correct we can do 
`${MAKE} -C libs/prelude install IDRIS2?=../../${TARGET} IDRIS2_PATH=${DESTDIR}${IDRIS2_BOOT_PATH}`
instead of
`${MAKE} -C libs/prelude install IDRIS2?=../../${TARGET} IDRIS2_PATH=${IDRIS2_BOOT_PATH} DESTDIR=${DESTDIR}`

If so I agree - I can try that.

src/Idris/Package.idr
```diff
@@ -365,7 +366,11 @@ install pkg opts -- not used but might be in the future
          -- Make the package installation directory
          let installPrefix = prefix_dir (dirs (options defs)) </>
                              "idris2-" ++ showVersion False version
-         True <- coreLift $ changeDir installPrefix
+         maybeDestdir <- coreLift $ idrisGetEnv "DESTDIR"
+         let installPath = case maybeDestdir of
+              Just destdir => destdir </> installPrefix
+              Nothing => installPrefix
+         True <- coreLift $ changeDir installPath
              | False => throw (InternalError ("Can't change directory to " ++ installPrefix))
```
gallais, line 374: Here the error message still refers to `installPrefix`.
Rather than naming the modified path `installPath`, you may want to shadow `installPrefix` tbh.

gallais (commented):
This looks good!

We should probably try to add a CI target with an unusual `DESTDIR` to see
whether we are missing anything.

src/Idris/Package.idr
Qeenon: that's mirror thing... also I guess CI will fail for now because bootstrapping version has no this changes

chrrasmussen:
Hi! I might be missing something but to me it looks like the new environment variables `DESTDIR` and `IDRIS2_DESTDIR` are just new names for the existing environment variables named `PREFIX` and `IDRIS2_PREFIX`.

It should already be possible to set a new destination directory by using `IDRIS2_PREFIX=/another/path idris2 --install idris-api.ipkg`. See https://github.com/idris-lang/Idris2/blob/9be15e7e61db7eefb78d36f859ed4f9b24c7aadc/src/Idris/Driver.idr#L49-L52 and https://github.com/idris-lang/Idris2/blob/9be15e7e61db7eefb78d36f859ed4f9b24c7aadc/src/Idris/Env.idr#L22

`DESTDIR` is maybe a better name than `PREFIX`, but I think that if we do such a change, we should get completely rid of `PREFIX`.

Qeenon:
@chrrasmussen prefix is currently using as runtime install path, how it will be use used after installation, it's including idris packages, binaries and also libraries. I think very default system level value for prefix is `/usr`, where DESTDIR could be temporary directory from which you can `install` (move) files for later work, for example `/var/tmp/idris-5` so with `DESTDIR=/var/tmp/idris-5 make install` we want to make working installation inside temp directory, it will create `/var/tmp/idris-5/usr/...` and later by installing from temporary directory to system root `/` we expect everything to work properly.

Qeenon:
(actually maybe it's no need for idris --install because it can be handled only with makefiles)
I might try making it as optional workflow in another branch.

Qeenon:
There is second variant: https://github.com/idris-lang/Idris2/pull/1123
Honestly I prefer second variant unless in future `idris --install` will remember what was passed into `IDRIS2_PREFIX`

Issue #592, opened 2020-08-24
Add exponentiation for Integers


thomasdziedzic:
I would like to write something like:
`2^64` and get back `18446744073709551616` like in haskell

I know about the Nat version;
Nats have a power function defined under `Data.Nat` but this is slow for `power 2 64`

@ohad mentioned I should link this issue to https://github.com/idris-lang/Idris2/issues/29 as it might be easy to add this feature along with the Double constants.

mr-infty:
I've made a draft PR (#1084) that fixes this issue. I've chosen to ignore issue #29, as it seems unrelated. Incidentally, I'm still working on fixing the latter issue, but this will take a bit more time.

gallais:
Fixed by #1063 (cf. `Data.Monoid.Exponentiation`)

PR #1084, opened 2021-02-18
Added exponentation operator for numerical types.

mr-infty:
An exponentation operation `x ^ n` is added for all types implementing `Num` and natural numbers `n`. In particular, this fixes issue #592.

## Remarks regarding the implementation

The operator `(^)` is right-associative, following mathematical
conventions, with precedence level `10`.

It is implemented using repeating squaring. This is reasonably fast, e.g. computing `2 ^ 2 ^ 2 ^ 2 ^ 2` takes under 10 seconds.

## Justification for the draft PR

Currently, there is no exponentation operator defined (for any of the numerical types), even though there is a power function for natural numbers. This power function is very slow however.

## Problems with this draft

Even though this operator makes sense for any numerical type, it is defined in the module `Data.Nat.Views`, which seems a rather odd place. However, I couldn't place it in `Prelude.Num` because it uses the recursive half-view for natural numbers, and even if it wasn't, it would still result in a circular module dependency as `Prelude.Num` is imported in `Prelude.Types` (where `Nat` is defined).

I therefore welcome any suggestions for a better placement.

### DIFF
--- a/libs/base/Data/Nat/Views.idr
+++ b/libs/base/Data/Nat/Views.idr
@@ -37,3 +37,14 @@ halfRec n with (sizeAccessible n)
           in HalfRecEven _ (halfRec (S k) | acc (S k) (LTESucc (LTESucc (lteAddRight _))))
     halfRec (S (k + k)) | Access acc | HalfEven k
       = HalfRecOdd _ (halfRec k | acc k (LTESucc (lteAddRight _)))
+
+public export
+||| Raises a number to a power which is a natural number.
+(^) : Num ty => ty -> Nat -> ty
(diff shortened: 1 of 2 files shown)


gallais:
I have an implementation for arbitrary monoids in #1063

mr-infty:
> I have [an implementation for arbitrary monoids](https://github.com/idris-lang/Idris2/pull/1063/files#diff-ba4efad0a452e5f76025aef1ba7a12b14cc26bad7a37873c158ff204ff4ea17c) in #1063

Ah, my bad, I didn't see that!

PR #1089, opened 2021-02-19
More compose instances and one usage of them

buzden:
This is in fact a further push of #1055 and applying these ideas to replacing an ad-hoc function in the existing code to use more abstract instances.


stefan-hoeck:
Would it make sense to make the interface implementations `public export`? The definitions are very small and could be useful when proving monad laws elsewhere.

Russoul:
Now I feel embarrassed about my bulky implementation of monad composition I mentioned on slack 😆. But anyway, useful stuff !

buzden:
> Would it make sense to make the interface implementations `public export`? The definitions are very small and could be useful when proving monad laws elsewhere.

Well, I think it would.

PR #1091, opened 2021-02-19
Cleanup List1

### LINKED ISSUE #1076: Bad definition of `Data.String.lines`
# Steps to Reproduce

```idris
import Data.String

main : IO ()
main = do putStrLn $ show $ unlines (lines "")
          putStrLn $ show $ unlines (lines " ")
```

# Expected Behavior

```
""
" "
```

# Observed Behavior

```
""
" \n"
```

andylokandy:
- fixes #1076.
- cleanup `List.foldr1`, `List.foldl1`, `List1.foldr1`, `List1.foldl1`.
- add `List.foldr1By`, `List.foldl1By`, `List1.foldr1By`, `List1.foldl1By`. (name inspired by `List.spanBy`)
- add `List.toList1` that converts `List` to `List1` with a proof of `NonEmpty`
- `Data.String.lines` returns `List1`.
- `Parser.some`, `Parser.endBy1`, `Parser.sepBy1`, `Parser.sepEndBy1`, `Parser.someTill` return `List1`

The drawback is adding redundant `forget`s, which could be solved by implementing implicit conversion.


stefan-hoeck:
Would it make sense to publicly export the new List1 functions? Their implementations are short, and they might be useful in proofs as well.

andylokandy:
> Would it make sense to publicly export the new List1 functions? Their implementations are short, and they might be useful in proofs as well.

They used sub-functions which made it not useful in proof. I think `List` version is preferred in this case.

gallais:
Can we get this up to speed & backwards compatible (the released version of the
stdlib doesn't have `(:::)` IIRC)? The changes seem worth keeping.

gallais:
I believe it's on purpose that you can just run `idris2 --build idris2.ipkg`
with the previous version to build the new one.

andylokandy:
So the test must fail because it's breaking change....

andylokandy:
I'm not sure if it's a good idea to release a going-to-be-broken `List1. foldl1` and its cousins in 0.4.0. It should be fine that we're just breaking some development version between 0.3.0 and 0.4.0, an even not breaking if the libs is built before idris2 compiler, which is how rustc does.

andylokandy:
@gallais I've made it compile by the previous release.

PR #1093, opened 2021-02-19
[ new ] Proof search from 'Applications of Applicative Proof Search'

gallais:
I'm halfway through the paper and there's already some good content in there.
Also allowed me to find some bugs in base:

* `Exists`'s `type` was given type `_` instead of `?` which meant it was not considered a type
  (or at least complaining loudly)
* `Alternative`'s `(<|>)` was not lazy in its second argument, making any proof search an exhaustive
  search of the *whole* space AFAIU.

There's also some non-search related goodies like:

* Cantor's zig zag function exploring a plane `Stream (Stream a)` to give a single `Stream a`
* a bit of refactoring of `Data.List.Quantifiers`

I'm not exactly following the paper. For instance:

* `Prop` is tracking the types it needs generators for. This allows me to manufacture a single big
  generator at the end by iterating Cantor's zig zag to merge all of the smaller ones.
  This means we explore a more divers search space when nesting `exists`: on `(Nat,Nat)` for instance,
  instead of looking at `(0,0)`, `(0,1)`, ..., `(0,fuel)` before moving to `(1,0)`, `(1,1)`, ... we
  instead explore `(0,0)`, `(1,0)`, `(0,1)`, `(2,0)`, `(1,1)`, `(0,2)`, etc.

* I have a `Negates` interface so that instead of looking for a half decider for `Not p`, I look
  for a half decider for `q` such that `q -> Not p`. It is often much easier to look for positive
  evidence (`Not (All p xs)` vs. `Any (Not . p) xs`, `Not (LT a b)` vs. `GTE a b`, etc.)

* I am using a `Colist1 a` instead of a `Stream a` in my `Generator` interface. This makes it
  possible to generate test cases without any duplicates for finite types (and any container
  with values taken from a finite type).


libs/prelude/Prelude/Interfaces.idr
```diff
@@ -361,7 +361,7 @@ for_ = flip traverse_
 ||| Note: In Haskell, `choice` is called `asum`.
 public export
 choice : (Foldable t, Alternative f) => t (f a) -> f a
-choice = foldr (<|>) empty
+choice = foldr (\ x, xs => x <|> xs) empty
```
buzden, line 364: Despite `<|>` is lazy on the right argument, this implementation of `choice` is fully eager. This thing needs to be revised as soon as the problem with the trick of lazy folding is solved.

libs/contrib/Search/Generator.idr
```diff
@@ -0,0 +1,87 @@
...
+||| Put two generators together by exploring the plane they define.
+||| This uses Cantor's zig zag traversal
+public export
+(Generator a, Generator b) => Generator (a, b) where
+  generate = plane generate (\ _ => generate)
+
+||| Put two generators together by exploring the plane they define.
+||| This uses Cantor's zig zag traversal
+public export
+{0 b : a -> Type} -> (Generator a, (x : a) -> Generator (b x)) =>
```
buzden, line 66: Shouldn't the second auto parameter be `{x : a} -> Generator (b x)` instead of `(x : a) -> Generator (b x)`? I mean, even your `Generator (Vect n a)` takes `n` as implicit parameter.

gallais: > even your Generator (Vect n a) takes n as implicit parameter.

That's down to the parser not allowing anything else. I don't have strong opinions one
way or the other: Idris is writing the code putting these together, not me.

buzden: I've updated my comment seemingly after you've answered, so I'll repeat: once I had a problem the compiler feeding `((x : a) -> X x) => ...` parameter when there was no problem with alternative one `({x : a} -> X x) => ...`

PR #1094, opened 2021-02-19
Introduce IDE Protocol Versioning.

jfdm:
Before, the IDE Protocol allowed clients to query the version of Idris being used.
When there was a single version of Idris this was the same as querying the version of the IDE Protocol being used.
Various clients (Emacs) uses this to keep track of the version of the IDE Protocol it is using.

With Idris2, and now different protocol implementations and extensions, such querying is not useful.

We introduce a notion of IDE Protocol versioning separate from the compiler version number.
This will allow IDE Clients to differentiate between versions of the protocol used by different versions of Idris.

This is a reasonable first step in supporting compatibility between clients, versions of Idris, and versions of the IDE Protocol.

By default Idris1 will have protocol version 0 and Idris2 protocol version 1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Changes since Idris 2 v0.3.0
 
 REPL/IDE mode changes:
 
+* Extended the IDE Protocol to introduce IDE protocol versions.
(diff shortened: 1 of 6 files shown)


jfdm:
Right, so it turns out the Idris protocol already does this. When following the code, I thought it should the version number from git and not the actual version.

PR #1095, opened 2021-02-20
 [ fix #758 ] desugar non-binding sequencing in do blocks to (>>)

### LINKED ISSUE #758: Linear functions lose their linearity annotation when wrapped in a monad
# Steps to Reproduce

```idris
module MonadFun

-- introduce linear bind and linear pure
interface LMonad m where
  pure : (1 _ : a) -> m a
  (>>=) : (1 _ : m a) -> (1 _ : (1 _ : a) -> m b) -> m b

fail : LMonad m => (1 _ : m ((1 _ : a) -> b)) -> (1 _ : a) -> m b
fail ma a = MonadFun.do f <- ma
                        ?what_is_f
                        MonadFun.pure (f a)
```

# Expected Behavior

Typechecks.

# Observed Behavior

Does not typecheck. It fails with error
```
Error: While processing right hand side of fail. Trying to
use linear name a in non-linear context.

MonadFun.idr:19:42--19:43
    |
 19 |                         MonadFun.pure (f a)
    |
```

If we ask the type for hole `what_is_f` we get

```
 0 b : Type
 0 m : Type -> Type
 1 a : a
 1 ma : m ((1 _ : a) -> b)
 1 f : a -> b
hot : ?m ?a
```

Which indicates that `f` has lost its linearity annotation.

However, if we wrap each linear function in its own constructor everything works as expected

```idris
infixr 1 #->

data (#->) : Type -> Type -> Type where
  LFun : (1 f : (1 _ : a) -> b) -> a #-> b

fail : LMonad m => m (a #-> b) #-> a #-> m b
fail = LFun $ \ma => LFun $ \a => do (LFun f) <- ma
                                     ?what_is_f
                                     MonadFun.pure (f a)
```

Here, `what_is_f` reports `f` having type `(1 _ : a) -> b` correctly

gallais:
Instead of desugaring do blocks using `>>=` only i.e. elaborating

```idris
do x <- p
   q
   r
```

to

```idris
p >>= \ x => q >>= \ _ => r
```

we make sure that non-binding sequencing uses `(>>)` and get:

```idris
p >>= \ x => q >> r
```

This allows us to have a special `>>` that, for instance, is compatible
with a linear bind by enforcing that its first argument returns `()`,
a value easily consumed.

### DIFF
--- a/libs/base/Data/Buffer.idr
+++ b/libs/base/Data/Buffer.idr
@@ -311,7 +311,7 @@ concatBuffers xs
          let cumulative = reverse revCumulative
          Just buf <- newBuffer totalSize
               | Nothing => pure Nothing
-         traverse (\(b, size, watermark) => copyData b 0 size buf watermark) (zip3 xs sizes cumulative)
+         traverse_ (\(b, size, watermark) => copyData b 0 size buf watermark) (zip3 xs sizes cumulative)
          pure (Just buf)
     where
         scanSize : (Int, List Int) -> Int -> (Int, List Int)
--- a/libs/base/System/File.idr
+++ b/libs/base/System/File.idr
@@ -246,8 +246,7 @@ fEOF (FHandle f)
 export
 fflush : HasIO io => (h : File) -> io ()
 fflush (FHandle f)
-    = do primIO (prim__flush f)
-         pure ()
+    = ignore $ primIO (prim__flush f)
 
 export
 popen : HasIO io => String -> Mode -> io (Either FileError File)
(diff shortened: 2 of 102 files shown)


andrevidela:
This seems like quite a disruptive change, doesn't it break backward compatibility too? Not having to bind resulting values from in do block was a nice quality of life feature that we lose.

gallais:
> doesn't it break backward compatibility too?

Yes. On the other hand, it helps inference and you get a warning when
you're silently throwing away a non-unit result (you can use `ignore`
to throw it away explicitly).

Issue #1029, opened 2021-02-04
Multi-line strings proposal


andrevidela:
# Current Status

This issue splits off from #1004 so that we can keep track of its progress separately here.

ToDo:

- [x] Implement multi-line strings with the correct trimming behaviour #1097 
- [x] add documentation
- [x] add to changelog
  - [x]  explain why old single quote behavior is not allowed to span multiple lines

# Summary

We propose to introduce a special syntax for string literals that span multiple lines. They allow to insert larges pieces of text in a singular literal while remaining compatible with string interpolation (#555), raw-strings (#1027) and preserving the natural indentation level of the surrounding program.

# Motivation

Strings are ubiquitous in programming and allow programmers to communicate through text. Another use for strings is to carry data that will be used by other systems that read strings. Typical examples of this are:

- Generate some HTML headers
- Print out some ascii-art.
- Write out manually an HTTP request.
- Output files given a known template.

What every example has in common is to write down a string literal across multiple lines. Additionally, some of them would benefit from string interpolation #555 while others would benefit from avoiding character escapes. The first two examples would benefit from avoid escaping characters, while the last two would benefit from string interpolation.

Here are each example with existing string syntax

### HTML header

```idris
let header = "<head>\n    <script type=\"text/javascript\" src=\"./tracking.js\"></script>\n</head>"
```
or
```idris
let header = "<head>\n"
          ++ "    <script type=\"text/javascript\" src=\"./tracking.js\"></script>\n" 
          ++ "</head>"
```

in order to show the layout of the file using vertical space.

### ascii-art

This example is taken from the idris compiler itself

```idris

banner : String
banner = "     ____    __     _         ___                                           \n" ++
         "    /  _/___/ /____(_)____   |__ \\                                          \n" ++
         "    / // __  / ___/ / ___/   __/ /     Version " ++ showVersion True version ++ "\n" ++
         "  _/ // /_/ / /  / (__  )   / __/      https://www.idris-lang.org           \n" ++
         " /___/\\__,_/_/  /_/____/   /____/      Type :? for help                     \n" ++
         "\n" ++
         "Welcome to Idris 2.  Enjoy yourself!"
```

As you can see this suffers from two problems: the `\` needs to be escaped, and the version number needs to be inserted. In this proposal I will not suggest to fix both of them at the same time but will provide a solution for each of them individually, leading to two different compromises that have different properties.

### Write out manually an HTTP request.

```idris
testRequest : String
testRequest = 
[...]

PR #1097, opened 2021-02-20
Implement multi-line string

andylokandy:
Implements https://github.com/idris-lang/Idris2/issues/1029. For example, the idris2 banner can be rewritten with multi-line string: 

## Before

```idris
banner : String
banner = "     ____    __     _         ___                                           \n" ++
         "    /  _/___/ /____(_)____   |__ \\                                          \n" ++
         "    / // __  / ___/ / ___/   __/ /     Version " ++ showVersion True version ++ "\n" ++
         "  _/ // /_/ / /  / (__  )   / __/      https://www.idris-lang.org           \n" ++
         " /___/\\__,_/_/  /_/____/   /____/      Type :? for help                     \n" ++
         "\n" ++
         "Welcome to Idris 2.  Enjoy yourself!"
```

## After

```idris
banner : String
banner = #"""
          ____    __     _         ___
         / // __  / ___/ / ___/   __/ /     Version \#{ showVersion True version }

     """#
```

## Known issue (Update: Solved)

In this PR, multiline string is not compatible with the '#' escape introduced by raw string. This is because the syntax chosen for multiline will be ambiguous if '#' is allowed. For example, `#""""""#` can be interpreted as a raw string equals to `"\"\"\"\""` and also can be interpreted as an multiline string (although it is invalid in this example because it's single-line).


andrevidela:
Note that the known issue is addressed in #1029 by forcing multi line string delimiters to being different lines and forcing the opening delimiter to have a line break after it

andylokandy:
@andrevidela My example might be inaccurate. Let's think this one instead:

```
#"""

   """#
```

andylokandy:
> single line raw strings should not span multiple lines. Is that it?

Exactly.

andrevidela:
Ah, I understand, I guess we could afford this kind of lookahead because we only need to read until the end of the line, but that would make the lexing even more complex than it already it.

I Suggest we leave this be for now and make a note of the progress on #1029

andylokandy:
What about this: only lex the multiline begin quote if `"""` is followed by \n

andrevidela:
Ah yes that's actually much simpler

mattpolzin:
I imagine some of the "multiline" strings made up of single line strings (even in this PR) can't be moved to the new syntax in the same version as the syntax is introduced, huh?

I noticed there's no entry in the changelog in this PR, and this change is certainly worth calling out!

andylokandy:
> I imagine some of the "multiline" strings made up of single line strings (even in this PR) can't be moved to the new syntax in the same version as the syntax is introduced, huh?

yeah, we have to wait until the next minor release.

> I noticed there's no entry in the changelog in this PR, and this change is certainly worth calling out!

Because the raw string, multiline string and interpolated string are highly related, I'd like to write the changelog together in an independent pr.

src/Idris/Parser.idr
```diff
@@ -776,23 +776,50 @@ mutual
...
+      toLines (x::xs) line acc
           = case x.val of
-                 Right s => StrLiteral (boundToFC fname x) s
-                 Left tm => StrInterp (boundToFC fname x) tm
+                 Left tm => toLines xs ((StrInterp (boundToFC fname x) tm)::line) acc
+                 Right (str:::[]) => toLines xs ((StrLiteral (boundToFC fname x) str)::line) acc
+                 -- FIXME: calculate the precise FC so as to improve error report for invalid indentation.
+                 Right (str:::strs@(_::_)) => toLines xs [StrLiteral (boundToFC fname x) (last strs)]
+                                                         ((map (\str => [StrLiteral (boundToFC fname x) str]) (List.drop 1 $ reverse strs))
+                                                            ++ (List.reverse ((StrLiteral (boundToFC fname x) str)::line)::acc))
```
andrevidela, line 822: why is this going through each line right to left and then reversing the list instead of left to right?

andrevidela (commented):
Looks good, error messages could be improved but we can open issues for those. Ideally I would like to also have a nicer error message when starting a multine string with non-whitespace characters `"""hello` instead of "bracket not properly closed" but I think it's out of scope for this PR. Good job!

src/Idris/Parser.idr
andylokandy: before a line wrap is met, the incoming interpolation and literal can be pushed to the left, which should be efficient.

src/Idris/Parser.idr
```diff
@@ -776,23 +776,50 @@ mutual
...
+                                     Left err => fatalError err
+                                     Right pstrs => pure $ pstrs
+           pure $ PString (boundToFC fname b) b.val
     where
-      interpBlock : Rule PTerm
-      interpBlock = interpBegin *> (mustWork $ expr q fname idents) <* interpEnd
+      toPStr : (WithBounds $ Either PTerm (List1 String)) -> Either String PStr
+      toPStr x = case x.val of
+                      Right (str:::[]) => Right $ StrLiteral (boundToFC fname x) str
+                      Right (_:::strs) => Left "Multi-line string is expected to begin with \"\"\""
```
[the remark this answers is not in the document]
andylokandy: But since this error is caught in the parser, it'll be hard to move the error point backward. Besides, this error points to the first wrapped line of a single-line string, so I think it may be reasonable?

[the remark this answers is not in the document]
andylokandy: Another solution is to postpone the check to `desugar`.

src/Idris/Parser.idr
andrevidela: Is there a mesurable performance gain? If there isn't I'd suggest sticking with the more familiar pattern of splitting the head and tail of the list rather than the prefix and the last element

src/Idris/Parser.idr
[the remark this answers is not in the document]
andrevidela: Either way is fine, if we don't want to deal with it now we can just open an issue just like for the "FIXME"

src/Idris/Parser.idr
andylokandy: I changed them into `snoc`s.

src/Compiler/ES/ES.idr
```diff
@@ -66,19 +66,19 @@ addSupportToPreamble name code =
...
+      "}" ++
+      "function __prim_stringIteratorToString(_, str, it, f) {" ++
+      "  return f(str.slice(it));" ++
+      "}" ++
+      "function __prim_stringIteratorNext(str, it) {" ++
+      "  if (it >= str.length)" ++
+      "    return {h: 0};" ++
+      "  else" ++
+      "    return {h: 1, a1: str.charAt(it), a2: it + 1};" ++
+      "}"
```
gallais, line 81: This removes a lot of newlines making the generated code less readable.
Can we use `unlines` instead?

andrevidela:
Looks good to me! I've opened #1133 and #1132 to track the issues mentioned in the `FIXME`s

PR #1098, opened 2021-02-20
`Validated` data structure was added

buzden:
This is, in fact, an error-accumulating flavor of `Either`.

It is intended to be used for applicative data creation (e.g. with idiom brackets like `[| MkData validatedF1 validatedF2 ... validatedFn |]`) where each component can be checked and as a result no failed checks should lost.

This kind of structures is usually named `Validation` or `Validated` and I chose the latter because 1) to my feeling the first one describes a process rather than a processed data 2) we already have a module named `Validation` (in `Control`) describing composable effectful processes of validation.

### DIFF
--- a/libs/contrib/Data/Validated.idr
+++ b/libs/contrib/Data/Validated.idr
@@ -1,0 +1,155 @@
+module Data.Validated
+
+import Data.List1
+
+import Decidable.Equality
+
+%default total
+
+||| `Validated` is like an `Either` but accumulates all errors with semigroup operation.
+public export
+data Validated e a = Valid a | Invalid e
+
+--- Instances of standard interfaces ---
+
+public export
+(Eq e, Eq a) => Eq (Validated e a) where
+  Valid x   == Valid y   = x == y
+  Invalid e == Invalid f = e == f
+  _ == _ = False
+
+export
+(Show e, Show a) => Show (Validated e a) where
+  showPrec d $ Valid   x = showCon d "Valid" $ showArg x
+  showPrec d $ Invalid e = showCon d "Invalid" $ showArg e
+
+public export
+Functor (Validated e) where
+  map f $ Valid x   = Valid $ f x
+  map _ $ Invalid e = Invalid e
+
+public export
+Bifunctor Validated where
+  bimap _ s $ Valid x   = Valid   $ s x
+  bimap f _ $ Invalid e = Invalid $ f e
+
+||| Applicative composition preserves invalidity sequentially accumulating all errors.
+public export
+Semigroup e => Applicative (Validated e) where
+  pure = Valid
+
+  Valid f    <*> Valid x    = Valid $ f x
+  Invalid e1 <*> Invalid e2 = Invalid $ e1 <+> e2
+  Invalid e  <*> Valid _    = Invalid e
+  Valid _    <*> Invalid e  = Invalid e
+
+-- There is no `Monad` implementation because it can't be coherent with the accumulating `Applicative` one.
+
+||| Semigroup operation selects the leftmost valid value.
+||| If both sides are invalid, errors are accumulated.
+public export
+Semigroup e => Semigroup (Validated e a) where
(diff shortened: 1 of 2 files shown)


stefan-hoeck:
I have been using `Validated` a lot in Haskell and Scala. It's highly useful. One thing the keeps coming up is the need to convert between `Either` and `Validated`, depending on whether one needs the Monad or the error accumulating Applicative. I wonder, therefore, whether it would make more sense to add a new named Applicative to `Either` instead of introducing a new datatype for this.

buzden:
> I wonder, therefore, whether it would make more sense to add a new named Applicative to `Either` instead of introducing a new datatype for this.

Well, you know, I ran into the following problem with this. Actually, this is a part of a bigger problem.

Consider we have two applicative actions that may have some error:

```idris
f1 : Monad m => Int -> m $ ValidatedL String Int
f1 5 = pure $ pure 0
f1 _ = pure $ oneInvalid "fail 1"

f2 : Monad m => Int -> m $ ValidatedL String Int
f2 6 = pure $ pure 0
f2 _ = pure $ oneInvalid "fail 2"
```

It's syntactically easy to compose them sequentially:
```idris
c2 : Monad m => m (ValidatedL String Int)
c2 = (f1 1 *> f2 1) @{Applicative.Compose}
```

Then, you can run it and the composition works as it should:
```idris
r2 : Bool
r2 = runIdentity c2 == Invalid ["fail 1", "fail 2"]
```

For this, we could use special named instance for `Applicative (Either e)` that accumulates an error.

But what if we want to sequentially compose three such things? Syntax
```idris
(f1 1 *> f2 1 *> f3 1) @{Applicative.Compose}
```
would apply `Applicative.Compose` only to the outermost `*>` operator, leaving the rest application non-deep.

The same problem would be for idiom-brackets with several parameters.

Ideally what would be good is a something like a namespace-defining syntax (`with <name> <expr>` or `<name>.do`) but for named instances, not only the namespaces. Then, it would be possible to write something like `with @{Applicative.Compose} f1 1 *> f2 1 *> f3 1` or `with @{Either.Accumulating.Applicative} [| MkData checked1 checked2 checked3 |]`.

So, such a feature could be a nice replacement for the separate data type. Until then, well, separate data structure is a classic (poor man?) alternative.

buzden:
> Edit: As an alternative, it might be useful to still have the bind operator `(>>=)` for `Validated` to use it in a `do` block (without error accumualtion).

This is dangerous because it would lead to potential refactorings that are not meant to change semantics but actually do. Like, to rewrite `a *> b` to `do a; b` -- despite it's not guaranteed by the compiler, it's not meant to mean different things because `Applicative` and `Monad` interfaces are defined in such a way.

stefan-hoeck:
> But what if we want to sequentially compose three such things? Syntax
> 
> ```idris
> (f1 1 *> f2 1 *> f3 1) @{Applicative.Compose}
> ```
> 
> would apply `Applicative.Compose` only to the outermost `*>` operator, leaving the rest application non-deep.

This example with `Applicative.Compose` is actually the reason, why I have been asking about named instances in the first place. In  #1089 the suggestion is to introduce a named instance and here a new data type is introduced. This seems to be somewhat inconsistent, as the limitations you describe apply to to both cases, no? So, we probably should handle both cases (`Validated` and functor composition) the same.

buzden:
> This seems to be somewhat inconsistent, as the limitations you describe apply to to both cases, no?

I don't see an inconsistency since suggested data structure can be used independently with the named `*.Compose` instances. So, having a `Validated` data structure would allow us to compose separate data checks into a checked creation of composed data using an applicative composition (like an example at the beginning with `[| MkData validatedF1 validatedF2 ... validatedFn |]`).

Additional instances that are suggested #1089, indeed suffer from the problem with composition of three elements. But this is a separate problem because suggested instances allow simplification and generalisation at least in some simple cases (i.e. have their own separate value even having some problems). There is a field of further improvements and PRs for more complex situations (like a composition of three or more elements).

PR #1112, opened 2021-02-23
[ WIP, not ready for review ] Push constructors, methods and projections into a nested namespace

### LINKED ISSUE #628: Bugs around records
# Steps to Reproduce
Typecheck these two programs:
```
module Main

namespace SuperDog
   public export
   record Dog where
      constructor MkDog
      supername : String
      age : Int
      weight : Int

namespace OrdinaryDog
   public export
   record Dog where
      constructor MkDog
      name : String
      age : Int
      weight : Int
```

```
module Main

data Data = MkData Int Int

map : (Int -> Int) -> Data -> Data
map f = record {arg $= f}

```
# Expected Behavior
I expect the first one to typecheck fine, the second one to fail
# Observed Behavior
The first one fails with:
```
Error: While processing type of name. Ambiguous elaboration. Possible results:
    Main.OrdinaryDog.Dog
    Main.SuperDog.Dog

Fld.idr:12:4--19:5
 12 |    public export
 13 |    record Dog where
 14 |       constructor MkDog
 15 |       name : String
 16 |       age : Int
 17 |       weight : Int

```

The second one passes the typechecker

Russoul:
This is still work in progress. Essentially, it's feature complete. But elaboration of nested (local) definitions is highly broken in master. As a result, tests that involve local definitions fail. The major culprit is the `TTImp.TTImp.definedInBlock` function, which is very hacky. And the elaboration of local definitions in general (`Local.idr`) which proceeds by forward substitution of `Nested` (mangled) names for user-defined local names prior to actual ~~disambiguation and~~ qualification of the latter. I intend to deal with (purge) `defineInBlock` and elaboration of locals in a separate PR (it is almost done too).

A taste of what is broken currently in master:
```idris
-- Works
data Main.Global : Type where
  ConsGlobal : Global

nested : () -> ()
nested x = let
 --Doesn't work
  data Main.Nested : Type where
    ConsNested : Nested
  in ()
```
In general, qualified names in local definitions lead to type checking failures.
But since we want to put constructors, methods, projections into an inner namespace,  we need that proper handing of qualified names in local definitions.

What the PR does:

- Every constructor, method and projection is pushed into a new nested namespace, named after the type constructor.
- Every data, interface and record definition without a user supplied constructor name defaults to having a constructor `Mk`.
- Fixes #628 
- function, data, interface and record definitions can be provided with a namespace to shift their names into (previously that only worked with pattern matching definitions)

Two playground files:
`Test2.idr`
```idris
module Test2

-- These forward declarations don't have to be written out
-- here for the main test file to type check.
-- It's enough to just import this file.
data ForwardDeclData : Type
data ForwardDeclRecord : Type
data ForwardDeclInterface : Type

```
`Test.idr`
```idris
import Test2

data IsJust : Maybe a -> Type where
  IsJust : IsJust (Just x)

record Info a b where
  first : a
  second : b

makeInfo : a -> b -> Info a b
makeInfo = Info.Mk

record Info' a b where
  constructor MkInfo'
  first : a
  second : b

makeInfo' : a -> b -> Info' a b
makeInfo' = MkInfo'

interface Show' a where
  show' : a -> String

Show'String : Show' String
Show'String = Show'.Mk id

interface Show'' a where
  constructor Show''
  show'' : a -> String

Show''String : Show'' String
Show''String = Show'' id

data Test2.ForwardDeclData : Type where
  Foo : ForwardDeclData

record Test2.ForwardDeclRecord where
  sayHiFromTest2 : IO ()

interface Test2.ForwardDeclInterface where
  sayHiFromTest2' : IO ()

runRecord : ForwardDeclRecord -> IO ()
runRecord = sayHiFromTest2

accessible : ForwardDeclData
accessible = Foo

ForwardDeclInterface where
  sayHiFromTest2' = putStrLn "hi"

-- BROKEN
nested : Nat -> ()
nested x = let
   data Nested = Cons0 | Cons1

   func : Nested -> ()
   func Cons0 = ()
   func _ = ()

  in func

```

Implementation has to rely on reflected `%hide` calls in a few `src` files. The necessity comes from the fact that we want to keep backward compatibility with the previous release. But since the previous release has a different namespacing scheme, we can't use regular name qualification (e.g `Prelude.Monad.(>>=)` vs `Prelude.(>>=)`) and `%hide`. 

Depends on #1065 
Needs #1065 to be in a release in order to pass the 'self-hosting from the previous release' step.
At the moment, successfully builds the libraries (and all the tests except those with locals) and does self-hosting.
Needs refactoring, so please don't review yet.

The draft is intended for introductory purposes and discussions of the design choices: 
 - Is `Mk` technically a good default name ?
 - Should we refrain from adding a repetitive nested namespace ? E.g don't do `Data.List1.List1` ?
 - Should we assign default names to interface implementations, akin to the newly added `Mk` for constructors ?


ohad:
This is properly cool! Thanks!

Can you try to give a taste as to what's challenging in the nested case? 
(I can imagine it is challenging, but I suspect you have a much better concrete idea!)

Russoul:
Yeah, so the current way of dealing with nested names (any kind of nested names) is by predicting (`definedInBlock`) what names are going to be added inside a local block. That's done in `Local.idr`. The names returned by `definedInBlock` are substituted with their mangled (wrapped in `Nested`) counterparts inside the headers of `ImpDecl`s, representing the nested block. Mangled names also come with a list of indices into the local environment (stored together as `NestedNames` in `TTImp.idr`). Next we elaborate the local block, like a global block of definitions, but with `NestedNames` in mind. Further in the elaboration procedure, inside `App.idr`, references to the old non-mangled names are replaced with the mangled ones, applied to the proper subenvironment, courtesy of `NestedNames`.

The downside of pre-handling definitions in a local block in `Local.idr` is the duplication of work, already carried out by corresponding definition elaborators. The duplicated work being the namespace resolution (name qualification).

Instead of pre-substituting mangled names in the headers of `ImpDecl` in `Local.idr`, I propose to postpone this work to the corresponding definition elaborators, populating `NestedNames` with an extra field, telling whether we are at the top level or nested. Definition elaborators can then handle mangling and filling in of `NestedNames` with the data at hand. Rendering `definedInBlock` redundant.

I tried to fix `definedInBlock` and adjust it to the new namespacing rule, but that didn't go smoothly...

Russoul:
The other thing I really want to mention is that `Name` is evil ! It is too loose and uninformative. I added an experimental alternative, it can be found in `Name.idr`. Names are represented as flattened subparts in an `HVect`. I'll need to refactor that part too. I'd appreciate some comments on those :)

Russoul:
@ohad Feel free to ask for more info and details. It takes time to look through the code and comprehend the peculiarities. So I'd be happy to share and save you some time.

ohad:
Well, my general take is that we shouldn't be lifting local definitions to the top level at all, but that would require us to re-design the type-theory (nothing major theoretically, just a substantial amount of code shuffling).

So I'm not surprised this requires fiddling. 

From my limited experience with interface resolution, there might be subtle interaction with type-checking/elaboration.
Specifically, a nested data declaration whose correctness relies on us knowing something specific about a variable. 

Something like this might stress-test it:

```idris
let U : Nat
    U = 0
in let
      data Foo : (a : Type) -> a -> Type where
        MkFoo : (i : Fin U) -> Foo Void (case i of {})
```

Russoul:
Let's close it for now, because it still needs more thought and a massive rebase...

Issue #1104, opened 2021-02-22
Include laziness information in CExp, Lifted etc.


Z-snails:
### Motivation
I have been trying to write a GRIN backend for Idris2, however it is way more complicated than it needs to be because laziness is always turned into a lambda in ANF and LambdaLift form. This prohibits memoization and more efficient backend dependant implementations.

### Implementation ideas
This could be done with `Force` and `Delay` constructors with `LazyReason`
Or with the function application constructors (eg. `AAppName : FC -> Name -> Maybe LazyReason -> List AVar -> ANF` where `Nothing` represents strict).

I am happy make a PR for this, but I wanted to know how this should be done first.

melted:
An alternative solution would be to compile from NamedCExp which has force and delay (or CExp I guess, but I prefer to have the names instead of indices). Is there any reason that would be a bad fit?

melted:
Deleted comment, I missunderstood the issue.

gallais:
> I have been trying to write a GRIN backend for Idris2,

This is very exciting!

> This could be done with Force and Delay constructors with LazyReason
  Or with the function application constructors 

I'd rather have annotations on lambdas & applications so that backend writers
who do not want to think about this yet can get started quickly and simply
ignore the annotation and rely on the current encoding of thunks as functions.

Z-snails:
I'm currently adding laziness annotations to `LambdaLift` and I have a question. Should there be a `Force` function (like the `eval` function in GRIN), or should each function assume values could be lazy and force them first (like Haskell/STG)?

There is potential for some optimisations (reduced number of function calls) if the backend know exactly when a value goes from lazy to strict, but I have no idea how much difference that would actually make (it will be a known function call which should be well optimised). Whereas the second option is slightly easier when writing a new backend.

PR #1113, opened 2021-02-23
Lazy annotations

Z-snails:
Added `LazyReason` to `Defer` and `Force` in `CExp`
Added laziness annotations to `Lifted` and `ANF` (specifically any function application that can be fully applied)
Also added newtype-ness to `ANFDef` `MkACon`
No backends have been changed (yet) (other than to support modified `getCompileData`)
See #1104


Russoul:
This is a breaking change for all backends, so tagging @ziman @karroffel @bamboo @chrrasmussen - out-of-tree backend authors, just in case. Sorry if I forgot to mention someone.

It would be nice to have a list of external backends on the wiki. As we have for Idris 1. Both authors and end users would benefit from it :)
(Not in this PR of course)

Z-snails:
There is an easy fix for backends, change `getCompileData` to `getCompileData False` and then you can assume `lazy` is always `Nothing` and just discard it.

PR #1122, opened 2021-02-25
Give App1 implementation of (>>)

claymager:
App1 is not a monad, and was missed in the change to desugaring of non-binding sequences.

This fixes that.

### DIFF
--- a/libs/base/Control/App.idr
+++ b/libs/base/Control/App.idr
@@ -197,6 +197,11 @@ namespace App1
   (>>=) = bindApp1
 
   export
+  (>>) : {u : _} -> (1 _ : App1 e ()) ->
+         (1 _ : App1 {u} e b) -> App1 {u} e b
+  ma >> mb = ma >>= (\() => mb)
+
+  export
   pure : (x : a) -> App1 {u=Any} e a
   pure x =  MkApp1 $ \w => MkApp1ResW x w
 
--- a/libs/base/Control/App.idr
+++ b/libs/base/Control/App.idr
@@ -196,15 +196,15 @@ namespace App1
           (1 k : Cont1Type u a u' e b) -> App1 {u=u'} e b
   (>>=) = bindApp1
 
-  delay : {u_act : _} -> (1 _ : App1 {u=u_k} e b) -> Cont1Type u_act () u_k e b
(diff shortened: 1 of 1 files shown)


claymager:
Or it did. Looks like I need to play around with the Usage argument first

claymager:
My motivating example was from the Linear Resources tutorial, where
```idris
storeProg : Has [Console, StoreI] e => App e ()
storeProg = app1 $ do
     store <- connect
     app $ putStr "Password: "
     ?what_next
```
failed without a monad instance. It now complains about a unification failure - One vs Any (from app)

gallais:
Look at the type of `>>=`: it has both a `u` and a `u'` but your version
only has one `u` (and a usage inferred for the `App e ()` argument).

The fix is probably to do something similar to what I did in `Control.Linear.LIO`:

https://github.com/idris-lang/Idris2/blob/d2eeb7ce867cf3bbfa49b45123e9c18964fe64ae/libs/contrib/Control/Linear/LIO.idr#L110-121

libs/base/Control/App.idr
```diff
@@ -196,10 +196,15 @@ namespace App1
           (1 k : Cont1Type u a u' e b) -> App1 {u=u'} e b
   (>>=) = bindApp1
 
-  export
-  (>>) : {u : _} -> (1 _ : App1 e ()) ->
-         (1 _ : App1 {u} e b) -> App1 {u} e b
-  ma >> mb = ma >>= (\() => mb)
+  delay : {u_act : _} -> (1 _ : App1 {u=u_k} e b) -> Cont1Type u_act () u_k e b
+  delay mb = case u_act of
+                  One => \ () => ?ars
```
gallais, line 201: FYI you've left a hole

claymager:
I chose 'delay' for consistency, but if it has an actual meaning in LinearIO, I'd prefer something like 'decideUsage.'

libs/base/Control/App.idr
claymager: oops, thanks

gallais:
Sorry, the :+1: was approving `delay` not asking for you to change it. I would expect
a `decideX` function to be something that returns a `Dec` proof.

PR #1128, opened 2021-02-26
[ base ] A property of interaction between `zipWith` and `index`

buzden:
Rewriting rules for `index i (zipWith ...)` were added for `Vect` and `Stream`.

Also, implementations of `Zippable` were made to be publicly exported (actually, some of them were originally `public export` and were dispublished when introducing `Zippable`).

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -330,7 +330,7 @@ intersectAll = intersectAllBy (==)
 -- Zippable --
 ---------------------------
 
-export
+public export
 Zippable List where
   zipWith _ [] _ = []
   zipWith _ _ [] = []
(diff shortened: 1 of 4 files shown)


gallais (commented):
AFAICT no proof of property uses this kind of naming scheme in the libs.
We should perhaps adopt an official style guide.

buzden:
> AFAICT no proof of property uses this kind of naming scheme in the libs.
> We should perhaps adopt an official style guide.

Oh, sure, sorry

buzden:
I rebased

> We should perhaps adopt an official style guide.

By the way, I haven't heard and didn't manage to find any *official style guide*, only a custom in the standard lib.

PR #1134, opened 2021-02-27
Version numbers in .ipkg files

edwinb:
There's been a `version` field for a while, but it didn't do anything. So, now, it does. Briefly:

* The version field is dot-separated numbers (kind of like semver) and when the package is installed, it gets installed into a directory named after the package name and version number
* Dependencies (the `depends` field) in an `ipkg` file can now express version constraints (lower/upper bound etc)
* If the version number is missing, it defaults to 0. Maybe this field should be a requirement, once we get in the habit of using it.
* Otherwise, things work as before, with a deprecation warning for old-style version strings (which never had any effect anyway)

Correspondingly, I've added a bit to the way Idris looks for packages:

* When adding a package directory to the path, to decide which one to use, it looks for directories with a name which satisfies the given dependency constraint. If there's more than one, it takes the one with the highest version number.
* (New thing) It looks in a `depends` subdirectory, local to the current project, for packages, before looking globally (packages installed under the `PREFIX`). If it finds something in `depends`, it looks no further. This allows installing dependencies locally to a project.

This doesn't change anything else - just the way Idris names and locates packages, and it will still successfully locate packages that don't have version numbers on them (for now). I've updated the package documentation accordingly.


src/Parser/Rule/Package.idr
```diff
@@ -47,6 +89,13 @@ stringLit = terminal "Expected string"
                                  StringLit str => Just str
                                  _ => Nothing)
 
+export
+integerLit : Rule Integer
+integerLit = terminal "Expected string"
```
andylokandy, line 94:
```suggestion
integerLit = terminal "Expected integer"
```

edwinb: Oops, well spotted!

PR #1136, opened 2021-02-28
Flush standard out after writing prompt to it

nmeum:
On Unix-like operating systems `stdio.h` is usually line-buffered. As
`putStr` uses `fputs(3)` from `stdio.h` internally, output will be written to
standard out only after a newline character is written to the buffer. Since
the prompt does not contain a newline, it will only be written to
standard output after the user presses return. I encountered this issue
on [Alpine Linux](https://alpinelinux.org) which uses [musl libc](https://musl.libc.org) (instead of glibc). However, I
believe this issue is likely also reproducible with glibc. This commit
fixes this issue by flushing standard output after writing the prompt to
it.


nmeum:
Looks like some tests do not expect the prompt to be flushed? Is that intended?

Russoul:
Sorry for the late response. All the tests do not expect that. But I am sure it was not intended.
So the fix should include the new golden files with reordered lines. The diff is going to be fairly long, I suspect.

nmeum:
I currently don't have the time on my hands to manual go through all of these failing tests and adjust the expected output file accordingly. Also: I am using idris2 on Alpine Linux (which uses musl instead of glibc) where many other test cases didn't pass for some other reason I haven't investigated yet. I have allowed edits by maintainers for this PR, so if the tests pass locally for you (without my patch) maybe write a script to make the output of this branch the new golden files and just push the changes to my branch?

gallais:
There's already a script: `make test` will ask you whether you want to update
each golden file that leads to a failure.

nmeum:
Updated the golden files using `yes y | make test` (had to adjust `node017` and `chez0017` manually though because the expected file is generated using a script for these test cases). Lets hope it passes now 🤗

nmeum:
Tests pass :tada: Please take another look.

tests/base/data_bits001/expected
```diff
@@ -58,5 +59,4 @@
 [0, 8, 16]
 [0, 16, 32]
 [0, 32, 64, 33]
-1/1: Building BitOps (BitOps.idr)
-Main> Main> Bye for now!
```
gallais, line 62: Oooh so this was the cause of the much dreaded `Main> Main>`!
Nice!

gallais (approved):

PR #1138, opened 2021-02-28
Fix package paths in for Nix

claymager:
Now that idris2 expects package version numbers, this tells nix to use them too.

It's assuming that `base`, `contrib`, `network`, and `prelude` version numbers follow the main version. Judging by `Release/CHECKLIST`, that's good enough for now.


guilhermehas:
#1154 I will add this change in my pull request

PR #1146, opened 2021-03-01
Check current holes only at end of elaboration

### LINKED ISSUE #1140: Local blocks should delay solving auto implicits
This problem was found by someone on the FP Discord.

Minimal example:
```idris
data A = A0
interface Cool a where cool : a -> a
implementation Cool A where cool x = x

please : A
please = cool (let x : String; x = "hello" in A0)
```

#### Steps to Reproduce

Use a local block in a situation where not every type with typeclass constraints has been inferred yet.

The easiest way is in `do`-notation; here is the original example:
```idris
data A = A0

giveA : HasIO io => io A
giveA = pure A0

please : HasIO io => io ()
please = do
     value_a <- giveA

     let a : A 
         a = value_a 
     
     pure ()
```
Here, it hasn't yet unified the `?io` hole from `giveA` with the `io` type from `please` by the time it typechecks the declaration of `a`, so it's unable to figure out that it has a monad implementation.

#### Expected Behavior

It should typecheck fine.

#### Observed Behavior

Error from minimized example:
```
Error: While processing right hand side of please. While processing type of please,x. Can't find an implementation for Cool ?a.

t.idr:6:10--6:50
 2 | interface Cool a where cool : a -> a
 3 | implementation Cool A where cool x = x
 4 | 
 5 | please : A
 6 | please = cool (let x : String; x = "hello" in A0)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Error from the example using `do`-notation:
```
Error: While processing right hand side of please. While processing type of please,a. Can't find an implementation for Monad ?m.

t.idr:8:6--8:22
 4 | giveA = pure A0
 5 | 
 6 | please : HasIO io => io ()
 7 | please = do
 8 |      value_a <- giveA
          ^^^^^^^^^^^^^^^^
```

edwinb:
Fixes #1140 - when checking holes were solved, we checked all of them,
but there may still be some open if there's a local definition.

### DIFF
--- a/src/Core/UnifyState.idr
+++ b/src/Core/UnifyState.idr
@@ -576,12 +576,12 @@ checkDelayedHoles
 -- not guarded by a unification problem (in which case, report that the unification
 -- problem is unsolved) and it doesn't depend on an implicit pattern variable
 -- (in which case, perhaps suggest binding it explicitly)
-export
 checkValidHole : {auto c : Ref Ctxt Defs} ->
                  {auto u : Ref UST UState} ->
-                 (Int, (FC, Name)) -> Core ()
-checkValidHole (idx, (fc, n))
-    = do defs <- get Ctxt
+                 Int -> (Int, (FC, Name)) -> Core ()
+checkValidHole base (idx, (fc, n))
+  = if base > idx then pure () else
+      do defs <- get Ctxt
          ust <- get UST
          Just gdef <- lookupCtxtExact (Resolved idx) (gamma defs)
               | Nothing => pure ()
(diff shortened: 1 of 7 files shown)


src/Core/UnifyState.idr
```diff
@@ -576,12 +576,12 @@ checkDelayedHoles
...
 -- (in which case, perhaps suggest binding it explicitly)
-export
 checkValidHole : {auto c : Ref Ctxt Defs} ->
                  {auto u : Ref UST UState} ->
-                 (Int, (FC, Name)) -> Core ()
-checkValidHole (idx, (fc, n))
-    = do defs <- get Ctxt
+                 Int -> (Int, (FC, Name)) -> Core ()
+checkValidHole base (idx, (fc, n))
+  = if base > idx then pure () else
```
gallais, line 583: We have `unless` btw

edwinb: Oh do we? You'd think I'd know these things. Also I don't know why I didn't think of `when` in any case. I'll change it just for tidiness.

Issue #965, opened 2021-01-20
Improper handling of shadowing on the LHS


gallais:
```idris
foo : {a : Type} -> Void -> {a : a} -> Void
foo _ {a = c} = ?foo_hole
```

currently gives the context

```
c : Type
a : c
=======================
?foo_hole : Void
```

but should give a context

```
a : Type
c : a
=======================
?foo_hole : Void
```

because the partially applied function `foo _` has type `{a : a} -> Void` so
the `{a = c}` binder cannot tackle anything but this first implicit argument
of type `a`.

gallais:
This is made worse by the new general named-application syntax that makes it impossible
to distinguish whether we are talking about implicit or explicit arguments because both
may use braces.

```idris
foo : (a : Void) -> {0 a : Type} -> a
-- The two following implementations of the previous function have the same meaning:
foo {a = a} v = absurd v
foo {a = v} {a = b} = absurd v
```

andrevidela:
I didn't realise named binding also worked for explicit arguments, that looks serious

buzden:
> I didn't realise named binding on the lhs with `{}` also worked for explicit arguments, that looks serious

I think it was added in #607 and test `basic049` tried to check (besides all) how named matching on explicit arguments worked with shadowing.

PR #1147, opened 2021-03-02
[refactor] Use named applications for better readability

ska80:


gallais:
I don't know if we should merge this given that fixing #965 will quite probably
involve breaking this syntax in favour of one that makes explicit implicit vs.
name application.

edwinb:
Given that this has been blocked a while, and the discussion has stopped, and there's an unresolved question about how this impacts #965, I'm going to close this for now. If that's the wrong call, and there's some new information, please feel free to reopen (new PRs get most attention anyway, older ones have a tendency to disappear!)

PR #1151, opened 2021-03-03
[ fix #1148 ] Support hyphenated package names

### LINKED ISSUE #1148: New versioning system can't find packages with a hyphen in the package name.
With the new versioning system for Idris2 packages (#1134), the directories of packages with a hyphen in their name (for example "elab-util") can no longer be found. The culprit seems to be [the following line of code](https://github.com/idris-lang/Idris2/blob/master/src/Idris/SetOptions.idr#L47), which splits a directory's name at the first hyphen, treating everything on the left as the package name and everything on the right as the version number.

We should therefore either

- fix the code to support package names with hyphens
- disallow package names with hyphens, throwing an error when parsing .ipkg files with such an invalid name

stefan-hoeck:


src/Idris/SetOptions.idr
```diff
@@ -43,13 +44,15 @@ candidateDirs dname pkg bounds
...
-               (name, ver) => case strM ver of
-                                   StrNil => (name, MkPkgVersion [0])
-                                   StrCons _ ver =>
-                                      (name, MkPkgVersion (toVersionNum ver))
-               _ => (str, MkPkgVersion [0])
+    getVersion str =
+      -- split the dir name into parts concatenated by "-"
+      -- treating the last part as the version number
+      -- and the initial parts as the package name
+      case reverse $ split (== '-') str of
```
gallais, line 51: Rather than reverse-reverse, I would introduce `unsnoc : List1 a -> (List a, a)`
that returns the beginning of a list & its last element. This will make the
intention clearer.

stefan-hoeck: Sounds reasonable. However, I need some guidance in the proper order of doing this. Eventually, `unsnoc` should go to `Data.List1` in base. But this will not be available to the compiler until the next minor release, right? So a copy of `unsnoc` should also go to `Libraries.Data.List.Extra`, probably with an annotation that it should be removed with the next release?

gallais: `Libraries.Data.List1` but yeah that's the idea.

gallais: You may need to `import Libraries.Data.List1 as Lib` and use `Lib.unsnoc`
to disambiguate between the two equal definitions.

src/Idris/SetOptions.idr
```diff
@@ -43,13 +44,15 @@ candidateDirs dname pkg bounds
...
-               (name, ver) => case strM ver of
-                                   StrNil => (name, MkPkgVersion [0])
-                                   StrCons _ ver =>
-                                      (name, MkPkgVersion (toVersionNum ver))
-               _ => (str, MkPkgVersion [0])
+    getVersion str =
+      -- split the dir name into parts concatenated by "-"
+      -- treating the last part as the version number
+      -- and the initial parts as the package name
+      case Lib.unsnoc $ split (== '-') str of
```
Russoul, line 51: Am I wrong to say that the whole deal with hyphenated package names is ambiguous right now, as we also support a naming strategy with missing version strings (for backward compatibility).
Then package my-awesome-package is ambiguous. However, if a version may only be formed of dot separated numbers, then we can parse that without ambiguity (assuming my-package-1 is not a valid package name root)

stefan-hoeck: Good catch. I'll have a look.

Russoul: That should do it, thanks !

gallais: Good catch!

PR #1152, opened 2021-03-03
[ new ] Lemmas: append preserves List membership.

donovancrichton:
Some lemmas that show appending doesn't impact on a proof that
an element is a member of a List.

### DIFF
--- a/libs/contrib/Data/Elem/Extra.idr
+++ b/libs/contrib/Data/Elem/Extra.idr
@@ -21,11 +21,11 @@ elemAppLeft (x :: xs) ys (There prf2) = There $ elemAppLeft xs ys prf2
 
 ||| Proof that an element is still inside a list if we prepend to it.
 public export
-elemAppRight :  (xs, ys : List a) -> (prf : Elem x xs) -> Elem x (ys ++ xs)
-elemAppRight (x :: xs) [] Here = Here
-elemAppRight (x :: xs) [] (There prf2) = There prf2
-elemAppRight xs [] prf = prf
-elemAppRight xs (y :: ys) prf = There $ elemAppRight xs ys prf
+elemAppRight :  (ys, xs : List a) -> (prf : Elem x xs) -> Elem x (ys ++ xs)
+elemAppRight [] (x :: xs) Here = Here
+elemAppRight [] (x :: xs) (There prf2) = There prf2
+elemAppRight [] xs prf = prf
+elemAppRight (y :: ys) xs prf = There $ elemAppRight ys xs prf
 
 ||| Proof that membership on append implies membership in left or right sublist.
 public export
@@ -55,7 +55,7 @@ notElemAppLeft xs ys prf val = prf $ elemAppLeft xs ys val
 
 ||| Proof that x is not in (xs ++ ys) implies proof that x is not in ys.
 public export
-notElemAppRight : (xs, ys : List a)
+notElemAppRight : (ys, xs : List a)
                -> (prf : Not (Elem x (xs ++ ys)))
                -> Not (Elem x ys)
-notElemAppRight xs ys prf val = prf $ elemAppRight ys xs val
+notElemAppRight ys xs prf val = prf $ elemAppRight xs ys val
--- a/libs/contrib/Data/Elem/Extra.idr
+++ b/libs/contrib/Data/Elem/Extra.idr
@@ -25,16 +25,10 @@ elemAppLorR : (xs, ys : List a)
            -> Either (Elem k xs) (Elem k ys)
 elemAppLorR [] [] prf = absurd prf
 elemAppLorR [] _ prf = Right prf
-elemAppLorR (x :: xs) [] prf =
-  let eq   : (xs ++ [] = xs)  = appendNilRightNeutral xs
-      prf' : Elem k (x :: xs) = replace {p = \g => Elem k (x :: g)} eq prf
-  in Left prf'
+elemAppLorR (x :: xs) [] prf = 
+  Left rewrite sym $ appendNilRightNeutral xs in prf
 elemAppLorR (x :: xs) _ Here = Left Here
-elemAppLorR (x :: xs) ys (There prf) =
-  let iH : Either (Elem k xs) (Elem k ys) = elemAppLorR xs ys prf
-  in case iH of
-      (Left l) => Left $ There l
-      (Right r) => Right r
+elemAppLorR (x :: xs) ys (There prf) = mapFst There $ elemAppLorR xs ys prf
 
 
 ||| Proof that x is not in (xs ++ ys) implies proof that x is not in xs.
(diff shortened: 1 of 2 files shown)


libs/contrib/Data/Elem/Extra.idr
```diff
@@ -0,0 +1,61 @@
+module Data.Elem.Extra
+
+import Data.List
+import Data.List.Elem
+
+%default total
+
+||| Convenience lemma showing that prepending nil to xs is just xs.
+public export
+appendNilLeftNeutral : (xs : List a) -> ([] ++ xs) = xs
```
gallais, line 10: This should be moved to a more sensible location as it has nothing to do with `Elem`.

libs/contrib/Data/Elem/Extra.idr
```diff
@@ -0,0 +1,61 @@
...
+||| Proof that an element is still inside a list if we append to it.
+public export
+elemAppLeft : (xs, ys : List a) -> (prf : Elem x xs) -> Elem x (xs ++ ys)
+elemAppLeft (x :: xs) ys Here = Here
+elemAppLeft (x :: xs) ys (There prf2) = There $ elemAppLeft xs ys prf2
+
+
+||| Proof that an element is still inside a list if we prepend to it.
+public export
+elemAppRight :  (xs, ys : List a) -> (prf : Elem x xs) -> Elem x (ys ++ xs)
```
gallais, line 24: I'd rather keep `xs` and `ys` in the same order as in the resulting type.
Otherwise it's always pretty confusing to remember to swap the arguments
around when using `elemAppLeft` vs. `elemAppRight`.

libs/contrib/Data/Elem/Extra.idr
donovancrichton: Is it too much to have it go into Data.List next ```appendNilRightNeutral```?
If so, I'll find a spot in contrib/Data.List to put it :)

libs/contrib/Data/Elem/Extra.idr
donovancrichton: Sounds good to me, I'll make the change and try again!

libs/contrib/Data/Elem/Extra.idr
gallais: That seems like a sensible place to put it.

gallais: Btw, you should not need to pattern-match on `xs` here: it should always be
true by computation only.

buzden: Is there any use of this function? I mean, it's obvious to the compiler and it's not used in newly added lemmata.

donovancrichton: Ahh whoops, I forgot that I rewrote things to not use it....well looks like I'll remove it again!

gallais: It can be useful as documentation.

libs/contrib/Data/Elem/Extra.idr
```diff
@@ -0,0 +1,52 @@
...
+public export
+elemAppLorR : (xs, ys : List a)
+           -> (prf : Elem k (xs ++ ys))
+           -> Either (Elem k xs) (Elem k ys)
+elemAppLorR [] [] prf = absurd prf
+elemAppLorR [] _ prf = Right prf
+elemAppLorR (x :: xs) [] prf =
+  let eq   : (xs ++ [] = xs)  = appendNilRightNeutral xs
+      prf' : Elem k (x :: xs) = replace {p = \g => Elem k (x :: g)} eq prf
+  in Left prf'
```
buzden, line 31: By the way, this case could be written simpler:
```idris
elemAppLorR (x :: xs) [] prf = Left rewrite sym $ appendNilRightNeutral xs in prf
```

libs/contrib/Data/Elem/Extra.idr
```diff
@@ -0,0 +1,52 @@
...
+elemAppLorR (x :: xs) _ Here = Left Here
+elemAppLorR (x :: xs) ys (There prf) =
+  let iH : Either (Elem k xs) (Elem k ys) = elemAppLorR xs ys prf
+  in case iH of
+      (Left l) => Left $ There l
+      (Right r) => Right r
```
buzden, line 37: This case also could be written simpler:
```idris
elemAppLorR (x :: xs) ys (There prf) = mapFst There $ elemAppLorR xs ys prf
```

libs/contrib/Data/Elem/Extra.idr
donovancrichton: I'm of two minds about this. On the one hand I think it's useful for people to see that rewrites are just an ordinary expression, on the other I also think there are not very many examples of `replace` floating about.
Should I be aiming for the minimal text possible when submitting proofs?

libs/contrib/Data/Elem/Extra.idr
donovancrichton: Again this comes down to whether or not it's better to have a terse proof, or one with (arguably) clear steps. Do we have a general consensus on this?

libs/contrib/Data/Elem/Extra.idr
buzden: > I think it's useful for people to see that rewrites are just an ordinary expression

My personal opinion is that standard libraries are not the best place for putting education code snippets. If you want clarify that `rewrite` is just a specially tuned `replace` expression, I think, you should add more clarification of it to the documentation rather than using it in the library just to show.

> Should I be aiming for the minimal text possible when submitting proofs?

Well, I don't know. I think that the proof that is understandable to most educated people is the best. *Personally* I think that short one with `rewrite` reads better than several `let`s -- but I'm sure there are people who think differently. Again, *personally* I think that `let`s add clarification only when expression is really big or where there are some similar expressions with completely different semantics -- in this case you'd want to attach a name to them. In all other cases, I think, `let`s do not add anything to readability and even disturb.

libs/contrib/Data/Elem/Extra.idr
buzden: I've written most of thoughts above, but for this particular example I can comment (keep in mind, that all of that is only a matter of taste):
- the name `iH` does not clarify anything; those that could would be `resultForTail` or something like this;
- the type if `iH` does not clarify anything either since it is exactly the type of result of the recursive call;
- pattern match of `Right` case does not change the type (so, there is not need to repack);
- the only thing the `Left` case is doing is an application of `There` to the what's in `Left`.

So, there are two simple operations -- recursive call for the tail and adding `There` for the `Left` case. The second one is what exactly `mapFst` created for (in case of `Either`, of course), so to all familiar with this function simple `mapFst` would communicate better than realizing that in this pattern match one case simply acts as identity and the second one maps over a particular function. The special name for a recursive call to a tail does not add anything just because there are no other calls to anything else.

And that's why I think that shorter piece of code here tells what's going on better.

PR #1153, opened 2021-03-03
Record local hints in delayed elaborators

edwinb:
We might not have set up search problems yet when delaying an elaborator, so we need to know what the local hints were at the point of delay.

### DIFF
--- a/src/Core/UnifyState.idr
+++ b/src/Core/UnifyState.idr
@@ -84,12 +84,13 @@ record UState where
   dotConstraints : List (Name, DotReason, Constraint) -- dot pattern constraints
   nextName : Int
   nextConstraint : Int
-  delayedElab : List (Nat, Int, Core ClosedTerm)
+  delayedElab : List (Nat, Int, NameMap (), Core ClosedTerm)
                 -- Elaborators which we need to try again later, because
                 -- we didn't have enough type information to elaborate
                 -- successfully yet.
                 -- 'Nat' is the priority (lowest first)
                 -- The 'Int' is the resolved name.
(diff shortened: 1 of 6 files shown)


edwinb:
Does the linter really need to care about test source files? Sometimes I just take other people's programs directly, unedited, and tidying them up to keep the linter happy is just busywork in those circumstances. Also, it's not impossible that we might test some inputs with odd spacing deliberately.

Russoul:
@edwinb you are a Vim user, AFAIK.
This might save you some time in the future:
```

function! ShowTrailingSpaces(...)
  let @/='\v(\s+$)|( +\ze\t)'
  let oldhlsearch=&hlsearch
  if !a:0
    let &hlsearch=!&hlsearch
  else
    let &hlsearch=a:1
  end
  return oldhlsearch
endfun


function! <SID>StripTrailingSpaces()
    let l = line(".")
    let c = col(".")
    %s/\s\+$//e
    call cursor(l, c)
endfun

" Highlight trailing whitespace
command! -bar -nargs=? ShowTrailingSpaces call ShowTrailingSpaces(<args>)

" Remove trailing whitespace across current buffer
command! -bar -nargs=? StripTrailingSpaces call <SID>StripTrailingSpaces()

"Or remove all trailing whitespace on write, if you want.
augroup removeTrailingWhitespace
   autocmd!
   autocmd BufWritePre * :call <SID>StripTrailingSpaces()
augroup END

```

edwinb:
Right, I can set up vim to do this, and that's what I did - the real question is whether the linting here is adding anything.

PR #1154, opened 2021-03-03
Added new nix functionalities

guilhermehas:
One of the new nix functionalities is to run emacs with Idris 2 using a single command:
```bash
nix run idris-lang/Idris2#emacs-with-idris idrisCode.idr
```
Another nix functionality is to create a nix template project with the command:
```bash
nix flake new mypkg -t github:idris-lang/Idris2
```


stefan-hoeck:
I really like these additions especially since Idris2 is still lacking a build tool / package manager. I have the following questions / suggestions:

- It would be great if the new `buildIdris.nix` would take an additional argument of other Idris package dependencies (or does it already?).
- Given the above, would it then be possible to enter a nix-shell and have the required Idris packages in scope (probably via the new local `depends` folder (see #1134), or by adjusting $IDRIS_PACKAGE_PATH) to get an isolated development environment? If this is possible, we should document this properly and include at least one Idris package dependency in the `flake.nix` template as an example. Probably even two: One, which is itself a nix flake and one which is not.

Please correct me, if the above assumptions are not correct. I am a nix user, but am new to nix flakes, and have not written many derivations myself.

vrescobar:
Hi, just commenting on a useful addition I got to my nix-shell: the rlwrap and an alias in the shellHook:

```
alias idris2="rlwrap -s 1000 idris2 --no-banner"
```
In that way, I can have all the usual deadline goodies like command line history. cursor navigation, killring shortcuts such as Ctrl+A Ctrl+K Ctrl+U etc it really does more comfortable the idris2 repl in user mode (and it should work as good with scripts, but the wrapper is there).

By the way, thanks for packaging idris2 on nix, it was a true time-saver.

guilhermehas:
I have already made both suggestions.
It is now possible to reference libraries and install them globally.

gallais:
I was going to ask the person responsible for the original nix code
to opine but I see that it's you. :D

So let's go! :rocket:

PR #1155, opened 2021-03-04
[ refactor ] Multiline error report

### LINKED ISSUE #1133: Reporting single-quoted multi-line strings
The following program return the correct error but at the wrong line:

```
hello : String
hello = "
    hello
    "

main : IO ()
```

The error is reported as

```
Error: Multi-line string is expected to begin with """.

 hello : String
 hello = "
   hello
 "
 main : IO ()
 ^
```

but should be reported as

```
Error: Multi-line string is expected to begin with """.

 hello : String
 hello = "
         ^
```

andrevidela:
- Parser errors can now take bounds for better diagnostics
- fix #1133


andrevidela:
ready for review @gallais

tests/idris2/perror008/expected
```diff
@@ -0,0 +1,8 @@
+1/1: Building MultilineError (MultilineError.idr)
```
andylokandy, line 1: This duplicates https://github.com/idris-lang/Idris2/blob/1d692dae6b418e00ed30d28b904c463e029d61d7/tests/idris2/perror007/expected#L82-L88

andylokandy:
I'm still confused by why `WithBounds tok`, which could be `NoBound`, enforce, while you can return `tok` when bound is not provided.

andrevidela:
It's explained in the commit message. Essentially, `WithBounds` used to always carry bounds but it also had a flag that indicates if the bounds are valid or not. This change makes it so that we _know_ when a term has valid bounds or not. With the additional invariant that bounds are only valid if our grammar consumes its input. Indeed I  assume that the token does not come from the input if the grammar is not guaranteed to consume. This is not quite the case because `peek` and `alt` both can return valid bounds while not being guaranteed to consume, but it's enough for most uses of the `bound` constructor. Bounds from non-consuming grammars can be recovered with clever use of `position`.

Almost all signatures return `WithBounds True` which means the parsed terms need to have valid bounds but that does not mean the flag is unnecessary, it means that every term that does not have valid bounds needs to be either merged with one which has (like in https://github.com/idris-lang/Idris2/pull/1155/files#diff-28f7a243932138226d35676705da183ae3ff559d1683ef8afb779e850004f378R692-R693) or needs to be given bounds (like in https://github.com/idris-lang/Idris2/pull/1155/files#diff-28f7a243932138226d35676705da183ae3ff559d1683ef8afb779e850004f378R855)

tests/idris2/interface016/expected
```diff
@@ -11,4 +11,4 @@ TwoNum.idr:4:7--4:8
 
 Possible correct results:
     conArg (implicitly bound at TwoNum.idr:4:3--4:8)
-    conArg (implicitly bound at TwoNum.idr:2:1--4:8)
+    conArg (implicitly bound at TwoNum.idr:2:1--5:1)
```
gallais, line 14: Line 5 does not exist so it's a bit strange to see it mentioned.

andrevidela: I suspect this is because of how block bounds are parsed at end of input. Let me see what's going  on

src/Idris/Parser.idr
```diff
@@ -1501,28 +1502,27 @@ import_ fname indents
...
-    = do b <- bounds (do doc    <- option "" documentation
-                         nspace <- option (nsAsModuleIdent mainNS)
-                                     (do keyword "module"
-                                         moduleIdent)
-                         imports <- block (import_ fname)
-                         pure (doc, nspace, imports))
-         ds      <- block (topDecl fname)
-         (doc, nspace, imports) <- pure b.val
-         pure (MkModule (boundToFC fname b)
+    = do p1 <- position
```
ShinKage, line 1505: Why the switch to `position` inplace of `bounds`? The former returns the position of the next token, and probably it is the cause of the relaxed bounds in some test cases, as mentioned also by @gallais

andrevidela: This was to avoid computing the bounds on all the following parser which are not guaranteed to consume. I tried removing the position and computing the bounds anyways (by returning `b ** WithBounds b tok` on `option`) but I still get block with bounds outside of the file in `"interface016"`

andylokandy:
> It doesn't even capture the invariant properly since if either the bounds are valid the result should have valid bounds, we could update this to carry proofs about Either but at this point it seems the original definition of mergeBounds is better suited.

Is there any concrete case where the proof is used?

andrevidela:
Basically everywhere its using `mergeBounds` in the parser the parser expects valid bounds but either arguments aren't guaranteed to have bounds. Typically when parsing `do` blocks, we have a symbol `do` which always has bounds, but the block might not have bounds and be empty, but the entire parsed term has bounds because we merge its bounds with the bounds of the keyword `do`: https://github.com/idris-lang/Idris2/pull/1155/files#diff-28f7a243932138226d35676705da183ae3ff559d1683ef8afb779e850004f378R693

andylokandy:
I still can't figure out why the type-level proof for valid bound is necessary. (1) `mergeBounds` works before this PR (2) I can't find anything that uses `Withbounds False`  in this PR.

andrevidela:
- The type-level proof is necessary in order to have a correct implementation of `.getBounds`.
- `.getBounds` is necessary in order to get the bounds of a symbol and forward it to `failLoc` which gives new bounds to an error
- This is necessary because with the previous implementation, bounds for errors were reported with the next token to be parsed
- Intuitively you would think that the old implementation of `Fail` could have taken a `WithBounds tok` (instead of `Bound`) and manually add it as the next token, however, this does not work because the type `tok` does not necessarily match with the type of the token parsed. Typically, symbols return `()` which is not `PTerm`.
- Therefore, we need to use `Bounds`, therefore we need a reliable way to get a `Bound`, therefore we need an indication of when they are valid or not, and therefore the `Bool` flag

>  I can't find anything that uses Withbounds False in this PR.

Every calls to the constructor `NoBounds` is of type `WithBounds False`. Typically, the `position` primitive of the parser returns `WithBounds False Bounds`

zhongzc:
I think it should be fine to return an empty (col 0, row 0) `Bound` in case no bound is given, just like the irrelevant in before. Because it's rare that a term has no bound and `eoi` in `blockEntry` can be assigned the bound to the last row and last col plus one.

andrevidela:
Yes this is correct, technically speaking you don't need the additional invariant because you can return a sentinel value just fine. In fact this is what most software does in practice, programming languages like Go or C do this routinely since they do not have ADTs. This is also what the previous implementation did by returning bounds of `(-1, -1)`.

However I don't believe this is the right move when developing software. We have the opportunity to both _document and enforce_ an invariant that we know should result in inaccurate diagnostics if broken (those are also particularly hard to notice since they don't result in runtime error). Maybe you're worried about the runtime performance, but it should be the same. That's because the flag simply moved position in the struct that the codegen uses for `WIthBounds`, it's now automatically handled by the compiler when matching on it, rather than checked manually by the programmer. If you are really worried about the runtime performance we can try to erase the flag but since every operation on it is O(1) I do not expect any measurable gain in performance

andylokandy:
I agree with @zhongzc, but I rather worry about the balance between the assumption explicitness and the development efficiency. As Idris is more powerful in expression ability than other ones, but we should still think twice when we are able to use the power -- is it necessary? does it bring us runtime cost? is it harder to understand? In this case, I think the overwhelming syntax noise is not deserved by the only one (maybe not any) special case. Not only that, I think the special case is still handled well by the internal dynamical `isIrrelavant` field in `WithBounds` which is just hiding the type-level `Bool` into runtime.

andrevidela:
I couldn't figure out what was wrong with the out-of-files bounds so I've reverted the changes to `WithBounds` @gallais

src/Libraries/Text/Bounded.idr
```diff
@@ -2,63 +2,85 @@ module Libraries.Text.Bounded
...
-  (MkBounded val ir sl sc el ec) == (MkBounded val' ir' sl' sc' el' ec') =
-    val == val' && ir == ir' && sl == sl' && sc == sc' && el == el' && ec == ec'
+  (MkBounded val ir bd) == (MkBounded val' ir' bd') =
+    val == val' && ir == ir' && bd == bd'
 
 export
 Show ty => Show (WithBounds ty) where
-  showPrec d (MkBounded val ir sl sc el ec) =
+  showPrec d (MkBounded val ir (MkBounds sl sc el ec)) =
     showCon d "MkBounded" (concat [showArg ir, showArg val, showArg sl, showArg sc, showArg el, showArg ec])
```
gallais, line 53: Should probably introduce `Show Bounds` using `showCon "MkBounds" (...)`
and call that rather than unpacking the `Bounds` record here.

andylokandy:
@andrevidela A bug breaking the CI has been fixed by https://github.com/idris-lang/Idris2/pull/1189. You need to rebase on the master.

PR #1158, opened 2021-03-05
Fix multiline string literals not properly merged

andylokandy:


src/Idris/Desugar.idr
```diff
@@ -424,10 +424,10 @@ mutual
 
       -- merge neighbouring StrLiteral
       mergeStrLit : List PStr -> List PStr
-      mergeStrLit [] = []
       mergeStrLit ((StrLiteral fc str1)::(StrLiteral _ str2)::xs)
-          = (StrLiteral fc (str1 ++ str2)) :: xs
+          = mergeStrLit $ (StrLiteral fc (str1 ++ str2)) :: xs
```
gallais, line 428: hmmmmm

This looks like the introduction of a quadratic slowdown.
Can we instead accumulate a `List String` and use `fastConcat (reverse acc)`?

gallais: I am suggesting something like this:

```
mergeStrLit : List PStr -> List PStr
mergeStrLit = loop [] where
  cons : List (FC, String) -> List PStr -> List PStr
  cons pieces xs = case reverse pieces of
    [] => xs
    (fc, x) :: xs => StrLiteral fc (fastConcat (x :: map snd xs))

  loop : List (FC, String) -> List PStr -> List Pstr
  loop acc [] = cons acc []
  loop acc (x@(StrInterp _ _) :: xs) = cons acc (x :: loop [] xs)
  loop acc (StrLiteral fc str :: xs) = loop ((fc, str) :: acc)
```

you could structure that differently (e.g. with `Maybe (FC, List1 String)`
to avoid the `map snd` shenanigans but the idea is the same.

andylokandy: I've pushed a shorter version, do you think its performance is the same as the loop version above?

gallais: Yeah I think it should be equivalent (although Idris won't see that
it's total because of the call to `span`). The following would work
too & without the annoying `cons` or `Maybe` nonsense:

```idris
mergeStrLit : List PStr -> List PStr
mergeStrLit = loop where

  loop : List PStr -> List PStr
  loopAcc : FC -> List String -> List PStr -> List PStr

  loop [] = []
  loop (StrLiteral fc str :: xs) = loopAcc fc [str] xs
  loop (x :: xs) = x :: loop xs
 
  loopAcc fc acc (StrLiteral _ str :: xs) = loopAcc fc (str :: acc) xs
  loopAcc fc acc xs = StrLiteral fc (fastConcat (reverse acc)) :: loop xs
```

andylokandy: Seems that totality is not required in desugar...

PR #1161, opened 2021-03-06
Namespace checks for resugaring

### LINKED ISSUE #634: Case-splitting Data types with a 'MkPair' constructor with two or more arguments results in a sugared pair pattern.
This is either a bug or has the makings of a potential feature.

# Steps to Reproduce

Given the following Idris2 program:

```idris
data Foobar = MkPair Int Char

getChar : (foo : Foobar) -> Char
```
1. Introduce an initial match clause.

```idris
getChar : (foo : Foobar) -> Char
getChar foo = ?getChar_rhs
```
2. case splitting on `foo` results in syntax sugar value constructor for pairs being introduced.
```idris
getChar : (foo : Foobar) -> Char
getChar (x, y) = ?getChar_rhs_
```

# Expected Behavior

Case-splitting on `foo` should present a pattern of the form `(MkPair x y)`.

# Observed Behavior

The resulting equation for `getChar` fails to type-check as `(x,y)` is seen as having type `(Int,Char)` than type `Foobar`.

```
- + Errors (1)
 `-- Feature.idr line 7 col 8:
     While processing left hand side of getChar. Sorry, I can't find any elaboration which works. All
     errors:
     If Builtin.Pair: When unifying Type and Foobar.
     Mismatch between: Type and Foobar.
     
     Feature.idr:7:9--7:15
        |
      7 | getChar (x, y) = ?getChar_rhs_
        |         ^^^^^^
     
     If Builtin.MkPair: When unifying (?a, ?b) and Foobar.
     Mismatch between: (?a, ?b) and Foobar.
     
     Feature.idr:7:9--7:15
        |
      7 | getChar (x, y) = ?getChar_rhs_
        |         ^^^^^^
```
# Comment

A question: Is the expected and observed behaviour correct here?

As we have overloading for `Nil` and `(::)` to obtain list notation, Should we thus allow overloading of `MkPair` to obtain pair notation?

Note that case splitting works as expected when there is one parameter.

```idris
data Foobar : Type where
  MkPair : Char -> Foobar

getChar : (foo : Foobar) -> Char
getChar (MkPair x) = ?getChar_rhs_
```

Having two or more results in:

```idris
data Foobar : Type where
  MkPair : Char -> Int -> Int -> Foobar

getChar : (foo : Foobar) -> Char
getChar ((x, y) z) = ?getChar_rhs_
```

RaoulHC:
Previously constructors which clashed with some builtin constructors
would get resugared incorrectly. `Nil` and `::` still get resugared
outside of the builtin namespace, allowing list syntax for different
types.

The linear002 test now expects and uses Unit explicitly, rather than
relying on this behaviour.

Fixes #634

### DIFF
--- a/src/Idris/Resugar.idr
+++ b/src/Idris/Resugar.idr
@@ -113,30 +113,37 @@ mutual
   ||| Put the special names (Nil, ::, Pair, Z, S, etc) back as syntax
   ||| Returns `Nothing` in case there was nothing to resugar.
   sugarAppM : PTerm -> Maybe PTerm
-  sugarAppM (PApp fc (PApp _ (PRef _ nm) l) r) =
-    case nameRoot nm of
-      "Pair"   => pure $ PPair fc (unbracket l) (unbracket r)
-      "MkPair" => pure $ PPair fc (unbracket l) (unbracket r)
-      "DPair"  => case unbracket r of
-        PLam _ _ _ n _ r' => pure $ PDPair fc n (unbracket l) (unbracket r')
-        _                 => Nothing
-      "Equal"  => pure $ PEq fc (unbracket l) (unbracket r)
-      "==="    => pure $ PEq fc (unbracket l) (unbracket r)
-      "~=~"    => pure $ PEq fc (unbracket l) (unbracket r)
-      "::"     => case sugarApp (unbracket r) of
-        PList fc xs => pure $ PList fc (unbracketApp l :: xs)
-        _           => Nothing
-      _        => Nothing
+  sugarAppM (PApp fc (PApp _ (PRef _ (NS ns nm)) l) r) =
+    if builtinNS == ns
+       then case nameRoot nm of
+         "Pair"   => pure $ PPair fc (unbracket l) (unbracket r)
+         "MkPair" => pure $ PPair fc (unbracket l) (unbracket r)
+         "DPair"  => case unbracket r of
+            PLam _ _ _ n _ r' => pure $ PDPair fc n (unbracket l) (unbracket r')
+            _                 => Nothing
+         "Equal"  => pure $ PEq fc (unbracket l) (unbracket r)
+         "==="    => pure $ PEq fc (unbracket l) (unbracket r)
+         "~=~"    => pure $ PEq fc (unbracket l) (unbracket r)
+         _        => Nothing
+       else if nameRoot nm == "::"
+               then case sugarApp (unbracket r) of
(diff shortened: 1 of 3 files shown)


Russoul:
I think it would be very useful _and_ harmless to have every syntactic sugar construction be compatible with user namespaces.

It could be very handy in situations when one wants to build a custom Prelude with sensible sugar for common types.
I believe it won't harm users, because no one would write, e.g. another `Pair` type, with a meaning differing from the canonical `Pair` type (unless, again, they are writing a replacement for Prelude).

RaoulHC:
I see, I guess in that case you'd want to make sure these constructs get desugared in a namespace independent way, as `::` and `Nil` presumably are?

Russoul:
Yes, that was my intention. Hopefully others agree on this too. There are a few corner cases regarding some of the constructs.
E.g. the disambiguator may get stuck occasionally at the `(,)` syntax standing for both `Pair` and `MkPair`.
But that's another story: if someone wants to write a custom Prelude from the ground up, they may as well choose another syntax for some of the types.

But having an option to stick with familiar Idris syntax surely doesn't hurt :)

RaoulHC:
After much sleuthing I've got this behaviour to sorta work (I did pick up this issue to learn the code base after all :) ), though it looks like I'll find out why it's failing to self-host now.

It required two fixes to the ambiguity resolving:
* Making sure that the metadata is set to the successful elaboration used, which I'm fairly sure was an oversight. A `typeat` test looking inside a pair couldn't find variables without this as all the metadata had been reverted to before the elaboration.
* Stopping adding an ambiguity try when falling back to the unique default, which I'm less sure is correct, but it does work, without it expressions such as `(a, b, c, d :: ds, e)` fail at the namespace resolution of `::`, as the ambiguity depth was too large.

I've also changed the `Pair` in `Data.Contrib` to `CPair`, but there may be a better new name for it.

gallais:
> I've also changed the `Pair` in `Data.Container` to `CPair`,
  but there may be a better new name for it.

If this change breaks this sort of code, I'm not sure it's worth adding tbh.

The original bugfix was to avoid resugaring too many things as `(,)` to not
break code and now we're not desugaring precisely enough and that breaks more code.

RaoulHC:
Changed this back to the original change (though I've left the unnamespaced desugaring on a branch [here](https://github.com/RaoulHC/Idris2/tree/desugar-unnamespace) should anyone wish to look at it).

Russoul:
Turns out, I was wrong about: 

> no one would write, e.g. another `Pair` type

Let's go with the conservative solution then.

Issue #1105, opened 2021-02-22
RFC - Create a Golden package


mattpolzin:
The code needed to run the Idris 2 test suite was (somewhat recently) duplicated from its place in the compiler source files into the contrib package. At face value, this is nice. I just started to use Golden in a personal project and immediately saw value in not having to roll my own test framework for testing. 

The duplication means we have an opportunity for unintentional code drift between the two copies of Golden, though. 

If we fully move Golden to contrib, we have an additional compiler dependency on contrib (whereas it makes sense to me to move away from the compiler or even it's test suite depending on the contrib package). So, I propose we move Golden into a new package alongside network, base, contrib, etc. and let the test suite depend on this new package. I'd call the package "golden" but other bike shed painters might prefer a name like "testing". 

If folks like the idea of a new package, I'm happy to tackle the PR that creates the package and removes the code currently living in contrib and src.

andylokandy:
Besides the golden test, I'd like to introduce an intrusive test framework as rust does. Because every golden test brings 3 more files which make the test too heavy and discourages us to write unit tests. It may look like:

```idris
module MyBoolOps

not : Bool -> Bool
not True = False
not False = True

%test
testNot : Test
testNot = do
  assert $ (not . not) True == True 
  assert $ not False == True
```

mattpolzin:
@andylokandy a unit test framework would definitely be nice. 

Do you think this motivates a single package ("testing") where `golden` is a module and unit test functionality stems from another module in the same package? I think that might be nice. A single package dependency anywhere that testing of either kind is going to be performed.

andylokandy:
Maybe good or not... It's indeed very convenience of providing as many functionaloties (including test framework, various parsers, benchmark framework, getops) in stdlib at this immature stage of idris. But once backward compatibility has became important, these things have better move out of std, for instance, to an official package in package manager registry.

mattpolzin:
In this latest comment I think you are talking pretty exclusively about the differences between the testing framework being located inside this git repository or not, as opposed to whether it should live in its own package or not, right?

I’m currently suggesting it move to its own package but remain in this repository. The advantage I care about most being that it can easily be used by the compiler project _and_ third parties without it needing to be maintained in both the compiler source code and the `contrib` package. 

I see this as potentially a stepping stone to it eventually being a standalone project pulled from a package registry but the two steps seem separable to me.

andylokandy:
[...]

PR #1162, opened 2021-03-06
Add a test package to the Idris 2 project

mattpolzin:
Closes https://github.com/idris-lang/Idris2/issues/1105.

This PR adds a `test` package to the Idris 2 project. It removes the recently duplicated `Golden` testing functionality from `tests/Lib.idr` and `libs/contrib/Test/Golden.idr` in favor of `libs/test/Test/Golden.idr`.

Removing the duplicated code removes opportunity for drift (for the record, the two files were already different as of this PR, though the differences were trivial to sort out).

Putting the resulting `Golden.idr` into a new testing package brings testing to the forefront and creates a home for modules that enable other styles of testing (unit and property-based, to name two non-mutually-exclusive examples).

There are changes in this PR to the way that Makefile steps depend on each other. Most prominently, building the test binary for the Idris 2 project is removed from the `all` step and `install` is added as a dependency of that test binary build step. This was done because the test binary is built and run against the version of Idris 2 installed on the system and at the same time building the test binary depends on libraries that might have been updated by the `make all` command.

The need to do this became apparent because the test binary cannot be built without the `test` package which is totally new as of this PR and therefore has not been installed before. Even without hitting a build failure like this, running tests without first installing newly built Idris files can result in misleading test failures (false negatives).

In our CI workflows, we always run `make install` before `make test` anyway, so in practice this has no impact on the guarantees made by CI. On the other hand, it eliminates the potential for someone running tests locally to mistakenly run `make all && make test` (with no install in between) which can result in failing tests cases that are false negatives.


package.nix
```diff
@@ -31,7 +31,7 @@ stdenv.mkDerivation rec {
   # TODO: Move this into its own derivation, such that this can be changed
   #       without having to recompile idris2 every time.
   postInstall = let
-    includedLibs = [ "base" "contrib" "network" "prelude" ];
+    includedLibs = [ "base" "contrib" "network" "prelude" "testing" ];
```
zhongzc, line 34: The difference between the module name `Test` and the package name `testing` increases the learning cost.

mattpolzin: Fair point -- and after all, we have "network" and not "networking" as precedent for package naming.

tests/tests.ipkg
```diff
@@ -1,5 +1,5 @@
 package runtests
 
-depends = contrib
+depends = contrib, test
```
gallais, line 3: Do we still need the dependency on `contrib`?

mattpolzin: Oh gosh, this comment is in tests. The answer must be no. Good catch.

edwinb:
I realise I'm far too late on this one, but... I don't think `make test` should `install` first. I know that there's a problem with the libraries that `make test` depends on, but we should fix that instead. The problem with installing whenever you test is, if you run tests during development - which you should be doing regularly - and happen to have some problem that causes tests to fail, you've just installed a broken version which you're then stuck with (unless you've been careful in some other ways).

We need to sort out the test script to use the right libraries, certainly, but it shouldn't involve installing first.

gallais:
Plus it's annoyingly slow to reinstall every time.

mattpolzin:
Agreed; especially in hindsight, I solved the problem with testing against libraries different than the installed versions the wrong way around. If no one else has beat me to it, I’ll try to get this sorted out soon.

PR #1177, opened 2021-03-11
[ new ] Support overloaded floating point literals

stefan-hoeck:
So far, there seems to be no way to overload floating point literals. I thought they might be useful, so this PR adds a new `%doubleLit` pragma. I added a test case and it seems to work OK, but I still might have overlooked something, so please review carefully.

### DIFF
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -122,8 +122,8 @@ HasNames e => HasNames (TTCFile e) where
           = pure $ Just $ MkRewriteNs !(full gam e) !(full gam r)
 
       fullPrim : Context -> PrimNames -> Core PrimNames
-      fullPrim gam (MkPrimNs mi ms mc)
-          = pure $ MkPrimNs !(full gam mi) !(full gam ms) !(full gam mc)
+      fullPrim gam (MkPrimNs mi ms mc md)
+          = pure $ MkPrimNs !(full gam mi) !(full gam ms) !(full gam mc) !(full gam md)
 
 
   -- I don't think we ever actually want to call this, because after we read
@@ -160,8 +160,11 @@ HasNames e => HasNames (TTCFile e) where
           = pure $ Just $ MkRewriteNs !(resolved gam e) !(resolved gam r)
 
       resolvedPrim : Context -> PrimNames -> Core PrimNames
-      resolvedPrim gam (MkPrimNs mi ms mc)
-          = pure $ MkPrimNs !(resolved gam mi) !(resolved gam ms) !(resolved gam mc)
+      resolvedPrim gam (MkPrimNs mi ms mc md)
+          = pure $ MkPrimNs !(resolved gam mi)
+                            !(resolved gam ms)
+                            !(resolved gam mc)
+                            !(resolved gam md)
 
 -- NOTE: TTC files are only compatible if the version number is the same,
 -- *and* the 'annot/extra' type are the same, or there are no holes/constraints
@@ -337,7 +340,9 @@ updatePrimNames : PrimNames -> PrimNames -> PrimNames
 updatePrimNames p
     = record { fromIntegerName $= ((fromIntegerName p) <+>),
                fromStringName $= ((fromStringName p) <+>),
-               fromCharName $= ((fromCharName p) <+>) }
+               fromCharName $= ((fromCharName p) <+>),
+               fromDoubleName $= ((fromDoubleName p) <+>)
+             }
 
 export
 updatePrims : {auto c : Ref Ctxt Defs} ->
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -123,7 +123,7 @@ HasNames e => HasNames (TTCFile e) where
 
       fullPrim : Context -> PrimNames -> Core PrimNames
       fullPrim gam (MkPrimNs mi ms mc md)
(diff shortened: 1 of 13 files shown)


src/TTImp/Elab/Ambiguity.idr
```diff
@@ -49,7 +49,8 @@ expandAmbigName mode nest env orig args (IVar fc x) exp
                         fi <- fromIntegerName
                         si <- fromStringName
                         ci <- fromCharName
-                        let prims = mapMaybe id [fi, si, ci]
+                        di <- fromDoubleName
+                        let prims = mapMaybe id [fi, si, ci, di]
```
gallais, line 53: This whole block could be refactored as a single call to `getPrimitiveNames`.

src/TTImp/Elab/Ambiguity.idr
```diff
@@ -82,15 +83,17 @@ expandAmbigName mode nest env orig args (IVar fc x) exp
 
     -- If there's multiple alternatives and all else fails, resort to using
     -- the primitive directly
-    uniqType : Maybe Name -> Maybe Name -> Maybe Name -> Name ->
```
gallais, line 85: Can we please take a `PrimNames` record instead of one `Maybe Name` per prim?

gallais (commented):
Looks good. Suggesting 2 refactorings.

src/TTImp/Elab/Ambiguity.idr
stefan-hoeck: I'm not sure this can be refactored as requested, as the values `fi`, `si` and so on are use further below in a call to `uniqueType`. But probably I just don't see the obvious.

gallais: If we refactor `uniqType` too like I suggested then it's not an issue. :D

stefan-hoeck: The thing that confuses me is that `Core.Context.getPrimitiveNames` returns a list of `Name`s, but that's not what we need after refactoring `uniqType` the way you suggested. Anyway, I tried to clean this up (hopefully in accordance to what you had in mind). Sorry for being thick. :D

gallais: Looks good!

src/Core/Binary.idr
```diff
@@ -122,8 +122,8 @@ HasNames e => HasNames (TTCFile e) where
           = pure $ Just $ MkRewriteNs !(full gam e) !(full gam r)
 
       fullPrim : Context -> PrimNames -> Core PrimNames
-      fullPrim gam (MkPrimNs mi ms mc)
-          = pure $ MkPrimNs !(full gam mi) !(full gam ms) !(full gam mc)
+      fullPrim gam (MkPrimNs mi ms mc md)
+          = pure $ MkPrimNs !(full gam mi) !(full gam ms) !(full gam mc) !(full gam md)
```
buzden, line 126: Why not `[| MkPrimNs (full gam mi) (full gam ms) (full gam mc) (full gam md) |]` instead?

stefan-hoeck: Thanks, buzden. Rebased and fixed this a part of other refactorings.

src/Core/Binary.idr
```diff
@@ -160,8 +160,11 @@ HasNames e => HasNames (TTCFile e) where
           = pure $ Just $ MkRewriteNs !(resolved gam e) !(resolved gam r)
 
       resolvedPrim : Context -> PrimNames -> Core PrimNames
-      resolvedPrim gam (MkPrimNs mi ms mc)
-          = pure $ MkPrimNs !(resolved gam mi) !(resolved gam ms) !(resolved gam mc)
+      resolvedPrim gam (MkPrimNs mi ms mc md)
+          = pure $ MkPrimNs !(resolved gam mi)
+                            !(resolved gam ms)
+                            !(resolved gam mc)
+                            !(resolved gam md)
```
buzden, line 167: The the same note as for `fullPrim` applies here.

stefan-hoeck: Well, personally, I find this version more appealing in this case, since the idiom brackets will not fit on a single line, so we do - in my opinion - not really reduce the visual noise here. But if there are strong reasons to change this, I'll be happy to include them.

buzden: Well, these two forms definitely compile to different code (i.e. `>>=` is not used when idiom brackets are used). But since it was previously defined with !-syntax, there are no strong reasons to change it.

gallais: It'd be a better refactoring to define `record PrimNames' (a : Type)` and
then have `PrimNames = PrimNames Name` so that you may define a `Traversable`
instance for `PrimNames'` and replace all of these `MkPrimNs !(f ...) !(f ...)`
with simply `traverse f primNs`.

stefan-hoeck: gallais, that doesn't work here since `Core` seems not to have an `Applicative` instance. Comments in `Core.idr` suggest that this is due to performance reasons. Therefore, I'm not sure whether this is worth the hassle as implementating `Traversable` for `PrimNames'` is quite verbose itself, since the most straightforward implementation using idiom brackets results in ambiguity errors. Still, if you think this will pay out in the long run, I'll change `PrimNames` as suggested.

gallais: I don't think it's necessary. I was just saying I'd rather see that
than a change from `MkP !(...) !(...)` to `[| MkP (...) (...) |]`.

PR #1184, opened 2021-03-14
[ makefile ] Build idris2 with local libraries

andylokandy:
Signed-off-by: Andy Lok <andylokandy@hotmail.com>

Build the libraries first and then the compiler, and use the local libraries instead of the global ones. 

This help:

1. build compiler by more versions of globally installed bootstrap, as long as the bootstrap compiler accept the grammar that the new compiler is using; 
2. not to break the compiler by changes on libraries. Without this PR, we can never accept #1091;
3. move some common things in `Libraries` to `contrib` with less concern.
4. unify the behavior with `make bootstrap` which has been using the local libraries. 

### Drawback

- `make` takes more time because it always checks and tries to rebuild the local libraries in case any new changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,8 +53,8 @@ all: support ${TARGET} libs
 
 idris2-exec: ${TARGET}
 
-${TARGET}: src/IdrisPaths.idr
-	${IDRIS2_BOOT} --build ${IDRIS2_APP_IPKG}
+${TARGET}: src/IdrisPaths.idr libs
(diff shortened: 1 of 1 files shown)


gallais:
> Without this PR, we can never accept #1091;

That sounds dodgy. We can:
* move the internals to a local version of the lib
* change `base`
* do a minor release
* move the internals back to relying on `base`

I don't think this is likely to get accepted as it would force us to keep `base`
compatible with the previously released version of the compiler which means it'd
be a pain to add any kind of new builtin type or function.

andylokandy:
Sounds reasonable. So I'll try to solve #1091 with this trick.

PR #1189, opened 2021-03-15
[ fix ] Change one implementation according to recent lib change

buzden:
A confusing thing: PR #1098 was checked as OK with CI some time ago, before the changes in the standard library. Then, the signature of the `Alternative` changed but CI wasn't rerun, thus this was merged and failing in the `master` now.

This fixes it.

I decided to repeat the implementation in order to get an advantage of the second lazy argument to be not evaluated when the first one is `Valid`


buzden:
@gallais Sorry for mentioning you, but could you, please, consider this, just to make `master` build green again.

gallais:
Thanks for the quick fix!

PR #1192, opened 2021-03-15
IdrisScript

petithug:
Permits to use Idris as a script language.


andylokandy:
Do you have any example for that?

petithug:
I use that to run tests without a recompilation step each time I add a test:

````
#!/usr/bin/env -S idris2 -p test -p contrib --exec main
import System
import Test.Golden

sipTests : TestPool
sipTests = MkTestPool []
  [ "error001"
  , "torture001"
  , "rfc3261001"
  , "rfc3261002"
  ]

httpTests : TestPool
httpTests = MkTestPool []
  [
  ]

main : IO ()
main =
  do (_ :: exe :: _) <- getArgs
       | _ => putStrLn "Executable must be passed as first parameter"
     case exe of
       "../../../build/exec/siptests" => runner [ testPaths "tests/sip" sipTests]
       "../../../build/exec/httptests" => runner [ testPaths "tests/http" httpTests]
       _ => putStrLn $ "Unknown executable: " ++ exe
  where
    testPaths : String -> TestPool -> TestPool
    testPaths dir = record { testCases $= map ((dir ++ "/") ++) }
````

**Comment by gallais** [comment]
I've added a test case for it but the patch seems partially broken on my
machine: I get a `File Not Found` error before the correct `Hello world!`
output.

**Comment by petithug** [comment]
I'll look into that tomorrow.

**Comment by chrrasmussen** [comment]
@petithug Can you explain a bit more about how you intend to use this feature?

What is the advantage of mixing shell commands and Idris code instead of having two separate files: One Idris module and one shell script?

To me this feature seems to solve a very specific use-case and the current implementation is not very cross-platform friendly. I have the following concerns:

- Embedding a shell command like this will probably not work on platforms like Windows.
- The embedded shell command expects the filename of the current file to be appended, i.e. the command on its own is incomplete.
- Mixing shell commands and Idris code does not play nice with syntax highlighting (more generally, might break IDE integrations).
- What if one wants to use literate Idris code instead of regular Idris code?


Some possible alternatives:

- If the point is to run the tests automatically, I would rather like to see a feature where we can run something like: `idris2 --test project.ipkg --watch` (`--watch` should be optional)
  * Today one can use an external tool, like [fswatch](https://github.com/emcrisostomo/fswatch) (See example below).
- If the point is to run a shell command at compile-time, I would rather like to see support for type providers (#859).

(I know these features are not easy to implement.)

---

## Example using two separate files

`run.sh`:
```sh
#!/bin/sh

idris2 -p contrib --exec main Main.idr
```

`Main.idr`:
```idris
main : IO ()
main = putStrLn "Running tests..."
```

Run the following commands:
- `chmod +x run.sh`
- `./run.sh`


### Optionally, automatically run tests on file change

Make sure [fswatch](https://github.com/emcrisostomo/fswatch#installation) is installed.

`watch.sh`:
```sh
#!/bin/sh

./run.sh # Initial run (Before any files have changed)
fswatch --one-per-batch Main.idr | xargs -n1 -I "{}" ./run.sh
```

Run the following commands:
- `chmod +x watch.sh`
- `./watch.sh`

`fswatch` can also watch directories, but make sure that the directory does not include the `build` directory because then each run will trigger a new run.

Command manuals:
- [fswatch](http://emcrisostomo.github.io/fswatch/doc/1.14.0/fswatch.html/Invoking-fswatch.html#fswatch-Options)
- [xargs](https://man7.org/linux/man-pages/man1/xargs.1.html)

---

As a side-note:

When I first saw this pull request I thought it added support for passing arguments to the built executable when using the `--exec` option: `idris2 --exec main Main.idr -- arg1 arg2`
Where `arg1 arg2` becomes available to the `System.getArgs` function.

Today one would need to do something like this: `idris2 -o foo Main.idr && ./build/exec/foo arg1 arg2`

petithug:
This PR is what it is.  It was on my stack of patches and I thought that it could be useful, but if that's not the case I'll withdraw it.  I refuse to argue with people who think it is fine to give unsolicited advice.

chrrasmussen:
Sorry, it was not my intention to give unsolicited advice. It is not up to me to decide what goes into Idris 2, but I wanted to express my concerns. I thought suggesting some alternative solutions would add some value to the discussion. I will keep this in mind in the future.

petithug:
The error is in fact unrelated to that.  To reproduce:
- Write a simple Idris program named Main.idr, with "main : IO (); main = pure ()" as content
- remove the build/ directory
- execute with: idris2 ./Main.idr --exec main

The problem is the "./", as Idris will try to create a directory named "build/ttc/.", but build/ttc does not exist yet, hence the failure.

gallais:
> Sorry, it was not my intention to give unsolicited advice.

I found the discussion useful as I am not myself well versed in
these kind of things. For every new feature we have to weight the
benefit vs. maintenance burden and this definitely helped.

The fact that this won't work across all platforms is definitely a
big minus as we'll inevitably get bug reports from people working
on an unsupported platform and the onus will be on us to port the
feature.

Given that it seems possible to get essentially the same behaviour
without patching Idris and by using an auxiliary file that is 2
lines long, I'm not sure this should get merged.

petithug:
Sure.

PR #1195, opened 2021-03-16
[ contrib ] Some functions for lazy list + fix for `foldlM`

buzden:
Sometimes I find myself wanting to switch from usual to lazy lists and missing some functions.

Also, the usual implementation of `foldlM` through `foldl` is very bad for shortcutting monads used with the lazy list. That's why I suggest to move `foldlM` inside the `Foldable` interface to make an effective specialised implementation to be definable. A test for that is included.


edwinb:
Sorry this has been a while. Thanks for the additions.

PR #1196, opened 2021-03-16
[ fix #1193 ] Fine tune REPL's typesearch

### LINKED ISSUE #1193: TypeSerach in REPL doesn't accept non-Pi type
TypeSerach in REPL doesn't accept non-Pi type 

# Steps to Reproduce

```
> idris2
Main> :s Nat
Could not parse input as a type signature.
```

# Expected Behavior

`:s Nat` shows all constant function that returns `Nat`.

gallais:
* allow non-Pi types as search targets
* do not return defs whose type is `Erased`
* do not return private definitions


gallais:
I think the right notion is `isVisibleIn` so I'll use that instead.

Issue #601, opened 2020-08-27
Idris 2 hast a very short proof of Fermat's Last Theorem


mr-infty:
# Steps to reproduce:

    module Main
    
    %default total
    
    minusZeroEqualsZero : negate 0.0 = 0.0
    minusZeroEqualsZero = Refl
    
    minusInftyEqualsPlusInfty : recip (negate 0.0) = recip 0.0
    minusInftyEqualsPlusInfty = cong recip minusZeroEqualsZero
    
    f : Double -> Bool
    f x = x < 0
    
    trueEqualsFalse' : f (recip (negate 0.0)) = f (recip 0.0)
    trueEqualsFalse' = cong f minusInftyEqualsPlusInfty
    
    fInftyEqualsFalse : f (recip 0.0) = False
    fInftyEqualsFalse = Refl
    
    fMinusInftyEqualsTrue : f (recip (negate 0.0)) = True
    fMinusInftyEqualsTrue = Refl
    
    trueEqualsFalse : True = False
    trueEqualsFalse = trans (sym fMinusInftyEqualsTrue) $ trans trueEqualsFalse' fInftyEqualsFalse
    
    voidIsInhabited : Void
    voidIsInhabited = uninhabited trueEqualsFalse
    
    infix 10 ^
    (^) : Nat -> Nat -> Nat
    (^) x Z = 0
    (^) x (S n) = (x^n) * x
    
    fermatsLastTheorem : (n : Nat) -> (3 <= n = True) ->
                         (x,y,z : Nat) -> (x <= 1 = True) -> (y <= 1 = True) ->
                         (x^n + y^n = z^n) -> Void
    fermatsLastTheorem n pf_n x y z pf_x pf_y pf = voidIsInhabited
    
    main : IO ()
    main = do putStrLn "I have discovered a truly remarkable proof of this theorem that even fits in this margin."

# Expected behaviour

You shouldn't be able to prove that `-0.0` equals `0.0`.

gallais:
:fireworks: 

@wenkokke will be happy to see that Idris' Float are extremely powerful.

RyanGlScott:
[...]

PR #1203, opened 2021-03-17
[ fix #1200 ] Switch scheme backend to flonum functions

### LINKED ISSUE #1200: Syntax for complex numbers
Idris 2 supports complex numbers, but lacks an easy way to write them. Presently, the only way I know to get the imaginary unit `i` is with,
`sqrt (-1)`
or,
`log (-1) / pi`
Since stringification will happily print `0.0 + 1.0i` after evaluating `log (-1) / pi`, and since `the Double (cast "0.0+1.0i")` will return successfully, it stands to reason that it should work the same way in source code. Most other types have a bijection between parsing and printing, and it stands to reason that the complex part of the double type should also.

Of course it is possible to write expressions that evaluate to complex numbers, but that lacks the nice property of the printed form of types being valid source code. Idris would be best if it only had one Double parsing behavior, and not two subtly distinct behaviors, one for parsing source, and the other for `cast`.

Atomotron:
Previously, `sqrt(-1.0)` would cause Chez scheme to return a complex type, which is unsound because Idris would believe it to be a Double. This change invokes specialized floating point functions which do not return complex numbers under any circumstances.

All tests pass with the Chez Scheme target after this change. I haven't tried any other targets, and I don't know how well floating point operations are covered.

Another reasonable TODO would be to add regression tests for this bugfix, because future targets may introduce a similar error. For example, Python's `math.sqrt` throws an exception when passed `-1.0`. 

See [issue #1200.](https://github.com/idris-lang/Idris2/issues/1200) for discussion.


Atomotron:
The check failed because Racket doesn't have flexp in the prelude.

Atomotron:
Tested with racket, chez and gambit codegen. Gambit doesn't need a special import, but Racket needed to have an import added to the codegen's boilerplate.

gallais:
This should at least test that the answers are consistent across chez & racket.

gallais:
Lol, on mac specifically the chez primitives return different values.

```diff
30c30
< -1.5574077246549023
---
> -1.557407724654902
34c34
< 1.5574077246549023
---
> 1.557407724654902
38c38
< -0.5235987755982989
---
> -0.5235987755982988
40c40
< 0.5235987755982989
---
> 0.5235987755982988
47c47
< 1.0471975511965979
---
> 1.0471975511965976
50c50
< -1.1071487177940904
---
> -1.1071487177940906
52c52
< -0.4636476090008061
---
> -0.46364760900080615
54c54
< 0.4636476090008061
---
> 0.46364760900080615
56c56
< 1.1071487177940904
---
> 1.1071487177940906
```

tests/idris2/evaluator005/Issue1200.idr
```diff
@@ -0,0 +1,9 @@
+module Issue1200
+
+%default total
+
+main : IO ()
+main = traverse_ printLn $ do
+  f <- [(1/), log, sin, cos, tan, asin, acos, atan, sqrt, floor, ceiling]
+  x <- [-2,-1,-0.5,0,0.5,1,2]
+  pure (f x)
```
ohad, line 9: Can change to `traverse_ printLn $ compare expected $ do...` where:
```idris
data Result = UpTo Double Double | PosInf | NaN ...
||| check whether the double is in the list of results
check : List Result -> Double -> Bool

expected : List [Result] -> List Double -> List Bool
expected results = map check . zip results
```

gallais: The point made in #623 is that different results in different backends will lead
to issues (runtime crashes in the best cases) so I don't think we want to write a
test case that is explicitly accepting of such a broken state of affairs.

I wonder whether this could be a version problem. Brew tends to install newer versions
AFAICT. I'll check after our weekly meeting.

ohad: I see. #623 says the result is non-deterministic, based on optimisation levels and architecture, so as @wenkokke said, we're going to get different behaviour between type-checking time and runtime, which you can bootstrap to runtime inconsistencies.

The #601 shows that the incongruence of floats means they're unsafe for type-checking time anyway.

Maybe all floating point operations should be monadic?

Atomotron: It is worse than merely compile-time versus runtime. The same compiled program can arrive at different results for the same floating point computation, on the same computer, run at different times. That's because the settings of the FPU can be changed at runtime! For example, control-F for "6.2.3.3  Rounding Control" in the [AMD x86_64 architecture manual](https://www.amd.com/system/files/TechDocs/24592.pdf). That's right - the x87 rounding mode is set by a couple of register bits, and can be changed any time!

This leads us to a serious dilemma. Since floating point operations depend on a part of CPU state, there is no such thing as a pure floating point function. I guess that means it should be monadic like @ohad said (I am a beginner with functional programming concepts so I can't say much about it). Nonetheless, I _deeply_ and _sincerely_ hope that we can keep writing code that processes Doubles in a natural way.

Atomotron:
I think it might be a good idea to allow for small differences in the test results. I say this because a complete solution to floating point nondeterminacy will require a lot of work. In the meantime, it would be good to take a short-term marginal decrease in the number of bugs (by solving the complex numbers problem) while awaiting the research necessary to solve all floating-point problems in a proof engine. 

To my understanding, Floq is not based on machine floats, but is instead a well-specified emulation. A practical programming language would probably need machine floats, and although [there has been some good work on that](https://drops.dagstuhl.de/opus/volltexte/2019/11062/pdf/LIPIcs-ITP-2019-7.pdf), the problem of nondeterminacy has not been resolved. The linked paper suggests and declines a monadic interface for rounding modes and doesn't bring up transcendental functions, although I highly recommend it for a review of approaches to arithmetic.

Accepting small amounts of nondeterminacy is not ideal, but it has a history of working well enough to survive by. To pick one example, LAPACK [runs its tests based on thresholds.](https://github.com/Reference-LAPACK/lapack/blob/c26d01b8e4d04d3d1b394cd6d1d6ec9335a4c30b/TESTING/LIN/dorhr_col02.f#L259) It may be an acceptable stopgap measure.

gallais:
This is a net improvement over the current situation so I'll just turn off
the failing test case and merge if the build goes through. We can revisit
this error in the future if there is a need.

PR #1207, opened 2021-03-18
[ fix #710 ] Enforce assumptions about capitalised idents

### LINKED ISSUE #710: Namespace name starting with lowercase letter doesn't work
# Steps to Reproduce

The following code doesn't work:
```
namespace a
  export
  tf : Int -> Int -> Int
  tf x y = x + y

namespace b
  export
  tf : Int -> Int -> Int
  tf x y = x + y

t2 : Int
t2 = a.tf 1 2
```

Changing the namespace names to A,B fixes the problem:

```
namespace A
  export
  tf : Int -> Int -> Int
  tf x y = x + y

namespace B
  export
  tf : Int -> Int -> Int
  tf x y = x + y

t2 : Int
t2 = A.tf 1 2
```

# Expected Behavior

Namespaces names a,b work correctly for Idris 1

gallais:
Given we keep getting tripped up by this, here we go:

* Namespaces
* Data names
* Record names
* Data constructor names (except for operators)
* Record constructor names (except for operators)
* Interface constructor names (except for operators)

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1166,10 +1166,10 @@ usingDecls fname indents
                          commit
                          symbol "("
                          us <- sepBy (symbol ",")
-                                     (do n <- option Nothing
+                                     (do n <- optional
                                                 (do x <- unqualifiedName
                                                     symbol ":"
-                                                    pure (Just (UN x)))
+                                                    pure (UN x))
                                          ty <- typeExpr pdef fname indents
                                          pure (n, ty))
                          symbol ")"
@@ -1290,7 +1290,7 @@ ifaceDecl fname indents
                                      (do symbol "|"
                                          sepBy (symbol ",") name)
                          keyword "where"
-                         dc <- option Nothing (Just <$> recordConstructor indents)
+                         dc <- optional (recordConstructor indents)
                          body <- assert_total (blockAfter col (topDecl fname))
                          pure (\fc : FC => PInterface fc
                                       vis cons n doc params det dc (collectDefs (concat body))))
@@ -1304,10 +1304,10 @@ implDecl fname indents
                          let opts = mapMaybe getRight visOpts
                          col <- column
                          option () (keyword "implementation")
-                         iname <- option Nothing (do symbol "["
-                                                     iname <- name
-                                                     symbol "]"
-                                                     pure (Just iname))
+                         iname <- optional (do symbol "["
+                                               iname <- name
+                                               symbol "]"
+                                               pure iname)
                          impls  <- implBinds fname indents
                          cons   <- constraints fname indents
                          n      <- name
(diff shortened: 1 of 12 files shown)


gallais:
I'm not sure why the error is being reported at the beginning of the next token.
Do I need to change my implementation of `capitalisedName`?
cc @andrevidela @andylokandy

alexhumphreys:
Ha I tried making this change just for Namespaces and got nowhere. Pretty educational to see how it's done!

andrevidela:
@gallais It's because `fail` reports the error at the location at the upcoming token since your parser `identPart` has already consumed part of the input, the error is reported for the next token.

What I would suggest is to use `bound` for `identPart` and retrieve the bounds of the token, then pass them to your predicate functions and use the same mechanism as `fatalLoc` in the parser to set the location of the error. There is no `failLoc` for a recoverable error so you'll need to write it.

The end result would look something like this:

```
isCapitalisedIdent : WithBounds String -> SourceEmptyRule ()
 isCapitalisedIdent str =
   let err : SourceEmptyRule ()
           = failLoc str.bounds ("Expected a capitalised identifier, got: " ++ str.val)
           ...

capitalisedName : Rule Name
 capitalisedName = do
   id <- bounds identPart
   isCapitalisedIdent id
   isNotReserved id
   pure (UN id.val)
```

buzden:
I think not only record and data constructors, but also constructor of interfaces should also follow the same rule.
I mean, even with you modifications code
```idris
interface Lalala where
  constructor cons
  x : Int -> Int
```
parses but shouldn't.

buzden:
Wow. I found myself defining a `public export %inline` functions with trivial bodies of `constr = Constr` after renaming the data constructors from `constr` to `Constr` to adopt this PR. I need lower-cased names because data constructors were used in a DSL. 

There were original problems with lowercase namespace names and lowercase module names, since `a.b` was ambiguous when `a` could be a namespace or module name.

But what was the main point of making namely _data constructors_ to start with the upper case? I mean, of course, pattern matching became easier since no need to pattern match a lower-case data constructor as `Type.constr`, now just `Constr`, but still.

gallais:
`record` define their own namespace named after their type constructor
so these need to be uppercase. For uniformity & anticipating @Russoul's
work on datatypes defining their own namespaces, I decided to have capitals
everywhere.

buzden:
I'm talking about data constructors, not type constructors. Like
```idris
data D : Type -> Type where
  c1 : D Int
```

`c1` is not going to define a namespace even in #1112, so why to require it to start with an upper-case letter?

buzden:
I'm talking about data constructors, not type constructors. Like
```idris
data D : Type -> Type where
  c1 : D Int
```

`c1` is not going to define a namespace even in #1112, so why to require it to start with an upper-case letter?

Issue #1185, opened 2021-03-14
Test suite improvements


andrevidela:
The test suite for Idris2 is extremely important since it ensures that regressions are tracked and new features are developped without impeding on existing functionality. However there is still room for improvement:

## Feature requests

- [ ] Pick the number of threads to use when running the tests
- [ ] Run sequentially and stop at the first failure
- [ ] List which tests failed at the end of a test run when running non-interactively

CodingCellist:
Re: Running sequentially and stopping at the first failure. Would it be useful to have that as a separate feature, or is it enough that that should be doable using `--threads 1`?

andrevidela:
I think it needs to happen as a separate feature in non-interactive mode

gallais:
Can we add

- [ ] Option to ring a bell when you've found a mismatch

? Even in interactive mode, there are so many tests now that I tend to look away
and check github while they're going. It'd be nice to be warned if there is something
that requires my attention.

gallais:
In #1394 I

* List the failing tests at the end
* Provide an option `--failure-file PATH` to write the list to a file
* Provide an option `--only-file PATH` to read the list of tests to run from a file

Combining the two, you can easily define a `make test` that always writes the list to e.g. `failures`
and a `make retest` that will read the list from `failures` meaning you should be able to quickly iterate
until you've solved all of the failing test cases by repeatedly running `make retest`.

fabianhjr:
Regarding "ctrl+c", I think that is a rough edge of the futures, in principle they shouldn't be capturing those interrupts but there is also no code to deal with interrupts in a (IO a). DX

fabianhjr:
As of recently there is a new signals library! https://github.com/idris-lang/Idris2/pull/1458

Will try to learn it to improve the CTRL+C Behavior.

mattpolzin:
@fabianhjr i played with the test framework a bit after adding the signal library and hit some weirdness I didn't fully sort out yet -- I'll be curious to see if you get something working or at least come away with ideas as to what is not quite going right. 

I did find one legitimate bug so far, fixed in upcoming https://github.com/idris-lang/Idris2/pull/1528

PR #1209, opened 2021-03-19
[ re #1185 ] Add primitive for obtaining number of processors

CodingCellist:


libs/base/System/Info.idr
```diff
@@ -14,3 +14,11 @@ codegen = prim__codegen
 export
 isWindows : Bool
 isWindows = os `elem` ["windows", "mingw32", "cygwin32"]
+
+%foreign "C:idris2_getNProcessors, libidris2_support"
+prim__getNProcessors : PrimIO Int
+
+export
+getNProcessors : IO Int
```
gallais, line 22: Should we make this a `Nat`?

support/c/windows/win_utils.c
```diff
@@ -78,3 +78,98 @@ int win32_modenv(const char* name, const char* value, int overwrite) {
...
+
+    // shortcut to a function (?)
+    LPFN_GLPI glpi;
+    glpi = (LPFN_GLPI) GetProcAddress( GetModuleHandle(TEXT("kernel32"))
+                                     , "GetLogicalProcessorInformation"
+                                     );
+    // to keep track of whether we're at the end of the array
+    DWORD byteOffset = 0;
+    
+    // repidiately try to malloc and retrieve the information until we have a
```
gallais, line 112: typo: repeatedly

libs/base/System/Info.idr
CodingCellist: I don't think so, since the C functions can return -1 to indicate an error.
We should maybe do something clever with error reporting (e.g. `Either`), but it's difficult due to the different ways Linux, Mac, and Windows do readable error messages and error codes.

gallais: Then a `Maybe Nat`? Basically a safe wrapper for the primitive.

tests/base/system_info001/NumProcessors.idr
```diff
@@ -1,6 +1,7 @@
 import System.Info
 
 main : IO ()
-main = do np <- getNProcessors
+main = do Just np <- getNProcessors
+            | Nothing => putStrLn "ERROR: Couldn't get number of processors!"
           putStrLn $ (show np) ++ " processors"
```
gallais, line 6: Argh, sorry about that.

CodingCellist: No problem at all. Thanks for implementing the `Maybe Nat` wrapper! ^^

PR #1210, opened 2021-03-19
[ install ] Check if 'realpath' exists for Chez and Racket backends

ska80:


src/Compiler/Scheme/Racket.idr
```diff
@@ -326,18 +326,27 @@ startRacket : String -> String -> String -> String
...
-    , racket ++ "\"`dirname \"$DIR\"`\"/\"" ++ target ++ "\" \"$@\""
+    , "if ! command -v \"$REALPATH\" >/dev/null                     "
+    , "then                                                         "
+    , "    echo \"$REALPATH is required for Racket code generator.\""
+    , "    exit 1                                                   "
+    , "fi                                                           "
+    , ""
+    , "DIR=$($REALPATH \"$0\")"
+    , "export LD_LIBRARY_PATH=\"$(dirname \"$DIR\")/\"" ++ appdir ++ "\":$LD_LIBRARY_PATH\""
+    , racket ++ "\"$(dirname \"$DIR\")\"/\"" ++ target ++ "\" \"$@\""
```
benhormann, line 349:
```suggestion
    , "DIR=$(dirname \"$($REALPATH \"$0\")\")"
    , "export LD_LIBRARY_PATH=$DIR/" ++ appdir ++ ":$LD_LIBRARY_PATH"
    , racket ++ "\"$DIR/" ++ target ++ "\" \"$@\""
```
This removes a few superfluous quotes too. (Unless `appdir` and `target` are already shell quoted?)
BTW, you've changed to "Exit" in existing comments but not in new ones.

---

To reviewers: note the change in order, as was done in #709 for Chez.

ska80: Thanks, I applied your suggestions but I left `appdir` and `target` quoted just in case.

> BTW, you've changed to "Exit" in existing comments but not in new ones.

Could you point out the line where this is a case?

ska80:
Still don't understand why the shell linter throws an error.

benhormann:
> Still don't understand why the shell linter throws an error.

From the diffs in the log, it expects different indentation. The diffs add hard tabs... ☹️
Also, these files haven't been changed since super-linter was added.

If it flagged `bootstrap/idris2-{,rkt}boot`, why didn't it flag `tests/chez/*/run`?

---

@gallais [.editorconfig](https://github.com/idris-lang/Idris2/blob/64af41c/.editorconfig) isn't specifying indentation for files without an extension.

gallais:
`expected` is explicitly excluded [here](https://github.com/idris-lang/Idris2/blob/64af41c298dabd473f47f42a559d1d4281c19eff/.github/linters/.ecrc)

I don't know about run.

benhormann:
@gallais I don't see how `bootstrap/*` relates to `expected`?

I also don't see why the bootstrap scripts needs to call `realpath`? During the bootstrap stages, it could just use `$(dirname "$0")`.

gallais:
The `run` files very much have to respect the conventions AFAIK.
You can get shouted at for trailing newlines at the end of the files.

Hence my mention of `expected` to answer your question because AFAIK the
run files are accepted (and if they are not it's because they were added
before the linter was setup and haven't been touched since):

> why didn't it flag tests/chez/*/run?
> @gallais .editorconfig isn't specifying indentation for files without an extension.

ska80:
@gallais I think I have managed to satisfy the shell linter.  Though, I had to add the `.sh` suffix to bootstrap scripts so the shell linter could use indentation settings from `.editorconfig`.

benhormann:
FYI, I'm working on a patch to replace the realpath dependency with pure shell. It removes the `uname` branching, so is less lines in total.

Issue #1101, opened 2021-02-21
Overflowing Int literals


stefan-hoeck:
Seems like `Int` literals are not properly truncated (the BitsXY primitives seem to work correctly).

# Steps to Reproduce

Run

```idris
main : IO ()
main = do printLn (the Int 10000000000000000000000000000)
          printLn (the Int 1 * 10000000000000000000000000000)
```

# Expected Behavior

Prints
```
4477988020393345024
4477988020393345024
```

# Observed Behavior

Prints
```
10000000000000000000000000000
4477988020393345024
```

gallais:
This is related to #1048: `Int` have no set meaning AFAIK and the
implementation is backend-dependent. This is obviously a bad situation
to be in and we should introduce `Int32` and `Int64` and decide which
one `Int` is an alias for.

stefan-hoeck:
OK, thanks. Look, I could give this a try and add Int32 and Int64 given that I have already been working on primitives regarding their bitops. However, I'm not sure I'm up to the task and will definitely need some guidance along the way (especially with the C backend: I am not much of a C programmer). Or is this already planned and scheduled by the core Idris team (in which case I'll happily step aside, :-) )?

Personally, I'd opt for making Int an alias for Int64 as this is how it is handled right now in the JS and Scheme backends. In JS, Int32 could be represented by `Number` (thus resolving #801) and Int64 by BitInt (as Int is right now).

stefan-hoeck:
Fixed in #1211

Issue #1048, opened 2021-02-09
Fix flaky Int


stefan-hoeck:
## The `Int` type right now
To the best of my knowledge, `Int` is implemented as a signed 64bit integer across all backends. Overflows are handled by using modular arithmetic to calculate the remainder modulo 2^63. This is, as far as I know, not specified anywhere in the documentation of Idris2, and it is not clear how much is already assumed about `Int`s behavior in the compiler and libraries. However, the actual implementation of `Int` has led to several issues recently. Here is a non-comprehensive list of some of them:

- Not all backends have native support for 64bit integers. This is for instance an issue in JS, where we are forced to use BigInt to represent `Int` instead of being able to use the more performant `Number` data type (see  #801).
- The smallest negative `Int` (-9223372036854775808) does actually not exist, as it is being truncated to 0.
- Bit shifts on integers do not behave as expected. It is, for instance, not possible, to left-shift a positive number into the negative range (see also #1033).
- It is not possible to test the highest bit of negative numbers without first switching to `Integer`, since `testBit` is not a primitive, and implementing it via logical `and` is not possible for the same truncation reasons given above.

I therefore would like to "fix" `Int` as it is right now, by choosing one of two possible pathways.

## Solution A: Leave implementations untouched but drop support for bitwise operations
So far, bitwise operations on `Int` are hardly used in the repo (the only exception being `System.Random` in contrib), so this would probably not hurt too much. We could argue that bitwise ops should be used on one of the `BitsXY` types anyway, where they are properly supported. While this would solve the issues with bitwise operations, it still would mean -9223372036854775808 would not exist (probably not a big issue).

## Solution B: Use bitmasks to truncate `Int`s on bitwise ops
This would reflect the behavior of native Word data types in case of overflows: Out of bounds bits are truncated, in case of positive numbers (bit 64 unset) by logical and-ing with (2^63-1), in case of negative numbers (bit 64 set) by logical or-ing with ((-1) << 63) (this is necessary, since negative numbers are typically represented as [two's complement](https://en.wikipedia.org/wiki/Two%27s_complement)). This would solve the issues described above, but might break code already relying on the actual truncation implementation.

## Addendum: `Int` on JS
[...]

PR #1211, opened 2021-03-19
[ refactor ] Cleanup integral primops

stefan-hoeck:
I'm in the process of adding Int32 and Int64 primitives as discussed in several issues and pull requests (#1101, #1048). One main obstacle is the implementation of all the primops, which turns out to be cumbersome and error prone especially when it comes to casts between integral types, the number of which scales with the square of the number of integer primitives. This is even more pronounced in the JS backends, where we plan to have two representations for integral types (Number and BigInt).

This PR adds new `Precision` and `IntKind` data types, which should simplify the task of adding the new primitives. So far, integral primops for the scheme backends have been reimplemented. If people agree that this is the way to go, I'll also adjust the JS implementations.


stefan-hoeck:
This should be ready for reviewing now. I just factored out some reoccuring patterns in the definition of integer operations. Once this gets accepted, I'll change some implementations in the backends to get better truncation behavior.

Please not: I did not touch the C backend. Someone else will have to look into that.

src/Compiler/ES/ES.idr
```diff
@@ -194,50 +194,64 @@ jsConstant (B32 i) = pure $ show i ++ "n"
...
+       -- than the one of the source, there is no need to cast.
+       (Just $ Unsigned m, Just $ Signed $ P n) =>
+         if P n > m then pure x else boundedInt (n-1) x
+
+       (Just $ Signed _, Just $ Unsigned $ P n) => boundedUInt n x
+
+       (Just $ Unsigned m, Just $ Unsigned $ P n) =>
+         if P n >= m then pure x else boundedUInt n x
+
+       _ => jsCrashExp $ jsString $ "invalid cast: + " ++ show from ++ " + ' -> ' + " ++ show to
```
gallais, line 244: Should this be a `throw`? The backends tend, too often, to happily generate garbage
rather than complaining that an error occurred and that makes debugging pretty painful.

tests/chez/chez015/Numbers.idr
```diff
@@ -9,7 +9,7 @@ large : (a: Type) -> a
 large Integer = 3518437212345678901234567890123
 -- int close to 2ˆ63
 -- we expect some operations will overflow
-large Int = 3518437212345678901234567890
+large Int = 9223372036854775800
```
gallais, line 12: It's unfortunate that you are changing the test case at the same time as you are refactoring
because now we can't tell whether the refactoring is innocuous.

src/Compiler/ES/ES.idr
stefan-hoeck: I'll look into this. Where do I find out more about `throw`?

tests/chez/chez015/Numbers.idr
stefan-hoeck: Well, I didn't like this test case for two reasons:

- the original `Int` is nowhere near 2^63, so either the comment or the literal is false
- the original `Int` is out of bounds. From #1101 we know, that right now Idris2 will not truncate this as it should, therefore is hard to figure out, what the point of this test was in the first place. Eventually (when we fix #1101), this needs to be changed anyway, doesn't it?

The new test case still makes sure that truncation behaves correctly. Still, if you think we should go back to the old test, I'll revert this change.

stefan-hoeck: OK, so here's the problem with this test case: It looks like this refactoring already fixes #1101, since the following line, which did not truncate when going from `Integer` to `Int` was replaced with the call to `castInt`, which does proper truncation for all combinations of integer types:

```idris
schOp (Cast IntegerType IntType) [x] = x
```

But with this change, the old version of the test fails, and it is the calls to `div` and `mod` that fail since

```
(a `mod` b) `mod` c/= (a `mod` c) `mod` b
```

in the general case.

src/Compiler/ES/ES.idr
stefan-hoeck: Should be fixed now.

gallais:
Beautiful refactoring :rocket:

PR #1225, opened 2021-03-24
[ fix #1220 ] Update arity of constuctors to reflect erased args

### LINKED ISSUE #1220: Inconsistent constructor arity
The arity of constructors is inconsistent between the constructor declaration (eg `MkACon`) and case patterns and application

# Steps to Reproduce
test.idr
```idris
main : IO Int
main = pure $ sum $ [1 .. 100]
```
`$ idris2 --dumpanf test.anf -o test test.idr`

# Expected Behavior

Arity of constructor declaration matches arity of constructor use in case patterns etc.

# Observed Behavior

Arity is inconsistent
```
...(%con Builtin.MkPair(v7, v8))...
Builtin.snd = [0, 1, 2]: %case v2 of { %conalt Builtin.MkPair(v3, v4) => v4 Nothing }
Builtin.fst = [0, 1, 2]: %case v2 of { %conalt Builtin.MkPair(v3, v4) => v3 Nothing }
Builtin.MkPair = Constructor tag Just 0 arity 4 newtype by Nothing
```

Z-snails:
fixes https://github.com/idris-lang/Idris2/issues/1220

I have checked a few test files and the arity matches in all cases except `IORes`
```
PrimIO.MkIORes = Constructor tag Just 0 arity 3 newtype by Just 1
```

Is this because `IORes` escapes type erasure or have I done something wrong?


buzden:
Maybe, a test case should be added for this fix, e.g. running your original example

gallais:
Looking at the way the code is structured, I think it would make more sense to
have this change be part of `eraseConArgs` where the `numArgs` information is
already available.

Chasing down where the arguments are actually dropped, I ended up with something
like this:

```diff
-dropPos epos (CCon fc c a args) = CCon fc c a (drop 0 args)
+dropPos epos (CCon fc c a args) = CCon fc c (a `minus` length epos) (drop 0 args)
```

Z-snails:
I don't understand your suggestion for how to fix the inconsistency, `a` is the tag not the arity. And `CCon` doesn't directly store the arity.

gallais:
Ok, my bad. I was trying to ensure that we perform the erasure & the analysis in
the same place so that they are kept in sync but apparently that's not possible. :/

PR #1233, opened 2021-03-26
[ fix #1230 ] Better error messages for out-of-scope identifiers

### LINKED ISSUE #1230: Improve out-of-scope error messages by listing existing similar names
# Steps to Reproduce

```haskell
record R where
  constructor MkR
  field : Nat

myRec1 : R
myRec1 = MkR 3
```
works. But the following example
```haskell
record R where
  constructor MkR
  field : Nat

myRec1 : R
myRec1 = MkR 3

mkRec2 : R
myRec2 = MkR 3
```
fails with error "No type declaration for Main.myRec2."
Tested on 71abc8e33

# Expected Behavior

Second example should work.

gallais:
To identify names close to the out of scope one the user inputed,
we use a modified version of the Levenshtein distance making it
more expensive to substitute an alpha / num character with one that
is a of a different category.

### DIFF
--- a/idris2api.ipkg
+++ b/idris2api.ipkg
@@ -106,6 +106,7 @@ modules =
     Libraries.Data.DList,
     Libraries.Data.IMaybe,
     Libraries.Data.IntMap,
+    Libraries.Data.IOMatrix,
     Libraries.Data.LengthMatch,
     Libraries.Data.List.Extra,
     Libraries.Data.List.Lazy,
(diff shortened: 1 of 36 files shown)


buzden:
Looks like a one-letter names shoudn't be searched with this algorithm because
- people usually see one-letter misprints easier than one letter inside a log indentifier;
- standard `S`, `Z` and etc. identifiers are constantly showing in every error involving a one-letter identifier which seems to be annoying.

PR #1234, opened 2021-03-26
[ performance ] add fastUnlines to base

stefan-hoeck:
I am working on an algorithm, which generates several tens of thousands lines of Idris2 code. Replacing all occurences of `unlines` with `fastUnlines` as presented in this PR, brought the running time down by two orders of magnitude (from > 50 s to < 0.5 s on my system). I therefore thought this might be useful for others as well.


gallais:
It'd be nice to have a mechanism to automatically substitute one for the other at compile time.

Issue #877, opened 2020-12-27
Improving messages expected in defining data types using lowcase identifier and correspondencing functions, errors in order to avoid ambiguity.


ghost:
# Steps to Reproduce

Version 0.2.1-b0e6cdc16

```idris
module Bug
data TeXNat = zero | succ TeXNat
dts : TeXNat -> String
dts n = case n of
  zero => "zero"
```

# Expected Behavior

Report can not define data types with lowercase identifier.

# Observed Behavior

```
alissa@stackage ~/scratch % idris Bug.idr
     ____    __     _         ___                                           
    /  _/___/ /____(_)____   |__ \                                          
    / // __  / ___/ / ___/   __/ /     Version 0.2.1-b0e6cdc16
  _/ // /_/ / /  / (__  )   / __/      https://www.idris-lang.org           
 /___/\__,_/_/  /_/____/   /____/      Type :? for help                     

Welcome to Idris 2.  Enjoy yourself!
Bug> :l Bug.idr
Loaded file Bug.idr
Bug> 
Bye for now!
```

buzden:
What's wrong with lowercase data constructors? You can even match on them:

```idris
module LowerCaseData

%default total

data TeXNat = zero | succ TeXNat
dts : TeXNat -> String
dts n = case n of
  LowerCaseData.zero   => "zero"
  LowerCaseData.succ _ => "non-zero"
```

I think everything works as expected.

alissa42:
> What's wrong with lowercase data constructors? You can even match on them:
> 
> ```idris
> module LowerCaseData
> 
> %default total
> 
> data TeXNat = zero | succ TeXNat
> dts : TeXNat -> String
> dts n = case n of
>   LowerCaseData.zero   => "zero"
>   LowerCaseData.succ _ => "non-zero"
> ```
> 
> I think everything works as expected.

That may become the problem of show the messages for ambiguous occurrences of `zero`, for syntax to use lowercase identifier to distinct type variables from data constructors?

buzden:
> That may become the problem of show the messages for ambiguous occurrences of `zero`, for syntax to use lowercase identifier to distinct type variables from data constructors?

Yes, it may. But I think that this is not a problem of what identifiers are allowed in data declarations. So, maybe error messages should be improved, for example printing the point where a name is defined when using the name in an error message.

gallais:
We should at least warn users when they define a datatype with
lowercase constructors as that breaks a lot of the assumptions
made in the compiler.

[...]

PR #1235, opened 2021-03-26
Restore an ability for data constructors to start with a lower-case char

buzden:
Following the discussion at #1207, _data constructors_ (like `c1` in the example below, unlike _type constructors_) do not define and are not going to define a namespace. Thus they still could start with an upper-case letter with no ambiguity.

```idris
data D : Type -> Type where
  c1 : D Int
```

### DIFF
--- a/src/Parser/Rule/Source.idr
+++ b/src/Parser/Rule/Source.idr
@@ -297,11 +297,11 @@ capitalisedName = do
 
 export
 dataConstructorName : Rule Name
-dataConstructorName = opNonNS <|> capitalisedName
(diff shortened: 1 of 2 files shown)


gallais:
I'm 100% opposed to this: it breaks existing assumptions about what
constructors look like and lead to puzzling errors cf. #877 where
the suggested solution is to use qualified name because otherwise
the constructor is treated as a variable binding.

buzden:
> This is not an unintended consequence.

Oh, now I see you point. I thought namespace problem was the only thing.

PR #1237, opened 2021-03-27
[ doc ] Custom backend cookbook

andorp:
During the preparation of my Bob2021 talk I wrote this document. It summarizes the thing I encountered with during the design/experiment/implementation of the STG backend. I hope this information could be useful for the community.

I encourage everyone to suggest improvements or just add extra commits to this PR as English is not my native language and I suspect there are many grammatical mistakes or possible Hunglish can be found in the text.

@edwinb I mentioned your SPLV20 talk at the beginning of the doc. I hope it's not a problem, I am also happy to remove that part if needed.

### DIFF
--- a/docs/source/backends/backend-cookbook.rst
+++ b/docs/source/backends/backend-cookbook.rst
@@ -20,11 +20,10 @@ The first codegen in the list will be set as the default codegen.
 Anyone who is interested in implementing a custom back-end needs to answer the
 following questions:
 
-- Which IR should be consumed by the custom back-end?
+- Which Intermediate Representation (IR) should be consumed by the custom back-end?
 - How to represent primitive values defined by the ``Core.TT.Constant`` type?
 - How to represent Algebraic Data Types?
 - How to implement special values?
-- Association between data and type constructors.
 - How to implement primitive operations?
 - How to compile IR expressions?
 - How to compile Definitions?
@@ -116,8 +115,8 @@ much as possible from it for the custom back-end.
 Another possibility is to implement a runtime that is capable of handling memory
 management and threading.
 
-Which IR should be consumed by the custom back-end?
----------------------------------------------------
+Which Intermediate Representation (IR) should be consumed by the custom back-end?
+---------------------------------------------------------------------------------
 
 Now lets turn our attention to the different intermediate representations (IRs)
 that Idris provides.
@@ -134,7 +133,7 @@ The question to answer here is: Which one should be picked?
 Which one fits to the custom back-end?
 
 How to represent primitive values defined by the ``Core.TT.Constant`` type?
------------------------------------------------------------------------
+---------------------------------------------------------------------------
 
 After one selects the IR to be used during code generation, the next question
 to answer is how primitive types should be represented in the back-end.
@@ -176,7 +175,8 @@ Strings are encoded in UTF-8.
 
 It is not always clear who is responsible for freeing up a ``String`` created by
 a component other than the Idris runtime. Strings created in Idris will
-always have value.
+always have value, unlike possible String representation of the host technology,
+where for example NULL pointer can be a value, which can not happen on the Idris side.
 This creates constraints on the possible representations of the Strings in the
 custom back-end and diverging from the Idris representation is not a good idea.
 The best approach here is to build a conversion layer between the string
@@ -264,11 +264,8 @@ in the IR:
   two parameters.
 - One for the ``Left`` constructor with arity of three.
   Three may be surprising, as the constructor only has one argument in Idris,
-  but we should keep in mind the type parameters for  the data type too.
-  Although the arguments associated with types can be erased in certain cases
-  and they are not real part of the constructor arguments, the number of real arguments needs to
-  be computed. See later in the 'How to compile IR expressions?' section.
-- One for the ``Right`` constructor with arity of three. Same as above.
+  but we should keep in mind the type parameters for the data type too.
+- One for the ``Right`` constructor with arity of three.
 
 In the IR constructors have unique names. For efficiency reasons,
 Idris assigns a unique integer tag to each data constructors so that constructor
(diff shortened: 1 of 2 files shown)


andorp:
Please checkout this branch and compile the doc to see if the formatting of the document reaches the standards of the rest of the documentation.

gallais:
> During the preparation of my Bob2021 talk

[Link for the curious](https://www.youtube.com/watch?v=WjFIcr2ruYc)

andorp:
@gallais Thanks :)

gallais:
That was a long read! I have a few questions:

> Strings created in Idris will always have value.

I don't understand what this means.

> Although the arguments associated with types can be erased in certain cases
  and they are not real part of the constructor arguments

Isn't this solved by #1225?

I don't understand the point the `Association between data and type constructors` section
is trying to make. It briefly talks about typing constructors (why?) before jumping to
the implementation of pattern-matching in a language without a pattern-matching construct.

> BelieveMe

My understanding of `BelieveMe` is that it really is meant to be implemented as the identity
on the underlying data. The fact you seem to be using it for other purposes (e.g. crossing FFI
boundaries & changing representations on the fly) suggests to me we should have a separate
primitive for that.

Similarly the discussion of type mismatches on the FFI looks to me like another wish for a
"backend do your magic" function. I would expect a typed language like Idris to simply fail
with a type error if the primitive bound by the user does not have the right precision for
the builtin type in question (e.g. float instead of double).


I've removed the bit about code responsible for the lives of human being and the comment about "mission critical" code as Idris2 is still a research compiler and it would be irresponsible to use it in production with such high stakes IMO.

andorp:
> That was a long read!

First of all: Thank You for all the work/reading/fixing. :)

> > Strings created in Idris will always have value.
> 
> I don't understand what this means.

In C, Java null pointer can be a value of the String type. We need to be careful about this egde-case.

> Isn't this solved by #1225?

Oh, than this is not relevant anymore.

> I don't understand the point the `Association between data and type constructors` section
> is trying to make. It briefly talks about typing constructors (why?)

This one comes from the particular constraints around how STG handles the datatypes. I started to feel that part is not relevant for any other back-end. Feel free to remove it.

> The fact you seem to be using it for other purposes (e.g. crossing FFI
> boundaries & changing representations on the fly) suggests to me we should have a separate
> primitive for that.

Yes. I am implementing the Idris primitive data types with the same representation as in GHC. I hope I can keep this bridge this simple between Idris and GHC runime.

> does not have the right precision for
> the builtin type in question (e.g. float instead of double).

I agree. I will have to do some magic conversion in these cases. I wonder how this is handled handled in other backends.
Definitely I will have to write conversion code for String handling in FFI.

> I've removed the bit about code responsible for the lives of human being and the comment about "mission critical" code as Idris2 is still a research compiler

Totally valid point.

stefan-hoeck:
Thanks for the great and thorough documentation! Just a minor thing: The acronym "IR" took me a moment to figure out, so it would be great if you could introduce it in unabbreviated form at the beginning of the document.

docs/source/backends/backend-cookbook.rst
```diff
@@ -0,0 +1,802 @@
...
+- ``NamedDef``
+- ``LiftedDef``
+- ``ANFDef``
+- ``VMDef``
+
+The question to answer here is: Which one should be picked?
+Which one fits to the custom back-end?
+
+How to represent primitive values defined by the ``Core.TT.Constant`` type?
+-----------------------------------------------------------------------
```
andorp, line 137:
```suggestion
---------------------------------------------------------------------------
```

docs/source/backends/backend-cookbook.rst
```diff
@@ -0,0 +1,802 @@
...
+The implementor should also consider how to transform the simple main expression
+which is represented in CExp.
+As Idris does not focus on memory management and threading. The custom back-end
+should model these concepts for the program that is compiled.
+One possible approach is to target a fairly high level language and reuse as
+much as possible from it for the custom back-end.
+Another possibility is to implement a runtime that is capable of handling memory
+management and threading.
+
+Which IR should be consumed by the custom back-end?
```
andorp, line 119:
```suggestion
Which Intermediate Representation (IR) should be consumed by the custom back-end?
```

docs/source/backends/backend-cookbook.rst
```diff
@@ -0,0 +1,802 @@
...
+---------------------------------------------------
```
andorp, line 120:
```suggestion
---------------------------------------------------------------------------------
```

docs/source/backends/backend-cookbook.rst
```diff
@@ -0,0 +1,802 @@
...
+- Comparison operations (``LT``, ``LTE``, ``EQ``, ``GTE``, ``GT``)
+- String operations
+  (``Length``, ``Head``, ``Tail``, ``Index``, ``Cons``, ``Append``,
+  ``Reverse``, ``Substr``)
+- Double precision floating point operations
+  (``Exp``, ``Log``, ``Sin``, ``Cos``, ``Tan``, ``ASin``, ``ACos``, ``ATan``,
+  ``Sqrt``, ``Floor``, ``Ceiling``)
+- Casting of numeric and string values
+- An unsafe  cast operation ``BelieveMe``
+- A ``Crash``operation taking a type and a string and creating a value at that
```
andorp, line 409:
```suggestion
- A ``Crash`` operation taking a type and a string and creating a value at that
```

docs/source/backends/backend-cookbook.rst
```diff
@@ -0,0 +1,802 @@
...
+on a data constructor, ``CASE`` picks a branch based on the constructor itself.
+An extra operation called ``PROJECT`` is introduced to explicitly extract a
+constructor's argument based on their position.
+
+There are no ``App`` or ``UnderApp``. Both are replaced by ``APPLY`` which
+applies only one value and creates a closure from the application. For erased
+values the operation ``NULL`` assigns an empty/null value for the register.
+
+How to implement the Foreign Function Interface?
+--------------------------------------------
```
andorp, line 600:
```suggestion
------------------------------------------------
```

docs/source/backends/backend-cookbook.rst
```diff
@@ -0,0 +1,802 @@
...
+  Any ``PrimIO`` defined computation will have this extra layer.
+  Pure functions shouldn't have any observable IO effect on the program state
+  in the host technology implemented runtime.
+  NOTE: ``IORes`` is also used when callback functions are registered in the
+  host technology.
+- ``CFWorld``
+  Represents the current state of the world. This should refer to a token that
+  is passed around between function calls.
+  The implementation of the World value should contain back-end
+  specific values information about the state of the Idris runtime.
```
andorp, line 647:
```suggestion
  specific information about the state of the Idris runtime.
```

docs/source/backends/backend-cookbook.rst
andorp: Changed in c78e5065fcfabcd334d0c73c6ff8e0c1d7ebff20

gallais:
This is a great contribution. Thanks!

andorp:
@gallais Thanks for merging this.
@zoli90 Thanks for the first proof-read.

PR #1240, opened 2021-03-27
[ re #1162 ] Test without install

mattpolzin:
This PR revisits a change made when the test package was [introduced](https://github.com/idris-lang/Idris2/pull/1162). When that package was added, I made installation of Idris & its included libraries a prerequisite for building & running tests. Doing this had the following three characteristics:

1. CI integrity was maintained because all CI workflows install prior to running `make test` anyway.
2. You could not accidentally run local tests requiring library updates (or additional libraries) against the wrong versions of those libraries. This ostensibly removed a vector for confusing test failures (false negatives).
3. You could not test changes to any Idris code or library code without first installing the changes which meant you could install a broken compiler in the process of testing whether the compiler was working!

The 3rd point is a strong negative, as noted [here](https://github.com/idris-lang/Idris2/pull/1162#issuecomment-808357936) -- the ability to accurately test library changes was traded for the ability to test compiler changes prior to installing them.

With that in mind, this PR aims to make it so that tests are run against both the latest build of the compiler and the latest build of the libraries packaged with it without installing either on the system.

### Compiler
The test execution processes (the `run` scripts for each test) already used the new build of the compiler. A change was needed so that the test build process (building the test binary) would use the new build of the compiler in all situations instead of the installed build of the compiler.

### Libraries
Libraries were a bit more complicated. Originally I attempted to get the test build & test execution processes to look for the newly built packages instead of the installed versions (via `IDRIS2_PACKAGE_PATH`). This did not work because Idris appears to prefer (or at least seek out first-) the installed libraries even when the newly built libraries are in the `IDRIS2_PACKAGE_PATH`. I could have continued down this path by exploring how to get Idris to look into the specified package path prior to the install path, but instead I came to the following simpler solution.

Idris will use the `IDRIS2_PATH` to locate TTC files with higher precedence than it looks for modules inside of  packages. This meant a change that included the build directories of all newly built packages in the environment for test execution (it was already in the environment for building the test binary) would pre-empt the need to change the way Idris seeks out packages.

This was the lowest overhead solve I came up with. I think there are other reasons why it might be beneficial to be able to force Idris to look to non-standard locations for packages before looking to installed locations, but that feels like a discrete and separable project as long as the approach taken in this PR does not have harmful side effects to folks' workflows.


gallais:
Thanks for looking into this! When I run `make test only=record001` (as an example, the fact
it's this test in particular does not matter), I get the following error:

```idris
make test only=record001
make[1]: Entering directory '/home/gxa1/projects/idris2/tests'
../build/exec/idris2 --build tests.ipkg
Uncaught error: INTERNAL ERROR: Can't find data file chez/support.ss in any of ["/home/gxa1/.idris2/idris2-0.3.0/support/chez/support.ss"]
```

I had nuked my `~/.idris2` directory prior to building the branch


I have tried taking a cue from the `IDRIS2_BOOT_TEST` part and adding:

```
# Library and data paths for test
IDRIS2_TEST_LIBS := ${IDRIS2_CURDIR}/lib
IDRIS2_TEST_DATA := ${IDRIS2_CURDIR}/support
```

to the `Makefile` and appending `IDRIS2_DATA=${IDRIS2_TEST_DATA} IDRIS2_LIBS=${IDRIS2_TEST_LIBS}` to
the `${MAKE} -C tests` call in the `test:` target to no avail.

gallais:
Ah! That's because I need to pass the same things in the `testbin` target too!

mattpolzin:
Oh! Interesting. So I had been accidentally still relying on some installed stuff even though I had tests passing. 

Good catch!

gallais:
I think that my current patch invalidates the `bootstrap-test` target
as I'm overriding variables it sets. I'm not sure how to fix that.

mattpolzin:
I did some more testing and made a few more tweaks.

I made sure the bootstrap and regular builds did not need to share any files by copying the support lib to `./lib` in the repo folder as part of the build phase (and copying from there on install) so that `IDRIS2_TEST_LIBS` could be changed from `/bootstrap/lib` to `/lib`.

> I think that my current patch invalidates the `bootstrap-test` [...] I'm overriding variables it sets

I fixed that by allowing the new variables you added to take on their current values if already set and passing them by name under `bootstrap-test`.

Then, I ran a few manual tests of the changes:
1. delete all contents of `.idris2` (install direcctory)
2. make clean && make distclean
3. make bootstrap
4. make boostrap-test
  --> Fails; can't find contrib, base, network packages.
5. make test
  --> Fails; can't find contrib, base, network packages.
6. make install
7. make test
8. make bootstrap-test
-- At this point, TTC version is out of date because only a bootstrap build has been performed so I do a clean build after making changes in the next step.
9. <edit compiler or library code> && make clean && make && make test
  --> Succeeds; Shows modification is part of testing but not yet installed.

So, my summary is:
Prior to this PR, a valid bootstrap or build test run required at least one initial install of Idris2 (otherwise packages are not found when running tests) and reinstall of Idris2 if libraries or TTC versions changed.

After this PR (in its current state) a bootstrap or build test run will require an initial install (at some point in the past) of Idris2 so that building of individual tests does not result in complaints about packages not being located, _but_ no reinstall is needed ever again to get valid test runs on library or compiler updates.

Personally, I think this PR now represents a solid improvement -- no need to wait for install before running tests (the state of the repo before this PR) and yet we retain the ability to test changes to the compiler and libraries. There is definitely room for improvement, but I vote this gets merged as-is first.

It's a bummer that the initial install is required because as best I can tell it is only required because Idris is looking for packages it needs even though it will find all of the TTC files it needs local to the build directory. I experimented a bit with fixing that as part of this PR initially but was unhappy with my approach of copying library files around (faking what the "install" step for each of prelude, base, contrib, etc. does) as a pretense for running tests because the files are all already in the build directory, just not in _exactly_ the same place or structure as Idris 2 would want them when it searches for packages.

gallais:
Your latest changes make a lot of sense once I see them written down!
I don't think I could have come up with them myself.

I agree that this is worth shipping as is. Thanks a lot for the PR!

PR #1243, opened 2021-03-28
Tune precedence of (===), (~=~) and (<+>)

Russoul:
Lower the precedence of `(===)` and `(~=~)` to match that of `==` (6). 
With the tuned precedence, arithmetic operators `(+)`, `(*)`, etc. behave as one would expect when combined with propositional equality. Examples of impacted expressions:
- `2 + 2 === 4` now parses as `(2 + 2) === 4`
- `2 :: Nil === [2]` now parses as `(2 :: Nil) === [2]`
- `[1, 2] ++ [3, 4] === [1, 2, 3, 4]` now parses as `([1, 2] ++ [3, 4]) === [1, 2, 3, 4]`

Also `(<+>)` is made to bind as tight as `(+)` (precedence 8).

None of the builtin operators seem to be negatively impacted by the change.

### DIFF
--- a/libs/prelude/Builtin.idr
+++ b/libs/prelude/Builtin.idr
@@ -114,7 +114,7 @@ data Equal : forall a, b . a -> b -> Type where
 
 %name Equal prf
 
-infix 9 ===, ~=~
+infix 6 ===, ~=~
 
 -- An equality type for when you want to assert that each side of the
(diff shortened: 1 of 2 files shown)


gallais:
Thanks! The precedence issue with `(===)` has been annoying me for a while
(just not enough to actually look into it!).

PR #1246, opened 2021-03-30
[ doc ] Fixed indentation of Lazy

gemmaro:
Fixed indentation in the conceptual definition of `Lazy` (like [`Data.Fin`](https://idris2.readthedocs.io/en/latest/tutorial/typesfuns.html#the-finite-sets)).


gallais:
Is this trying to suggest that `Force` is a constructor of `Lazy`?
Because that's not the case.

gemmaro:
I'm sorry. I was mistaken.

PR #1250, opened 2021-03-31
Ignore some REPL options in package commands

claymager:
Currently, `idris2 -p contrib --build mypkg.ipkg` enters the REPL, while ` --build mypkg.ipkg -p contrib` fails with an error. This PR makes both instances fail.

I also priveledged some REPL options to silently pass through package commands like `--build`, so users can alias idris2 with `--no-banner --no-colour`.

### DIFF
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -664,7 +664,11 @@ filterPackageOpts acc (SetCG f       ::xs) = filterPackageOpts (record {oopts $=
 filterPackageOpts acc (BuildDir f    ::xs) = filterPackageOpts (record {oopts $= (BuildDir f::)}     acc) xs
 filterPackageOpts acc (OutputDir f   ::xs) = filterPackageOpts (record {oopts $= (OutputDir f::)}    acc) xs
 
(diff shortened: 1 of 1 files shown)


gallais:
Looking at just this change I genuinely don't understand how:

* `idris2 -p contrib --build mypkg.ipkg` opens a REPL
* the proposed change turns that into an error

Do you have an insight you can share?

claymager:
`filterPackageOpts` processes all options in sequence. When it hit `-p contrib`, the recursive loop broke, returning either `MkPFR (Just _) _ True` (if it had already seen `--build mypkg.ipkg`), or `MkPFR Nothing _ True` (if it had not).

`processPackageOpts` only throws on `MkPFR (Just _) _ True`. With `Nothing`, Idris.Driver.stMain continues and opens the REPL.

gallais:
Ah I see it now. The `pkgDetails = ` part had escaped me. Thanks!

Btw: isn't it a bit strange that this `pkDetails` update overwrites
the content if it's already `Just (cmd, f)`? Wouldn't it be better
for us to collect *all* of the commands rather than only keeping the
last one we've seen?

PR #1253, opened 2021-03-31
Add Agda-like builtins

Z-snails:
Is there an easy way to check some property of the type without having to manually traverse the type each time?

Also is this reasonably implemented or should I use a constructor that already exists in `ImpDecl` etc.?


cypheon:
For someone not familiar with Agda (like myself), could you maybe add a small example, how this would be used?

From Agda's docs i reckon it is a way to specify a (potentially more efficient) RTS representation for a type (i.e. a generalisation of the "natToInt hack"), is that correct?

If so, then this could also help to get rid of some assumptions/hardcoded facts in the backend support code (e.g. that List.Nil hast tag "0", While List.:: has tag 1), that might change, when the type is modified?

Z-snails:
Here's an example:
```idris
data Nat : Type where
    Z  : Nat
    S : Nat -> Nat

%builtin Natural Nat

data Fin : Nat -> Type where
    FZ : Fin (S k)
    FS : Fin k -> Fin (S k)

%builtin Natural Fin
```
In the codegen phase (specifically when generating `CExp`) and occurances of `Nat` will be replaced by a primitive Integer.
This on its own reduce the memory usage from linear to (almost) constant, but also some functions (`+`, `*`, `minus`, more to come later) can be altered to use primitive Integer ones (which are also much much faster).

Here is some more details about the conversion (which I will add to the documentation later):
`Z` -> `0`
`S` -> `prim__add_Integer 1`
`natToInteger x` -> `x`
`integerToNat` -> `prim__integerToNat`
`plus` -> `prim__add_Integer`
`mult` -> `prim__mul_Integer`
`minus` -> `prim__sub_Integer`
```
case x of
    Z => zexp
    S y => sexp
```
->
```
case x of
    0 => zexp
    _ => let y = prim__sub_Integer x 1 in sexp
```

This is possible for any data type that has exactly 1 constructor with an arity (number of non-erased arguments) of 0 and 1 constructor with an arity of 1.

Z-snails:
I've finished the type checking part of this proposal, but I don't know how to get that information in Compiler.CompileExpr. Is there existing code to provide this 'analysis' to CompileExpr or do I need to add that myself (if so can anyone suggest a good way (possibly extensible?) to do it)?

src/Core/Context.idr
```diff
@@ -645,6 +645,42 @@ data Transform : Type where
...
+
+-- TODO: After next release remove nat from here and use %builtin pragma instead
+initBuiltinTransforms : BuiltinTransforms
+initBuiltinTransforms =
+    let natTy = NS typesNS (UN "Nat")
+        natZ = NS typesNS (UN "Z")
+        natS = NS typesNS (UN "S")
+    in MkBuiltinTransforms
+        { natTyNames = insert natTy (natZ, natS) empty
+        , natZNames = insert natZ () empty
```
gallais, line 680: Probably worth adding `singleton` to `Libraries.Data.ANameMap` and use it here.

src/TTImp/ProcessBuiltin.idr
```diff
@@ -0,0 +1,163 @@
...
+getRetTy : {vars : _} -> Term vars -> (vars ** Term vars)
+getRetTy (Bind _ _ _ tm) = getRetTy tm
+getRetTy tm = (vars ** tm)
+
+||| Get the first non-erased argument type.
+getFirstNETy : {vars : _} -> Term vars -> (vars ** Term vars)
+getFirstNETy (Bind _ _ b tm) = if isErased $ multiplicity b
+    then getFirstNETy tm
+    else (_ ** binderType b)
+getFirstNETy tm = (vars ** tm)
```
gallais, line 29: Shouldn't this catchall be `Nothing` instead?

src/TTImp/ProcessBuiltin.idr
```diff
@@ -0,0 +1,163 @@
...
+import Core.Metadata
+import Core.TT
+import Core.UnifyState
+
+import TTImp.TTImp
+
+||| Get the return type.
+getRetTy : {vars : _} -> Term vars -> (vars ** Term vars)
+getRetTy (Bind _ _ _ tm) = getRetTy tm
+getRetTy tm = (vars ** tm)
```
gallais, line 22: I suspect you're going to want to `subst` in case the `Bind` is a `Let`.

src/TTImp/ProcessBuiltin.idr
```diff
@@ -0,0 +1,163 @@
...
+
+||| Get the first non-erased argument type.
+getFirstNETy : {vars : _} -> Term vars -> (vars ** Term vars)
+getFirstNETy (Bind _ _ b tm) = if isErased $ multiplicity b
```
gallais, line 26: Same comment here about the different types of binders: you only want to return
the binder if it's not erased *and* it's a `Pi`.

src/TTImp/ProcessBuiltin.idr
Z-snails: It doesn't return a `Maybe`, are you suggesting it should return `Maybe`?
Also when it comes to the type system I basically just learnt it for this PR so you know better than me.

gallais: Yes I think it should as if you cannot find a `NETy` it's better to explicitly
return `Nothing` than some potential garbage.

At the call site the check on non-erased args hopefully told you that there ought 
to be a value so if you get `Nothing` back you can raise an internal error. This
way if we have a broken invariant somewhere we'll get an error rather than silently
continuing with the wrong domain.

src/TTImp/ProcessBuiltin.idr
Z-snails: Could you explain the `Let` constructor of `Core.TT.Binder` please?

Z-snails: Actually I think I get it, is it
`Bind fc x (Let _ _ val ty) scope` represents
```
let x : ty
    x = val in
scope
```

Z-snails:
For now I've only added constructors and case statements for now (although Nat still uses faster functions) I plan to add functions like plus and mult later, and documentation soon.

docs/source/reference/builtins.rst
```diff
@@ -0,0 +1,71 @@
...
+    data Fin : Nat -> Type where
+        FZ : Fin (S k)
+        FS : Fin k -> Fin (S k)
+    
+    %builtin Natural Fin
+
+works because the ``k`` is always erased.
+
+This doesn't work if the argument to the ``S``-like constructor
+is lazy (sometime known as ``CoNat``) because these can be infinite.
```
gallais, line 50: This is not exactly true. Unlike Haskell we make a distinction between `Lazy` and `Inf`.
In Idris `Lazy` only means that the argument is lazy, not that the value is allowed to be
infinite.

docs/source/reference/builtins.rst
```diff
@@ -0,0 +1,71 @@
...
+
+During codegen and occurances of ``Nat`` will be converted to the faster ``Integer`` implementation.
```
gallais, line 52: typos:
and -> any
occurances -> occurrence (singular because of "any")

src/Compiler/CompileExpr.idr
```diff
@@ -165,63 +165,99 @@ magic ms e = go ms e where
...
-  = do guard (nm == "S" && ns == typesNS)
+-- get all transformation from %builtin pragmas
+builtinMagic : Ref Ctxt Defs => Core (forall vars. CExp vars -> CExp vars)
+builtinMagic = do
+    defs <- get Ctxt
+    let b = defs.builtinTransforms
+    let nats = concat $ uncurry builtinMagicNat <$> values (natTyNames b)
+    pure $ magic $ natHack ++ nats
+  where
+    builtinMagicNat : Name -> Name -> List Magic
```
gallais, line 190: Can we add a record type `NatBuiltin` that stores these names? I am uneasy about having
multiple arguments with the same type and would rather the function relied on field names
than the arguments being passed in the right order.

gallais: Similarly I'd rather have `NameMap SUC` and `NameMap ZERO` than `NameMap ()` where
`SUC` and `ZERO` are both `data X = MkX`.

gallais (commented):
Nice! And documented too!

I have suggested some improvements to make sure we don't confuse one
constructor for the other when passing them around.

Russoul:
Wow, awesome project!

I wanted to ask: does it collapse unary representations of `Nat`-like terms into one literal yet?
For example: `S (S (S k))` to become `prim__add_Integer 3 k`

Z-snails:
The tests are failing because of the test that checks the backend actually converts `Nats` to `Integers`, which I've implemented as dumping `CExp` but this is very prone to failing if the order changes. Oddly it passes on my PC, so I can't easily update the golden result. Perhaps we should add a new variety of tests for the codegen/performance checks?

@Russoul Thanks. It doesn't currently group up nested 'S' constructors, but that could certainly be done! I'm going to take a break from this project once I've sorted out the issues, so someone else can have a shot at it.

gallais:
> Oddly it passes on my PC, so I can't easily update the golden result.

I diff'd the expected vs. given produced by the CI and all of the changes
are internal numbers for metas & line numbers. I think this was caused by
changes to `prelude` that were in `master` but this PR was not taking into
account.

This could have been solved by merging & then running `make test` but
I adopted the nuclear option: simply drop all of the internal names for
metavariables and all mentions of line & columns. This will make the test
more robust wrt future changes to `prelude`.

alissa-tung:
I am wondering should datatypes have automatic literary representation when using `%builtin Natural` pragma like what Agda had done?
```agda
data ℕ : Set where
  zero : ℕ
  succ : ∀ (n : ℕ) → ℕ
{-# BUILTIN NATURAL ℕ #-}

prop : ℕ
prop = 42
```

PR #1255, opened 2021-04-01
[base] Data.Nat comparisons are proof-irrelevant

ohad:
Proves that the arithmetic comparison relations (`LTE`, `GTE`, `LT`,
`GT`) have unique proofs. This means that we don't need to export
proofs of them.

Also add some docstrings mentioning this.

### DIFF
--- a/libs/base/Data/Nat.idr
+++ b/libs/base/Data/Nat.idr
@@ -63,10 +63,31 @@ Uninhabited (NotBothZero 0 0) where
   uninhabited LeftIsNotZero impossible
   uninhabited RightIsNotZero impossible
 
+||| `m LTE n`: the proposition `m` is less than, or equal to, `n`.
+||| Since every such proposition has a unique proof, proofs returning
+||| `LTE` don't need to be public.
 public export
 data LTE  : (n, m : Nat) -> Type where
   LTEZero : LTE Z    right
   LTESucc : LTE left right -> LTE (S left) (S right)
+
+||| `m LTE n` is a proposition: every two proofs are equal.
+||| We ford the lower and upper bounds in the inequality when further
+||| algebraic manipulations are needed.
+export
+irrelevantLTE : {auto 0 ford_lower : n1 = n2}
+             -> {auto 0 ford_upper : m1 = m2}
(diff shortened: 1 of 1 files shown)


libs/base/Data/Nat.idr
```diff
@@ -63,11 +63,32 @@ Uninhabited (NotBothZero 0 0) where
...
 data LTE  : (n, m : Nat) -> Type where
   LTEZero : LTE Z    right
   LTESucc : LTE left right -> LTE (S left) (S right)
 
+||| `m LTE n` is a proposition: every two proofs are equal.
+||| We ford the lower and upper bounds in the inequality when further
+||| algebraic manipulations are needed.
+export
+irrelevantLTE : {auto 0 ford_lower : n1 = n2}
+             -> {auto 0 ford_upper : m1 = m2}
```
gallais, line 79: I'm not sure what the point here is. The only proof `auto` will find is `Refl̀`
so we may as well just have the equality be homogeneous.

ohad: Exactly. The point is that if the bounds are judgementally equal, the type-checker will automatically continue.
But if they're not, you have an opportunity to help it. See the proof for `irrelevantLT` on line 135.

gallais: But you only need a proof in `irrelevantLT` because of the level of
indirection you imposed onto yourself! Should we make all of our
function this indirect e.g.

```idris
map : {auto _ : m = n} -> (a -> b) -> Vect m a -> Vect n b
```

?

ohad: I think we shouldn't do it for map.
Fording is quite noisy, and should be avoided wherever possible. 
My intuition is that if you end up having to use `irrelevantLTE`, it probably means that the type-checker is really stuck on some intentional aspect. Heterogeneity gives you more slack to deal with this intentional mismatch, and we want the default `base` tool to help you out there.

ohad:
Another consideration is the name. 
We have `irrelevantEq : (0 _ a = b) -> a = b` in `base`, which is a different notion of irrelevance. (Runtime irrelevance, as opposed to proof irrelevance.)

PR #1256, opened 2021-04-01
Functor-like interface for transformers + universal named lifting instances

buzden:


gallais (commented):
* The name looks dodgy to me
* Is that just saying that `t` is a functor in a category other than `Type`? Should we think about that?

gallais:
No I mean something like this:

```idris
interface
  Category objC arrC =>
  Category objD arrD =>
  Functor (f : objC -> objD) where
    map : arrC a b -> arrD (f a) (f b)
```

buzden:
I amended and rebased the original commits with the ideas of a generalised functor. (inside the `CI` (Ubuntu / Sphinx workflow) happens an HTTP 500 error (internal server error) for unknown reason)

By the way, to look from a generalised point of view, `MonadTrans t` interface seems to be no more than a natural transformation between `m` and `t m`, right?

buzden:
> No I mean something like this:
> 
> ```idris
> interface
>   Category objC arrC =>
>   Category objD arrD =>
>   Functor (f : objC -> objD) where
>     map : arrC a b -> arrD (f a) (f b)
> ```

Oh, I see. Yeah, that looks nice, I'll try that. I'm not sure there will be no problems with type inference, but whatever.

buzden:
> ```idris
> interface
>   Category objC arrC =>
>   Category objD arrD =>
>   Functor (f : objC -> objD) where
>     map : arrC a b -> arrD (f a) (f b)
> ```

It seems to be much harder than I expected. Interfaces obviously may have implicit arguments. It's even exercised in `contrib`'s `Category` being defined as `interface Category (0 cat : obj -> obj -> Type) | cat where` having `obj` to be an implicit parameter.

When we are defining `data` in a GADT form, syntax allows us to explicitly define implicit parameters.

However, syntax for interfaces does not allow neither to have them after the interface name, nor to have implicit parameters before the whole declaration, i.e. nothing of these is allowed:
```idris
interface F {0 arrC : objC -> objC -> Type} {0 arrD : objD -> objD -> Type} (0 f : objC -> objD) where ...
interface {0 arrC : objC -> objC -> Type} -> {0 arrD : objD -> objD -> Type} -> F (0 f : objC -> objD) where ...
```

This seems to be odd, actually: why if implicit parameters can exist, we can't specify them precisely when we need?

Or, at least, maybe during interface elaboration all unspecified names that are in constraints should mean also an implicit parameter of the interface. I mean, in your example, `arrC` and `arrD` are implicit for the interface, and we can't say explicitly that there are these implicit parameters without changing the meaning of `objC` and `objD`, and at the same time, mentioning of `arrC` and `arrD` in constraints does not automatically add them to a list of interface parameters -- as it would do for functions. E.g., function `f : Functor m => Int` does have `m` as an implicit parameter despite the fact that it is present only in a constraint.

Russoul:
@buzden Can you write out your original method of the interface, please. (I just want to look back at its signature).
I bumped into a similar signature in a paper but can't recall your original one for comparison.

buzden:
> @buzden Can you write out your original method of the interface, please. (I just want to look back at its signature).
> I bumped into a similar signature in a paper but can't recall your original one for comparison.

Do you mean before the rebase with `GenFunctor`? It was

```idris
interface MonadFunctor t where
  liftMap : Monad m => (m a -> m b) -> t m a -> t m b
```

Russoul:
Exactly! The signature corresponds to a functor action on arrows.
The arrows `f : m a -> m b` are taken from a sub-category of `Type`, whose objects lie in the image of the endofunctor `m` (which doesn't have to be a monad in this interpretation).
Cf. https://www.cs.ox.ac.uk/people/nicolas.wu/papers/Scope.pdf page 8 for more details.

buzden:
> No I mean something like this:
> 
> ```idris
> interface
>   Category objC arrC =>
>   Category objD arrD =>
>   Functor (f : objC -> objD) where
>     map : arrC a b -> arrD (f a) (f b)
> ```

The maximum that I could reach is something like
```idris
-- using `Caterogy` from `contrib`'s `Control.Category`

interface      
  Category {obj=objC} arrC =>      
  Category {obj=objD} arrD =>      
  GenFunctor objC objD arrC arrD f | f where      
    map : {0 a, b : objC} -> arrC a b -> arrD (f a) (f b)
```

but it's fairly hard to use ([see experiments](https://github.com/buzden/idris-playground/blob/4e8155bc5744fc39b690b11d26ae7cfc5bbda014/GenerFunctor.idr) with this kind of definition).

So, maybe the approach from the PR being not so general still is a bit more usable.

zseri:
```idris
public export
interface GenFunctor n (0 m : Type -> Type) | m where
  gmap : (n a -> n b) -> m a -> m b

public export %inline
Functor m => GenFunctor Prelude.id m where
  gmap = map
```
Does that mean that it would be impossible to define both `Functor` and `GenFunctor`? (guess: yes)
Does that (`| m`, functional dependency) mean that it would be impossible to define two different `GenFunctor` implementations with different `n` arguments? (unsure, probably yes)
Does that thus mean that if we define `GenFunctor`, we can't define `Functor`? (if the previous two statements are correct, then yes)

How does the current approach of this PR compare to:
```idris
GenFunctor Prelude.id m => Functor m where
  map = gmap
```
This would probably have the disadvantage that every thing which wants to support the `GenFunctor` interface needs to implement it explicitly, e.g. needs to be adjusted to make it usable. but imo/afaik it would allow implementing `GenFunctor` for more interfaces without breaking the possibility of using it via `Functor`... It's a bit sad that there is afaik no support for interface equivalence (instead of only "constrained specializations").

edwinb:
This looks to have stalled. It looks interesting, but in general I think we should be careful of what we put in `base`. Rightly or wrongly, we (or maybe I) have decided that `base` is a place for things that are now stable. So perhaps an experimental version could be put into `contrib`, or, given that it's not really visible there, it could be an external library that's listed on https://github.com/idris-lang/Idris2/wiki/Libraries then if it proves useful we can consider adding it.  For now, I think it's best to close this, sorry.

(There are some things in `base` that don't qualify under those rules, most notably `App`, which I'm going to demote to `contrib` at some point.)

PR #1265, opened 2021-04-02
[ new ] Add Bifoldable and Bitraversable interfaces to Prelude

stefan-hoeck:
Since the Prelude already has `Bifunctor` it makes sense to include these too.


libs/base/Data/These.idr
```diff
@@ -40,6 +40,26 @@ Bifunctor These where
...
+%inline
+public export
+Bifoldable These where
+  bifoldr f _ acc (This a)   = f a acc
+  bifoldr _ g acc (That b)   = g b acc
+  bifoldr f g acc (Both a b) = f a (g b acc)
+
+  bifoldl f _ acc (This a)   = f acc a
+  bifoldl _ g acc (That b)   = g acc b
+  bifoldl f g acc (Both a b) = f (g acc b) a
```
gallais, line 52: Shouldn't this be

```suggestion
  bifoldl f g acc (Both a b) = g (f acc a) b
```

if we're going left-to-right? I expect that depending on whether you implement
`Bifoldable t => t a b -> List (Either a b)` using `bifoldr` or `bifoldl` you
should get one list or its reverse.

libs/prelude/Prelude/Interfaces.idr
```diff
@@ -372,6 +372,145 @@ namespace Foldable
...
+||| of elements (as opposed to `Foldable`, which has one variety of element).
+||| Common examples are `Either` and `Pair`.
+public export
+interface Bifoldable p where
+  bifoldr : (a -> acc -> acc) -> (b -> acc -> acc) -> acc -> p a b -> acc
+
+  bifoldl : (acc -> a -> acc) -> (acc -> b -> acc) -> acc -> p a b -> acc
+  bifoldl f g z t = bifoldr (flip (.) . flip f) (flip (.) . flip g) id t z
+
+  binull : p a a -> Lazy Bool
```
gallais, line 385: Why are we suddenly taking `p a a` instead of `p a b`?

libs/prelude/Prelude/Interfaces.idr
```diff
@@ -372,6 +372,145 @@ namespace Foldable
...
+biall f g = bifoldl (\x,y => x && f y) (\x,y => x && g y) True
+
+||| Add together all the elements of a structure.
+public export
+bisum : (Bifoldable p, Num a) => p a a -> a
+bisum = bifoldr (+) (+) 0
+
+||| Add together all the elements of a structure.
+||| Same as `bisum` but tail recursive.
+export
```
gallais, line 449: I'm not convinced all of these should go in the prelude.
Remember that the criterion is: needed for almost any program.
This should probably go into `Data.Bifoldable` in `base`.

libs/prelude/Prelude/Interfaces.idr
stefan-hoeck: Thanks for catching this. Fixed.

libs/prelude/Prelude/Interfaces.idr
stefan-hoeck: Moved most utility functions to Data.Bifoldable in base.

PR #1266, opened 2021-04-02
[ RFC, WIP, don't-merge ] Sugar for dependent pairs with multiplicities

Russoul:
Adds 9 operators to complement `**`:
`X|Y`, where `X` and `Y` are one of these symbols: `0`, `1`, `*`.
Thus the augmented syntax of dependent pairs is:
`(t [: t] X|Y t [: t] X|Y ... t [: t] X|Y t)`
Other notable changes are:
- `Data.DPair.Subset` and `Data.DPair.Exists` are moved to `Builtin.DPair`
- The implicit argument of `Exists` is made explicit (to match `DPair` and `Subset`).
- Five extra plainly named (`DPairXY`) dependent pair data types are added to `Builtin.DPair`
- Better pretty-printing of dependent pairs

Syntax desugars correspondingly to one of the nine dependent pair instances in `Builtin.DPair`

**EDIT:**
I think I have a far better idea now... I'll write back.


Russoul:
During a conversation on slack with gallais, ohad and ziman, we decided it would be getter to try implement
an anonymous record structure, where we could also specify multiplicities.

So I think this PR can be considered to be superseded by the hypothetical implementation of the above.

PR #1268, opened 2021-04-03
Use file extensions when looking for exes on windows

melted:
Move pathLookup to Compiler.Common and use it on Node.


melted:
Yes, that looks cleaner.

PR #1269, opened 2021-04-03
Replace prim__getArgs with prim__getArgsCount and prim__getArg

melted:
This spares us from having to build Idris data structures in the foreign
code.

### DIFF
--- a/support/racket/support.rkt
+++ b/support/racket/support.rkt
@@ -415,7 +415,7 @@
 (define (blodwen-arg n)
   (cond
     ((= n 0) (path->string (find-system-path 'run-file)))
-     (< n (+ (vector-length (current-command-line-arguments) 1))
+    ((< n (+ (vector-length (current-command-line-arguments)) 1))
         (vector-ref (current-command-line-arguments) (- n 1)))
      (else "")))
 
--- a/support/racket/support.rkt
+++ b/support/racket/support.rkt
@@ -409,13 +409,15 @@
 (define (blodwen-clock-second time) (time-second time))
 (define (blodwen-clock-nanosecond time) (time-nanosecond time))
 
-
 (define (blodwen-arg-count)
-  (vector-length (current-command-line-arguments)))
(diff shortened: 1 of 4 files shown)


melted:
I don't understand the CI failure at all.

gallais:
The problem happens earlier already cf. build-bootstrap-racket:
https://github.com/idris-lang/Idris2/pull/1269/checks?check_run_id=2288909053#step:4:753

I'm not sure why though.

support/racket/support.rkt
```diff
@@ -409,14 +409,13 @@
...
 (define (blodwen-clock-nanosecond time) (time-nanosecond time))
 
-(define (blodwen-args)
-  (define (blodwen-build-args args)
-    (if (null? args)
-        (vector 0) ; Prelude.List
-        (vector 1 (car args) (blodwen-build-args (cdr args)))))
-    (blodwen-build-args
-      (cons (path->string (find-system-path 'run-file))
-            (vector->list (current-command-line-arguments)))))
```
gallais, line 419: By using `current-command-line-arguments` you accidentally threw away
`path->string (find-system-path 'run-file)` as the first argument.

I expect that somewhere Idris assumes it gets the name of the executable
first & then the args and that breaks that assumption?

You should probably have something like:

```racket
(define (blodwen-arg-count)
  (+ 1 (vector-length (current-command-line-arguments))))

(define (blodwen-arg n)
  (cond 
     [(= 0 n) (path->string (find-system-path 'run-file))]
     [(< (- n 1) (vector-length (current-command-line-arguments)))
         (vector-ref (current-command-line-arguments) (- n 1))]
     [else ""])
```

gallais (changes requested):
May have found the bug

support/racket/support.rkt
melted: Ok!

melted:
Ooops.

melted:
Parentheses are hard. This should work.

PR #1280, opened 2021-04-06
Support Multi-declarations

ohad:
Allow declaring multiple identifiers with the same visibility and type-signature in the same claim statement:

```idris
data Foo : Type where
  |||An A
  A ,
  ||| Just a B
  B : Foo


public export
foo1, foo2 : Foo
foo1 = A
foo2 = LocalB
  where
    LocalB, LocalA : Foo
    LocalB = B
    LocalA = let NestedLocalA, NestedLocalB : Foo
                 NestedLocalA = A
                 NestedLocalB = B
             in NestedLocalA
```


ohad:
Is it OK for this kind of desugaring to take place in the parser?

ohad:
Also, for review: are the new `WithBounds` OK?

gallais:
> Is it OK for this kind of desugaring to take place in the parser?

I'd say because it's the same kind of desugaring we already have in telescopes:
`(m, n, p : Nat) -> ...` becomes `(m : Nat) -> (n : Nat) -> (p : Nat) -> ...`.

src/Idris/Parser.idr
```diff
@@ -963,8 +965,8 @@ dataBody fname mincol start n indents ty
                                                dopts <- sepBy1 (symbol ",") dataOpt
                                                symbol "]"
                                                pure $ forget dopts)
-                         cs <- blockAfter mincol (tyDecl (mustWork dataConstructorName) "" fname)
-                         pure (opts, cs))
+                         cs <- blockAfter mincol (tyDecls (mustWork dataConstructorName) "" fname)
+                         pure (opts, concat $ map forget cs))
```
gallais, line 969:
```suggestion
                         pure (opts, concatMap forget cs))
```

gallais (approved):

PR #1283, opened 2021-04-07
Fix a linearity leak in network; expose error codes to the user

Russoul:
1) Fix a leak in `Control.Linear.Network.accept`, where it returns a `Res` with an unrestricted pair as its second component.
   This bug is probably a consequence of removing multiplicity subtyping & tweaking default data type multiplicities to unrestricted throughout the code base.
2) Expose the socket errors to the user as they _are_ helpful when one of the operations on sockets fails.

### DIFF
--- a/libs/contrib/Control/Linear/LIO.idr
+++ b/libs/contrib/Control/Linear/LIO.idr
@@ -49,6 +49,14 @@ data L : (io : Type -> Type) ->
             (1 _ : ContType io u_act u_k a b) ->
             L io {use=u_k} b
 
+public export
+L0 : (io : Type -> Type) -> (ty : Type) -> Type
+L0 io ty = L io {use = 0} ty
+
+public export
+L1 : (io : Type -> Type) -> (ty : Type) -> Type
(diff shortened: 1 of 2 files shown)


gallais:
Would it make sense to refactor the code to make the small state machine more
explicit by extracting:

```idris
data Action : SocketState -> Type where
  Listen : Action Bound
  Accept : Action Listening
  (...)

Next : Action state -> Maybe SocketError -> Type
```

And then use `x : (...) -> L io {use=1} (Res (Maybe SocketError) (Next X))` as each
action's type?

gallais:
Is there a single use case ever where we don't need to explicitly apply
`L` to `{use=...}`? At this point shouldn't we just accept that `L` should
take its use explicitly?

Russoul:
I haven't seen one... We can add an alias, like `L1`, and ship it together with `L`?

gallais:
Ah, I see, it needs to be implicit so that we may have the `default` value.
Hmmm. Then yeah shipping `L1` and `L0` as aliases is probably the easiest.

Issue #860, opened 2020-12-18
chez/chez014 test fails


kozross:
When I build (bootstrap from Chez), I get this:

```
chez/chez014: FAILURE
Expected:
Received: hello world!
Received: echo: hello world!
1/1: Building Echo (Echo.idr)
Main> Main> Bye for now!

Given:
[server] Failed to bind socket with error: 99
1/1: Building Echo (Echo.idr)
Main> Main> Bye for now!
```

gallais:
Are you [using arch](https://aur.archlinux.org/packages/idris2/) by any chance?
It looks like the libs are built out of order leading to this issue.

kozross:
Yes, I am, but this doesn't just affect the AUR package. I got this same error when I cloned from Github and followed the instructions.

kozross:
I've provided what additional info I thought necessary to move this along. Is there more I can provide so that this could be resolved, @gallais ?

gallais:
Ah, I hadn't noticed you were the one who had posted the upstream issue. :sweat_smile: 
I know next to nothing about the network code so it's probably better if
someone more knowledgeable steps in.

kozross:
I can confirm this issue is still present on 0.3.0.

ShinKage:
I don't know much about the network code, but the error is raised when the test code tries to bind a new socket on localhost:0 (which should mean a free port on localhost). Can you check if it is indeed possible to do so in your system?

kozross:
@ShinKage Not a network expert - how would I check this?

ShinKage:
@kozross Sorry, I must have involuntarily skipped the notification. I'm very far from a network expert myself but you can try to directly call the C primitives, simplifying all the checks done in the support library, in particular you can take a look at the two functions called by the test [here](https://github.com/idris-lang/Idris2/blob/master/support/c/idris_net.c#L76) and [here](https://github.com/idris-lang/Idris2/blob/master/support/c/idris_net.c#L129). They will probably fail again, but maybe you can get more helpful error descriptions since the Idris2 support code does not propagate the exact error number.

kozross:
99 is `EADDRNOTAVAIL`. What port is the server trying to start on in the test code?

gallais:
[Looks like `0`](https://github.com/idris-lang/Idris2/blob/master/tests/chez/chez014/Echo.idr#L13)

kozross:
Binding to port 0 is not allowed, I'm pretty sure.

[...]

PR #1284, opened 2021-04-07
Fix issues with use of unix sockets

RaoulHC:
This change adds logic to set up `sockaddr` correctly for connect and
bind, handles the `AF_UNIX` case for `getSockAddr` and adds a test for unix
sockets.

Mingw-w64 seems to be missing the `afunix.h` header even though windows does support unix sockets, so I've added the structure manually but I might be missing something. It looks like CI only builds windows but doesn't test it, should it be? Can add if so.

### DIFF
--- a/libs/network/Network/FFI.idr
+++ b/libs/network/Network/FFI.idr
@@ -40,6 +40,10 @@ prim__idrnet_sockaddr_family : (sockaddr : AnyPtr) -> PrimIO Int
 export
 prim__idrnet_sockaddr_ipv4 : (sockaddr : AnyPtr) -> PrimIO String
 
+%foreign "C:idrnet_sockaddr_unix,libidris2_support"
+export
+prim__idrnet_sockaddr_unix : (sockaddr : AnyPtr) -> PrimIO String
+
 %foreign "C:idrnet_sockaddr_ipv4_port,libidris2_support"
 export
 prim__idrnet_sockaddr_ipv4_port : (sockaddr : AnyPtr) -> PrimIO Int
--- a/libs/network/Network/Socket/Raw.idr
+++ b/libs/network/Network/Socket/Raw.idr
@@ -70,6 +70,7 @@ getSockAddr (SAPtr ptr) = do
 
       pure $ parseIPv4 ipv4_addr
     Just AF_INET6 => pure IPv6Addr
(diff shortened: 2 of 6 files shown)


gallais:
> It looks like CI only builds windows but doesn't test it, should it be? Can add if so.

It would be a good idea, I think.

Crossing my fingers this fixes #860.

gallais:
FYI this is the diff to add tests:
```diff
-        run: c:\msys64\usr\bin\bash -l -c "cd $env:PWD && make bootstrap && make install"
+        run: c:\msys64\usr\bin\bash -l -c "cd $env:PWD && make bootstrap && make install && make test INTERACTIVE=''"
```

I tried to push it myself but you did not grant maintainers the right to update your PR.

RaoulHC:
I had tried to make a windows self-host and test step yesterday but perhaps that was too bold attempt when I don't know actions or windows :sweat_smile:.

gallais:
Right now the issue is that the output is flushed after printing the prompt
following c725b11c891b9f9f1e25ead66d9fc75a91b69836 so you'll need to update
the golden file to change that (the easiest is probably to merge `master`
into this branch, rebuild, run `make test` again and accept the changes).

RaoulHC:
Looks like the unix socket test unfortunately doesn't work on Windows still. Will try and figure this out at least, but it does look like a number of tests are failing on Windows so it might not make sense to add tests for it to CI at the moment.

gallais:
>  but it does look like a number of tests are failing on Windows so
  it might not make sense to add tests for it to CI at the moment.
  
:+1

RaoulHC:
Eventually figured it out, it seems like errno in windows does not quite have the same semantics as in unix, and is overridden to 0 when we switch between scheme and c (while unix will never set the value to 0, stated [here](https://man7.org/linux/man-pages/man3/errno.3.html)). This resulted in network functions saying they were successful when they were not, and bind was failing as I had the socket located in `/tmp/` which doesn't exist on windows. I'll raise an issue for this and the windows ci.

You can see [here](https://github.com/RaoulHC/Idris2/runs/2318215478) that the unix socket test does pass now on windows.

gallais:
Issues #1293 and #1294 track the remaining windows-related issues.

PR #1287, opened 2021-04-09
Multiple build commands

claymager:
As noted in #1250, Idris can only handle one package command at a time. This PR allows multiple in sequence.


gallais:
Nice!

PR #1289, opened 2021-04-09
[ refactoring ] Some equivalent changes, mostly code styling nicened

buzden:
While digging into nuances of parser around interface parameters, I found that something in the code could be expressed more elegantly, which I suppose to do. No suggested changes are meant to change the behaviour.

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -53,6 +53,8 @@ plhs = MkParseOpts False False
 %hide Core.Core.(>>=)
 %hide Prelude.pure
 %hide Core.Core.pure
+%hide Prelude.(<*>)
+%hide Core.Core.(<*>)
 
 atom : FileName -> Rule PTerm
 atom fname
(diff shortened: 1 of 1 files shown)


src/Idris/Parser.idr
```diff
@@ -710,7 +711,7 @@ mutual
                            val <- expr pdef fname indents
                            pure (n, val))
            atEnd indents
-           (n, val) <- pure b.val
+           let (n, val) = b.val
```
andrevidela, line 714: I'm very excited about this change. I think there are other places where this pattern is used because the type inference used to give up on `let` inside do-notation. Have you found and tried to fix other instances of this pattern?

buzden: I tried and it didn't work in other places, still because of type inference

Russoul: Hm, interesting. Since when `let ... = b.val` started to work? 
I also once found myself completely lost in the parser code with that pesky problem being the cause of inference errors.

gallais: It could be that this one always worked but we got accustomed to program
defensively against this type of inference issue and just did not even try.
IME the boolean flag stating whether a parser is productive is always quite
hard to infer.

PR #1296, opened 2021-04-12
add tests to Windows CI

### LINKED ISSUE #1293: Add windows to CI
Many of the tests currently fail on windows (see [here](https://github.com/RaoulHC/Idris2/runs/2318215478)), these tests should be fixed and then windows tests can be added to the CI as it currently only bootstraps Idris.

# Steps to Reproduce
Run test suite on windows

# Expected Behavior
Test suite passes

# Observed Behavior
Test suite does not pass

melted:
Fixes #1293

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -45,4 +45,4 @@ jobs:
         run: |
           scheme --version
       - name: Bootstrap and install
-        run: c:\msys64\usr\bin\bash -l -c "cd $env:PWD && make bootstrap && make install"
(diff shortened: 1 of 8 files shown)


melted:
Looks good, brew failed on macOS, hard to blame this commit for that. Let's rerun that and see.

melted:
Looks like bintray returns 403. Brew is still depending on Bintray? I hope they are aware that bintray will shut down in a couple of weeks.

melted:
Ok, that's not something we can do anything about, more than hoping that Github and homebrew fixes stuff.

melted:
Merging, because that problem is not under our control.

PR #1298, opened 2021-04-13
Add an option to specify the Chez's '__collect_safe' calling convention

Russoul:
I followed [this wiki section](https://github.com/idris-lang/Idris2/wiki/Networking-primitives) by @abailly to circumvent the deadlocks you get when executing blocking foreign code in a multithreaded application.

The method used in this PR is essentially equivalent to what was proposed in the wiki: every foreign procedure generated out of `%foreign` FFI specifiers with the `C__collect_safe` header is passed to chez additionally tagged to have the `__collect_safe` calling convention. Internally, Chez wraps such foreign chunks of code in a pair of `Sdeactive_thread()/Sactive_thread()`.

This approach is more convenient than the manual insertion of this pair of calls into foreign code, because otherwise we'd need to depend on the Chez headers that contain those functions.

Unfortunately, the Ubuntu version the CI runs on ships a version of Chez which doesn't implement `__collect_safe` yet.
I tried to adjust the workflow file so it manually builds the latest master of Chez, but that seems to break bootstrapping: it suddenly starts to throw `unbound identifier` errors. I can't figure out why.

Anyway, I am leaving this patch here, because it seems to be the only (?) way to have a deadlock free app that has blocking FFI calls running on Chez RTS ?

#### TODO
- [ ] Fall back to default calling convention if the chez version is not recent enough


abailly:
FWIW, I just tried change on my application that uses networking code and now it works fine! I can have clients connected to the server and get responses. That's awesome :) Is there anything I can do to help? I would ♡ to see this reach the mainline.

Russoul:
@abailly That's great! Sure, I think gallais is ready to merge this as soon as we have a way to fall back to the regular calling convention (that is without `__collect_safe ` magic)
if the user's Chez version is not recent enough to support the calling convention.
So if you have time and will, you could help us with that :) I'd do that anyway, but later... So yeah, if you want it now, you are welcome to take over!

abailly:
I have been working on this code for about 2 years, so I can wait a bit more. But it would be a good segway into getting back working on Idris2 itself and I suppose quite a few people could be interested in having workable networking support. Can you point me in the right direction? Is this something that need to be checked for at code generation time? Or before that?

gallais:
Basically [this test](https://github.com/idris-lang/Idris2/pull/1298/files#diff-4a28367cf7a88811e6dda020a76467414d53518dd51f2a9ae91d57ba0d783d9eR226) needs to be changed to include a check that the version of chez being used
is recent enough that it supports this calling convention.

abailly:
I saw that part, but the test should be done at runtime, hence in Scheme isn't it?

gallais:
Right, I am willing to assume that the version of chez available at runtime
will be at least as good as the one available at compile time. We could make
this an environment variable rather than guesswork, mind you. This way the
makefile can default to guessing but if you want to compile with a specific
behaviour you can set the option.

abailly:
I have tried to add a more detailed error message but got confused with this error in CI : https://github.com/abailly/Idris2-SH/runs/2657025870?check_suite_focus=true#step:4:805
Obviously, the bootstrapped compiler does not know how to handle the new calling convention, but this will always be the case. How can I workaround that? Should I update the bootstrap compiler?

gallais:
If you look at the current PR you'll see a lot of 

```diff
- %foreign "C:idrnet_recv,libidris2_support"
+ %foreign "C__collect_safe:idrnet_recv,libidris2_support"
           "C:idrnet_recv,libidris2_support"
```

But after your merge some `foreign` functions end up with a single
`C__collect_safe` directive e.g.

https://github.com/abailly/Idris2-SH/commit/5b0949e52454c1d9efffb676cd262cf988ce708c#diff-35f6059ae0b6332f01415cc3bf99628c512dc55656f2962efe15c0655453f555L60

abailly:
Thanks, I thought this duplication was unintentional so I removed it.

abailly:
OK, now I have a green build, I should be able to start working and add that test :)

abailly:
Having a look at `Chez.idr`. IIUC, the idea would be to check the chez version, given we have a `findChez` function. Perhaps I could extract the version with
```
$ scheme --version
9.5.2
```
and then in the `cCall` function or above, check the version is at least `9.5` and either fallback to `C` calling convention or rather raise an error. Is this what you have in mind @gallais ? 

Then the other option, or possibly an added safety check, would be to `addDirective` to the Chez codegen and handle that directive to generate `__collect_safe` or not calling convention.

abailly:
An alternative would be to always generate the `__collect_safe` calling convention when it is explicitly required in a foreign definition, for chez, and leave it to either the runtime or the `compileToSO` function to catch the problem if the chez version is too low?

gallais:
We could have both:

* magically guess the version by using `chez --version`
* allow the user to override the magic with an explicit command line option (convenient if they're building on a machine with an old chez but deploying on a newer one for instance)

abailly:
I have completed the first part, namely to extract the version from `chez --version` but I see Windows and Nix failures in the CI: https://github.com/abailly/Idris2-SH/tree/fix-deadlocks I don't have a Windows machine around to test. I could test Nix tomorrow. Could these failures be related to the fact I am relying on symlinks created in the `Makefile` to import modules from Idris2 codebase?

abailly:
I am sorry, but I give up trying to make this work. I don't have the patience to troubleshoot issues with Nix or Windows.

Russoul:
No problem! I can only imagine how hard it is to figure out how to make it work on all platforms.

abailly:
This makes me sad because I was so happy you found a way to make basic server work in Idris2 on chez. But troubleshooting CI issues with nix and windows goes is not how I want to use the spare time I can devote to OSS beside work.

Russoul:
Yeah, I understand. But nothing really prevents you from enjoying the fix without any more effort. Just merge the patch locally :)

abailly:
Yeah I know. That's certainly what I will do :)

Russoul:
Continued in #1536. Thanks @fabianhjr!

fabianhjr:
Good news @Russoul and @abailly, your work got merged today! UwU

PR #1300, opened 2021-04-13
[ fix ] Create library dir if necessary

claymager:
I have been routinely clearing ~/.idris2, and Idris currently doesn't like installing libraries to a directory that doesn't exist.

This allows installation of libraries even if the user doesn't already have a `~/.idris2/idris2-${version}`.


claymager:
I have a test for this, but it's reliant on some other changes I haven't finalized yet.

claymager:
The linter is only complaining about `gen_expected.sh` in the new test, which is a direct copy  from `chez017` (and `node017`).

This is becoming a fairly common pattern (establishing a system-agnostic `pwd` to generate `expected`, which we add to a `.gitignore`). Should we add some preprocessing of `expected` to `Test.Golden`?

gallais:
Sorry it took so long

PR #1301, opened 2021-04-14
Shared C codegen

madman-bob:
Expose the RefC code for compiling CC object files and executables, and add the option to compile them as libraries.

This should result in no behavioural changes to the RefC backend. I'm exposing these functions to enable other backends to make use of them, such as Python.

I wasn't able to run `make install-api` on `master`, so this branch is instead based off the most recent tag (`v0.3.0`). From a quick glance, it appears the only merge issues are a matter of `import`s, but I'm not able verify if any particular solution is correct.

### DIFF
--- a/src/Compiler/RefC/CC.idr
+++ b/src/Compiler/RefC/CC.idr
@@ -1,6 +1,7 @@
 module Compiler.RefC.CC
 
 import Core.Context
+import Core.Context.Log
 import Core.Options
 
 import System
@@ -37,7 +38,7 @@ compileCObjectFile {asLibrary} sourceFile objectFile =
                        "-I" ++ fullprefix_dir dirs "refc " ++
                        "-I" ++ fullprefix_dir dirs "include"
 
-     coreLift $ putStrLn runccobj
+     log "compiler.refc.cc" 10 runccobj
      0 <- coreLift $ system runccobj
        | _ => pure Nothing
 
@@ -62,7 +63,7 @@ compileCFile {asShared} objectFile outFile =
                        "-L" ++ fullprefix_dir dirs "refc " ++
                        clibdirs (lib_dirs dirs)
 
-     coreLift $ putStrLn runcc
+     log "compiler.refc.cc" 10 runcc
      0 <- coreLift $ system runcc
        | _ => pure Nothing
 
--- a/src/Compiler/RefC/CC.idr
+++ b/src/Compiler/RefC/CC.idr
@@ -1,0 +1,65 @@
+module Compiler.RefC.CC
+
+import Core.Context
+import Core.Options
+
+import System
+
+import Idris.Version
+import Libraries.Utils.Path
+
+findCC : IO String
+findCC
+    = do Nothing <- getEnv "IDRIS2_CC"
+           | Just cc => pure cc
+         Nothing <- getEnv "CC"
+           | Just cc => pure cc
+         pure "cc"
+
+fullprefix_dir : Dirs -> String -> String
+fullprefix_dir dirs sub
+    = prefix_dir dirs </> "idris2-" ++ showVersion False version </> sub
+
+export
+compileCObjectFile : {auto c : Ref Ctxt Defs}
+                  -> (outn : String)
+                  -> (outobj : String)
+                  -> Core (Maybe String)
+compileCObjectFile outn outobj =
+  do cc <- coreLift findCC
+     dirs <- getDirs
+
+     let runccobj = cc ++ " -c " ++ outn ++ " -o " ++ outobj ++ " " ++
+                       "-I" ++ fullprefix_dir dirs "refc " ++
+                       "-I" ++ fullprefix_dir dirs "include"
+
+     coreLift $ putStrLn runccobj
+     0 <- coreLift $ system runccobj
+       | _ => pure Nothing
+
+     pure (Just outobj)
+
+export
+compileCFile : {auto c : Ref Ctxt Defs}
+            -> (outobj : String)
+            -> (outexec : String)
+            -> Core (Maybe String)
+compileCFile outobj outexec =
+  do cc <- coreLift findCC
+     dirs <- getDirs
+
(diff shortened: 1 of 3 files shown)


src/Compiler/RefC/CC.idr
```diff
@@ -0,0 +1,74 @@
...
+     dirs <- getDirs
+
+     let libraryFlag = if asLibrary then "-fpic " else ""
+
+     let runccobj = cc ++ " -c " ++ libraryFlag ++ sourceFile ++
+                       " -o " ++ objectFile ++ " " ++
+                       "-I" ++ fullprefix_dir dirs "refc " ++
+                       "-I" ++ fullprefix_dir dirs "include"
+
+     coreLift $ putStrLn runccobj
```
gallais, line 41: Should this `putStrLn` be a call to `log "compile.refc"` instead?
(You'll need to import `Core.Context.Log` to get access to it)

src/Compiler/RefC/CC.idr
```diff
@@ -0,0 +1,74 @@
...
+     let sharedFlag = if asShared then "-shared " else ""
+
+     let runcc = cc ++ " " ++ sharedFlag ++ objectFile ++
+                       " -o " ++ outFile ++ " " ++
+                       fullprefix_dir dirs "lib" </> "libidris2_support.a" ++ " " ++
+                       "-lidris2_refc " ++
+                       "-L" ++ fullprefix_dir dirs "refc " ++
+                       clibdirs (lib_dirs dirs)
+
+     coreLift $ putStrLn runcc
```
gallais, line 66: Same remark wrt `putStrLn` vs `log` here.

src/Compiler/RefC/RefC.idr
```diff
@@ -1044,27 +1031,18 @@ executeExpr c _ tm
...
+                   -> Core (Maybe String)
+generateCSourceFile defs outn =
+  do _ <- newRef ArgCounter 0
      _ <- newRef FunctionDefinitions []
      _ <- newRef TemporaryVariableTracker []
      _ <- newRef OutfileText DList.Nil
      _ <- newRef ExternalLibs []
      _ <- newRef IndentLevel 0
-     traverse_ (\(n, d) => createCFunctions n d) defs
+     _ <- traverse (\(n, d) => createCFunctions n d) defs
```
gallais, line 1045: What's the reason behind this change?

src/Compiler/RefC/RefC.idr
```diff
@@ -1073,34 +1051,29 @@ compileExpr ANF c _ outputDir tm outfile =
...
+     let outexec = outputDir </> outfile
 
-     coreLift $ putStrLn runccobj
-     0 <- coreLift $ system runccobj
-       | _ => pure Nothing
-     coreLift $ putStrLn runcc
-     0 <- coreLift $ system runcc
-       | _ => pure Nothing
-     pure (Just outexec)
+     _ <- coreLift $ mkdirAll outputDir
```
gallais, line 1069: Same question here about changing a call to `coreLift_` to `_ <- coreLift`.

src/Compiler/RefC/RefC.idr
```diff
@@ -1073,34 +1051,29 @@ compileExpr ANF c _ outputDir tm outfile =
      coreLift_ $ writeFile outn code
      coreLift_ $ putStrLn $ "Generated C file " ++ outn
```
gallais, line 1052: Same comment about `putStrLn` vs. `log`

src/Compiler/RefC/CC.idr
madman-bob: This line is unchanged from the original code, just moved. I'm happy to change it if that's the right thing to do, but I don't know why this line is as it is.

src/Compiler/RefC/RefC.idr
madman-bob: This change comes from merging `master`. You'd have to ask whoever made the original commit.

src/Compiler/RefC/RefC.idr
madman-bob: Same response.

src/Compiler/RefC/RefC.idr
ohad: `Core.Core.traverse` in this case is `Core (List Unit)`, and it seems that to sequence something it needs to return `Core Unit`.

src/Compiler/RefC/RefC.idr
ohad: Same response.

src/Compiler/RefC/RefC.idr
gallais: Right so I suspect that the conflict was resolved the wrong way.

src/Compiler/RefC/CC.idr
gallais: Right but as we are working on that part of the codebase, that's an opportunity
to clean it up a bit as we go. Debugging information should be communicated using
`log` rather than `putStrLn`. This way it can be turned off if we don't need the
info.

src/Compiler/RefC/RefC.idr
ohad: I see. Let me give it a go on a fresh branch.

ohad: No, manually inserting those changes on a fresh branch gives the same error. Let's chat.

src/Compiler/RefC/RefC.idr
```diff
@@ -1024,7 +1026,7 @@ footer = do
 export
 executeExpr : Ref Ctxt Defs -> (execDir : String) -> ClosedTerm -> Core ()
 executeExpr c _ tm
-    = do coreLift_ $ putStrLn "Execute expression not yet implemented for refc"
+    = do log "compiler.refc" 0 "Execute expression not yet implemented for refc"
```
gallais, line 1029: I think this one is a valid `putStrLn`: it's an error message displayed upon failure.

madman-bob: I'll take your word for it. Updated

PR #1303, opened 2021-04-14
Hermetic tests

claymager:
*As I understand the way Idris2 reads paths*, this solves #934.

All CI appears to run correctly and fail correctly, including the new (to us) `checkPhase` in Nix.

`make bootstrap-test` works as I expect, but I don't know how to make it fail without either a) deleting one of the packages I tell it to look for, or b) breaking a test case so `make test` also fails.

---

As a rough summary, currently, `make test` reads `~/.idris2/idris2-${VERSION}/` to see what packages are available, then imports packages preferentially through the local paths.

By setting IDRIS2_PREFIX to a dir containing symlinks to those files instead, we bypass the need to have packages installed. This allows us to run `make test` without having a `~/.idris2`, and lets us use nix for the full test suite*.

The `make bootstrap` already used a variant IDRIS2_PREFIX, so I co-opted that for `make bootstrap-test`.

---

\* Exept for `racket`, which still has a few issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ export IDRIS2_BOOT_PATH := "${IDRIS2_CURDIR}/libs/prelude/build/ttc${SEP}${IDRIS
 export SCHEME
 
 
-.PHONY: all idris2-exec ${TARGET} testbin support support-clean clean distclean FORCE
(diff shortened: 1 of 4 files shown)


gallais:
This looks great. Thanks!

Issue #1290, opened 2021-04-10
Paths in ipkg files are relative to $(pwd), not to the file


claymager:
Given `mypkg.ipkg`
```
package mypkg
version = 0.1.0

sourcedir = "src"
modules = MyLib
```
with the directory structure
```
./
  mypkg/
    mypkg.ipkg
    src/
      MyLib.idr
```

calling `idris2 --build mypkg/mypkg.ipkg` fails with error: `MyLib not fould`.

I'd expect Idris to know that I meant relative to the location of `mypkg.ipkg`.

PR #1304, opened 2021-04-15
Fixes for processPackage

claymager:
While fixing (#1290), I noticed another issue: because we store options in a mutable Ref and allow for multiple package commands per call to idris2, the first ipkg commands can influence those that come later.

I found a solution for this, but don't know the `Ref`s well enough to know which of `Ref Ctxt`, `Ref Syn`, and `Ref ROpts` to protect, so this PR currently saves them all.


gallais:
Sorry about the conflict introduced by #1299

About the patch itself: I wonder what the right behaviour is in case `op` fails
with an error? Should we catch it, reset the Ref, and rethrow the error like we
would e.g. close a file handle in a `withFile (...)`?

claymager:
For `withRef Ctxt`, yes, and that probably needs to be broken out into its own function.
(What is the state after `withRef Ctxt . setWorkingDir "foo"` runs?)

Might as well make `withSyn` and `withROpts` behave the same way.

src/Idris/Package.idr
```diff
@@ -536,30 +536,36 @@ processPackage opts (cmd, file)
...
-                                | errs => coreLift (exitWith (ExitFailure 1))
+          do Just (dir, filename) <- pure $ splitParent file
+               | Nothing => throw $ InternalError "Tried to split empty string"
+             True <- pure $ isSuffixOf ".ipkg" filename
+               | False => coreLift . putStrLn $ "Packages must have an '.ipkg' extension: " ++ show file ++ "."
+             setWorkingDir dir
+             Right (pname, fs) <- coreLift $ parseFile filename
+                                      (do desc <- parsePkgDesc filename
+                                          eoi
+                                          pure desc)
```
gallais, line 547:
```suggestion
             Right (pname, fs) <- coreLift $ parseFile filename (parsePkgDesc filename <* eoi)
```

src/Idris/Package.idr
```diff
@@ -536,30 +536,36 @@ processPackage opts (cmd, file)
...
-                          pure ()
-              Install => do [] <- build pkg opts
-                               | errs => coreLift (exitWith (ExitFailure 1))
-                            install pkg opts
-              Typecheck => do [] <- check pkg opts
-                                | errs => coreLift (exitWith (ExitFailure 1))
+          do Just (dir, filename) <- pure $ splitParent file
+               | Nothing => throw $ InternalError "Tried to split empty string"
+             True <- pure $ isSuffixOf ".ipkg" filename
+               | False => coreLift . putStrLn $ "Packages must have an '.ipkg' extension: " ++ show file ++ "."
```
gallais, line 542: We're not exiting with a failure anymore?

```suggestion
            let True = isSuffixOf ".ipkg" filename
                 | _ => do coreLift $ putStrLn ("Packages must have an '.ipkg' extension: " ++ show file ++ ".")
                           coreLift (exitWith (ExitFailure 1))
```

src/Idris/Package.idr
```diff
@@ -536,30 +536,36 @@ processPackage opts (cmd, file)
...
+             setWorkingDir dir
+             Right (pname, fs) <- coreLift $ parseFile filename
+                                      (do desc <- parsePkgDesc filename
+                                          eoi
+                                          pure desc)
+                | Left err => throw err
+             pkg <- addFields fs (initPkgDesc pname)
+             maybe (pure ()) setBuildDir (builddir pkg)
```
gallais, line 550: Why are we removing the more idiomatic call to `whenJust`?

```suggestion
            whenJust (builddir pkg) setBuildDir
```

src/Idris/Package.idr
```diff
@@ -536,30 +536,36 @@ processPackage opts (cmd, file)
...
+                                install pkg opts
+                  Typecheck => do
+                    [] <- check pkg opts
+                      | errs => coreLift (exitWith (ExitFailure 1))
+                    pure ()
+                  Clean => clean pkg opts
+                  REPL => do
+                    [] <- build pkg opts
+                       | errs => coreLift (exitWith (ExitFailure 1))
+                    runRepl (map snd $ mainmod pkg)
```
gallais, line 567: Given that in this case we open a repl, does it make sense to require that there is only
one package file passed as an argument?

src/Idris/Package.idr
```diff
@@ -545,7 +545,7 @@ processPackage opts (cmd, file)
              Right (pname, fs) <- coreLift $ parseFile filename (parsePkgDesc filename <* eoi)
                 | Left err => throw err
              pkg <- addFields fs (initPkgDesc pname)
-             maybe (pure ()) setBuildDir (builddir pkg)
+            whenJust (builddir pkg) setBuildDir
```
gallais, line 548: Oops looks like I messed up indentation here on the online editor. :x

src/Idris/Package.idr
claymager: Not really, but that gets complicated. `idris2 --build foo.ipkg bar.ipkg` also makes sense to me.

However, `idris2 --repl foo.ipkg --repl bar.ipkg` acts exactly the same as `idris2 --repl foo.ipkg`, which probably not the right behaviour.

Should certain options, like `--repl`, `--build`, `-p`, etc; take `n>=1` arguments?

src/Core/Context.idr
```diff
@@ -2143,6 +2143,14 @@ getWorkingDir
               | Nothing => throw (InternalError "Can't get current directory")
          pure d
 
+export
+withCtxt : {auto cc : Ref Ctxt Defs} -> Core a -> Core a
```
gallais, line 2147: I would expect a `withX` function to take an `(x : X)` and locally override the value
of a global `X` with this `x` à la 

```idris
withVerbosity : Int -> Core a -> Core a
```

to escalate the verbosity level of debug message; or provide a sanitized version of
this `x` à la

```idris
withFile : FilePath -> (Handle -> IO a) -> IO a
```

that makes sure to close the handle in case an exception is raised.

Should this be called `saveX` or something similar instead?

claymager: I would argue that it does sanitize the`x`.
```
{c : _}  ->
(op: {c : _} -> Core a) ->
Core a
```
Both happen to have the same type, and we don't force the `Core a` to depend on the Ref, but that's what I see it doing: providing a locally mutable resource to `op`.

gallais: ok

src/Core/Context.idr
```diff
@@ -2143,6 +2143,14 @@ getWorkingDir
               | Nothing => throw (InternalError "Can't get current directory")
          pure d
 
+export
+withCtxt : {auto c : Ref Ctxt Defs} -> Core a -> Core a
+withCtxt = wrapRef Ctxt resetCtxt
+  where
+    resetCtxt : Defs -> Core ()
```
Russoul, line 2150: Sorry that I am writing this only now: I think we should `branch` the context here _and_ keep the new `timings`.
If we don't branch the context, some changes, that might have been made during the callback, will *persist* after the reversal to the original version of `Defs`. That happens because definitions are stored a mutable array - its state will persist across `get/put` calls.
Here is the array I am talking about: https://github.com/idris-lang/Idris2/blob/d6583048d8682b28bb9d611ca44c6dac957cd6b8/src/Core/Context.idr#L330

Am I right about this, @gallais ?

Russoul: For reference, this is how Edwin handles backtracking:
https://github.com/idris-lang/Idris2/blob/71abc8e33b17adba61173f3ce1c7564a1d9ce4dc/src/TTImp/Elab/Check.idr#L465-L484

Issue #1277, opened 2021-04-06
Performance problems of totality checker when many type parameters and recursion


buzden:
I discovered that addition of a fifth type parameter to a datatype in my codebase leads to great slowdown of compilation. A code that compiled several seconds starts to compile more than several hours.

Commenting out three existing recursive calls leads to compiling withing several minutes; four commented out recursive calls leads to compiling for only several seconds.

# Steps to Reproduce

To show the problem, I tried to extract a simple example from my existing code.
It compiles for only two minutes.

```idris
module LoooongRecur

%default total

data TTy : Nat -> Nat -> Nat -> Nat -> Bool -> Type where

  TA : TTy p1 r1 i1 i2 False ->
       TTy i1 i2 i1 u2 False ->
       TTy i1 i2 p2 b1 b2    ->
       TTy p1 r1 p1 i2 False

  TB : TTy v1 r1 v2 r2 bt ->
       TTy v1 r1 v1 r2 (be && bt)

  TC : TTy p1 r1 v2 r2 False ->
       TTy v2 r2 v3 r3 b2    ->
       TTy p1 r1 v3 r3 b2

n : Nat -> Nat
n = (+ 2)

showInd : (indent : Nat) -> TTy p1 r1 v3 r3 br -> String
showInd i (TA x1 x2 x3) = if False
                            then (if False then show' i else show' i)
                            else showInd (n i) x1 ++ show' (n i)
                            where
                              show' : Nat -> String
                              show' i = showInd (n i) x3 ++ if False then "" else showInd (n i) x2
showInd i (TB x) = showInd (n i) x ++ if False then "" else showInd (n i) x
showInd i (TC x y) = if False then showInd i y else showInd i x ++ showInd i y
```

On my machine running
```
rm build/ -r && time idris2 --check LoooongRecur.idr
```
gives
```
1/1: Building LoooongRecur (LoooongRecur.idr)

real    2m3.630s
user    2m2.868s
sys     0m0.639s
```

Every thing is important: removing a type parameter or a single data constructor leads to instant compiling. Removing recursive calls, or branching, or a parameter of a data constructor leads to significant speed up.

# Expected Behavior

Fast compilation or at least independence of compilation time when somewhat small change in code.

# Observed Behavior

Very slow compilation times.

Being run with `--log 10` option, it shows that all the slow work happens in the termination checker.

jfdm:
I have not looked at this, but adding some more information, when running Idris2 with the `--timing` flag we see:

[...]

PR #1312, opened 2021-04-16
Report time taken to check totality.

jfdm:
This is a noisy update to the information reported by `--timing`. 

Recent work of mine has suggested bottlenecks with compilation are in the totality checker. The Issue tracker has seen related issues, for example see #1277. 

Although noisy, the output can be grepped away giving the structure in the logs.

### DIFF
--- a/src/TTImp/ProcessDecls.idr
+++ b/src/TTImp/ProcessDecls.idr
@@ -96,7 +96,7 @@ checkTotalityOK n
 
     checkTotality : FC -> Core (Maybe Error)
     checkTotality fc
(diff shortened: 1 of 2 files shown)


src/TTImp/ProcessDecls.idr
```diff
@@ -87,6 +87,7 @@ checkTotalityOK n
          case treq of
             PartialOK => pure Nothing
             CoveringOnly => checkCovering fc (isCovering tot)
+
```
gallais, line 90: I don't think this was intended.

gallais (approved):

PR #1314, opened 2021-04-17
Generalise error handling for arbitrary Refs; fix #1313

### LINKED ISSUE #1313: Namespaces are not closed properly when there are errors inside
# Steps to Reproduce

Consider some code with a split namespace (i.e. a namespace declared in several chunks) and a type error in one chunk:

```idris
namespace X

  export
  f : Int -> Int

namespace Y

  g : Int -> Int -> Int
  g x y = x + f g

namespace X

  f 5 = 6
  f x = x - 1
```

# Expected Behavior

Only an error inside the namespace `Y` in a function `g`.

# Observed Behavior

Not only the main error, but also an induced one appears:

```
Error: While processing right hand side of g. When unifying Int -> Int -> Int and Int.
Mismatch between: Int -> Int -> Int and Int.

.../SplitNamespace.idr:9:17--9:18
 5 |
 6 | namespace Y
 7 |
 8 |   g : Int -> Int -> Int
 9 |   g x y = x + f g
                     ^

Error: No type declaration for Main.Y.X.f.

.../SplitNamespace.idr:13:3--13:10
 09 |   g x y = x + f g
 10 |
 11 | namespace X
 12 |
 13 |   f 5 = 6
        ^^^^^^^
```

From this it's known that namespace `Y` didn't end when it contains an error.

# Discussion

Well, this could be seen only as an annoying additional induced errors, but there exists the case which is harder.

I didn't manage to reproduce it in a small example yet, but there are some cases when **only** induced error is shown (i.e., no original error!). And this is very misleading since it looks like compiler forgets its own rules of namespacing and indents.

Russoul:
Throughout the Idris 2 codebase there are mainly two instances of use of the try-catch mechanics. 
One can be found in parts of the unification, while the other is used in the `RawImp` elaboration pipeline.
Both groups of functions (error handlers) differ only in the `Ref` arguments they take, so that implies code duplication.
In order to fix #1313 I needed to handle yet another set of `Ref`s . To avoid any more code duplication and to eliminate the existing one this PR generalises the error handing functions to act on arbitrary list of mutable references. 
That also makes it possible to insert ad-hoc one-off try-catch[-finally] blocks for a set of mutable references, not yet covered by the default two groups, when appropriate. 

Shall I include a test case for #1313 ?


src/Core/Core.idr
```diff
@@ -715,3 +715,32 @@ condC : List (Core Bool, Core a) -> Core a -> Core a
 condC [] def = def
 condC ((x, y) :: xs) def
     = if !x then y else condC xs def
+
+public export
+data RefList : List (Type, Type) -> Type where
```
gallais, line 720: Is this `All (uncurry Ref) las`?

src/Core/Core.idr
```diff
@@ -715,3 +715,32 @@ condC : List (Core Bool, Core a) -> Core a -> Core a
...
+  Nil : RefList []
+  (::) : {l : _} -> Ref l a -> RefList ls -> RefList ((l, a) :: ls)
+
+0
+GetTy : Ref l a -> Type
+GetTy _ = a
+
+0
+RefListTy : RefList xs -> Type
```
gallais, line 729: Is this `All snd las`?

src/Core/Error.idr
```diff
@@ -0,0 +1,53 @@
+module Core.Error
+
+import Core.Core
+import Core.Context
+
+export
+tryError : {auto c : Ref Ctxt Defs} ->
+           {auto refs : RefList ls} ->
```
gallais, line 8: I'm not sure how this `auto` is supposed to be inferred.
Should it (or `ls`?) be an explicit argument instead?

gallais (commented):
Probably worth getting in sync with #1304 too as these 2 PRs target the same concerns AFAICT.

Russoul:
@gallais, please see my (late) comment on #1304 in case you missed it.

edwinb:
This looks like a neater approach than the current one. If there's a test case that demonstrates this fixes #1313 then I'll merge this.

However, since it's got a bit stale and older PRs have a tendency to disappear (which is mostly my fault, sorry) I'm still going to close it for now. Please could you reopen once the details are resolved, then I'll merge?

PR #1315, opened 2021-04-18
[ base ] `Not (a == b)` implies `not a == b`

prozacchiwawa:
… so it can be used in proof rewriting.  There may be methods of doing things like this that work better, are more terse etc, but I'm new and learning.  This certainly might not be the right way of doing what I want but I find myself wanting the pattern of using bool conditions and the proof of them in proofs downstream often (in both Yes and No cases), maybe also doing it wrong.


gallais (commented):
You can just make the whole definition be `invertContraBool_` and inline
`fEf` and `tEt` tbh. The name is not ideal but I'm not sure what else to
call this function.

I'm not sure it's worth introducing a new module just for that one function, surely it'd be at home in `Data.Bool`.

I don't think we need a test case for this basic function.

prozacchiwawa:
Ok I'll move things around.

gallais:
Thanks!

prozacchiwawa:
Thank you for taking the time to review the pr.

PR #1316, opened 2021-04-19
[ REPL ] Improving :doc (fixity, totality, colours)

alissa-tung:
I am not sure whether these modifications are suitable for work. Sorry for bothering.


unsafe-coe:
Screenshot_20210419_142610
fixed

src/Idris/DocString.idr
```diff
@@ -167,7 +191,18 @@ getDocsForName fc n
...
+                    Just infixStr => infixStr
+                    Nothing       => ""
+             getPreFixDocn <- getPreFixDoc n
+             let prefixes = case getPreFixDocn of
+                    Just prefixStr => prefixStr
+                    Nothing        => ""
+             let fixes = if infixes == "" && prefixes == ""
+                            then ""
+                            else "\nFixity Declarations:\n" ++ infixes ++ prefixes
+             pure (doc ++ extra ++ fixes)
```
gallais, line 205: Rather than adding additional newlines that are not needed if there is not extra / fixity
information, it would make more sense to use `unlines` here.

gallais (changes requested):
Can we add a test case rather than a screenshot of a test case?

Also it seems to me that the screenshot demonstrates a flaw: `:doc (-)`
only returns the binary function `Prelude.-` but the fixity information
attached corresponds to both `(-)` and the syntactic sugar for `negate`.
This should not be the case.

src/Idris/DocString.idr
unsafe-coe: I am afraid I do not know how to use `unlines` there, sorry…

unsafe-coe: So should we also print the prefix version docs for `(-)` or just remove the prefix information?

unsafe-coe: the trim NEWLINE was dealt with so former changes on tests was revert

to see what now it looks like at [add test](https://github.com/idris-lang/Idris2/compare/master...unsafe-coe:master#diff-359d888a70b3df58bad8867a5dfe6d7c65d5625b531e33085738c0ff1d0b73ad)

unsafe-coe:
The current PR implementation is verbose and unbeatable messy, I do not think I am qualified to dealt with it now. To take a journey into code base of Idris 2 is really a pedagogic and fun thing todo which had cured me a lot. Anyway, it would be better to turn this into a feature requesting issue. I wish the natïve proto would be helpful if someone who want to give a hand on but do not familiar with Idris 2 but experienced with functional programming and language implementation.

gallais:
I think this was good work. Some things may not be in the best place but
I really like that you even went to the trouble of adding the documentation
for `negate` and `=` that were lacking.

Let me take a look and try to cleanup the `Idris.DocString` module.

gallais:
Hmmm. The tests pass successfully on my machine.
I guess some of the TTC files may be out of sync.
I'll rebuild from scratch.

gallais (approved):

alissa-tung:
Very happy to see these changes! I have learnt a lot from them.

dotpat:
maybe `(->)` worth a doc too...

PR #1318, opened 2021-04-19
REPL commands: load-package & fuzzy-search

Russoul:
This PR:
- Adds `Maybe.filter` to `base`
- Fixes the definition of `Path.extension` both in `contrib` & `idris2`
- Adds a way of loading every single global definition in a package from the REPL (`lp` command)
- Adds a simple fuzzy-search algorithm, see the description below, invoked from the REPL (`fs` command)

Syntax of patterns of the fuzzy-search command is:
`_ -> B` or `B` (shortcut for the previous variant) or `A -> B`, where `A` and `B` are spines of (global) names and/or type literals.
E.g. 
  The first form: `_ -> Vect length`
  The second form: `Int List Type`
  The third form: `List -> Vect length`
What it does:
  For every visible global definition, try to match the given pattern against the type of that definition.
  The matching succeeds (and the definition is printed) if:
  - Every name in `A` (if any) occurs, cumulatively, within the "negative" subterms of the type, i.e. in the domain of top-level Pis.
  - Every name in `B` (if any) occurs within the "positive" subterm of the type, i.e. the chunk that happens to be the codomain of the last top-level Pi of the type.
  
Number of occurrences of a name can be arbitrary. In that case at least that quantity must be present in the target type (negatively or positively)  
`_` stands for the empty set of names (always matches)

In tandem with the load-everything-from-a-package command, fuzzy-search may prove to be useful when discovering the codebase.

Comments and suggestions on how to improve/complement the search are very welcome :)


src/Idris/REPL.idr
```diff
@@ -932,6 +937,200 @@ process NOP
...
+  res <- case errs' of
+    [] => pure (pretty "Done")
+    onePlus => pure $ vsep !(traverse display onePlus)
+  pure (Printed res)
+ where
+  unfoldTree : {root : _} -> Tree root -> Core (List String)
+  unfoldTree (MkTree files subtrees) = do
+    (map toFilePath files ++) . concat <$> flip Core.traverse subtrees \(_ ** treeIO) => do
+      tree <- coreLift treeIO
+      unfoldTree tree
```
gallais, line 967: This could be refactored in terms of `depthFirst`.

gallais (commented):
As mentioned on slack, I'm not sure the filesystem is the right level of
abstraction for this. Surely we only want to explore the modules that are
actually exported by the package & these are already listed in the package
file (and so show up together with the corresponding filepath in the `PkgDesc`
produced by `Idris.Package`).

Russoul:
The problem is that package file information is not available when you are loading that package as a dependency.
Idris only adds the relevant path to its internal list of package directories. See these lines:
https://github.com/idris-lang/Idris2/blob/5946442dc25aadd44d11955753c2d47c6f5027f6/src/Idris/SetOptions.idr#L157-L159
`PkgPath` holds a string, parsed when you enter e.g. `idris2 -p contrib ...` in a CLI session (` p = "contrib"` in that case).
`addPkgDir` does nothing more than just seeking for the path matching the given package name among the known paths to directories with `ttc` files.

The directory that stores package `ttc` files doesn't contain an `.ipkg` either. You can verify that yourself by browsing your `~/.idris2/idris2-0.3.0/any-package-name`

So I guess we are only left with the option to deal with the filesystem directly, unless I am missing something of course.

libs/contrib/System/Directory/Tree.idr
```diff
@@ -63,6 +68,14 @@ export
 emptyTree : Tree root
 emptyTree = MkTree [] []
 
+namespace Tree
+  export
+  setRoot : Tree root -> (0 root' : Path) -> Tree root'
+  setRoot (MkTree files sub) root' = assert_total $
```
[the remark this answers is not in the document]
Russoul: Dropped `setRoot` in favour of `toRelative`, as you suggested.
The one for the tree is now the identity.

src/Idris/REPL.idr
Russoul: Done

src/Idris/REPL.idr
```diff
@@ -949,6 +955,187 @@ process NOP
...
+  pure (Printed res)
+ where
+  toPaths : {root : _} -> Tree root -> IO (List String)
+  toPaths tree =
+    depthFirst (\x => map (toFilePath x ::) . force) tree (pure [])
+
+-- B or _ -> B or A -> B
+-- where A and B are spines of references.
+process (FuzzyTypeSearch expr) = do
+  let Just (neg, pos) = parseExpr expr
```
gallais, line 988: Can we put this in a separate module? e.g. `Idris.REPL.FuzzySearch`

gallais (commented):
I see. Thanks for the explanation! One last thing wrt housekeeping.

src/Idris/REPL.idr
Russoul: Done. Also did a few refactorings around the new namespace. See the commit message

gallais: Nice!

PR #1321, opened 2021-04-20
Allow implicit and multiplicities in parameter block arguments

### LINKED ISSUE #1242: Allow implicit parameters and multiplicities in parameter blocks
Recently I've been needing parameter blocks a lot and I noticed that I could not declare them with a custom multiplicity nor could I make the arguments implicit. While I suspect the multiplicity problem is a simple oversight, I think there is a bit more to talk about implicit parameter blocks.

Ideally I would like something like this:

```
parameters {auto m : Monoid a} 
           {0 proof1 : {v : a} -> (<+>) m v (neutral m) === v}
           {0 proof2 :  {v : a} -> (<+>) m (neutral m) v === v}
           {0 proof3 : (<+>) m ((<+>) m a b) c = (<+>) m a ((<+>) m b c)}
```

Now the question is, should we support the following?

```
parameters {auto m : Monoid a
           ,0 proof1 : {v : a} -> (<+>) m v (neutral m) === v}
           ,0 proof2 :  {v : a} -> (<+>) m (neutral m) v === v}
           ,0 proof3 : (<+>) m ((<+>) m a b) c = (<+>) m a ((<+>) m b c)}
```

Since this is how explicit parameters are declared?

The arguments for supporting it is that it would be consistent with explicit parameter declaration. The argument for _not_ supporting it is that nowhere else you can use this syntax to declare parameters, and I would argue this is also a bit awkward for explicit parameters since `(a, b c, : Nat)` is forbidden when declaring parameter arguments.

Additionally supporting it would mean that we should allow this interleaving of explicit and implicit arguments:

```
parameters {a : Ty, b : Sy} (f : a %% b, g : a %% b) {check : (x : Ty) -> f $$ x = g $$ x, auto m : Monad c}
```

I don't have a strong opinion either way but I would like to know what people think about it. I would tend toward not supporting the current `,`-syntax but that would break backward compatibility.

@Russoul  maybe we can unify that with your proposal for a new syntax for implicit/auto parameters?

andrevidela:
Fix #1242 

Since this is a breaking change in the syntax I propose to add a new flag `--old-parameter-blocks` which will allow parsing the old syntax so that we don't break bootstrapping.

Where in the past you could only do:

```
parameters (arg1 : T1, arg2 : T2, arg3 : T3)
```

Now you can do:

```
parameters {0 arg1 : T1} (1 arg2 : T2) (arg3 : T3)
```


Russoul:
Do we really need the flag ? We could wait until the next release before rewriting the existing uses of parameter blocks. And support both syntaxes until then ?

gallais:
> I propose to add a new flag --old-parameter-blocks which will allow
  parsing the old syntax so that we don't break bootstrapping.

We only have 2 such blocks in the whole of the compiler AFAICT.
I think it would be simpler to inline the parameters and to make
a not to bring the blocks back once a minor version is released.

src/Idris/Parser.idr
```diff
@@ -1344,14 +1344,35 @@ recordDecl fname indents
...
+                           symbol ")"
+                           ds <- assert_total (nonEmptyBlock (topDecl fname))
+                           pure (ps, ds))
+           (ps, ds) <- pure b.val
+           pure (PParameters (boundToFC fname b) ps (collectDefs (concat ds)))
+
+  newParamDecls : FileName -> IndentInfo -> Rule PDecl
+  newParamDecls fname indents
+      = do b <- bounds (do keyword "parameters"
+                           commit
```
gallais, line 1370: You can't commit because you can fail here and want to try `oldParamDecls`.

andrevidela:
test

andrevidela:
I changed the parser to support both syntaxes @gallais

src/Idris/Parser.idr
```diff
@@ -1345,13 +1345,28 @@ recordDecl fname indents
...
-    = do b <- bounds (do keyword "parameters"
-                         commit
-                         ps <- map concat (many $ typedArg fname indents)
-                         ds <- assert_total (nonEmptyBlock (topDecl fname))
-                         pure (ps, ds))
-         (ps, ds) <- pure b.val
-         pure (PParameters (boundToFC fname b) ps (collectDefs (concat ds)))
+    = do b1 <- bounds (keyword "parameters")
+         commit
+         args <- bounds (newParamDecls fname indents <|> oldParamDecls fname indents)
```
gallais, line 1350: You can `commit` again after `args` right? If the block does not parse, there is no
point in trying to parse it again by interpreting the parameters as old-style params.

andrevidela: Ah yes good point!

andrevidela:
I just noticed the changelog is wrong now

PR #1323, opened 2021-04-21
C Codegen Primitive FFI

madman-bob:
Adds support for foreign functions returning primitive types in the `RefC` backend.

I'm currently having issues with segfaults, so this is a draft for now.


src/Compiler/RefC/RefC.idr
```diff
@@ -131,23 +98,22 @@ where
...
 cConstant IntegerType = "i64"
 cConstant StringType = "string"
 cConstant CharType = "char"
 cConstant DoubleType = "double"
 cConstant WorldType = "f32"
-cConstant (B8 x)   = "(Value*)makeInt8("++ show x ++")" --"(constant #:type 'i64 #:val " ++ show x ++ ")"
-cConstant (B16 x)  = "(Value*)makeInt16("++ show x ++")" --"(constant #:type 'i64 #:val " ++ show x ++ ")"
-cConstant (B32 x)  = "(Value*)makeInt32("++ show x ++")" --"(constant #:type 'i64 #:val " ++ show x ++ ")"
-cConstant (B64 x)  = "(Value*)makeInt64("++ show x ++")" --"(constant #:type 'i64 #:val " ++ show x ++ ")"
+cConstant (B8 x)   = "(Value*)makeInt8("++ show x ++")"
```
gallais, line 113: Are these unsigned int types? In this case we should probably call the function `makeUInt8`.
Same question for the rest.

tests/refc/strings/TestStrings.idr
```diff
@@ -0,0 +1,43 @@
...
+    putStrLn $ show $ length helloWorld
+
+    putStrLn $ reverse helloWorld
+    putStrLn $ substr 1 2 helloWorld
+    putStrLn $ show $ assert_total $ strIndex helloWorld 1
+
+    putStrLn $ strCons 'a' "bc"
+    putStrLn $ show $ strUncons "abc"
+
+    putStrLn $ fastPack ['p', 'a', 'c', 'k']
```
gallais, line 27: Can we get some `fastPack` & so on tests on empty strings too? Last time around we had a bug
precisely when the input string was empty.

gallais (commented):
Looks very good to me! A cosmetic comment & one for test coverage.

src/Compiler/RefC/RefC.idr
madman-bob: They are stored internally as signed. I was considering changing them to unsigned, and renaming to `makeBits8` to more closely match the Idris types they represent, but this PR is already quite big.

I'm planning another PR for bigint support - I could include it in that.

tests/refc/strings/TestStrings.idr
madman-bob: Good call. Upon closer inspection I made an off-by-one error in `fastConcat`.

PR #1324, opened 2021-04-22
[ fix ] broken unicode parsing in JSON

stefan-hoeck:
JSON encoding and decoding of escaped unicode characters in contrib is broken in several ways. Here's the fix plus a test.

As an aside: Whoever implemented the erroneous version was working under the assumption that the `Cast` implementation from `Int` to `String` would behave correctly for hexadecimal strings, i.e. that the following would hold: 

```idris
the Int (cast "0x2") = 2
```

This doesn't work, though. Whether this is yet another bug, I do not know.


gallais:
Perfect!

PR #1327, opened 2021-04-22
[base] Minor: change the order of terms in the IntegerView

ohad:
This way it is similar to the `S n` view of natural numbers


gallais:
Should `IntRec` also be modified in a similar way?

Could you add a comment explaining this design choice so that it's not just
reverted later on without realising it's done on purpose?

ohad:
> Should `IntRec` also be modified in a similar way?

I guess -- I haven't worked with `IntRec` yet, so I don't know how this fans out, but sure!

> Could you add a comment explaining this design choice so that it's not just
> reverted later on without realising it's done on purpose?

Done.

PR #1329, opened 2021-04-24
[ new ] Add Contravariant and Const to base

stefan-hoeck:
I'd like to have a small lens library ala Haskell's microlens for Idris2. For this we need some utilities that are general enough to go to base. This PR adds several rather unrelated things, but they are so small that I thought it not to be worth to split them across several PRs. They come in distinct commits, though, so if one of the additions turns out to be controversial, I'll happily move that part to its own PR.

This PR adds
  
- some missing interface implementations for `Control.Monad.Identity`
- `Control.Applicative.Const` (the `Const` functor)
- interface `Data.Contravariant` (contravariant functors)
- `Data.Morphisms.Op` (the contravariant version of `Morphism`)

I didn't include `Profunctor` as that might be big enough to deserve its own PR.


libs/base/Data/Contravariant.idr
```diff
@@ -0,0 +1,48 @@
...
+
+||| This is an infix alias for `contramap`.
+public export %inline
+(>$<) : Contravariant f => (a -> b) -> f b -> f a
+(>$<) = contramap
+
+||| This is an infix version of `contramap` with the arguments flipped.
+public export %inline
+(>$$<) : Contravariant f => f b -> (a -> b) -> f a
+(>$$<) = flip contramap
```
Russoul, line 28: Shouldn't this be named `>&<`, since we have `<&>` in the Prelude for flipped `map` already?

stefan-hoeck: Ohh, I didn't even know about `<&>`. Good catch. Should be fixed now.

stefan-hoeck:
I rebased this to clean up the commit history (found two ugly copy-paste errors).

libs/base/Control/Applicative/Const.idr
```diff
@@ -0,0 +1,112 @@
+module Control.Applicative.Const
+
+import Data.Contravariant
+import Data.Bits
+
+public export
+record Const (a : Type) (b : Type) where
+  constructor MkConst
+  value : a
```
gallais, line 9: Can we call this projection `runConst` to match the overall convention?

stefan-hoeck: Done.

PR #1330, opened 2021-04-24
PrettyPrint: Fix SimpleDocTree.fromStream dropping parts of documents

cypheon:
When flattening the `SimpleDocTree` created from a `SimpleDocStream`, the
first part of a concatenated doc is sometimes dropped, depending on the
result of the recursive call to `flatten`.

Can be reproduced by:
`idris2 -p idris2` (Same bug is in corresponding module in `contrib`):

```idris
:module Libraries.Text.PrettyPrint.Prettyprinter.Doc
:module Libraries.Text.PrettyPrint.Prettyprinter.SimpleDocTree
fromStream $ SText 5 "first" (SText 6 "second" SEmpty)
```

Current result: `STText 6 "second"`

Result after this fix: `STConcat [STText 5 "first", STText 6 "second"]`


gallais:
Good catch!

PR #1331, opened 2021-04-24
[ fix #1328 ] print infix functions enclosed in grave accents

### LINKED ISSUE #1328: Cmd `:t` does not print infix function applications enclosed in grave accents
# Steps to Reproduce
```idris
module Main

data InSet : (a, b : _) -> Type where
infix 4 `InSet`

prop : x `InSet` a
prop = ?undef
```

# Expected Behavior
```idris
Main> :t undef 
 0 x : _
 0 a : _
undef : x `InSet` a
```

# Observed Behavior
```idris
Main> :t undef 
 0 x : _
 0 a : _
undef : x InSet a
```

alissa-tung:
may close #1328

### DIFF
--- a/src/Core/Name.idr
+++ b/src/Core/Name.idr
@@ -76,16 +76,6 @@ isUserName (DN _ n) = isUserName n
 isUserName _ = True
 
 export
-||| Test whether a user name begins with an operator symbol.
-isOpName : Name -> Bool
-isOpName n = let Just n = userNameRoot n
-                 | Nothing => False
-             in let n' = unpack n
-             in case n' of
-                  []     => False
-                  c :: _ => c `elem` (unpack ":!#$%&*+./<=>?@\\^|-~")
-
-export
 isUN : Name -> Maybe String
 isUN (UN str) = Just str
 isUN _ = Nothing
--- a/src/Core/Name.idr
+++ b/src/Core/Name.idr
@@ -76,6 +76,16 @@ isUserName (DN _ n) = isUserName n
 isUserName _ = True
 
 export
+||| Test whether a user name begins with an operator symbol.
+isOpName : Name -> Bool
+isOpName n = let Just n = userNameRoot n
+                 | Nothing => False
+             in let n' = unpack n
+             in case n' of
+                  []     => False
+                  c :: _ => c `elem` (unpack ":!#$%&*+./<=>?@\\^|-~")
+
+export
 isUN : Name -> Maybe String
 isUN (UN str) = Just str
 isUN _ = Nothing
--- a/src/Idris/Pretty.idr
+++ b/src/Idris/Pretty.idr
@@ -157,6 +157,10 @@ mutual
       appPrec = User 10
       leftAppPrec : Prec
       leftAppPrec = User 9
+      prettyOp : OpStr -> Doc ann
+      prettyOp op = if isOpName op
+        then pretty op
+        else Chara '`' <+> pretty op <+> Chara '`'
 
       go : Prec -> PTerm -> Doc IdrisAnn
       go d (PRef _ n) = pretty n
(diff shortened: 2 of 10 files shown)


src/Core/Name.idr
```diff
@@ -75,6 +75,16 @@ isUserName (NS _ n) = isUserName n
...
 
+export
+||| Test whether a user name begins with an operator symbol.
+isOpName : Name -> Bool
+isOpName n = let Just n = userNameRoot n
+                 | Nothing => False
+             in let n' = unpack n
+             in case n' of
+                  []     => False
+                  c :: _ => c `elem` (unpack ":!#$%&*+./<=>?@\\^|-~")
```
Russoul, line 86: Hm, if for some reason this string containing valid operator symbols is changed in the lexer, it will be harder to debug why printing is broken again :)
Should we decouple it and place somewhere accessible to both the lexer and the pretty-printer ?

alissa-tung: Thank you for reviewing! I think that would be nice to unify them. I will commit some changes but still not sure where is the suitable place. 0.0

Russoul: That works

alissa-tung:
I also noticed another pattern which occurrences in tests like this:
```idris
Issue1328A> Issue1328A.inDef0 : Nat
inDef0 = fromInteger 42 + fromInteger 21
```
Do `:printdef` does put nat constant literals under `fromInteger`. I am wondering if there are any use cases to show they are critical (or not)?

Russoul:
What do you mean by "critical"? Whether it's fine to omit them when printing?

alissa-tung:
> What do you mean by "critical"? Whether it's fine to omit them when printing?

Yep. About whether it's fine to omit them when printing and when is safe for doing it, plus when the cast operations are explicitly needed.

Russoul:
`fromInteger` can be replaced with a custom user-defined function, via a pragma.
But all in all, I don't think one can somehow enter a literal in the surface language and have it elaborated to something other than `fromInteger x` or similar. 
So I think printing `F x` as `x` where `x` is a literal and `F` is a special conversion function for literals  is safe and backward compatible.

But I don't remember where `F` is stored in the type-checker. You can trace it from the parser.

alissa-tung:
> `fromInteger` can be replaced with a custom user-defined function, via a pragma.
> But all in all, I don't think one can somehow enter a literal in the surface language and have it elaborated to something other than `fromInteger x` or similar.
> So I think printing `F x` as `x` where `x` is a literal and `F` is a special conversion function for literals is backward compatible.
> 
> But I don't remember where `F` is stored in the type-checker. You can trace it from the parser.

Thx a lot!

src/Idris/Pretty.idr
```diff
@@ -276,10 +278,22 @@ mutual
...
       go d (PPrefixOp _ op x) = parenthesise (d > appPrec) $ pretty op <+> go startPrec x
-      go d (PSectionL _ op x) = parens (pretty op <++> go startPrec x)
-      go d (PSectionR _ x op) = parens (go startPrec x <++> pretty op)
+      go d (PSectionL _ op x) = parens (prettyOp <++> go startPrec x)
+        where prettyOp : Doc ann
+              prettyOp = if isOpName op
+                then pretty op
+                else Chara '`' <+> pretty op <+> Chara '`'
+      go d (PSectionR _ x op) = parens (go startPrec x <++> prettyOp)
+        where prettyOp : Doc ann
```
gallais, line 293: Can we lift `prettyOp` as a definition on the same level as `go` so
that we don't need to have 3 copies of the code?

src/Idris/Syntax.idr
```diff
@@ -596,10 +598,19 @@ mutual
     showPrec d (PDotted _ p) = "." ++ showPrec d p
     showPrec _ (PImplicit _) = "_"
     showPrec _ (PInfer _) = "?"
-    showPrec d (POp _ op x y) = showPrec d x ++ " " ++ showPrec d op ++ " " ++ showPrec d y
+    showPrec d (POp _ op x y) =
```
gallais, line 601: Same remark here: please lift `showPrecOp : Prec -> Name -> String` out and refactor
each use site as `showPrec d x ++ " " ++ showPrecOp d op ++ " " ++ showPrec d y`

gallais (commented):
Looking good, some cleanups suggested

src/Idris/Pretty.idr
alissa-tung: thx, fixed

src/Idris/Syntax.idr
alissa-tung: fixed. this patch leads to some tests failed for the expected `x::xs` turns out to be `x :: xs` or `xs++ys` to `xs ++ ys`. I think these failures are kind of ... hmm ... work as expected. should I accept those new values?

alissa-tung: well, I had done diff in the wrong direction, nevermind

src/Parser/Lexer/Source.idr
```diff
@@ -238,6 +238,16 @@ export
...
 
+export
+||| Test whether a user name begins with an operator symbol.
+isOpName : Name -> Bool
+isOpName n = let Just n = userNameRoot n
+                 | Nothing => False
+             in let n' = unpack n
+             in case n' of
+                  []     => False
+                  c :: _ => isOpChar c
```
gallais, line 249: This way we avoid unpacking the whole string:

```suggestion
isOpName n = fromMaybe False $ do
   n <- userNameRoot n
   c <- fst <$> strUncons n
   guard (isOpChar c)
   pure True
```

alissa-tung: til :)

PR #1335, opened 2021-04-26
Semantic highlighting

ohad:
First stab at collecting, and sending, semantic highlighting information on IDE mode.

This involved adding new fields to Metadata, and extracting the information as the file progresses through the parser, desugaring, and elaboration. 

Since parsing, desugaring, and elaboration introduce new names that don't correspond to regions in the sourcefile, we introduced a new constructor `Core.FC.MkVirtualFC : FC`. 

Since the semantic information is stored in the `.ttm` file, it can be extracted by external tools (using the `idris2-api` library). For example, [Katla](http://www.github.com/ohad/katla) generates LaTeX code listings.

We leave to future PRs:
+ semantic highlighting of `(interactive)` terms (ones from a REPL, or arising from IDE commands)
  This might involve propagating this information through unelaboration and resugaring.
+ sending additional information about names (namespace, types of names).
+ semantic annotations of holes and `_`. We might want to send the semantic role of the root operation the hole has been solved for, or the full solution to a hover text bubble.
+ Highlight headers (the `import` statements at the top of each module). This is done by a different part of the parser, but should be straightforward.
+ Output semantic information as we progress through the file (eg, once after parsing takes place, and once more after type-checking completes). A straightforward first stab would be to send the information twice. Also straightforward, but slightly more fiddly, is to keep track of what was sent and only send the new stuff. Maintaining two `PosMap`s should do.


src/Idris/Parser.idr
```diff
@@ -491,7 +491,7 @@ mutual
     <|> do start <- bounds (symbol "[")
            listExpr fname start indents
     <|> do b <- bounds (decoratedSymbol fname "!" *> simpleExpr fname indents)
-           pure (PBang (boundToFC fname b) b.val)
+           pure (PBang (virtualiseFC $ boundToFC fname b) b.val)
```
gallais, line 494: I think we need to keep the semantics that `virtualiseFC` is only used on things
that do not exist in the source code. Same remark for the virtualise on `PCase`
and the various do blocks.

gallais: I have had a look at `checkCase` and let's just go with virtualising the argument to
`PCase` because this is nasty...

src/Idris/Elab/Interface.idr
```diff
@@ -125,6 +126,10 @@ mkIfaceData {vars} fc vis env constraints n conName ps dets meths
                                                   (mkDataTy fc ps))
                                   opts [con])
   where
+
+    fc : FC
```
gallais, line 130: This should probably be changed because of #1356

gallais:
AFAIK we've now covered the whole language.

If people want to try the branch and look at some code exercising more niche corners
of the language and try to find missing semantic highlighting, please do!

ohad:
Interface implementations leave some weird residue. For example:
```idris
data Foo : Type where

%logging "ide-mode.highlight" 21

Eq Foo where
```

has (among other stuff):
```
LOG ide-mode.highlight:8: getNameType is trying to add something for: Prelude.Basics.Bool (Prelude/EqOrd.idr:16:22--16:26)
LOG ide-mode.highlight:7: getNameType is adding type: Prelude.Basics.Bool
```
which adds the annotation:
```
LOG ide-mode.highlight:20: highlighting at ("Prelude/EqOrd.idr", ((15, 21), (15, 25))): Prelude.Basics.Bool
As: type
```

ohad:
Shall we start thinking about merging this?

gallais:
I still need to fix the issue with nested withs.

ohad:
> Especially given we've been very lucky to avoid merge conflicts so far given we've touched
> 100+ files. :D

Yes, large numbers are against us on this one.
Does anything needs to happen? Do we need to poke someone? (say, @edwinb ?)

gallais:
I'm not planning to merge anything until next week and to merge this one first.

I've advertised the PR on our team's chat and told people we'd be grateful
for any bug reports so either none come in and we can ship straight away or
some arise and if they're easy we can quash them quickly & then merge.

PR #1336, opened 2021-04-26
Lambdalift optimisation: drop unused variables from outer scopes

cypheon:
This is my attempt at dropping unused variables during the "LambdaLift" compilation phase, which reduced the generated "lifted defs" for at least one real-world example (Idris 2 itself) by about 35%.
Usage bookkeeping is currently done via a `Vect n Bool` but could theoretically be moved to an IOArray (all the usage tracking code is in `Core`). In practice it turns out not to matter that much, since lambda lifting takes only about 1 sec for all of Idris 2 (as opposed to 0.9 secs without the usage bookkeeping).

### DIFF
--- a/src/Compiler/LambdaLift.idr
+++ b/src/Compiler/LambdaLift.idr
@@ -158,6 +158,66 @@ unload fc _ f [] = pure f
 -- only outermost LApp must be lazy as rest will be closures
 unload fc lazy f (a :: as) = unload fc Nothing (LApp fc lazy f a) as
 
+record Used (vars : List Name) where
+  constructor MkUsed
+  used : Vect (length vars) Bool
+
+initUsed : {vars : _} -> Core (Used vars)
+initUsed {vars} = do
+  pure $ MkUsed (replicate (length vars) False)
+
+contractUsed : {vars : _} ->
+               (Used (x::vars)) ->
+               Core (Used vars)
+contractUsed (MkUsed (_::rest)) = pure $ MkUsed rest
+
+contractUsedMany : {vars : _} ->
+                   {remove : _} ->
+                   (Used (remove ++ vars)) ->
+                   Core (Used vars)
+contractUsedMany {remove=[]} x = pure x
+contractUsedMany {remove=(r::rs)} x = contractUsedMany {remove=rs} !(contractUsed x)
+
+markUsed : {vars : _} ->
+           (idx : Nat) ->
+           {0 prf : IsVar x idx vars} ->
+           Used vars ->
+           Core (Used vars)
+markUsed {vars} {prf} idx (MkUsed us) = do
+  let newUsed = replaceAt (finIdx prf) True us
+  pure $ MkUsed newUsed
+    where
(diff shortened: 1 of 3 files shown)


src/Compiler/LambdaLift.idr
```diff
@@ -234,6 +297,129 @@ mutual
   liftExp (CErased fc) = pure $ LErased fc
   liftExp (CCrash fc str) = pure $ LCrash fc str
 
+  usedVars : {vars : _} ->
```
gallais, line 300: Can we not pass a `Used vars` around and accumulate the values in it rather than
creating one for each `LLocal` and then merging all of these arrays as we come out
of nodes with multiple arguments?

src/Compiler/LambdaLift.idr
```diff
@@ -158,6 +158,66 @@ unload fc _ f [] = pure f
...
 
+record Used (vars : List Name) where
+  constructor MkUsed
+  used : Vect (length vars) Bool
+
+initUsed : {vars : _} -> Core (Used vars)
+initUsed {vars} = do
+  pure $ MkUsed (replicate (length vars) False)
+
+contractUsed : {vars : _} ->
```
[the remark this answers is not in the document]
cypheon: I think I removed the implicit argument where possible (some places actually _do_ need it, but most didn't)

src/Compiler/LambdaLift.idr
```diff
@@ -158,6 +158,66 @@ unload fc _ f [] = pure f
...
+    finIdx {idx=S x} (Later l) = FS (finIdx l)
+
+mergeUsed : {vars : List Name} ->
+           Used vars ->
+           Used vars ->
+           Core (Used vars)
+mergeUsed {vars=[]} (MkUsed []) (MkUsed _) = pure $ MkUsed []
+mergeUsed {vars=(_::_)} (MkUsed (x::xs)) (MkUsed (y::ys)) = do
+  MkUsed rest <- mergeUsed (MkUsed xs) (MkUsed ys)
+  pure $ MkUsed ((x || y)::rest)
```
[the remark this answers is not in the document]
cypheon: The `Core` thing was left over from when I thought I might need an `IOArray` to achieve acceptable performance. It seems `Vect` performs quite well (at least well enough), so I removed the `Core` wrappers. `mergeUsed` might even be completely unnecessary once I get to implement changes regarding your third comment.

src/Compiler/LambdaLift.idr
cypheon: Should be possible (and will probably avoid some unnecessary allocations), I'll get to that one a bit later and update the PR again

cypheon:
Thanks for your feedback, very helpful suggestions

src/Compiler/LambdaLift.idr
cypheon: Done. This change also made the code a bit simpler, because mergeUsed is no longer needed at all.

edwinb:
Looks great, thanks!

PR #1337, opened 2021-04-26
[ fix ] print infix operators with parens

gallais:
I noticed that e.g. List's (::) was not printed correctly in `:doc`
so here's a fix.


ziman:
#1337 :tada:

Issue #1241, opened 2021-03-28
Add a REPL shortcut for `:t` with implicits shown


youngmysteriouslight:
# Steps to Reproduce

```idris
data V : Nat -> Type

fpiei : (f : {e : Nat} -> V e) -> V e
fp0ei : (0 f : {e : Nat} -> V e) -> V e
fpie0 : (f : {0 e : Nat} -> V e) -> V e
fp0e0 : (0 f : {0 e : Nat} -> V e) -> V e
fmiei : {e : Nat} -> (f : V e) -> V e
fm0ei : {e : Nat} -> (0 f : V e) -> V e
fmie0 : {0 e : Nat} -> (f : V e) -> V e
fm0e0 : {0 e : Nat} -> (0 f : V e) -> V e
```
(naming convention: `fpie0` stands for "argument `f` is polymorphic, its parametricity is infinity, parametricity of `e` is zero", `fm0ei` stands for "monomorphic `f` with zero parametricity, parametricity of `e` is infinity")

Run in REPL
```
Main> :t fpiei
Main.fpiei : V e -> V e
Main> :t fpie0
Main.fpie0 : V e -> V e
Main> :t fmiei
Main.fmiei : V e -> V e
Main> :t fmie0
Main.fmie0 : V e -> V e
Main> :t fp0ei
Main.fp0ei : (0 _ : V e) -> V e
Main> :t fp0e0
Main.fp0e0 : (0 _ : V e) -> V e
Main> :t fm0ei
Main.fm0ei : (0 _ : V e) -> V e
Main> :t fm0e0
Main.fm0e0 : (0 _ : V e) -> V e
```

# Expected Behavior

`:t` shows different types distinguishably.

# Observed Behavior

4 different types are represented with `V e -> V e`.

youngmysteriouslight:
The example points out that `:t` ignores implicit parameters. That can confuse while debugging since `V e -> V e` is interpreted as `{e : ?} -> V e -> V e` but never `({e : ?} -> V e) -> V e`.

I think the problem might be in Resugar.idr module which begins with the following note:
https://github.com/idris-lang/Idris2/blob/ec77ad21abfcc335971c8170e3dbe94372f86431/src/Idris/Resugar.idr#L20
```
-- Convert checked terms back to source syntax. Note that this is entirely
-- for readability therefore there is NO GUARANTEE that the result will
-- type check (in fact it probably won't due to tidying up names for
-- readability).
```

[...]

PR #1338, opened 2021-04-26
[ REPL ] Added commands :ti and :opts (#1241)

ionathanch:
`:ti` does the same thing as `:t`, but shows the type as if `showimplicits` were set. The value of `REPLOpt.ShowImplicits` is unchanged.

`:opts` shows the current values of the options that can be set or unset with `:set`/`:unset`. (This already existed in `REPLCmd.GetOpts`, idk why it just wasn't in `Parser.parserCommandsForHelp`.)

I can add to or change the names of the commands; I don't think `:t+` can be parsed, and I don't know if we also want a `:typei`?

Also fixed the spacing of the help message.


gallais:
Did you forget to implement `:opts`?

ionathanch:
It was already implemented (see https://github.com/idris-lang/Idris2/blob/master/src/Idris/REPL.idr#L831), I just added `:opts` to the parser

PR #1348, opened 2021-04-27
[ refactor ] type inference and evaluation for the REPL

JoeyEremondi:
This creates an `inferAndNormalize` function in `Idris/REPL.idr` that, given an evaluation mode and a `PTerm`, produces the term's type, its normal form, and the resugared version of its normal form. The idea is that other REPL commands and IDE actions, like refining a hole, will likely need this information, so the functionality can be shared.

### DIFF
--- a/src/Idris/REPL.idr
+++ b/src/Idris/REPL.idr
@@ -644,9 +644,6 @@ loadMainFile f
            [] => pure (FileLoaded f)
            _ => pure (ErrorsBuildingFile f errs)
 
-emode : REPLEval -> ElabMode
-emode EvalTC = InType
-emode _ = InExpr
 
 nfun : {auto c : Ref Ctxt Defs} ->
   {vs : _} ->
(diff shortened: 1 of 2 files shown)


JoeyEremondi:
I've separated elaboration out of `inferAndNormalize`, so there's separate `inferAndElab` and `inferAndNormalize` functions. The elaboration is shared between the handling of `Eval` and `Check` REPL cases.

src/Idris/REPL.idr
```diff
@@ -644,6 +644,63 @@ loadMainFile f
...
+||| Infer the type of a PTerm and elaborate it to a Term
+||| Returns a pair containing the term and its type
+inferAndElab : {auto c : Ref Ctxt Defs} ->
+  {auto u : Ref UST UState} ->
+  {auto s : Ref Syn SyntaxInfo} ->
+  {auto m : Ref MD Metadata} ->
+  {auto o : Ref ROpts REPLOpts} ->
+  ElabMode ->
+  PTerm ->
+  Core (Term [], Term [])
```
gallais, line 663: Can we get a `TermAndType` record type so that these can be accessed by name?
They have exactly the same type so it's very easy to use the wrong one.

src/Idris/REPL.idr
```diff
@@ -644,6 +644,63 @@ loadMainFile f
            [] => pure (FileLoaded f)
            _ => pure (ErrorsBuildingFile f errs)
 
+
+nfun : {auto c : Ref Ctxt Defs} ->
```
gallais, line 648: As a local name `nfun` is probably fine but as a toplevel function I'd rather
have something a bit more descriptive. Maybe `replEval` as it seems to be
computing the semantics of `REPLEval`?

JoeyEremondi:
Alright, both changes made. Let me know if the infix pattern matching for the record types provides less clarity than it gives, I could just switch it to using the accessors if that's better.

Issue #1182, opened 2021-03-13
Comparison operations on Nat should be O(1)


stefan-hoeck:
Right now, arithmetic functions like addition and multiplication on `Nat` are optimized to be O(1) by the compiler. It would be great, if this would also be the case for comparison operators like `(==)` and `(<=)` and probably even `lte` and friends from `Data.Nat` in base. Right now, these seem to be linear. For instance, building and running the following example takes 20 seconds on my system:

```idris
main : IO ()
main = printLn (the Nat 10000000000 == 10000000000)
```

while the following is instantaneous (which I think is already pretty cool :-) )

```idris
main : IO ()
main = printLn . the Nat $ 10000000000 * 10000000000
```

Feel free to close this issue, if this is not feasible or not desired.

Z-snails:
I think it would be better if every `Nat`-like type could be optimised by the compiler.
One way this could be done is like Agda's Builtin pragma, which would mean types like `Fin` are as fast as `Nat`.
Alternatively the compiler could try checking every data type but that might slow down compilation too much.

gallais:
The hard thing is that the definition used in the compiler should work
for open terms too. I suppose we could have some special logic to patch
these defined names on the fly so that evaluation would be using a
different implementation for closed terms. Checking whether a term is
closed could be expensive though.

Z-snails:
I don't know the codebase very well except the backends, but currently `Nat`s are optimised when going to `CExp` (are they optimised elsewhere?). I don't know if `CExp` is used during typechecking but if it isn't surely we can add that optimisation when going to `CExp` before passing that to the backend?

Z-snails:
I'm about to start work on adding O(1) `Nat` `Ord` and `Eq`, but I need to add some more constructors to `BuiltinType`, does anyone have an opinion on using a `String` instead (fewer ttc version changes) or extend the `BuitlinType` ADT (nicer?)?

stefan-hoeck:
@Z-snails have you seen #1357?

Z-snails:
Yep, but %transform works on `Term` but nats are only transformed to `Integer`s in `CExp` so there'll be some time where it is wrong. Unless there is something else you mean?

Z-snails:
It's easier to optimise natToInteger and integerToNat then define (or %transform) other operations to use that, so that's what I'll do. Is that what you meant?

stefan-hoeck:
[...]

PR #1357, opened 2021-04-30
A couple of optimisations

edwinb:
A couple of changes that make Idris typecheck faster, and will be useful in general:

* Use the `%transform` syntax to pick fast runtime versions of things where possible. This isn't documented properly yet (or, I think, at all) so I'll get to that later. It's basically rewrite rules of the form `%transform "label" lhs = rhs` where `lhs` is a function application with variable names, and `rhs` is what it rewrites to.
* Add a flag to check when any logging has been enabled, so we only have the overhead of checking the categories if we're actually logging. The above change does make logging faster, but this makes it cheaper still.

On my machine, checking the Idris 2 source (with `--check`) is now done in around 1m26 (down from 1m45) on the current code base.

### DIFF
--- a/libs/base/Data/String.idr
+++ b/libs/base/Data/String.idr
@@ -18,25 +18,6 @@ partial
 foldl1 : (a -> a -> a) -> List a -> a
 foldl1 f (x::xs) = foldl f x xs
 
--- This function runs fast when compiled but won't compute at compile time.
--- If you need to unpack strings at compile time, use Prelude.unpack.
-%foreign
-  "scheme:string-unpack"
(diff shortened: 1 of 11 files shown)


stefan-hoeck:
Is it possible to do this also for `Eq` and `Ord` operations on `Nat`? They are linear right now, but could be made O(1) if going via `Integer` instead. (See #1182)

edwinb:
Hmm, yes, I bet there's quite a few things we could do! I've only added things that showed up in a profile but maybe there's more Nat optimisations we should do straight away. I'll save them for another patch though.

edwinb:
The fail is because I took out a log and forgot to update the tests (or think about what they should be). Will fix...

Issue #1358, opened 2021-05-01
Idris2 process holds lots of useless memory during codegen stage


ziman:
As this crop-out shows, Idris2 hangs on a gigabyte memory during code generation and Chez compilation.
![image](https://user-images.githubusercontent.com/316753/116779305-d36d8e00-aa75-11eb-8f25-3da445d9fb2e.png)
The useless idris2 memory can be seen as the big blue stripe in the right half of the plot.

This is very unfortunate because the memory is really useless. It's served its purpose, the code has been generated and the only thing left to do before exiting is calling Chez to do its job.

If this memory were gone, we could do the following things:
* Generate code on older hardware / less powerful platforms because the maximum overall memory usage happens in this stage.
* Make codegen compilation a lot more parallelised, and thus (hopefully) a lot faster.

I tried to invoke the Chez compiler at the end of `Idris.Package.processPackage`, after `withCtxt` was done, but even after FFI-ing out to `(collect)`, the Idris process kept hanging on to the memory.

It's a shame we can't do this so I'm filing this issue for the record and future reference, at least.

ziman:
The fact that GC does not have much effect after `withCtxt` might have something to do with https://github.com/idris-lang/Idris2/pull/1304/files#r617554834.

locriacyber:
I think this is fixed in main branch.

PR #1359, opened 2021-05-01
[ new ] Add `chez-sep` codegen

ziman:
This patch adds a separately compiling Chez codegen, named `chez-sep`.

Shortcomings/future work:
* Paralellise the `.ss` to `.so` compilation step. Could be done entirely in Chez, since this step is a single Chez script.
* [Stop holding lots of useless memory during code generation](https://github.com/idris-lang/Idris2/issues/1358)
* The compiler looks only at `.hash` files to asses whether it needs to generate code. If the `.ss` or `.so` is broken or missing, it'll break the build. This won't happen normally but it will happen if you're doing development on codegens that mess with the codegen directory and then quit midway through. For now, the easiest fix is to `rm -rf build/exec`.
* Check that the codegen works on non-Linux platforms.
* CI tests of the codegen.
* Check that the `Makefile` works with `chez-sep`-generated `idris2` on non-Linux platforms.
* Sort out `extraRuntime`, which is now ignored in the Separate Chez backend.
* Add `Hashable` and `HasNamespaces` implementations for the remaining IRs.
* Benchmark the performance of the separately compiled programs.
* Try out the Chez whole-program optimisation feature with separate compilation. Maybe we can get the runtime performance benefits without the disadvantages of having to compile a single huge `.ss` file.
* Try out the Chez `.so` concatenation feature to pack up all compiled modules in one file.


ziman:
For the record, this is how to bootstrap if you want to switch your entire installation to Idris2 built exclusively by `chez-sep` (not necessary for building programs with `chez-sep`, of course):
* `idris2 --build idris2.ipkg` -- build Idris that contains `chez-sep`
* `make install` -- install Idris containing `chez-sep` into `$PATH`
* add `options = "--codegen chez-sep"` in `idris2.ipkg`
* `idris2 --build idris2.ipkg` -- build Idris using `chez-sep`
* `make install` -- install separately compiled Idris

stefan-hoeck:
It would be great to have this documented somewhere (probably via a separate PR): What are the benefits of this, when to use it, when not to use it. I have only superficially been following this development and would find it useful to have some quick notes about why I should give this a try.

ziman:
Note to self: besides Stefan's good idea, update `CHANGELOG.md`.

edwinb:
I just tried this, but it looks like it needs to be brought up to date with the new integer types before we can merge. I got errors in Core.Hash, which were easy to fix, but then complaints about unbound identifiers, which I can't immediately see how to fix.

src/Compiler/Scheme/ChezSep.idr
```diff
@@ -0,0 +1,338 @@
...
+writeFileCore fname content =
+  coreLift (writeFile fname content) >>= \case
+    Right () => pure ()
+    Left err => throw $ FileErr fname err
+
+readFileCore : (fname : String) -> Core String
+readFileCore fname =
+  coreLift (readFile fname) >>= \case
+    Right content => pure content
+    Left err => throw $ FileErr fname err
```
gallais, line 145: Can we move these to `Core.Core`?

src/Compiler/Scheme/ChezSep.idr
```diff
@@ -0,0 +1,338 @@
...
+chezNS ns = case showNSWithSep "-" ns of
+  "" => "unqualified"
+  nss => nss
+
+-- arbitrarily name the compilation unit
+-- after the alphabetically first namespace contained within
+chezLibraryName : CompilationUnit def -> String
+chezLibraryName cu =
+  case SortedSet.toList cu.namespaces of
+    [] => "unknown"  -- this will never happen because the Tarjan algorithm won't produce an empty SCC
```
gallais, line 157: Should Tarjan produce a `List1` then?

src/Compiler/Scheme/ChezSep.idr
```diff
@@ -0,0 +1,338 @@
...
+    ns::_ => chezNS ns
+
+touch : String -> Core ()
```
gallais, line 160: Same: can we move this to `Core.Core`? Maybe even create `Core.Core.File` if we have a lot of file manipulation functions.

src/Compiler/Scheme/ChezSep.idr
```diff
@@ -0,0 +1,338 @@
...
+
+||| Make Windows start scripts, one for bash environments and one batch file
+makeShWindows : String -> String -> String -> String -> Core ()
+makeShWindows chez outShRel appDirSh targetSh
+    = do let cmdFile = outShRel ++ ".cmd"
+         Right () <- coreLift $ writeFile cmdFile (startChezCmd chez appDirSh targetSh)
+            | Left err => throw (FileErr cmdFile err)
+         Right () <- coreLift $ writeFile outShRel (startChezWinSh chez appDirSh targetSh)
+            | Left err => throw (FileErr outShRel err)
+         pure ()
```
gallais, line 279:
```suggestion
         writeFileCore cmdFile (startChezCmd chez appDirSh targetSh)
         writeFileCore outShRel (startChezWinSh chez appDirSh targetSh)
```

src/Compiler/Scheme/ChezSep.idr
```diff
@@ -0,0 +1,338 @@
...
+  pure (Just outShRel)
+
+||| Chez Scheme implementation of the `executeExpr` interface.
+||| This implementation simply runs the usual compiler, saving it to a temp file, then interpreting it.
+executeExpr : Ref Ctxt Defs -> (tmpDir : String) -> ClosedTerm -> Core ()
+executeExpr c tmpDir tm
+    = do Just sh <- compileExpr False c tmpDir tmpDir tm "_tmpchez"
+            | Nothing => throw (InternalError "compileExpr returned Nothing")
+         coreLift_ $ system sh
+         pure ()
```
gallais, line 333: No need for a trailing `pure ()`

src/Core/TTC.idr
```diff
@@ -749,6 +749,7 @@ TTC CDef where
 export
 TTC CG where
   toBuf b Chez = tag 0
+  toBuf b ChezSep = tag 1
```
gallais, line 752: You should bump the version number in `Core.Binary`

gallais (commented):
This is pretty much ready to merge.

We should definitely open a new issue to track the future work.

Makefile
```diff
@@ -143,6 +143,7 @@ endif
 	install support/c/${IDRIS2_SUPPORT} ${PREFIX}/lib
 	mkdir -p ${PREFIX}/bin/${NAME}_app
 	install ${TARGETDIR}/${NAME}_app/* ${PREFIX}/bin/${NAME}_app
+	-rm -f "${PREFIX}/bin/${NAME}_app"/*.ss  # SS files interfere with separately compiled idris
```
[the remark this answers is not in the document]
melted: It allows it to fail without aborting the build.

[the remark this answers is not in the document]
gallais: Isn't that the point of `-f`?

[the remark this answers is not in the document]
ziman: No. `-f` means "don't ask for confirmation interactively". `-rm` means "if there's no such file, don't abort the whole make".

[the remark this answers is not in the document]
gallais:
```shell
       -f, --force
              ignore nonexistent files and arguments, never prompt
```

src/Compiler/Scheme/ChezSep.idr
```diff
@@ -0,0 +1,338 @@
...
+  , "                           (load-shared-object \"ws2_32.dll\")]"
+  , "  [else (load-shared-object \"libc.so\")]"
+  , unlines ["  (load-shared-object \"" ++ escapeString lib ++ "\")" | lib <- libs]
+  , ")"
+  ]
+
+schFooter : String
+schFooter = "(collect 4)\n(blodwen-run-finalisers)\n"
+
+startChez : String -> String -> String -> String
```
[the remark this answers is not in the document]
ziman: Sure, I factored out `startChezPreamble`.

src/Compiler/Scheme/ChezSep.idr
ziman: Great idea. Done.

src/Compiler/Scheme/ChezSep.idr
ziman: Well spotted, thanks!

src/Core/TTC.idr
ziman: Thanks! Done.

ziman:
@stefan-hoeck I added [a short blurb in the changelog](https://github.com/idris-lang/Idris2/pull/1359/commits/e22b0e91646c6aa05ecfa9df135c86b3d534cab3). Not sure if that's the best place but it looks reasonable to me, at least. If you have better ideas or would prefer a more detailed description, please let me know.

src/Compiler/Separate.idr
```diff
@@ -294,12 +295,12 @@ getCompilationUnits {def} defs =
...
     --
     -- Each SCC will become a compilation unit.
-    components : List (List Namespace)
+    components : List (List1 Namespace)
       = List.reverse $ tarjan nsDeps  -- tarjan generates reverse toposort
 
     -- Maps a namespace to the compilation unit that contains it.
     nsMap : SortedMap Namespace CompilationUnitId
-      = SortedMap.fromList [(ns, cuid) | (cuid, nss) <- withCUID components, ns <- nss]
+      = SortedMap.fromList [(ns, cuid) | (cuid, nss) <- withCUID components, ns <- toList nss]
```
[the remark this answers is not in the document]
ziman: Oh, right, `toList` is a `Foldable` thing. Uh, thanks for catching this!

Yes, I think I'd support that change to `Foldable`.

Makefile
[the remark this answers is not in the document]
ziman: How portable is that, though? Can we rely on this on multiple platforms?

Makefile
```diff
@@ -149,7 +149,7 @@ endif
 	install support/c/${IDRIS2_SUPPORT} ${PREFIX}/lib
 	mkdir -p ${PREFIX}/bin/${NAME}_app
 	install ${TARGETDIR}/${NAME}_app/* ${PREFIX}/bin/${NAME}_app
-	-rm -f "${PREFIX}/bin/${NAME}_app"/*.ss  # SS files interfere with separately compiled idris
+	$(RM) -f "${PREFIX}/bin/${NAME}_app"/*.ss  # SS files interfere with separately compiled idris
```
ska80, line 152: The default value for the `RM` variable is `rm -f`.  And you have missed `-` before `$(RM)`.

ziman: Other occurrences of `$(RM)` don't use `-` (for example in the `clean` target) so I removed it. Should I put it back in?

ska80: If your intention was to continue even if it is unable to remove files then you should put `-` back.

PR #1360, opened 2021-05-01
Update docs to reflect usage of >> in do syntax

RaoulHC:
As of  #1095 do syntax desugars using both `>>=` and `>>`, and this wasn't mentioned in the docs.

Took me a little bit to notice that change when trying to use do syntax with something that had only implemented `>>=`.

### DIFF
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -277,12 +277,18 @@ defined as follows:
     interface Applicative m => Monad (m : Type -> Type) where
         (>>=)  : m a -> (a -> m b) -> m b
 
+There is also a non-binding sequencing operator, defined for ``Monad`` as:
+
+.. code-block:: idris
+
+    v >> e = v >>= \_ => e
+
 Inside a ``do`` block, the following syntactic transformations are
 applied:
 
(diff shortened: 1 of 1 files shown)


docs/source/tutorial/interfaces.rst
```diff
@@ -282,7 +282,8 @@ applied:
 
 - ``x <- v; e`` becomes ``v >>= (\x => e)``
 
-- ``v; e`` becomes ``v >>= (\_ => e)``
+- ``v; e`` becomes ``v >> e``, where ``>>`` is the non-binding sequencing
+  operator defined as ``v >>= \_ => e``.
```
buzden, line 286: This definition is correct only for `>>` defined for `Monad`, but is not correct in the general case. Since `do`-notation works syntactically, not only for `Monad`, this should be reformulated.

RaoulHC: Updated with a line above defining `>>`. Wasn't sure whether to include the type signature, or whether that over complicates the explanation, given it's now defined as `(>>) : Monad m => m () -> Lazy (m b) -> m b`.

edwinb:
Thanks for the update!

PR #1363, opened 2021-05-01
O(1) natToInteger for any 'Nat'-like type

Z-snails:
Still very much a draft, but I'd like some feedback.

How important is it that we check the function is actually correct? Currently `%transform` doesn't so maybe this doesn't need to either.

Edit:
no longer a draft.


Z-snails:
Do I need to wait for the next release before adding `%builtin` pragmas to prelude/base? Would this break bootstrapping?

stefan-hoeck:
As far as I know, yes, we have to wait until the next minor release before being able to use `%builtin` in the libraries. And thanks again, for working on this!

src/Compiler/CompileExpr.idr
```diff
@@ -208,6 +209,10 @@ builtinMagic = do
         , MagicCCon cons.succ 1
              (\ fc, [k] => CApp fc (CRef fc (UN "prim__add_Integer")) [CPrimVal fc (BI 1), k])
         ] -- TODO: add builtin pragmas for Nat related functions (to/from Integer, add, mult, minus, compare)
+    natToIntMagic : (Name, NatToInt) -> Magic
+    natToIntMagic (fn, MkNatToInt {arity, natIdx}) =
```
gallais, line 213: Please don't use named pattern-matching in the compiler: the syntax will probably have
to change in a backwards incompatible way in the future because of #965.

src/Core/Context.idr
```diff
@@ -649,22 +650,11 @@ data Transform : Type where
...
-    ||| 'SLike : {0 _ : index} -> NatLike [index] -> NatLike [f index]'
-    ||| 'ZLike : {0 _ : index} -> NatLike [index]'
-    BuiltinNatural : BuiltinType
-    -- All the following aren't implemented yet
-    -- but are here to reduce number of TTC version changes
-    NaturalPlus : BuiltinType
-    NaturalMult : BuiltinType
-    NaturalToInteger : BuiltinType
-    IntegerToNatural : BuiltinType
+    MkBuiltinType : String -> BuiltinType
```
gallais, line 653: Can you comment on this change? We're moving from an enumerated type that can be checked
for exhaustivity / may raise warnings if a pattern is mistyped to something stringly-typed.
This seems like a net loss.

src/TTImp/ProcessBuiltin.idr
```diff
@@ -129,45 +185,82 @@ checkCons c cons ty fc = case !(foldr checkCon (pure (Nothing, Nothing)) cons) o
...
     addBuiltinNat n $ MkNatBuiltin {zero, succ}
 
+||| Check a `%builtin NaturalToInteger` pragma is correct.
+processNatToInteger :
+    Ref Ctxt Defs =>
+    Defs -> FC -> Name -> Core ()
+processNatToInteger ds fc fn = do
+    log "builtin.NaturalToInteger" 5 $ "Processing %builtin NaturalToInteger " ++ show fn ++ "."
+    [(n, _, gdef)] <- lookupCtxtName fn ds.gamma
+        | [] => throw $ UndefinedName fc fn
```
gallais, line 236: Here you can use `undefinedName fc fn` instead of `throw $ UndefinedName fc fn` and Idris
will try to suggest names that look similar in case the user made a typo.

src/TTImp/ProcessBuiltin.idr
```diff
@@ -129,45 +185,82 @@ checkCons c cons ty fc = case !(foldr checkCon (pure (Nothing, Nothing)) cons) o
...
+    let Just tyCon = getTypeCons arg
+        | Nothing => throw $ GenericMsg fc
+            $ "No type constructor found for non-erased arguement of " ++ show n ++ "."
+    Just _ <- getNatBuiltin tyCon
+        | Nothing => throw $ GenericMsg fc $ "Non-erased argument is not a 'Nat'-like type."
+    let arity = length $ getTypeArgs gdef.type
+    let Just natIdx = getNEIndex arity gdef.type
+        | Nothing => throw $ InternalError "Couldn't find non-erased argument."
+    addNatToInteger n (MkNatToInt {arity, natIdx})
+    pure ()
```
gallais, line 253: No need to finish a block with `pure ()`

gallais (commented):
This looks good to me. A few suggestions.

For the future though, I wonder whether we could have a system where we'd be
able to write an approximate type that focuses only on non-erased arguments
and implement matching once and for all for these.

Eg.


```idris
natToInteger : NatLike -> Integer
add : NatLike -> NatLike -> NatLike
```

and then we'd check that, discounting erased arguments, the types of the
bound primitives do look like that, that it's the same `NatLike` in all
positions & that we get back a list of the indices corresponding to these
positions.

src/Core/Context.idr
Z-snails: My thinking was fewer ttc version changes because they are really annoying, especially when you're working with multiple compilers with different versions, but I can change it back if that would be safer/better. Although I reckon we can stick with just BuiltinNatural, NaturalToInteger and IntegerToNatural and define the rest in terms of those, so that wouldn't actually be too many ttc version changes now I think about it.

src/TTImp/ProcessBuiltin.idr
Z-snails: Leftover from when I was defining it to get it to typecheck, and I forgot to remove it...

Z-snails:
I was thinking about having some sort of view of `Term`s representing types as a list of pi bound arguments and their multiplicites and another view of `Term` as a type constructor, but I never got round to doing that. Probably someone who knows QTT better than me should do that.

src/Core/Context.idr
gallais: It is a bit annoying when rebuilding the compiler but I think it's worth it.

src/TTImp/ProcessBuiltin.idr
```diff
@@ -129,45 +185,81 @@ checkCons c cons ty fc = case !(foldr checkCon (pure (Nothing, Nothing)) cons) o
...
-    {auto u : Ref UST UState} ->
+    Ref Ctxt Defs =>
     NestedNames vars -> Env Term vars -> FC -> BuiltinType -> Name -> Core ()
 processBuiltin nest env fc type name = do
     ds <- get Ctxt
     case type of
         BuiltinNatural => processBuiltinNatural ds fc name
-        _ => throw $ InternalError $ "%builtin " ++ show type ++ " not yet implemented."
+        NaturalToInteger => processNatToInteger ds fc name
+        IntegerToNatural => throw $ InternalError "%builtin NaturalToInteger not yet implemented."
```
gallais, line 264:
```suggestion
        IntegerToNatural => throw $ InternalError "%builtin IntegerToNatural not yet implemented."
```

Issue #1367, opened 2021-05-01
Documentation has spacing issues in Chrome


alebahn:
# Steps to Reproduce
When viewing generated documentation at https://www.idris-lang.org/docs/idris2/current/base_docs/docs/Data.Nat.html, I noticed that the spacing is inconsistent with some spaces between functions and arguments completely missing
![image](https://user-images.githubusercontent.com/232895/116795047-b21d9980-aa97-11eb-9e2a-9b826d4653ac.png)

I am using Google Chrome Version 83.0.4103.106 (Official Build) (64-bit) on Arch Linux

# Expected Behavior
The indentation between the function and the type should be consistent. There should be spaces between a function and it's arguments (e. g. CmpNat x x)

# Observed Behavior
The spacing is inconsistent and there are missing spaces (e. g. CmpNatxx)

PR #1368, opened 2021-05-02
Some tiny HTML doc fixes (including a fix for #1367)

cypheon:
This includes fixes for the layout (weird indentation and missing spaces) in WebKit based browsers (e.g. Chrome mentioned in #1367), and a small fix to correctly escape operator names and text included in docstrings.


gallais:
Thanks for the quick fix!

PR #1369, opened 2021-05-03
[ fix #1366 ] NType is not an NTCon

### LINKED ISSUE #1366: INTERNAL ERROR: Called `getCons` on something that is not a Type constructor
Today I started getting a new kind of error when checking the docs of [elab-util](https://github.com/stefan-hoeck/idris2-elab-util) against a recent version of Idris2. I tried to break this down to a smaller example, but so far to no avail. Anyway, when I typecheck the docs of the package mentioned above (`idris2 --type-check elab-util-docs.ipkg`) I get the following error:

```
11/15: Building Doc.Generic3 (src/Doc/Generic3.md)
Error: While processing right hand side of implGenericDocGeneric3Crazy. INTERNAL
ERROR: Called `getCons` on something that is not a Type constructor
```

Note that the above module typechecked fine until very recently.

I'll try an break this down to something smaller tomorrow. But probably someone already knows what the culprit might be here.

gallais:


stefan-hoeck:
Thanks for the quick fix. I'll have time tonight to check this against elab-util and another project of mine where this occured. If you are interested in the results, I can report back here.

gallais:
Please do!

stefan-hoeck:
So, I tested this against several projects of mine and it works like a charm. In the `elab-util` docs I had to generate an additional `impossible` clause as described your comment in #1366. Thanks again, for fixing this so quickly.

PR #1372, opened 2021-05-03
[ fix #1370 ] use the lambda's type for eta-expansion

### LINKED ISSUE #1370: Problem with Eta expansion
I have a puzzling error trying to port https://github.com/statebox/idris-ct/blob/master/idris2/Idris/TypesAsCategory.idr to Idris2. I can't seem to get a term for the following type, Refl ought to work and is actually what is returned by Proof Search, but leads to an error, which in any case does not seem to be correct. The expected type of the term is '\x => f x = x', so eta-expansion.

# Steps to Reproduce
https://gist.github.com/cxandru/0f591ba8f0e9b3176a34fe13bee9c52c
```Idris
identity : (a : Type) -> a -> a
identity a = id

rightIdentity :
  (a, b : Type )
  -> (f : a -> b)
  -> (identity b) . f = f
rightIdentity a b f = Refl
```
# Expected Behavior
Typechecks
# Observed Behavior
Error: While processing right hand side of rightIdentity. When unifying f = f and identity b . f = f.
Mismatch between: b and a.
# Discord link
https://discord.com/channels/827106007712661524/836903399932690482/838722292040466433

gallais:
Not sure why we were not doing this from the very beginning so pinging @edwinb for a review.

### DIFF
--- a/src/Core/Unify.idr
+++ b/src/Core/Unify.idr
@@ -1228,28 +1228,6 @@ mutual
       isDelay (NDelayed _ _ _) = True
       isDelay _ = False
 
-  -- Try to get the type of the application inside the given term, to use in
-  -- eta expansion. If there's no application, return Nothing
-  getEtaType : {vars : _} ->
-               {auto c : Ref Ctxt Defs} ->
-               {auto u : Ref UST UState} ->
-               Env Term vars -> Term vars ->
-               Core (Maybe (Term vars))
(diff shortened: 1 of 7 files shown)


edwinb:
Hmm, that does just seem to be a long way of getting the type from the lambda, so I don't know why `getEtaType` was there. Although I also wonder why it got the wrong answer, but I didn't dig into the original issue.

PR #1375, opened 2021-05-04
[ fix #801 ] Use Number for up to 32 bit integers on JS backends

### LINKED ISSUE #801: JavaScript Backend: Ints are considered to be BigInts

stefan-hoeck:
Signed and unsigned integer types of up to 32 bit precision are now represented as `Number`s (instead of `BigInt`s) on Javascript backends with the following caveats:

- For `Int32` and `Bits32`, multiplication goes via `BigInt` to avoid introducing rounding errors in case an intermediary result lies beyond `Number.MAX_SAFE_INTEGER`.
- For `Bits32`, all bitwise operations go via `BigInt`, since bitops on `Number` behave correctly for 32bit signed integers only. This leads to all kinds of nastiness, introducing negative numbers whenever bit 32 is set before or after one of these operations.

For `Bits32`, this means that quite a lot operations go via `BigInt` and back. I still think it's better to use `Number` instead of `BigInt`, as interacting with the FFI is less cumbersome. I'll happily change it back to `BigInt`, if people prefer it that way.


stefan-hoeck:
As discussed on discord, I added a new set of tests for the primitive bitops in chez and node. I left the old tests for now to make sure this PR doesn't introduce new bugs.
I also fixed left and right shifts for `Int32` as there were several unexpected corner cases in how JS treats these operations.

This should now be read for review.

gallais:
One thing I'm not too clear about is how the conversion from `BigInt` to `Number` works for `Bits32`.
Isn't it the case that the max `Bits32` will overflow when cast to a signed `Number`? Or is the cast
purely a bitwise operation?

stefan-hoeck:
`Number` in JS is a 64bit IEEE double (and Idris2 `Double` maps directly to `Number`). As such, it can represent integers in the range [-(2^53-1),2^53-1] without rounding errors. Therefore, there is no risk of precision errors nor overflows when going from a properly truncated 32bit unsigned `BigInt` back to `Number`. It is during multiplication of 32bit signed or unsigned integers that we risk leaving the safe range, that's why this operation goes via `BigInt` for both 32bit types.

On top of this floating point `Number` representation, JS implements bitwise operations, but they behave predictably only up to 32bits, and in that case, they use a twos complement representation, that's why 32bit signed integers can use the `Number` bitops directly, while 32bit unsigned integers behave rather erratically.

See also [this link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) about the safe integer range of `Number`.

gallais:
Ah, ok. I thought you were limiting to `< 32` because it was 32 bits number. Ok let's go then!

PR #1381, opened 2021-05-05
[ docs ] Improve HTML Output

cypheon:
Data constructor docstrings where not wrapped in a `<dd>` tag, and therefore
where not indented correctly.

As a relict of the REPL output, several `<br>` tags where introduced,
where they are not needed or even permitted. This led to some spacing
issues (sometimes the docstring was closer to the next term than to the
one above that it actually described).

To counter the removed forced newlines, some extra margin is added below
each declaration.

As a side-effect, this also makes the W3 "Nu Html Checker" happy (at least for _some_ pages ;)

### DIFF
--- a/src/Idris/DocString.idr
+++ b/src/Idris/DocString.idr
@@ -145,18 +145,19 @@ getDocsForName fc n
       let root = nameRoot n in
       if isOpName n then parens (pretty root) else pretty root
 
-    getDConDoc : Name -> Core (List (Doc IdrisDocAnn))
+    getDConDoc : Name -> Core (Doc IdrisDocAnn)
     getDConDoc con
         = do defs <- get Ctxt
              Just def <- lookupCtxtExact con (gamma defs)
-                  | Nothing => pure []
+                  | Nothing => pure Empty
              syn <- get Syn
              let [(n, str)] = lookupName con (docstrings syn)
-                  | _ => pure []
+                  | _ => pure Empty
              ty <- resugar [] =<< normaliseHoles defs [] (type def)
-             pure $ pure $ vcat $
-               annotate (Decl con) (hsep [dCon (prettyName n), colon, prettyTerm ty])
-               :: reflowDoc str
+             pure $ vcat
+               [ annotate (Decl con) (hsep [dCon (prettyName n), colon, prettyTerm ty])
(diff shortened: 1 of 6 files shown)


src/Idris/DocString.idr
```diff
@@ -242,7 +243,7 @@ getDocsForName fc n
                TCon _ _ _ _ _ _ cons _
                    => do let tot = [showTotal n (totality d)]
                          cdocs <- traverse (getDConDoc <=< toFullNames) cons
-                         let cdoc = case concat cdocs of
+                         let cdoc = case cdocs of
```
gallais, line 246: This used to collapse a list of empties to an empty list which we are not doing anymore?

cypheon: I noticed that `getDConDoc` was always either returning an empty list or a singleton, so I changed that to just return a Doc (and `Empty` in case no documentation could be found), so no `concat` should be necessary anymore.
I'm not sure if `getDConDoc` can ever return `Empty`, since we are passing in a fully qualified Name for a data constructor which is known to exist, so the lookup should never fail.

cypheon: For a type with no data constructors (tested with `Void`) this works as expected and does not show the "Constructor(s):" header

gallais: What if we have a single constructor? Do we get everything on a single line?

cypheon: Ok, I found several Names for which `getDConDoc` returns empty (e.g. `Control.ANSI.MkDString`). I'll investigate an update this PR

cypheon: Two insights:
a) For single-constructor types, everything is shown on one line
b) For records, no constructor is shown, ever (but this was the case before this change, so it seems I didn't break anything that was working before).

It seems it's not possible to add a docstring to the constructor of a record, is that correct?

gallais: > It seems it's not possible to add a docstring to the constructor of a record, is that correct?

Looking at the parser, that looks correct.
This is definitely something we could change.

cypheon: I just added documentation output for record constructors, and added test cases for single-constructor types and records

PR #1383, opened 2021-05-06
[ close #1384 ] built-in Snoc-lists [< 1, 2, 3 ]

### LINKED ISSUE #1384: Built-in SnocList notation
Following up on a discussion on the Idris Community Discord server:

It's often useful to have lists that are nested to the _left_ instead of to the right:
```idris
||| Snoc lists.
public export
data SnocList a =
  ||| Empty snoc-list
  Lin

  | ||| A non-empty snoc-list, consisting of the rest of the snoc-list and the final element.
  (:<) (SnocList a) a

%name SnocList sx, sy, sz
infixl 7 :<
```
with the list-like notation expanding in the opposite direction:
```idris
A,B : SnocList Nat
A = [< 1, 2, 3 ] 
B = Lin :< 1 :< 2 :< 3

Same : A = B
Same = Refl
```

Just like the list literals can be overloaded by defining `Nil` and `(::)`, this proposal is about overloading an analogous pair of names. Since we use lists all over the place, whichever proposal we converge on has to be non-breaking, at least initially. 

While a `SnocList` contains the same information as the corresponding reversed list, anecdotal experience shows that when working with both lists and their reverse in the same program is confusing, and having two different types that represent the same collection of values helps dealing with this confusion.

Pitfalls:

* I think with this change we should put a `SnocList` datatype in the Prelude.
* Even though we're currently using `Empty` in the stdlib (`base/Data.List.Views.SnocList`), it's a fairly common datatype constructor, so we sticking it into Prelude might break existing code.

Since syntactic changes always cause a lot of discussion, I propose that suggestions for alternative concrete syntax / architecture should be accompanied by a (draft) pull request. We will list them here:

1. [Here](https://github.com/idris-lang/Idris2/pull/1383) is one proposal with `SnocList, Lin, :<, [< 1, 2, 3 ]` that also provides  `[> 1, 2, 3]` alongside `[1,2,3]` for `List` to help with readability.

ohad:
Based on discussions on the developers Discord server and library code
by @jfdm. 

See [feature proposal](https://github.com/idris-lang/Idris2/issues/1384), and [test/idris2/basic060](https://github.com/idris-lang/Idris2/blob/783a7c5c49e0990564ab6bd346ac9284bb482063/tests/idris2/basic060/Snoc.idr) summarises this concrete proposal:

```idris
import Data.SnocList

A, B, C, D, E : SnocList Nat
A = [<]

B = Lin :< 1 :< 2 :< 3

C = [< 1, 2, 3]

D = [< 1, 2, 3] :< 4

E = [< 1, 2, 3] <>< [4, 5, 6]     -- [< 1, 2, 3, 4, 5, 6]

X : List Nat
X = [< 1, 2, 3] <>> [> 4, 5, 6]   -- [  1, 2, 3, 4, 5, 6]
```


Russoul:
Fish and chips but no `Lin`? :sweat_smile:

ohad:
I used `Empty` becase `base/Data.List.Views.SnocList` uses `Empty`, but apart from consistency with this part of the library, I don't have a preference. Feel free to push a downstream commit to rename it if you feel strongly about `Lin`.

Russoul:
No, it's good :) I like it anyway :+1:

ziman:
> [...] but I also like meaningful names without puns.

Oh, so true! :)

ziman:
Could we have `Nil` for snoc-lists, too? That's only one step of disambiguation (compared to something like reusing `::`), and it should be always quite clear which you mean (?). Also, if you have an empty list, perhaps it matters less in which direction it goes.

Russoul:
Oh, `Empty` also comes up in contrib's SortedMap and in specialised maps in Idris2 src, like IntMap and NameMap

ohad:
I worry about using the same symbol (and name) for two data types that are almost certainly always going to be used in the same context. Seems like a disambiguation disaster waiting to happen.

libs/base/Data/SnocList.idr
```diff
@@ -15,13 +15,13 @@ sx <>< (x :: xs) = sx :< x <>< xs
 ||| 'chips': Action of snoc-lists on lists
 public export
 (<>>) : SnocList a -> List a -> List a
-Empty     <>> xs = xs
+Lin       <>> xs = xs
 (sx :< x) <>> xs = sx <>> x :: xs
 
 ||| Reverse the snoc-list and make a list.
 public export
 toList : SnocList type -> List type
```
gallais, line 23: I don't think we should have local functions called `toList` because that clashes with the
`Foldable` definition. I have had issues with this in the past in `Data.List1` (hence the
function from `List1 a` to `List a` being called `forget`).

ohad: I'm happy to drop it, it was in @jfdm 's library, but since we can recover it as `toList sx = sx <>> []`, we can just drop it.

gallais: `sx <>> []` reverses the list whereas  ̀toList` preserves the ordering.

ohad: Oh, do you think we should swap `cast` and `asList`?

My principle was that `cast` should maintain the logical structure.

tests/ideMode/ideMode005/expectedH
```diff
@@ -0,0 +1 @@
+000018(:protocol-version 2 0)
```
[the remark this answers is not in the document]
ohad: Yes, I rounded 32 to 30 in my head. Silly me.

gallais:
I was just about to merge but should we get an entry in the CHANGELOG and
some doc in `docs/source/tutorial/typesfuns.rst` first?

libs/prelude/Prelude/Ops.idr
```diff
@@ -11,6 +11,8 @@ infixr 4 ||
 
 -- List and String operators
 infixr 7 ::, ++
+infixl 7 :<, <><
+infixr 6 <>>
```
gallais, line 15: Isn't it a bit strange to declare fish & ships in `prelude` even though they're only defined in `base`?

src/Idris/Parser.idr
```diff
@@ -407,6 +407,26 @@ mutual
...
 
+  snocListExpr : FileName -> WithBounds () -> IndentInfo -> Rule PTerm
+  snocListExpr fname s indents
+      = {- TODO: comprehension -}
+        do mHeadTail <- option Nothing $ do
+             hd <- many $ do x <- expr pdef fname indents
+                             b <- bounds (symbol ",")
+                             pure (x <$ b)
+             tl <- expr pdef fname indents
+             pure $ Just (hd, tl)
```
gallais, line 418:
```suggestion
        do mHeadTail <- optional $ do
             hd <- many $ do x <- expr pdef fname indents
                             b <- bounds (symbol ",")
                             pure (x <$ b)
             tl <- expr pdef fname indents
             pure (hd, tl)
```

Issue #1382, opened 2021-05-05
Idris can't resolve interfaces in particular cases


Russoul:
# Steps to Reproduce
Load the snippet:
```Idris
----------- Real case example ----------
data ListM : (Type -> Type) -> (Type -> Type) where
  Nil : ListM m a
  (::) : a -> m (ListM m a) -> ListM m a

ListT : (Type -> Type) -> (Type -> Type)
ListT m a = m (ListM m a)

-- Works
[Named] Monad m => Semigroup (ListT m a) where
  xs <+> ys = do
    case !xs of
      [] => ys
      x :: xs => pure (x :: (<+>) @{Named} xs ys)

-- m (ListM m a) ~ ListT m a
-- Doesn't work
Monad m => Semigroup (ListT m a) where
  xs <+> ys = do
    case !xs of
      [] => ys
      x :: xs => pure (x :: (<+>) xs ys)

-- Error(s) building file XXX.idr: While processing right hand side of <+>. Can't find an implementation for Semigroup (m (ListM m a)).

-- XXX.idr:22:29--22:40
--  18 | Monad m => Semigroup (ListT m a) where
--  19 |   xs <+> ys = do
--  20 |     case !xs of
--  21 |       [] => ys
--  22 |       x :: xs => pure (x :: (<+>) xs ys)
--                                   ^^^^^^^^^^^

--------------------------------------

----- Contrived simplified example -----

interface Dummy (0 m : Type) where

Functor f => Dummy (f (f a)) where

[NamedDummy] Functor f => Dummy (f (f a)) where

ok : Functor f => Dummy (f (f a))
ok = NamedDummy

fail : Functor f => Dummy (f (f a))
fail = %search

-- While processing right hand side of fail. Can't find an implementation for Dummy (f (f a)).

-- XXX.idr:48:8--48:15
--  44 | ok : Functor f => Dummy (f (f a))
--  45 | ok = NamedDummy
--  46 |
--  47 | fail : Functor f => Dummy (f (f a))
--  48 | fail = %search
--              ^^^^^^^

--------------------------------------

```
# Expected Behavior
No type-checking errors
# Observed Behavior
Two type-checking errors as shown in the snippet.

Russoul:
After giving it a bit more thought, the problem likely boils down to
```Idris
func : {0 f : Type -> Type} -> f (f a) -> a

unify : {0 f : Type -> Type} -> (x : f (f a)) -> a
unify x = func x

[...]

PR #1386, opened 2021-05-06
[prelude] Add explicit constructor to every interface

Russoul:
As was shown in #1382 Idris has a hard time figuring out interface constraints in particular cases. The problem is exacerbated when you want another interface to depend on an interface like in #1382 - it appears that this simply can't be accomplished at the moment. 
Because Idris won't be able to tell that the sub-interface constraint is satisfied. Even `using` doesn't make any difference. 

The only solution I am aware of is to explicitly pass the instance of the parent interface to the sub-interface constructor.

This PR exposes constructor of every interface in Prelude.


stefan-hoeck:
Wow, I didn't even know one could do that (giving explicit constructors to interfaces, I mean). Is this documented somewhere? Otherwise I think it should be.

Russoul:
Yeah, it's one of those hidden features of Idris. Almost sure it isn't documented anywhere.

andrevidela:
Could we document it somewhere while we're at it? maybe here? https://github.com/idris-lang/Idris2/blob/master/docs/source/tutorial/interfaces.rst

andrevidela:
Looks good to me!

PR #1388, opened 2021-05-07
[ refactoring ] Tiny changes following up the idris-lang/Idris2#830

buzden:
Some zeroes in signatures, one simpler implementation and formatting.

### DIFF
--- a/libs/base/Data/Fin.idr
+++ b/libs/base/Data/Fin.idr
@@ -229,7 +229,7 @@ namespace Equality
 
   ||| The actual proof used by cast is irrelevant
   export
-  congCast : {n, q : Nat} -> {k : Fin m} -> {l : Fin p} ->
+  congCast : {0 n, q : Nat} -> {k : Fin m} -> {l : Fin p} ->
              {0 eq1 : m = n} -> {0 eq2 : p = q} ->
              k ~~~ l ->
              cast eq1 k ~~~ cast eq2 l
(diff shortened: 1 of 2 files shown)


libs/base/Data/Fin.idr
```diff
@@ -229,17 +229,16 @@ namespace Equality
...
              cast eq1 k ~~~ cast eq2 l
   congCast eq = transitive (castEq _) (transitive eq $ symmetric $ castEq _)
 
   ||| Last is congruent wrt index equality
   export
-  congLast : {m, n : Nat} -> (0 _ : m = n) -> last {n=m} ~~~ last {n}
-  congLast {m = Z} {n = Z} eq = reflexive
-  congLast {m = S _} {n = S _} eq = FS (congLast (succInjective _ _ eq))
+  congLast : {m : Nat} -> (0 _ : m = n) -> last {n=m} ~~~ last {n}
+  congLast Refl = reflexive
```
gallais, line 241: My idea here was that `congLast` would reduce to canonical constructors whenever the
nats are constructor-headed without needing to pattern match on the equality.
Maybe not exactly necessary. :thinking:

buzden: Does this make sence having the fact that `congLast` is not `public export`? I changed the implementation without doubt only because it's not exposed outside.

gallais: Good point. Let's go with the updated version.

PR #1390, opened 2021-05-08
Some optimisations on case compilation

edwinb:
Firstly, the 'case of case' transformation, which lifts out case blocks from the scrutinee of another case block. This has a small positive effect on run time,
Secondly, compiling lists as scheme lists, which has a bigger positive effect on runtime.

This second one will require codegen authors to update their code for the new `ConInfo` flag, which gives information about the constructor (whether it's a nil or cons or something else) and can be used to compile a constructor to the list type native to whatever the code generator uses. Or, you can safely ignore it.

Currently, we only compile `Prelude.List` specially, but this should soon be extended to all list shaped things. They're easy enough to spot, I just don't have the time right now. (EDIT: I did it anyway and it shaved off a couple more seconds)

These two changes mean that, on my machine, the Idris 2 code typechecks about 10 seconds quicker - the best time I got was about 1m20.

This has required updating the bootstrap code, because we can't currently have a separate support.ss/rkt for the bootstrap and normal builds. We probably ought to fix that, but... one thing at a time!


stefan-hoeck:
Oh, having updated the bootstrap code, does this mean the new bootstrap compiler will know about the new `IntX` types already?

edwinb:
It will, but the bootstrap compiler is only for convenience and new installs. We still have to worry about the previous version.

PR #1391, opened 2021-05-09
Add stdlibs to bootstrap compiler

stefan-hoeck:
This change gives the bootstrap compiler its own set of stdlibs and adjusts the CI to no longer build the libs from older compilers. This allows us to support new functionality from the compiler in the libs immediately without having to wait for a new minor release. Of course, with each update of the bootstrap compiler, its stdlibs need to be updated as well.

Right now, `contrib` is also in `bootstrap/libs`. This is due to a transitive dependency, as `network` still depends on `contrib`. However, There is only one module in `network` depending on a module from `contrib`: `Control.Linear.Network`. Since this is not needed in the compiler sources, I suggest to move `Control.Linear.Network` to `contrib`, and thus get rid of a dependency of the compiler on `contrib` once and for all.

Before I do that (as part of this PR), I'd like to get some opinions, whether people think this is the way to go.

Edit: To demonstrate that this actually allows us to add bleeding edge stuff, I included interface implementations for the new Int types here. Before this PR gets accepted, I either remove those, or rebase in such a way, that the two modifications can be cleanly merged as distinct commits.


Russoul:
I think that instead of copying over the sources we should ship their compiled `ttc` artefacts instead, if that's possible.
This would be more on par with how the bootstrapping process has been done up until now, IMO.

stefan-hoeck:
I considered this as well. Two reasons that made me decide against using the ttc artifacts: 1) Their size (on my machine, the ttc of the prelude alone are 4.2 M of data, 2) in the rare case it might be necessary to fix an earlier bug in the bootstrap libs. This will be much easiuer (I guess), if they are still present as source files.

However, I don't have a strong opinion on this.

stefan-hoeck:
As described above, I moved `Control.Linear.Network` to contrib. Therefore `contrib` now depends on `network` instead of the other way round. This is now ready for reviewing.

benhormann:
How about gutting `Idris.Driver` to create `Idris.Bootstrap`, so that it isn't the whole REPL?

It could be as minimal as a function that takes a cg, gets relevant paths from env / cmdline, and builds the ipkg.
Perhaps create two mains (or a template) so as to hardcode the cg, to reduce the bootstrap size even further?

PoC:
```sh
(cd src
  sed 's/Driver/Boot&/; /import.*IDEMode/d; s/if not ideSocket/do coreLift $ exitWith (ExitFailure 42)/; /then do/,/file file/d; /replIDE/d' Idris/Driver.idr > Idris/BootDriver.idr
  sed -E 's/Driver|Main/Boot&/' Idris/Main.idr > Idris/BootMain.idr
  idris2 Idris/BootMain.idr --build-dir ../bootstrap/build --output-dir ../bootstrap -o idris2 --verbose
  cd ../libs/prelude
  ../../bootstrap/idris2 --build prelude.ipkg --build-dir ../../bootstrap/build/prelude
  #todo: base.ipkg
)
```

stefan-hoeck:
While I agree that the bootstrap compiler could be optimized further, in this PR I just wanted to decouple the libraries bootstrap uses from the ones the current compiler uses in order to speed up adding new features from the compiler to the libs. However, I'm not even sure, people in charge agree with this decision. So let's see what @edwinb and @gallais think before we move this any further.

stefan-hoeck:
Rebased to resolve conflicts.

gallais:
One inconvenient thing I can see happening is us:
* introducing a non-backwards compatible change in userland libs
* releasing a minor version
* the new dev version does not compile with the libs of the released minor version

edwinb:
I'm not going to change the restriction that Idris has to build from the previous version, sorry. If someone has the previous version installed, they should be able to download the new tarball and install it without problems.

I know people are always keen to using the latest new features when working on the compiler, but it's best to resist. We do try to release fairly often (even though it's probably not as often as I'd like at the minute) so it doesn't take that long. New features are also the ones most likely to have problems, and if we end up relying on them, we could get ourselves into a mess.

This is awkward for primitives that are expected to compute at compile time, which end up requiring two rounds of releases to be fully implemented due to the need for them to work at the type level. That's a shame, but fortunately we don't need to do it often.

stefan-hoeck:
> I'm not going to change the restriction that Idris has to build from the previous version, sorry. 

The content / goal of this PR was not to change this restriction, and CI still would make sure that this restriction holds. However, I agree about the stability of new features and us being careful not to push ahead too quickly. So, for me it's perfectly fine to close this PR. Probably an alternative as suggested by gallais (having an `experimental` lib that's only built by the dev compiler) would be acceptable?

stefan-hoeck:
So, I moved implementations for the new primitives to an [external library](https://github.com/stefan-hoeck/idris2-experimental). I'll happily move stuff from there to an `experimental` module in this repository as suggested by gallais, if people like to have this stuff closer to the official Idris2 repo. Closing this PR now.

PR #1393, opened 2021-05-10
Choose foreign string based on priority

melted:
Instead of taking the first applicable string, the backends submit an
ordered list of preferred prefixes and get the first match.

### DIFF
--- a/src/Compiler/Common.idr
+++ b/src/Compiler/Common.idr
@@ -342,23 +342,20 @@ exists f
              | Left err => pure False
          closeFile ok
          pure True
-
--- Parse a calling convention into a backend/target for the call, and
--- a comma separated list of any other location data.
+-- Select the most preferred target from an ordered list of choices and
+-- parse the calling convention into a backend/target for the call, and
+-- a comma separated list of any other location data. For example
+-- the chez backend would supply ["scheme,chez", "scheme", "C"]. For a function with
+-- more than one string, a string with "scheme" would be preferred over one
+-- with "C" and "scheme,chez" would be preferred to both.
 -- e.g. "scheme:display" - call the scheme function 'display'
 --      "C:puts,libc,stdio.h" - call the C function 'puts' which is in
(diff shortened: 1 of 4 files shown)


src/Compiler/Common.idr
```diff
@@ -342,30 +342,38 @@ exists f
...
 export
-parseCC : String -> Maybe (String, List String)
-parseCC "" = Nothing
-parseCC str
-    = case span (/= ':') str of
-           (target, "") => Just (trim target, [])
-           (target, opts) => Just (trim target,
-                                   map trim (getOpts
-                                       (assert_total (strTail opts))))
+parseCC : List String -> List String -> Maybe (String, List String)
```
gallais, line 356: Can we document the meanings of the two arguments that happen to have the same type?

src/Compiler/Common.idr
```diff
@@ -342,30 +342,38 @@ exists f
...
+    hasTarget target str = case span (/= ':') str of
+                            (targetSpec, _) => targetSpec == target
+    findTarget : String -> List String -> Maybe (String, List String)
+    findTarget target [] = Nothing
+    findTarget target (s::xs) = if hasTarget target s
+                                  then case span (/= ':') s of
+                                        (t, "") => Just (trim t, [])
+                                        (t, opts) => Just (trim t, map trim (getOpts
+                                                                  (assert_total (strTail opts))))
+                                  else findTarget target xs
```
gallais, line 376:
```suggestion
    findTarget : String -> List String -> Maybe (String, List String)
    findTarget target [] = Nothing
    findTarget target (s::xs)
      = do let (t, rest) = mapFst trim (span (/= ':') s)
              let True = t == target
                     | False => findTarget target xs
              pure (Just (t, map trim (getOpts (assert_total (strTail opts)))))
```

gallais (commented):
Argh got merged before I could submit the review

Issue #1295, opened 2021-04-12
Add TAB autocompletion for idris2 cli arguments


gallais:
As the title says. If I double TAB after writing `firefox -`, I get:

```shell
--allow-downgrade        --headless               --new-window             --safe-mode
--browser                --help                   --no-remote              --screenshot
--debug                  --jsconsole              -P                       --search
--debugger               --jsdebugger             --preferences            --setDefaultBrowser
--debugger-args          --kiosk                  --private-window         --start-debugger-server
--devtools               --migration              --profile                --sync
--display=               --MOZ_LOG=               --ProfileManager         --version
--first-startup          --MOZ_LOG_FILE=          --recording              --wait-for-jsdebugger
--full-version           --new-instance           --recording-output       --window-size
--g-fatal-warnings       --new-tab                --remote-debugging-port  
```

It'd be nice to get the same for `idris2 -`.

jfdm:
Maybe we should be more specific and state: Let Idris2 generate TAB autocompletion instructions. Most shells require these for CLI. IIRC one of the more popular CLI libraries for Haskell has a `--bash-completion` flag to generate the instructions.

For reference, see the instructions for the Stack tool: 

https://docs.haskellstack.org/en/stable/shell_autocompletion/

cypheon:
With zsh, you can already get pretty far with just `compdef _gnu_generic idris2`, which parses the `--help` output and provides completion (and description) for all documented arguments.

stefan-hoeck:
I started implementing this in a [separate project](https://github.com/stefan-hoeck/idris2-complete). You can already give it a try as described in the project's README. Once I'm happy with it, I'll add the functionality to idris2 itself.

Right now, `bash` and `zsh` are supported as these are the shells I use myself.

stefan-hoeck:
OK, this is ready for experimenting. The following things are being autocompleted (depending on context):

- command line options
- available codegens
- installed packages
- log levels
- directories

If people think, this is the way to go, I'll try and add this functionality to Idris2 itself.

PR #1397, opened 2021-05-11
[ new ] Tab completion

stefan-hoeck:
As requested in #1295. The idea is very similar to what [stack does](https://docs.haskellstack.org/en/stable/shell_autocompletion/).

Edit: Somebody please fix the markdown in `INSTALL.md`. Lessons learned: Don't touch anything, or the linter starts drooling :-)
Edit2: "Somebody" turned out to be me...


gallais:
There seems to be a bug when you only have `idris2 `:

```shell
$ idris2 
Display all 260 possibilities? (y or n)
-?                        chez                      directory.                GAMBIT_GSC                import                    mode                      port                      string
(0                        chez)                     Disables                  GAMBIT_GSC_BACKEND        in                        mode;                     --prefix                  supported
0.3.0-656988f1c           Chez                      display                   GAMBIT_GSI                
(...)
```

As soon as you double tab on `idris2 -` instead you only get the options.

stefan-hoeck:
This should be fixed not. TAB completion is not super fast, since Idris2 startup time is around 0.5 s (on my system) even when invoking `idris2 --version`, but we will see if this is an issue for the fast typers (haha) in the community. There's still the alternative of having a much smaller standalone program, but this will not get the full Idris2 environment (for instance to list additional codegens) and maintenance would be harder.

stefan-hoeck:
Rebased and using `knownTopics` now for log completion. But now a question: Is it correct, that the loglevel is specified using the following syntax:

```sh
> idris2 --log unify.binder:5
```

Should we therefore support the completion of the `:5` part as well?

gallais:
Note that any prefix "m.n" of an existing "m.n.p.q" topic is a valid topic too.
Not sure we handle that properly at the moment.

> Should we therefore support the completion of the :5 part as well?

I think it'd be annoying to explicitly list the levels available for each topic.

stefan-hoeck:
Should work now.

PR #1400, opened 2021-05-11
Some compilation/code generation improvements

edwinb:
Mostly this is about better inlining of case blocks, but also some tweaks for compiling records and option types,giving a small improvement to generated code overall.


edwinb:
A curious thing - one of the node tests has a stack overflow exception, but no failure. This might have been tripped by this patch, but I suspect it's a thing that just might happen rather sooner on node than scheme based back ends. Can someone who knows about node take a look please?

stefan-hoeck:
Yes, node is much more susceptible to stack overflows. I'll have a look and report back here.

stefan-hoeck:
So, it is the call to `mapMaybe` in `tests/node/newints/IntOps.idr` that's causing the stack overflow. With a recent master, `mapMaybe` is properly tailcal optimized, but with `tweak-cg` this is no longer the case. Should I dig some more, or is this already enough for you to know where the problem lies?

edwinb:
Oh, that's interesting. I'll dig into this a bit to see if the case inlining needs to be a bit more careful although I don't have time right now. It's odd that it would change the tail call properties but perhaps I'll understand once I take a look at the generated case trees.

stefan-hoeck:
Just a quick heads-up: I'm slowly ploughing (is that a word?) through the ES backend code. Wanted to understand how it works anyway... @edwinb : how long until your patience runs out and you pull the plug on the JS backends? This might take a couple o days...

edwinb:
@stefan-hoeck I'm not going to pull the plug that quickly, I'm busy with some other things in the next couple of days anyway. Looking at the test, I'm a bit surprised it ever worked on node given the size of the list, but maybe there was something about the non-inlined version that meant it didn't quite fill up the stack. My currently inclination is that we should fix the test and merge this, but we should still consider whether the JS back end should be maintained separately.

stefan-hoeck:
OK, so here's the issue: With nested case expressions, only the outer-most level got properly tailcall optimized in the JS backends. I have a fix for this. We could either disable the failing node test for this PR and I submit the fix in a separate PR, or you replace the content of `Compiler.ES.Imperative.idr` with the [code here](https://gist.github.com/stefan-hoeck/1712d140819784d323a9e58af7117b0d).

Please note: The fix is quick and dirty. I plan to open an new issue about the state of the JS backend soonish, since the whole code requires some cleanup and documentation.

buzden:
> or you replace the content of `Compiler.ES.Imperative.idr` with the [code here](https://gist.github.com/stefan-hoeck/1712d140819784d323a9e58af7117b0d)

You can also make yourself a pull request to Edwin's `tweak-cg` branch, i.e. to have a PR to a PR.

edwinb:
Thanks! I'm trying it out now... fingers crossed!

edwinb:
It certainly worked for me locally, let's see why CI says.

edwinb:
Sorted! Thanks for your help @stefan-hoeck

Issue #1124, opened 2021-02-25
Statically enforce that all the logging options are documented


gallais:
At the moment we have quite extensive (and everyday expanding) logging options
to get Idris to print more debug output. However these are not documented anywhere
and the best way to discover them is by `grep`ping the source code. This is less
than ideal.

In [`Idris.Env`](https://github.com/idris-lang/Idris2/blob/master/src/Idris/Env.idr#L46) we use a nifty trick to ensure that all of the environment variables we
attempt to observe in the codebase are thoroughly documented.

It would be nice to have the same enforcement for logging options.

NB: there are 450+ uses of `log` in the codebase so it'll take a bit of time to
collect all of the existing topics. The documentation part can be left for a future
PR: having an exhaustive list would already be pretty useful!

AdamHarries:
I think I could tackle this. 

With the environment variables we don't need to deal with hierarchies of names, but as I understand it with logs we need to be able to deal with them (e.g. `elab`, `elab.with` etc). Do we want users to have to write out every possible "path" of log contexts/names, or do we want a smarter way of doing it (e.g. checking membership of a tree structure of some sort).

gallais:
There are two aspects to this:

* Declaring all of the topics of interest & enforcing the log functions
  can only be called with one such topic
* Generating the documentation

I think you are right that it would be good to print the whole thing not as a list
but as a hierarchical structure. We could do so by reusing (parts of) our existing
functions in [`Core.Options.Log`](https://github.com/idris-lang/Idris2/blob/master/src/Core/Options/Log.idr) to generate a trie of documentation strings & printing it nicely.

It's probably worth writing this printer generically for any [`Libraries.Data.StringTrie`](https://github.com/idris-lang/Idris2/blob/master/src/Libraries/Data/StringTrie.idr)

This trie could also be used to efficiently perform the check needed in the first
part of the problem.

AdamHarries:
Declaring topics of interest is what I was specifically referring to, but generating documentation is something I'd not considered, and definitely something that should go into the design. 

To clarify, what I was asking was whether "declaring" a topic of interest should be done in a "list like" way, i.e. 

```
topics : List String 
topics = [
    "core", 
    "core.elab",
    "syntax"
]
-- etc
```

or if it should be done in a "tree like" way, i.e. 

```
-- or equivalent structure
record LogTree where 
    constructor MkTree
    name : String 
    children : List LogTree

[...]

PR #1402, opened 2021-05-12
[ towards #1124 ] Statically enforced logging

stefan-hoeck:
This adds a type level check to most log calls to make sure that only known logging topics are used. Exceptions are allowed when topics are coming from client code, for instance during elab reflection.

The list of logging topics is now provably exhaustive (see also discussion in #1397), but the topics are not yet documented, nor is there any functionality for pretty printing all known topics.

### DIFF
--- a/src/Core/Context/Log.idr
+++ b/src/Core/Context/Log.idr
@@ -13,7 +13,9 @@ import System.Clock
 export
 logTerm : {vars : _} ->
           {auto c : Ref Ctxt Defs} ->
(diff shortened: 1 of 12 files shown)


stefan-hoeck:
As an alternative, I tried also using an enum type for known log topics. The effect on build times and performance was negligible. You can find the corresponding branch [here](https://github.com/stefan-hoeck/Idris2/tree/safe_log_enum).

PR #1406, opened 2021-05-13
Some small changes to get Idris 2 working on FreeBSD

cypheon:
Just a few small changes are required to enable building and installing on FreeBSD. The code then passes all tests on my machine.

For successful bootstrapping, the bootstrap code would need to be updated as well (built by the new compiler version). I think this can be done in a later PR (or just the next release).

The most "invasive" thing of this PR is the special-casing of "libc" FFI for the Racket backend, as it changes the codegen on all OSs. But I'm quite convinced, that this is the correct thing to do, as stated by the Racket documentation ([The Racket Foreign Interface > Loading Foreign Libraries](https://docs.racket-lang.org/foreign/Loading_Foreign_Libraries.html#%28def._%28%28lib._ffi%2Funsafe..rkt%29._ffi-lib%29%29)) (emphasis mine):

> If path is #f, then the resulting foreign-library value represents all libraries loaded in the current process, including libraries previously opened with ffi-lib. **In particular, use #f to access C-level functionality exported by the run-time system** (as described in Inside: Racket C API). The version argument is ignored when path is #f.

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -72,6 +72,7 @@ schHeader chez libs
     "; @generated\n" ++
     "(import (chezscheme))\n" ++
     "(case (machine-type)\n" ++
+    "  [(i3fb ti3fb a6fb ta6fb) #f]\n" ++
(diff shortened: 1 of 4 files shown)


melted:
That libc thing is probably useful for windows too.

PR #1407, opened 2021-05-13
[ fix #621 ] add warnings for shadowed global definition

### LINKED ISSUE #621: Cannot prove equality between two definitionally equal terms
# Steps to Reproduce

Load the following program with Idris2:
```idris
a1 : Char
a1 = 'a'

a2 : Char
a2 = 'a'

whyNot : a1 = a2
whyNot = Refl
```
# Expected Behavior

The terms `a1` and `a2` are definitionally equal, so I expect to be able to prove they are equal.

# Observed Behavior
```
1/1: Building crash (crash.idr)
crash.idr:8:10--8:14:While processing right hand side of whyNot at crash.idr:8:1--10:1:
When unifying a2 = a2 and a1 = a2
Mismatch between:
	a2
and
	a1
at:
8	whyNot = Refl
	         ^^^^
```

gallais:


Russoul:
Is this infamous case handled properly now?
```
namespace Hoo

   f : Either a b -> Either b a
   f (Left a) = Right a
   f (Right b) = Left b

   -- bug: f should *not* have been generalised!
   natural : (xs : Either a b) -> ((f . f) . map g) xs === (map g . (f . f)) xs
   natural = ?natural_1
```
i.e. it should emit a warning that `f` is shadowed in the signature of `natural`

gallais:
Uh, thanks for the test case. The nested namespace somehow prevents the warning from being raised.

Russoul:
It's yours (the test case). It's a shame that I can't recall what was the reason behind it. (When I had a go at the `generalise` branch of yours)

gallais:
I see. The problem is that we process the file in small units in `processDecls`.
For each set of mutually defined things we desugar & then elaborate them all
together before moving on to the next small unit.

And so when we define `a1 : Char`, it is then part of the global context already
when we check `a2 : Char`, and then both are in scope when we reach #621's `whyNot`
and so during desugaring we will see that `a1` and `a2` are shadowed.

But a `namespace` block is seen as a separate unit altogether and so it is fully
desugared in one go, then fully elaborated. And so when we are desugaring `neutral`,
we have not yet added `Hoo.f` to the global scope and we do not detect it is shadowing.

We have a similar problem with mutual definitions:

```idris
mutual

  g : h === h
  g = Refl

  h : g === g
  h = Refl
```

does not raise a warning

gallais:
This refactoring fixes the issue for toplevel-ish namespaces (i.e. as long as they're
at the top level or nested under other namespaces then it's fine) but does not solve
the issue for `mutual` blocks, `parameters` block, or `namespace` ones nested under
such blocks.

For these we would need a more invasive change that is an approximation of scope checking.
I'm not sure it's worth bothering as I would hope that people using advanced features like
this would perhaps have had this warning drilled into their heads already. :thinking:

Russoul:
Sounds fair!

gallais:
We may actually be able to handle `mutual` blocks the same way without too much work.
Parameters are really getting in the way though so I'll probably leave these be.

PR #1408, opened 2021-05-13
[ docs ] modify Racket bootstrap instructions.

gh-name:
See https://github.com/idris-lang/Idris2/issues/1410


gh-name:
https://github.com/idris-lang/Idris2/issues/1410#issuecomment-842261752

PR #1409, opened 2021-05-13
[ totality ] Make `Language.Reflection` modules have `%default total`

buzden:
As soon as I have a total function that uses `Elab`'s monadic stuff, it's sad that totality checker is unsatisfied.

The original implementations could be preserved but with totality assertions, so why not to inline it all to have an equivalent implementation that satisfies the totality checker?

### DIFF
--- a/libs/base/Language/Reflection.idr
+++ b/libs/base/Language/Reflection.idr
@@ -3,6 +3,8 @@ module Language.Reflection
 import public Language.Reflection.TT
 import public Language.Reflection.TTImp
 
+%default total
+
 ||| Elaboration scripts
 ||| Where types/terms are returned, binders will have unique, if not
 ||| necessarily human readabe, names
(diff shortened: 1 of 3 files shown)


buzden:
Wow.
```idris
f <*> a = Bind f \f' => map f' a
```
works well.

But 
```idris
f <*> a = Bind f $ flip map a
```
leads to a "bad elab" error.

But
```idris
f <*> a = Bind f (<$> a)
```
still works... It looks strange.

dotpat:
hmmm…is this `<&>`?

buzden:
> hmmm…is this `<&>`?

Yeah. I mean, it does not matter whether to write `(a <&>)` or `(<$> a)`. The problem is when you use `flip` in the definition of `<*>`.

By the way, the same problem as with `flip` arises when you define
```idris
(<*>) f = Bind f . (<&>)
```

In this case you also have a "bad elaborator script" error.

PR #1411, opened 2021-05-13
[ fix ] missing log option

Z-snails:


gallais:
Ah, yes, the joy of merging non-orthogonal features. Thanks for the prompt fix!

PR #1412, opened 2021-05-13
Add builtin.IntegerToNatural to knownTopics

michaelmesser:
Fixes broken master. Check "Require branches to be up to date before merging" to prevent this from happening again.


gallais:
Cf #1411

PR #1422, opened 2021-05-16
Bugfix highlighting of `let` declarations

ohad:
Fixes @michaelmesser 's bug report:
```idris
module Test

import Data.List1
import Data.String

parseHeader : ()
parseHeader = let (_ ::: xs) = split (== ':') "test" in ()
```
which erroneously generates two annotations for the let:
```idris
-- (("FILE.idr", ((6, 18), (6, 19))), (keyword, Nothing))
--                vv
parseHeader = let (_ ::: xs) = split (== ':') "test" in ()
--                ^---------------------------------^
-- (("FILE.idr", ((6, 18), (6, 52))), (type, Just Data.List1.List1))
```


gallais (approved):
I guess we had a test case for a pattern matching do let but not a standalone let. :sweat_smile: 
I've added a test case for:
* irrefutable pattern matching let
* pattern matching let with alternatives
* `:=`-using let
* local definition

Everything looks good.

PR #1423, opened 2021-05-16
[ perf ] Move unnecessary `toFullNames` calls under a `logC`

Russoul:
I assume `toFullNames` is injective?


Russoul:
Oh, It was introduced in #1222. However, it is the only place in the unifier where we resort to `toFullNames`.
Recently @edwinb checked that name lookups appeared to be the most resource-intensive task that show up in the profiler.
@gallais, I wonder if it is possible to avoid those calls. I haven't grasped yet why those are necessary.

gallais:
As you've noticed it was a bug fix.

One thing you could do is test for equality and only perform `toFullNames` & a final
equality test if the equality test fails.

Russoul:
Oh thanks for looking into the cause of the problem again. The latest commit also seems like a viable option?

PR #1428, opened 2021-05-17
[ fix #1421 ] Use resolved names for the impossible LHS 

### LINKED ISSUE #1421: Order of Non-Overlapping Clauses in Function Defintiion Should Not Matter
# Steps to Reproduce

* Idris 2, version 0.3.0-881a5e7fb

clause_order_good.idr:
```idris
plus2 : Nat -> Nat
plus2 = S . S

plus2Injective : (x, y : Nat) -> Equal (plus2 x) (plus2 y) -> Equal x y
plus2Injective Z Z Refl = Refl
plus2Injective (S n) (S n) Refl = Refl
plus2Injective Z (S _) _ impossible
plus2Injective (S _) Z _ impossible
```

clause_order_bad.idr
```idris
plus2 : Nat -> Nat
plus2 = S . S

plus2Injective : (x, y : Nat) -> Equal (plus2 x) (plus2 y) -> Equal x y
plus2Injective Z Z Refl = Refl
plus2Injective Z (S _) _ impossible
plus2Injective (S _) Z _ impossible
plus2Injective (S n) (S n) Refl = Refl
```

Note that the only difference is the order of the clauses in `plus2Injective`, the contents of each clause is identical.  Note also that the clauses do not overlap at all.

# Expected Behavior

Both files type check and compile to very similar function.

# Observed Behavior

clause_order_good.idr is accepted.  clause_order_bad.idr is rejected with "Error: plus2Injective is not covering." and "Missing cases: plus2Injective (S _) (S _) _".

See also: https://www.reddit.com/r/Idris/comments/nd7kcj/q_why_does_function_clause_order_matter_here/

gallais:
Thanks to @Russoul's recent PR I had the fully qualified vs. `Resolved` name issue
in mind and was able to spot it after making the debug output easier to read.

I wonder whether we could improve the internal representation to avoid these type of
errors altogether...

Summary:

When generating the compile-time case trees that are used for coverage checking, we
generate dummy clauses for the impossible branches. We were not careful during this
generation and were using fully qualified names for the constructors rather than their
`Resolved` counterpart. As a consequence the case tree builder was not able to group e.g.

```idris
f (S ...) ps = rhs
f (S ...) qs impossible
```

as

```idris
f m e = case m of
  (S ...) => case e of
     ps => rhs
     qs => Impossible
```

because the first one looked like `f ($resolvedN ...) ...` internally while the other was
`f (Prelude.Types.S ...) ...)` and clearly `$resolvedN` and `Prelude.Types.S` are the
names of distinct constructors. :see_no_evil: 

So instead you'd get the nonsensical:

```idris
f m e = case m of
  (S ...) => case e of
     ps => rhs
  (S ...) => case e of
     qs => Impossible
```


Russoul:
One of the possible solutions to preserving the invariant is to decouple Resolved from Name. But that seems like a huge change. @ShinKage and I have wanted to do that forever. I also had an idea to propose a new `Name` structure but that would be too much on part, I already have a few unfinished Idris projects that need some time to be poured into them.
At the same time, problems with `Name` come up almost in any moderately sized modification, at least in my experience.

PR #1429, opened 2021-05-17
Believe me fix for refc

voltur01:
Hi,

I was playing with the RefC backend and noticed that even simple test programs crashed - apparently reference counting was not correct somewhere. A bit of debugging showed that a `Value*` was deleted right after `prim__believe_me()` returned a value it received without incrementing the reference counter.

This pull request adds a simple test that fails without the fix and succeeds with the fix. In addition, valgrind confirms that memory issue is resolved with this fix.


ohad:
Hi,

This is already fixed in #1323 (draft), and @madman-bob told me today he's wrapping up a few commits making substantial progress to a self-hosting RefC.
(Still some way to go for full self-hosting, but hopefully not too much.)

voltur01:
Hi Ohad,

Oh, good to know and thanks for the link - I shall have a look where the RefC is going.

Issue #1086, opened 2021-02-19
One-line exception in REPL : "Exception: variable DataC-45Strings-fastUnpack is not bound"


felipegchi:
# Steps to Reproduce
```idris
public export
simpleMd : String
simpleMd = "#eita"

simpleMdRes : String
simpleMdRes = (fixTkns revertTokens (lexMd simpleMd))

simpleLex : simpleMdRes = simpleMd 
simpleLex = Refl
```

and then, i reloaded it in REPL and i got an error saying that i cant unify the types BUT i saw that `simpleMdRes` was evaluating a strange weird code:

```idris
case case let (ts, (l, (c, str'))) =
                case case scan (Pred (\{arg:0} => intToBool (prim__eq_Char arg '['))) [] (fastUnpack "#eita") of
                     {Just (tok, rest) => Just (MkToken line col (fn (pack (reverse Char tok))),
                     (line + cast (countNLs a str tmap acc col line pred tok), (getCols a str tmap acc col line pred tok col, rest)));
                     Nothing => getFirstToken a str tmap acc col line pred ts str} of
                {Just (tok, (line', (col', rest))) => if pred tok
                                                        then (reverse ?a acc, (line, (col, [])))
                                                        else assert_total (tokenise pred line' col' (tok :: acc) tmap rest);
                Nothing => (reverse ?a acc, (line, (col, str)))}
          in (ts, (l, (c, pack str'))) of
     {(tokens, (_, (_, ""))) => Right tokens;
     (_, (line, (column, _))) => Left (line, column)} of
{Left x => "";
Right list => fn (map tok list)}
```

So i thought that i could just run `:exec putStrLn simpleMdRes` and then i got the issue.

# Expected Behavior
```
#eita
```

# Observed Behavior
```
Exception: variable DataC-45Strings-fastUnpack is not bound
```

gallais:
Can you please make sure the example is self-contained?
Your code won't load as is (even after commenting `simpleLex` out).

Chiyoku:
```
➜  Md git:(main) ✗ idris2 LexerTest.idr -p contrib
     ____    __     _         ___                                           
    /  _/___/ /____(_)____   |__ \                                          
    / // __  / ___/ / ___/   __/ /     Version 0.3.0-e13e069c2
  _/ // /_/ / /  / (__  )   / __/      https://www.idris-lang.org           
 /___/\__,_/_/  /_/____/   /____/      Type :? for help                     

[...]

Issue #845, opened 2020-12-13
Data.Strings.split needs an explicit Data.List1 import in its client codes.


andorp:
Using `Data.Strings.split` throws an error in the `idris2` repl and it does not compile either.

# Steps to Reproduce

```
import Data.Strings

main : IO ()
main = do
  printLn $ split (==',') "12,x,x,45,3,6,x,x,100,x"
```

# Expected Behavior

The given snippet should compile.
The error message is not telling what is the source of the error. I mean there is no source location rendered in the error message. It took a few minutes in the original program to figure out what went wrong. As a developer I would expect if I import a function from a different module, it should work without importing the module what the Dict.Strings depends on.

Maybe solutions for this problem (as I see this issue with a novice user's eye):

 * Data.Strings reexports Data.List1
 * Identify the sub expression which causes this problems and render its source location
 * During elaboration/type-checking Idris should load the missing modules from context

# Observed Behavior

```
1/1: Building Day13 (Day13.idr)

Error: While processing right hand side of main. Undefined name Data.List1.List1. 
Error(s) building file Day13.idr
```
Adding the explicit Data.List1 import fixes the problem.

gallais:
Idris should be able to refer to names in scope by transitive imports
even if us mere users are not.

PR #1434, opened 2021-05-18
Some TTC loading improvements

edwinb:
Fix one situation where ttcs were reloaded unnecessarily, and do a bit more metavariable inlining so there's fewer definitions we need to save out. Overall this took about 8s off building Idris, for me.

### DIFF
--- a/libs/contrib/System/Path.idr
+++ b/libs/contrib/System/Path.idr
@@ -1,6 +1,7 @@
 module System.Path
 
 import Data.List
+import Data.List1
 import Data.Maybe
 import Data.Nat
 import Data.Strings
--- a/src/Compiler/Common.idr
+++ b/src/Compiler/Common.idr
@@ -110,9 +110,9 @@ execute {c} cg tm
 -- If an entry isn't already decoded, get the minimal entry we need for
 -- compilation, and record the Binary so that we can put it back when we're
 -- done (so that we don't obliterate the definition)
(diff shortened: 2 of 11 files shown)


edwinb:
This is odd - I guess I should rebuild libraries and test this from scratch. I'll poke further. I have more things I want to do in this PR anyway so I've set it to draft for now.

edwinb:
Oh, I think that error is something else - it's because the inlining exposes names that probably should have been imported since they're referred to in the completely explicit definition.

gallais:
> it's because the inlining exposes names that probably should have
  been imported since they're referred to in the completely explicit
  definition.

Yeah I did not mean internal names, I meant out of scope names.
But they're only out of scope in userland IMO: the file itself
transitively depends on the right modules so these names should
be visible to the compiler when filling in implicit arguments.

That's the source of the error we have in #845 and also I suspect in #1086.

edwinb:
Aha, yes, that's a good point. So this PR just exposes another instance of that. Not sure how to resolve it to be honest, but I'll come back to it.

edwinb:
Hmm, this is awkward. On the name problem, the error happens during linearity checking, and we need the type of the missing name to do that successfully. So this (and therefore #845) won't be as easy to resolve as I'd hoped. For this particular PR, I'm just putting the required import in, but the undefined name problem needs a bit more thought.

PR #1438, opened 2021-05-19
Substantial reduction in TTC size

edwinb:
Write small integers out as 1 byte, not 8, at the cost of writing larger integers out in 9 bytes. Most integers fit in 1 byte, since it's string and list lengths from user-written code.

### DIFF
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -33,7 +33,7 @@ import Data.Buffer
 -- TTC files can only be compatible if the version number is the same
 export
 ttcVersion : Int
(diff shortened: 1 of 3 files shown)


buzden:
It looks like after this change reloading of a file which already has an old TTC version leads to "segmentation fault" error (when compiler is compiled with the chez backend).

gallais:
Is it per change because the `ttcVersion` is a small `Int` so it'll be (de/)serialised
using the new scheme which is incompatible with the previous one? :see_no_evil:

Issue #1432, opened 2021-05-18
pow 2.0 3.0 --> 7.999999999999998, pow 3.0 2.0 --> 9.000000000000002


glyh:
# Steps to Reproduce

```
$ idris2
pow 2.0 3.0 
pow 3.0 2.0
```

# Expected Behavior

Output: 8.0 and 9.0 

# Observed Behavior

Output 7.999999999999998 and 9.000000000000002

stefan-hoeck:
`pow` is not a primitive but is implemented in terms of `exp` and `log`, hence the rounding errors.

PR #1440, opened 2021-05-20
add doublePow to primitives

AliasQli:
This PR adds `prim__doublePow`. This solves [issue #1432](https://github.com/idris-lang/Idris2/issues/1432).


stefan-hoeck:
Thanks a lot for this! Please note the following: Right now, continuous integration tests on github will verify, if the libs are buildable by the bootstrap compiler, because the bootstrap compiler doesn't have its own set of standard libraries. Since bootstrap doesn't know about the new `prim__doublePow` primitive yet, we have to wait with adding this to the `prelude` until the next minor release (when bootstrap will be updated) or until #1391, which deals with exactly this issue, is merged.

@gallais : There hasn't been a lot of feedback on #1391. Do you know what the status there is?

One minor thing: If you change stuff in `Core.TTC`, you also need to increase `Core.Binary.ttcVersion`.

AliasQli:
Sorry, I misclosed the PR. Reopened again.

libs/prelude/Prelude/Types.idr
```diff
@@ -699,7 +699,7 @@ log x = prim__doubleLog x
 
 public export
 pow : Double -> Double -> Double
-pow x y = exp (y * log x)
+pow x y = prim__doublePow x y -- exp (y * log x)
```
gallais, line 702:
```suggestion
pow x y = exp (y * log x) -- prim__doublePow x y
```

src/Core/Binary.idr
```diff
@@ -33,7 +33,7 @@ import Data.Buffer
 -- TTC files can only be compatible if the version number is the same
 export
 ttcVersion : Int
-ttcVersion = 54
+ttcVersion = 55
```
gallais, line 36:
```suggestion
ttcVersion = 56
```

src/Core/Hash.idr
```diff
@@ -350,31 +350,33 @@ Hashable (PrimFn arity) where
...
 
     Cast f t =>
-      h `hashWithSalt` 35 `hashWithSalt` f `hashWithSalt` t
+      h `hashWithSalt` 36 `hashWithSalt` f `hashWithSalt` t
     BelieveMe =>
-      h `hashWithSalt` 36
-    Crash =>
       h `hashWithSalt` 37
+    Crash =>
+      h `hashWithSalt` 38
```
gallais, line 379: Can we just put `DoublePow` at the end and not shift everything by one?

AliasQli:
Still can't bootstrap via racket.

gallais:
Please don't update the bootstrap, there is no good reason to do so AFAICT

AliasQli:
Ok, I understand. But if so, I'll have to change it back to

```idris
pow x y = exp (y * log x)
```

Then why should I add `prim__doublePow`? Or is there another way to work around it?

gallais:
> But if so, I'll have to change it back to

This is indeed what I suggested in the review.

> Then why should I add `prim__doublePow`?

We are about to release a new version & the definition of `pow` can be updated once
the version with the new primitive has been released.

AliasQli:
Okay. Now those changes have been reverted.

gallais:
> Now those changes have been reverted.

You still have the updated bootstrap

edwinb:
I've resolved the conflicts and made a few other changes. I'll submit a new PR to check everything is okay with the latest version and close this one. Note that since inlining happens, this does need two rounds of build and install to actually get the right result from `pow`, but that's to be expected.

edwinb:
Oh, right, it'll also need the prelude updated and a wait for the next release. Adding new primitives can be awkward like that.

Issue #1176, opened 2021-03-11
Slow compile times for some case statements


alexhumphreys:
# Steps to Reproduce

Compile the following file:
```idris
data Expr
  = ENaturalLit Nat
  | ENaturalTimes Expr Expr

data Value
  = VNaturalLit Nat
  | VBoolLit Bool
  | VNaturalTimes Value Value

eval : Expr -> Either String Value
eval (ENaturalLit k) = Left "error"
eval (ENaturalTimes k j) =
  case (!(eval k), !(eval j)) of
       (VNaturalLit 0, u) => pure u
       (t, VNaturalLit 0) => pure t
       (VNaturalLit 1, u) => pure u
       (t, VNaturalLit 1) => pure t
       (VNaturalLit 2, u) => pure u
       (t, VNaturalLit 2) => pure t
       (VNaturalLit 3, u) => pure u
       (t, VNaturalLit 3) => pure t
       (VNaturalLit t, VNaturalLit u) => pure (VNaturalLit $ t * u)
       (t, u) => pure $ VNaturalTimes t u
```

# Expected Behavior

Type check in a reasonable amount of time.

# Observed Behavior

This takes about 20 seconds to type check on my machine.

# Notes

Changing/reordering/removing almost anything about the cases reduces the compile time to ~1s. For example, this is almost instant:
```idris
case (!(eval k), !(eval j)) of
     (VNaturalLit 0, u) => pure u
     (VNaturalLit 1, u) => pure u
     (VNaturalLit 2, u) => pure u
     (VNaturalLit 3, u) => pure u
     (t, VNaturalLit 0) => pure t
     (t, VNaturalLit 1) => pure t
     (t, VNaturalLit 2) => pure t
     (t, VNaturalLit 3) => pure t
     (VNaturalLit t, VNaturalLit u) => pure (VNaturalLit $ t * u)
     (t, u) => pure $ VNaturalTimes t u
```
This issue seems different to #797 where reordering the cases didn't effect compile time, but did effect startup time.

Adding `assert_total $ case ...` has no effect on compile time.

Some comments from Edwinb on the FP Slack:

[...]

PR #1448, opened 2021-05-22
Case tree heuristics

mattpolzin:
This PR adds a new ordering step to the process of choosing which pattern to match on next when building case trees. Put another way, it does not affect the order in which function clauses are evaluated, it affects the order in which function arguments are evaluated.

The new step applies 3 heuristics based on work presented in [Compiling Pattern Matching to Good Decision Trees](http://moscova.inria.fr/~maranget/papers/ml05e-maranget.pdf). Specifically, this PR introduces what the paper calls the `f`, `b`, and `a` heuristics (in that order).

The goal is to reduce the size of case trees (i.e. reduce the number of decisions needing to be made when averaged out over all possible paths through the tree).

This first PR takes a careful approach, applying the optimization only to runtime case trees. This is because I found the optimization step to cause problems when applied to compile time case trees and have not had time to dig into that problem yet. Given how much Idris does at compile time and how interactive editing is, I suspect that we will see a bigger win if we get this step worked into compile time case trees as well in the future. 

## Performance Metrics
I chose to measure file size of resulting compiled programs, run time, and for some tests, case tree node count. Results are based off of an upstream compiler commit that is several weeks old -- it was not practical for me to rerun tests after each new commit to the upstream project.

Before getting into the metrics I collected, I'll briefly mention the `mergeLim` example I wrote to prove the concept of these optimizations.

### `mergeLim` Sample
The following function is run in a large loop to generate performance numbers large enough to differentiate the performance of the upstream compiler and the compiler being tested with case tree optimizations.
```idris
mergeLim : List a -> List a -> Nat -> List a
mergeLim xs ys 0 = xs
mergeLim [] ys (S k) = ys
mergeLim xs [] (S k) = xs
mergeLim (x :: xs) (y :: ys) (S k) = x :: mergeLim xs (y :: ys) k
```
The `mergeLim` function was specifically formulated to put the `Nat` argument last despite the fact that matching on it first produces a more performant case tree.

### `SlowTree` Sample
The `SlowTree` code comes directly from https://github.com/idris-lang/Idris2/issues/1176.

### `pg-fn-diff` Sample
This is just the only full executable Idris 2 program I happened to have handy. It's not massive, but it is real-world. It is closed-source because I wrote it at work but it served as another data point for comparison nonetheless.

#### Node Count
|  | Upstream | Optimized |
|-|-----------|-----------|
| mergeLim | 28 | 10 |
| pg-fn-diff | 216 | 142 |
| idris compiler | 39218 | 33450 |

#### Filesize
|  | Upstream | Optimized |
|-|-----------|-----------|
| mergeLim | 160K | 156K |
| pg-fn-diff | 1.8M | 1.8M |
| idris2.ss | 9.77M | 11.29M |

#### Runtime
`mergeLim` was run at 10000000 iterations with a roughly 350 element list passed in twice and a limit of 600. An arbitrary printout of list length was used to prevent any downstream optimizing away of the call to `mergeLim`.

|  | Upstream | Optimized |
|-|-----------|-----------|
| mergeLim | 1m35.8s | 1m34.2s |
| Idris Compiler (compiling SlowTree) | 10.66s | 10.24s |
| Idris Compiler (compiling Self) | 2m43.18s | 2m54.98s |

Roughly 5.6s of the above 2m54.98s ends up being due to running the heuristics, so the remaining 6.2s is actually the compiled compiler having had case tree optimizations applied running more slowly (when compiling itself).

On the other hand, the `mergeLim` example shows that some programs do run faster when optimized this way, a result that is echoed by compiling `SlowTree` which is marginally faster despite heuristics taking some extra time to run.

### Conclusions
1. The difference in compile time caused by the additional work during case tree building _can_ be substantive for a codebase as large as the compiler.
2. When these optimizations make a difference to the number of case tree nodes needed to compile an app, the difference is always ostensibly good (fewer nodes).
3. Fewer nodes does not necessarily equal less large files; at times file size dropped with node count, at others (notably the compiler itself) the file size was _larger_ with these optimizations than without them -- perhaps there are other optimizations that are not as effective given certain case tree structures promoted by the heuristics in this PR.
4. Fewer nodes does not necessarily equal faster runtime (though sometimes it absolutely does); similar to file size, the compiler with optimized case trees from this PR's heuristics builds itself slightly more slowly than the upstream compiler even when heuristics are not being applied to the new build.
5. It is unclear how much compilation time might drop from running these heuristics on compile-time trees but it seems worth exploring more in the future (I just have not had time to dig into the compilation failures that surface when I apply these heuristics there).

With those conclusions in mind, I propose that these optimizations get merged behind the `-Xcase-tree-opt` flag I added while testing these heuristics. This would allow experimentation, hopefully leading to more data points from the community on more programs against more backends (does this particularly help Refc or JavaScript etc.?).

I won't be disappointed if instead we decide not to merge this at all in its current form, but I've run out of time for experimenting or gathering more data for the moment so there is no harm in opening a PR.

Regardless, I'd love feedback on the concept, implementation, or results thus far!


edwinb:
This is an interesting experiment, and it's next in my queue to look at. The conflicts look like they shouldn't take too much resolving fortunately, so I'll try to have a go with it soon.

It doesn't look like it adds too much complexity, which is nice, although your results suggest it doesn't really make much difference, which surprises me a bit. It is probably worth exploring in more depth, which is going to be easier if it's merged.

mattpolzin:
@edwinb if you'd like I'll fix the merge conflicts tonight since you're clearly cruising through PRs and have better things to do than fix small conflicts.

edwinb:
That would be great if you can, thanks! I was saving this for tomorrow since I'm probably going to play around with it a bit once the conflicts are resolved.

edwinb:
Great, thanks for fixing. I'll merge this and have a play with it. I'm curious as to why the generated Idris gets bigger but in general this is going to be worth experimenting with.

PR #1450, opened 2021-05-22
[ new ] `--install-with-src`; refactoring around `FC`s

Russoul:
Why:
* To implement robust cross-project go-to-definition in LSP. i.e you can jump to definition of any global name coming from library dependencies, as well as from the local project files.

What it does:
* Modify `FC`'s to carry `ModuleIdent` for .idr sources, file name for .ipkg sources or nothing for interactive runs.
* add `--install-with-src` to install the source code alongside the `ttc` binaries. The source is installed into the same directory as the corresponding `ttc` file. Installed sources are made read-only.
 As we install the sources pinned to the related `ttc` files we gain the versioning of sources for free.

#### Notes:
* The upstream stores loaded module names in `FC` as `src-folder/Path/To/Module.idr` whereas this modification changes it to
`Path.To.Module`. We do that because the information about the `src-folder` prefix becomes inaccessible right after we install the sources (we don't keep the prefix). Meaning that we no longer know what part of the full file name is the prefix and what part is the module ident.
* ~~This adds a breaking change to the IDE mode protocol: instead of sending file names in whatever command, it now sends module idents (i.e `Core.Name.Namespace` vs `src/Core/Name/Namespace.idr`) If we do care very much about the IDE protocol, I think we could retain the old behaviour.~~ Only `NameAt` command seems to be affected by the change.
Syntax highlighting data is sent the way it was before.
* The diff is big because practically any test that features intended failures has to ship with an updated golden value, because `FC` rendering is slightly changed, as described above.


src/Core/FC.idr
```diff
@@ -20,21 +22,75 @@ public export
...
+public export
+data OriginDesc : Type where
+  ||| Anything that originates in physical Idris source files is assigned a
+  ||| `PhysicalIdrSrc modIdent`,
+  |||   where `modIdent` is the top-level module identifier of that file.
+  PhysicalIdrSrc : (ident : ModuleIdent) -> OriginDesc
+  ||| Anything parsed from a package file is decorated with `PhysicalPkgSrc fname`,
+  |||   where `fname` is path to the package file.
+  PhysicalPkgSrc : (fname : FileName) -> OriginDesc
+  Virtual : (ident : VirtualIdent) -> OriginDesc
```
gallais, line 60: Does that mean we can now finally make `ModuleIdent` store a `List1`?
There are enough changes already in this PR but that's something to keep in mind!

src/Core/Metadata.idr
```diff
@@ -96,7 +98,7 @@ record Metadata where
        currentLHS : Maybe ClosedTerm
        holeLHS : List (Name, ClosedTerm)
        nameLocMap : PosMap (NonEmptyFC, Name)
-       sourcefile : String
+       sourceIdent : Either VirtualIdent (SourceFileType, FileName)
```
gallais, line 101: Why not just an `OriginDesc`? Is it because we want a `FileName` rather than a `ModuleIdent` in `PhysicalIdrSrc`?
Then we could define

```idris
data OriginDesc' : Type -> Type where      
      ||| Anything that originates in physical Idris source files is assigned a
      ||| `PhysicalIdrSrc modIdent`,
      |||   where `modIdent` is the top-level module identifier of that file.
      PhysicalIdrSrc : (ident : a) -> OriginDesc
      ||| Anything parsed from a package file is decorated with `PhysicalPkgSrc fname`,
      |||   where `fname` is path to the package file.
      PhysicalPkgSrc : (fname : FileName) -> OriginDesc
	  Virtual : (ident : VirtualIdent) -> OriginDesc

OriginDesc = OriginDesc' ModuleIdent
MetaOrigin = OriginDesc' FileName

Eq a => Eq (OriginDesc' a) where (...)
Functor OriginDesc' where (...)
```

and `compareOrigins` becomes `(pathToNS' wdir sdir) <$> finfo == ident`

src/Core/FC.idr
Russoul: Yes!

src/Core/Metadata.idr
Russoul: Yes, good idea. Let me do that

gallais:
Note that there is a script in `tests/ideMode/ideMode005` to regenerate the golden files.

Russoul:
Thanks for the info. I think I won't need that though: I reverted the printing of highlights to file names (versus module idents) to keep backward compatibility of IDE mode.

Russoul:
Okay, I think this is ready now :)

src/Idris/Driver.idr
```diff
@@ -178,7 +179,15 @@ stMain cgs opts
                  when (checkVerbose opts) $ -- override Quiet if implicitly set
                      setOutput (REPL False)
                  u <- newRef UST initUState
-                 m <- newRef MD (initMetadata $ fromMaybe "(interactive)" fname)
+                 origin <- maybe
+                   (pure $ Virtual Interactive) (\fname => do
+                     defs <- get Ctxt
+                     let wdir = defs.options.dirs.working_dir
+                     let sdir = defs.options.dirs.source_dir
+                     modIdent <- pathToNS wdir sdir fname
```
gallais, line 187: This is repeated in

* `src/Idris/IDEMode/SyntaxHighlight.idr`
* `src/Idris/ModTree.idr`
* `src/Idris/Package.idr` (twice)
* `src/Idris/REPL.idr`
* `src/TTImp/ProcessDecls.idr`
* `src/Yaffle/Main.idr`

so it should probably be factored out.

tests/idris2/reg033/expected
```diff
@@ -1,2 +1,21 @@
 1/2: Building DerivingEq (DerivingEq.idr)
-2/2: Building test (test.idr)
+Error: While processing right hand side of genEq. Can't find an implementation for FromString OriginDesc.
```
gallais, line 2: This looks like an unintended error

gallais (commented):
Finally went through the whole thing.

cypheon:
This will be amazing for documentation links! Linking to Names originating from other packages is not implemented right now (and would be difficult and maybe sometimes incorrect with just the information from the "old" `FC`), but with this change it should be almost straight-forward.

PR #1452, opened 2021-05-22
Update Emacs idris-mode doc

nickdrozd:


fabianhjr (approved):

docs/source/tutorial/interactive.rst
```diff
@@ -21,7 +21,7 @@ section, we explain how these features can be exploited by a text
 editor, and specifically how to do so in `Vim
 <https://github.com/edwinb/idris2-vim>`_. An interactive mode
 for `Emacs <https://github.com/idris-hackers/idris-mode>`_ is also
-available (though not yet updated for Idris 2).
+available, updated for Idris 2 compatability as of 23 February 2021.
```
gh-name, line 24: Since we had already use this fork in https://github.com/idris-lang/Idris2/blob/b457d15701f6f359d66c6e32af29561b44bfd81e/flake.nix#L6, if there is a need to introduce Emacs mode, IMO it would be better to change them in one shoot. Anyway this version isn't active yet. There are few requests left to be merged and still some bugs.

gallais:
Should we point at https://github.com/idris-community/idris2-mode too?

PR #1454, opened 2021-05-23
[contrib] A finger tree implementation

AliasQli:
A finger tree implementation.


libs/contrib/contrib.ipkg
```diff
@@ -89,6 +89,9 @@ modules = Control.ANSI,
 
           Data.Rel.Complement,
 
+          Data.Seq,
+          Data.Seq.Seq',
```
Z-snails, line 93: Why are there 2 of these?

libs/contrib/Data/Seq/Seq'.idr
```diff
@@ -0,0 +1,219 @@
...
+
+public export
+implementation Monoid (Seq' a) where
+  neutral = empty
+
+||| This implementation is differnt from that of Seq.
+public export
+implementation Applicative Seq' where
+  pure = singleton
+  fs <*> xs = concat $ map (\f => map f xs) fs
```
Z-snails, line 204: `concat . map` = `foldMap` (unless it's not `Prelude.concat`)

libs/contrib/contrib.ipkg
AliasQli: The first one has its length in its type, the second not.

libs/contrib/Data/Seq/Seq'.idr
AliasQli: Ok, I forgot.

edwinb:
This looks nice, although I'm now in two minds about merging this sort of contribution. In the past, we've always been fairly happy to add anything to `contrib` but we're now at a stage where there's a lot there that risks becoming unmaintained, and a lot of it is untested. So, would this be better put into its own library, which we can link from https://github.com/idris-lang/Idris2/wiki/Libraries?

My worry in general is that merging a new library also implies some kind of committment on our part to maintain it, and we already probably have too much to keep track of (as evidence of that: we're too slow at dealing with PRs!)

PR #1458, opened 2021-05-25
Simple signal handling

mattpolzin:
This adds some simple signal handling to the base library.

I chose to implement this as a collector that allows retrieving signals that have been sent to a process since the last time the process checked for signals. This allows Idris developers to handle signals as part of a run loop.

If someone else wants to expand the implementation to offer thread-based interrupt handlers, this PR's C code and Idris types representing the most common signals should offer a solid foundation.

I, for one, am excited to take this and implement graceful shutdown on `SIGINT` (Ctrl+C) for a couple of my utilities!

### DIFF
--- a/libs/base/System/Signal.idr
+++ b/libs/base/System/Signal.idr
@@ -114,6 +114,9 @@ prim__collectSignal : Int -> PrimIO Int
 %foreign signalFFI "handle_next_collected_signal"
 prim__handleNextCollectedSignal : PrimIO Int
 
+%foreign signalFFI "send_signal"
+prim__sendSignal : Int -> Int -> PrimIO Int
+
 %foreign "C:idris2_getErrno, libidris2_support, idris_support.h"
 prim__getErrorNo : PrimIO Int
 
@@ -177,3 +180,13 @@ export
 handleNextCollectedSignal : HasIO io => io (Maybe Signal)
 handleNextCollectedSignal =
   toSignal <$> primIO prim__handleNextCollectedSignal
+
+||| Send a signal to a process using a PID to identify the process.
+export
+sendSignal : HasIO io => Signal -> (pid : Int) -> io (Either SignalError ())
+sendSignal sig pid = do
+  res <- primIO $ prim__sendSignal pid (signalCode sig)
+  case isError res of
+       False => pure $ Right ()
+       True  => Left <$> getError
+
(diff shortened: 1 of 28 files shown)


libs/base/System/Signal.idr
```diff
@@ -0,0 +1,259 @@
...
+
+toSignal : Int -> Maybe Signal
+toSignal (-1) = Nothing
+toSignal x    = foldr matchCode Nothing codes
+  where
+    matchCode : (Int, Signal) -> Maybe Signal -> Maybe Signal
+    matchCode x (Just y) = Just y
+    matchCode (code, sig) Nothing = case x == code of
+                                         True  => Just sig
+                                         False => Nothing
```
gallais, line 115:
```suggestion
toSignal x    = lookup x codes
```

libs/base/System/Signal.idr
```diff
@@ -0,0 +1,259 @@
...
+||| Use `forever` as your fuel if you don't want or need to
+||| retain totality. Alternatively, pick a max number to
+||| retrieve and use `limit/1` as your fuel.
+export
+handleManyCollectedSignals : HasIO io => Fuel -> io (List Signal)
+handleManyCollectedSignals fuel = catMaybes <$> handleMany fuel
+  where
+    handleMany : Fuel -> io (List (Maybe Signal))
+    handleMany Dry = pure []
+    handleMany (More fuel) = [| handleNextCollectedSignal :: handleMany fuel |]
```
gallais, line 239: Shouldn't this stop at the first `Nothing`? This way if you don't know how many signals
you got but are willing to bet it's a finite number, `handleManyCollectedSignals forever`
would give you a finite list back rather than diverging.

libs/base/System/Signal.idr
mattpolzin: Right! Of course this is just lookup.

libs/base/System/Signal.idr
mattpolzin: Yeah, definitely did not want this to diverge. I was blinded by how clean it looked in those brackets.

I've fixed, but if you've got a suggestion to make my fix any tidier, let me know! I don't find my fix _ugly_, but it does do a bit more explicit wrapping/unwrapping than my (admittedly broken) first attempt.

mattpolzin: Thanks! I don't plan on adding `unfoldM` for `Fuel` or tweaking this function more before merge, but I always learn a lot and appreciate you taking the time to answer.

PR #1461, opened 2021-05-25
[ fix ] System.getArgs on node backend

stefan-hoeck:
As described by @Z-snails on [discord](https://discord.com/channels/827106007712661524/827109161857712128/846685919582748692), there were some number errors in the Node backend when building and running the Inigo package. The culprit were two FFI calls in module `System` in `base`, which didn't properly convert between `BigInt` and `Number`.


gallais:
Thanks!

PR #1471, opened 2021-05-26
[ fix ] use twos complement truncation for signed ints

stefan-hoeck:
As discussed with @madman-bob on discord (relevant quote can be found [here](https://github.com/idris-lang/Idris2/issues/1048#issuecomment-847722512)), the way signed limited precision integers are truncated right now in case of overflows or during casts is counterintuitive for programmers coming from other languages like C, Haskell, or Rust. This PR amends this by applying proper two's complement truncation in case of integer arithmetic overflows.

Only the Scheme and JS backends were adjusted in this PR.


stefan-hoeck:
I added the new integer tests from #1480 to the node and chez tests. However, tests for `div` and `mod` are not run right now, because these two backends do not (yet) use the Euclidian `div` and `mod`. I'd like to add those in a separate PR, since it is not yet clear whether this change will break stuff in the compiler.

PR #1472, opened 2021-05-27
Define and implement Relation interfaces

nickdrozd:
There are several flaws in the way binary relations are handled right now:

1. Ugly, ad-hoc, duplicated names. `lteRefl`, `factorReflexive`, `lteTransitive`, these names are terrible. It's clear what they mean, but they aren't named systematically. It would be nicer to say: this relation is reflexive and transitive, and that relation is reflexive and transitive, and so on.

2. Where binary relations are defined in a general way, the definitions are tied to a specific and limited chain of orderings. `Preorder` is transitive and reflexive, and `Equivalence` is a symmetric preorder. But how does a "tolerance relation", which is reflexive and transitive, fit into this scheme? Why does an equivalence relation have to be defined as a symmetric preorder instead of a transitive tolerance relation? Why bring a preorder into the definition of equivalence at all?

3. Following the previous point, it isn't possible to define general relationships among binary relations except insofar as they fit within the ordering scheme. A transitive and symmetric relation is euclidean, but how can that be proved?

The idea in this PR is to define basic binary relation types as interfaces which can be mixed and matched to define more complicated relations and orderings. It's both simpler and more flexible than the existing scheme. (The existing scheme dates back to the olden days of Idris 1, and doesn't seem to have had all that much thought put into it.)

I attempted this about a year ago in https://github.com/idris-lang/Idris2/pull/314, but at that time the type inference was still a little squirrely and so the implementation was somewhat flawed. Things have improved since then, so I dusted off the idea, and here it is again.


gh-name:
That's cool to see these changes. Since we will have `Reflexive` and `Transitive` interfaces, can we have a syntax sugar for all chain-reasonings, not just on propositional equality?

gallais:
We already have https://github.com/idris-lang/Idris2/blob/6f839240c5eac4c5c36bddd19f005415efeca3c2/libs/contrib/Syntax/PreorderReasoning/Generic.idr

nickdrozd:
I went through and updated all the remaining uses of the old relation / order system. There are just a few places where I haven't been able to figure out how to update:

1. `Data.Fin.Order`: prove that `FinLTE` is `Connex`. The proof is close at hand, I just can't figure out the last piece. Fresh eyes should be able to get it quickly.

2. `Decidable.Order.Strict`. The idea is to make "strict" versions of relations (`Irreflexive` for `Reflexive`, `Asymmetric` for `Symmetric`, etc). This PR changes not just the interfaces themselves, but also the way the relations are called. So this module will take a little more work to re-express the logic.

Other than that, it works!

nickdrozd:
An interesting paper on relations between relations: https://arxiv.org/pdf/1806.05036.pdf

libs/base/Control/Relation.idr
```diff
@@ -0,0 +1,79 @@
...
+interface Serial ty rel where
+  serial : {x : ty} -> (y : ty ** rel x y)
+
+public export
+interface Euclidean ty rel where
+  euclidean : {x, y, z : ty} -> rel x y -> rel x z -> rel y z
+
+public export
+interface (Reflexive ty rel, Symmetric ty rel) => Tolerance ty rel where
+
```
gallais, line 36: We can probably use

```suggestion

public export
interface (Transitive ty rel, Symmetry ty rel) => PartialEquivalence ty rel where
```

too

libs/base/Control/Relation.idr
```diff
@@ -0,0 +1,79 @@
...
+(Reflexive ty rel, Euclidean ty rel) => Symmetric ty rel where
+  symmetric {x} xy = euclidean {x} xy $ reflexive {x}
+
+public export
+(Reflexive ty rel, Euclidean ty rel) => Transitive ty rel where
+  transitive {rel} xy yz = symmetric {rel} $ euclidean {rel} yz $ symmetric {rel} xy
+
+public export
+(Transitive ty rel, Symmetric ty rel) => Euclidean ty rel where
+  euclidean {rel} xy xz = transitive {rel} (symmetric {rel} xy) xz
```
gallais, line 66: Should these be named? It feels like they'd get in the way during instance resolution.

libs/base/Control/WellFounded.idr
```diff
@@ -58,7 +59,7 @@ sizeAccessible x = Access (acc $ size x)
   where
     acc : (sizeX : Nat) -> (y : a) -> (size y `LT` sizeX) -> SizeAccessible y
     acc (S x') y (LTESucc yLEx')
-        = Access (\z, zLTy => acc x' z (lteTransitive zLTy yLEx'))
+        = Access (\z, zLTy => acc x' z (transitive {x = (S (size z))} {y = (size y)} {z = x'} zLTy yLEx'))
```
gallais, line 62: Does specifying `transitive {rel = LT}` work here?
That'd be more economical.

If so we may want to redefine `LT` to be a newtype wrapper.
With `Cast (LT x y) (LTE (S x) y)` and `Cast (LTE (S x) y) (LT x y)`
it may be nice enough to use if we get improved inference from it.

libs/contrib/Data/Nat/Order/Properties.idr
```diff
@@ -72,22 +73,24 @@ notLteIsnotlte a b not_a_lte_b = reflect (lteReflection a b) not_a_lte_b
...
-        <~ a     ...(b_lt_a)
-  in notLteIsnotlte a b not_a_lte_b
+-- GTIsnotlte a b b_lt_a =
+--   let not_a_lte_b : Not (a `LTE` b)
+--       not_a_lte_b = \a_lte_b =>
+--         irreflexive {rel = LT} $ CalcWith {leq = LTE} $
+--         |~ 1 + a
+--         <~ 1 + b ...(plusLteMonotoneLeft 1 a b a_lte_b)
+--         <~ a     ...(b_lt_a)
+--   in notLteIsnotlte a b not_a_lte_b
```
gallais, line 83: is this proof just missing?

libs/contrib/Data/Nat/Order/Strict.idr
```diff
@@ -1,25 +1,30 @@
...
-irreflexiveLTE (S a) (LTESucc a_lt_a) = irreflexiveLTE a a_lt_a
+%default total
+
+public export
+Irreflexive Nat LT where
+  irreflexive {x = 0} _ impossible
+  irreflexive {x = S k} (LTESucc prf) =
+    -- why can't this be inlined?
+    let rec = irreflexive {x = k} {rel = LT} in
+      rec prf
```
gallais, line 19: You can probably use `Data.Nat`'s  `view : (lt : LT m n) -> View lt` to make the definition cleaner.

nickdrozd:
I figured out the last `Fin` proof. The "strict" stuff is harder to untangle, but also less important. In any case, this PR is up-to-date and ready to rock.

gallais (commented):
Looks very good!

There are still quite a lot of annotations that I would not expect to be
necessary (either having to manually pass the `rel` or the `x` in `reflexive`
calls).

Probably worth looking into at some point.

libs/contrib/Data/Nat/Order/Properties.idr
nickdrozd: Forgot to fill it in. Added.

libs/base/Control/WellFounded.idr
nickdrozd: Simplified to use `{rel = LT}`

libs/base/Control/Relation.idr
nickdrozd: That sounds like a good idea, but I don't know how to do it. There were also some other relation laws that I wanted to added, but I couldn't get the compiler to accept them all. I found the compiler to be very finicky when it came to this part.

libs/base/Control/Relation.idr
nickdrozd: Added

libs/contrib/Data/Nat/Order/Strict.idr
nickdrozd: The same problem came up every way I tried with `view`. There's something funny going on here.

libs/base/Control/Relation.idr
gallais: You can write `[TSE] (Transitive ty rel, Symmetric ty rel) => Euclidean ty rel where (...)`
to name the instance `[TSE]`. You should then be able to use it in specific instance declarations
by writing `using TSE`, cf. the existing work on additive vs. multiplicative semigroups:

https://github.com/idris-lang/Idris2/blob/9c63e90fd215526c2e7a8a83cbb6ad6aa2e3f62c/libs/contrib/Data/Nat/Algebra.idr#L10-L12

libs/contrib/Data/Nat/Order/Strict.idr
gallais: This seems to work:

```idris
public export
Irreflexive Nat LT where
  irreflexive lt = case view lt of
    LTSucc lt => irreflexive {rel = Nat.LT} lt
```

For some reason, without the explicit qualifier Idris picks `Prelude.LT` over the `Nat` one.

stefan-hoeck:
Sorry for my chiming in here, but when I quickly looked at this PR, I found it hard to figure out what this is all about. Since this is to be in the base library, I think all these interfaces should be properly annotated with doc strings explaining in some detail, why they are useful and how to properly use them.

I am well aware that in my own PRs I myself did and do not always live up to this standard, but we should definitely strive to make the standard libs as accessible as possible for people new to Idris2, dependent types and compile time proofs. Failing to do so now, will just mean more work later on.

libs/contrib/Data/Nat/Order/Strict.idr
nickdrozd: In fact, just putting in `Nat.LT` fixes it without any need for `view`.

gallais: The view means you don't need to match on `x` and deal with an impossible case.

libs/base/Control/Relation.idr
nickdrozd: Okay, I got them named. One problem left, which is that I don't know how to use these proofs. For example, it's easy to prove that `Equal` is transitive and symmetric. It should be possible to use those facts along with `TSE` to prove it's also euclidean, but I don't know how to do that.

nickdrozd:
> Sorry for my chiming in here, but when I quickly looked at this PR, I found it hard to figure out what this is all about. Since this is to be in the base library, I think all these interfaces should be properly annotated with doc strings explaining in some detail, why they are useful and how to properly use them.
> 
> I am well aware that in my own PRs I myself did and do not always live up to this standard, but we should definitely strive to make the standard libs as accessible as possible for people new to Idris2, dependent types and compile time proofs. Failing to do so now will just mean more work later on.

Good point, I added docstrings to the new stuff. Take a look and let me know if it's clear enough.

stefan-hoeck:
> > Sorry for my chiming in here, but when I quickly looked at this PR, I found it hard to figure out what this is all about. Since this is to be in the base library, I think all these interfaces should be properly annotated with doc strings explaining in some detail, why they are useful and how to properly use them.
> > I am well aware that in my own PRs I myself did and do not always live up to this standard, but we should definitely strive to make the standard libs as accessible as possible for people new to Idris2, dependent types and compile time proofs. Failing to do so now will just mean more work later on.
> 
> Good point, I added docstrings to the new stuff. Take a look and let me know if it's clear enough.

Thank you very much for this. Personally, for me it would be helpful to have a concrete Idris2 example (like you did with `LTE` and `Nat`) for each of these relation interfaces. Other users might find this useful as well.

libs/base/Control/Relation.idr
gallais: Hmmm. I guess there isn't good syntax for it. We may want to open an issue for this.
Here are two possible solutions, each with their inconvenients:

```idris
%hint EqualEuclidean : Euclidean ty Equal
EqualEuclidean = TSE

Euclidean ty Equal where
  euclidean = euclidean {rel = Equal} @{TSE}
```

nickdrozd: That did it! The hint was not required.

nickdrozd:
> Personally, for me it would be helpful to have a concrete Idris2 example (like you did with `LTE` and `Nat`) for each of these relation interfaces. Other users might find this useful as well.

I added an example of a tolerance relation that is not transitive (although not proved to be intransitive). I think every relation either is implemented somewhere or does not have any natural examples. Even the [Wikipedia page for euclidean relations](https://en.wikipedia.org/wiki/Euclidean_relation) doesn't give a single example!

stefan-hoeck:
Thanks again, for adding these docstrings! Having examples where they make sense is perfectly fine.

libs/base/Control/Relation.idr
gallais: The `%hint` is so that it actually behaves like an implementation and is used by auto search!

libs/contrib/contrib.ipkg
```diff
@@ -76,6 +76,7 @@ modules = Control.ANSI,
           Data.Nat.Fib,
           Data.Nat.Order.Strict,
           Data.Nat.Order.Properties,
+	  Data.Nat.Order.Relation,
```
gallais, line 79: indentation is off

libs/contrib/Decidable/Order/Strict.idr
```diff
@@ -1,51 +1,57 @@
...
 EqOr spo a b = Either (a = b) (a `spo` b)
 
--- Can generalise to an arbitrary equivalence, I belive
-public export
-[MkPreorder] {spo : t -> t -> Type} -> StrictPreorder t spo => Preorder t (EqOr spo) where
-  reflexive a = Left Refl
-  transitive a _ c (Left  Refl) bLTEc        = bLTEc
-  transitive a b _ (Right aLTb) (Left  Refl) = Right aLTb
-  transitive a b c (Right aLTb) (Right bLTc) = Right $ Strict.transitive a b c aLTb bLTc
+-- -- Can generalise to an arbitrary equivalence, I belive
```
gallais, line 34: This file is full of commented out proofs that probably need to be revived.

gallais:
The force pushes make it hard to see what is new which means I need to re-check
the whole PR every time. Please try not to do that.

libs/contrib/Decidable/Order/Strict.idr
nickdrozd: I can't figure out how to get them up to date, and TBH I'm not totally sure what they do. I can cut out the commented stuff or leave it, but as it is now I don't know how to revive them.

ohad:
Sorry, this went under my radar and has now broken a substantial library I'm working on (a few thousands lines of code).

Making reflexivity take an implicit argument means we can no longer write anonymous reflexivity proofs, since we don't support anonymous functions with implicit arguments:
```idris
-- previously:
{ reflexivity = \x => the (rel x x) ?h
...}
```

nickdrozd:
@ohad Could you link to the code that's broken by this? Is it that you had implemented the old order relations and they can't be updated for the new relations?

gallais:
@nickdrozd This is the PR: https://github.com/frex-project/idris-frex/pull/31

PR #1478, opened 2021-05-27
module Data.SnocList.Elem

kasiaMarek:
added `Elem` for `SnocList` analogically to `Elem` for `List`


gallais:
This should build now that the CI uses 0.4.0 and thus can parse snoclists.

PR #1480, opened 2021-05-28
RefC Integer Support

madman-bob:
Improve integer support for the RefC backend.

- Fix issues caused by conflation of `Int` and `Bits` types.
- Add support for `IntN` types.
- Make `Integer` arbitrary precision.
- Add support for `Bits` maths operations.
- Change integer `div` and `mod` to be Euclidean.
- Add casting tests between all integer type pairs.
- Add maths operation tests on all integer types.
- Add bit-op tests on all integer types.
- Add comparison tests on all integer types.

The only change here I expect to be potentially controversial is the choice of Euclidean `div` and `mod`. See [Division and Modulus for Computer Scientists (Daan Leijen)](https://www.microsoft.com/en-us/research/publication/division-and-modulus-for-computer-scientists/) for a comparison of the various possible choices of integer `div` and `mod`.


stefan-hoeck:
We can't add `Data.Bits` implementations for `Bits64` and `Integer` to base yet, since the bootstrap compiler does not yet know about the necessary prim__xor functions. These implementations should therefore directly go to the test source file for the time being.

Also please note, that there are already extensive tests for integer primops and casts in the `chez` and `node` test suite. We should probably use the same tests for all backends (however, the tests from the other backends will not yet adhere to the two's complement overflow behavior until #1471 is merged). I'm sure your tests cover the necessary corner cases better than mine, so we could use yours in the other backends if you prefer.

madman-bob:
Ok, I've moved the interfaces you mentioned, in a separate commit to make it easier to revert later.

After taking a quick look, I find the `chez` and `node` tests quite difficult to read. As such, it's difficult to tell that it covers all cases. That said, perhaps my approach is too far in the other direction - trying to be overly general.

I agree in principle that the backends should use the same test files - perhaps even to the extent of having a `tests/backends-common` directory, to avoid duplication of the test files. In practice, it'll probably be a little fiddly to make sure we haven't dropped any corner cases.

PR #1481, opened 2021-05-28
[ cleanup ] Remove unnecessary Libraries.Utils.Either

fabianhjr:


src/Libraries/Utils/Either.idr
```diff
@@ -5,13 +5,3 @@ module Libraries.Utils.Either
 export
 mapError : (a -> c) -> Either a b -> Either c b
```
gallais, line 6: This is `mapFst` so we can probably get rid of that too.

fabianhjr: Done, removed the whole module (`Libraries.Utils.Either`)

PR #1483, opened 2021-05-29
Add `foldMap` to `Foldable`

Z-snails:
Tail recursion is very important for performance and `foldMap` can usually be implemented with tail recursion so use it for functions like `any`.

This doesn't appear to cause any performance regressions in the compiler.


libs/prelude/Prelude/Interfaces.idr
```diff
@@ -267,77 +267,96 @@ interface Foldable t where
...
+    monoid = MkMonoid @{MkSemigroup (\x, y => x && y)} True
 
 ||| Add together all the elements of a structure.
 public export
 sum : (Foldable t, Num a) => t a -> a
-sum = foldr (+) 0
+sum = concat @{(%search, monoid)}
+  where
+    monoid : Monoid a
+    monoid = MkMonoid @{MkSemigroup (+)} 0
```
buzden, line 339: Why not to have a top-level named implementations e.g. `[Additive] Num a => Monoid a`, like it's done in `Data.Nat` module in `base`?

Z-snails: My inital thinking was to not introduce more top levels definitions than necessary, but I guess that is only important if I export them. Each of the interfaces is only used once, so it won't make a difference to code size.
I'm happy to use top-level named instances, so I'll do that when I'm next programming.

libs/base/Data/Vect.idr
```diff
@@ -351,10 +351,18 @@ foldrImpl f e go (x::xs) = foldrImpl f e (go . (f x)) xs
...
 
   null [] = True
   null _ = False
 
+  foldMap f xs = go xs neutral
+    where
+      go : forall n. Vect n a -> m -> m
+      go [] z = z
+      go (x :: xs) z = go xs (z <+> f x)
+
```
buzden, line 365: Why not just `foldMap f = foldl (\acc, elem => acc <+> f elem) neutral` as soon as `foldl` is already tail-recursive?

libs/prelude/Prelude/Types.idr
```diff
@@ -390,6 +393,12 @@ Foldable List where
 
   toList = id
 
+  foldMap f xs = go xs neutral
+    where
+      go : List a -> m -> m
+      go [] z = z
+      go (x :: xs) z = go xs (z <+> f x)
+
```
buzden, line 401: Why not just `foldMap f = foldl (\acc, elem => acc <+> f elem) neutral` as soon as `foldl` is already tail-recursive?

libs/contrib/Data/List/Lazy.idr
```diff
@@ -98,6 +98,12 @@ Foldable LazyList where
   null []     = True
   null (_::_) = False
 
+  foldMap f xs = go xs neutral
+    where
+      go : LazyList a -> m -> m
+      go [] z = z
+      go (x :: xs) z = go xs (z <+> f x)
+
```
buzden, line 106: `foldMap` for lazy lists mustn't be tail-recursive as it forces a lazy list to be always fully evaluated. This leads to too eager behaviour in a case when you map values of a lazy list to some lazy type (e.g., when `m` is `Lazy Bool`) and them are folding over its monoid.

libs/base/Data/Vect.idr
Z-snails: This way it's using 1 fewer higher order functions which may be faster, I'll benchmark it soon.

Z-snails: Turns out it's almost exactly the same speed, so I'll use that suggestion.

Issue #1485, opened 2021-05-29
Incorrect parsing of let expressions with mixed local definitions and let binding


ohad:
# Steps to Reproduce
```idris
Works : Nat
Works = let n = 0
            k : Nat
            k = 1
        in n + k
                
Fails : Nat
Fails = let k : Nat
            k = 1
            n = 0
        in n + k        
```
# Expected Behavior
Both definitions should type-check --- the only difference is the order of the local definitions and the let-binding clauses.

# Observed Behavior
```
     While processing right hand side of Fails. No type declaration for Main.n.
     
      06 |                 
      07 | Fails : Nat
      08 | Fails = let k : Nat
      09 |             k = 1
      10 |             n = 0
                       ^^^^^
```

ohad:
Parsing collections of local definitions seems to be too greedy, and so we treat `n` as a local definition without an accompanying local claim, instead of another let-binding clause.

ohad:
Workarounds: 

If there is no dependency between the clauses, can re-order them so that local definitions appear last.

If there is a dependency, but it's not mutual, can split into two nested lets.

If there's a mutual dependency, promote the let-binding into a local definition by adding a corresponding claim.

gallais:
Following #691 you can use `:=` to force something to be a let-binding.

gallais:
I'll close this because, as I've mentioned above, I'd rather have users rely on the unambiguous `:=`.

PR #1487, opened 2021-05-31
[ doc ] Some documentation on `:=` syntax of `let` bindings

buzden:
I missed the original PR that adds this syntax and used to wondering what's it until #1485. Let's add some documentation about it.

UPD: A UFO flew and broke CI, thus rebased to restart it.

### DIFF
--- a/docs/source/tutorial/typesfuns.rst
+++ b/docs/source/tutorial/typesfuns.rst
@@ -1182,6 +1182,8 @@ Or even:
 More Expressions
 ================
 
+.. _sect-let-bindings:
+
 ``let`` bindings
 ----------------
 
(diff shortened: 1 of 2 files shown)


alissa-tung:
Can we give docs for Builtin.# btw and squash them？The current status is only data type are documented but not the constructor `:docs (#)`

buzden:
> Can we give docs for Builtin.# btw and squash them？The current status is only data type are documented but not the constructor `:docs (#)`

This is definitely irrelevant to this particular PR. Moreover, `:doc (#)` in REPL shows signatures of both `#` constructors and their signatures seem to be fairly self-explanatory.

gallais:
I threw in more justifications (e.g. `let ty : Type := v = v` vs. the unreadable
`let ty : Type = v = v`) and I've opened #1492 for @alissa-tung

PR #1489, opened 2021-05-31
[ base ] Some lacking implementations for `Uninhabited`

buzden:
There are a lot of situations where `Uninhabited` interface can be implemented for the standard types, but there's no such an implementation.

Since `Uninhabited` is effectively a `Not` type, but being an interface, it's convenient to use it in data structures where "not" is logically desired. But a lot of type in a standard library lacks implementations for some interesting (e.g., recursive) cases.

Couple of examples of data structures with logical "not"s as `auto`-parameters using `Uninhabited` are added as tests.


gallais (commented):
Shouldn't the test case go in `tests/base`? Should we have a test case at all?

buzden:
> Shouldn't the test case go in `tests/base`?

Well, I thought this test case is more about checking that multi-step `auto`-arguments search is working using provided implementations.

> Should we have a test case at all?

I'm not sure, but I thought what if some unintended change in the future would break it, then maybe we'd better have a test case.

edwinb:
Looks useful, thanks - I'm not sure I'd worry too much about where the test case is here.

PR #1492, opened 2021-06-01
[ doc ] for the linear pair constructor

gallais:


libs/prelude/Builtin.idr
```diff
@@ -72,7 +72,12 @@ infix 5 #
 ||| A pair type where each component is linear
 public export
 data LPair : Type -> Type -> Type where
-     (#) : (1 _ : a) -> (1 _ : b) -> LPair a b
+  ||| A linear pair of elements.
+  ||| If you have take one copy of the linear pair apart
```
mattpolzin, line 76:
```suggestion
  ||| If you take one copy of the linear pair apart
```

buzden:
Maybe, some text for `Builtin.DPair.Res.(#)` also should be added as soon as it's also a `(#)` constructor.

gallais:
I had a look at it but the definition is puzzling to me. It claims to be

> a dependent variant of LPair

so I don't understand why we have

```idris
data Res : (a : Type) -> (a -> Type) -> Type where
  (#) : (val : a) -> (1 r : t val) -> Res a t
```

rather than

```idris
data Res : (a : Type) -> (a -> Type) -> Type where
  (#) : (1 val : a) -> (1 r : t val) -> Res a t
```

buzden:
As far as I understand from Edwin's examples with the `network` library, the first part of `Res` is used for intermediate decision results and the second part is used for tracked (linear) resources. So, the definition is correct, but the description is not.

PR #1494, opened 2021-06-01
[ fix ] Memoize intermediary results in JS backends

stefan-hoeck:
See [discord](https://discord.com/channels/827106007712661524/841274390481600562/849316249677725816) for context.


gallais:
Can we add a `perf` test case to `node/` that only passes in a reasonable amount
of time if we don't break this?

stefan-hoeck:
> Can we add a `perf` test case to `node/` that only passes in a reasonable amount
> of time if we don't break this?

Just a quick question: In such a test, do we use some kind of timeout to prevent the test from running forever?

gallais:
Nope, cf. the `tests/idris2/perfX` tests

stefan-hoeck:
This should be ready now.

Issue #1493, opened 2021-06-01
make install-api fails on MacOS


avalonx:
from a fresh git clone of Idris2 Master or after doing make clean, I get the same error when trying to: make install-api after a successful make install.

Error is: line 22: 30246 Illegal instruction: 4  "$DIR/idris2_app/idris2.so" "$@"
make: *** [install-api] Error 132

gallais:
Do you have `zsh` installed?

avalonx:
Meaning uninstall my homebrew installed gnu coreutils (version of BASH) to default to the coreutils supplied by MacOS which uses the zsh version of bash since #1418 obsoletes the need for the former? Okay I’ll try that. Thanks! 

Meanwhile it also looks like there is a need to run make clean after doing make install, before running the make install-api (not documented in the instructions). It looks like doing that might have worked as it gave no error and ran through the full make process but I’ve run out of time to verify. I’ll update on that by tomorrow.

avalonx:
Here's the update.
1) gnu coreutils still required (installed via 'brew coreutils) as 'make bootstrap SCHEME=chez' still won't work without it. So I'm not sure what #1418 fixed.

2) Doing a 'make clean' before 'make install-api' seems to install the api components without errors.

3) However, it strangely creates ANOTHER 'idris2-0.30' directory under the existing 'idris2-0.30' directory and installs the api components into that. Is that how it is intended?
tree looks like this:
Idris2
--bin
--idris2-0.30
----//other stuff ...
----idris2-0.30
--------//api components

4) I've not compiled and tested the LSP server against it yet to know if it works. I'll do so next and update.

ska80:
> 1. gnu coreutils still required (installed via 'brew coreutils) as 'make bootstrap SCHEME=chez' still won't work without it. So I'm not sure what #1418 fixed.

Yes, that is correct.  The generated scheme file for bootstrapping still uses `realpath` from `coreutils`.  Once the next version of Idris is released the new generated scheme file won't use `realpath` anymore.

ska80:
> 3\. However, it strangely creates ANOTHER 'idris2-0.30' directory under the existing 'idris2-0.30' directory and installs the api components into that. Is that how it is intended?
>     tree looks like this:
>     Idris2
>     --bin
>     --idris2-0.30
>     ----//other stuff ...
>     ----idris2-0.30
>     --------//api components

The nested `idris2-0.30` is the API package and has the same version as  Idris itself as it should be.

[...]

PR #1495, opened 2021-06-02
Add 'make clean' before 'make install-api'

avalonx:
'make clean' is a needed step before doing 'make install-api'
https://github.com/idris-lang/Idris2/issues/1493

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -93,6 +93,7 @@ executable in `./build/exec`.
 You'll only need this if you're developing support tools, such as an external
 code generator. To do so, once everything is successfully installed, type:
 
+- `make clean`
 - `make install-api`
 
 The API will only work if you've completed the self-hosting step, step 4, since


gallais:
Isn't the case it's only needed if you build from bootstrap?

avalonx:
Yes. I thought to put a comment behind it to explain that, but for the sake of keeping the install doc simple and uncomplicated, I felt that it doesn’t hurt to instruct to do a ‘make clean’ even in the event there is nothing to clean. For the sake of brevity and the need to make sure the build is cleaned before someone attempts ‘make install-api’ it seemed best to just leave out the note.

technic:
Hi,
The correct order for me is described in the doc

> The API will only work if you've completed the self-hosting step, step 4,

So actually after `make clean` you need to do `make all && make install` again, and then `make install-api`
I also had same issue today, that I didn't look at the docs carefully. I suggest that we add the following sentence in the begining of the step 4:

This step requires that you have completed step 4. Then you should type `make install-api`

Because this postfix precondition format is really causing troubles since I assume most people just copy paste and execute commands as soon as they see them.

avalonx:
That doesn’t work for me… ‘make install’ without first doing ‘make bootstrap SCHEME=chez’ (or ‘make bootstrap-racket’) will fail. And whenever you do ‘make clean’, you have to do the bootstrap again before you can do ‘make install’. 

However, the ‘make install-api’ does not need the bootstrap and in fact will not work if the build is not cleaned (such as happens after doing the bootstrap and ‘make install’)

Hence why I feel the install doc needs to be amended to state that a ‘make clean’ should come before ‘make install-api’ to avoid a failing  api installation.

technic:
@avalonx this does not make sense to me, why should one bootstrap twice? 
For me in the completely fresh system the order is the following

1) Here put correct `scheme` interpreter
```
make bootstrap SCHEME=scheme
make install
```

2) Make idris2 available in the path, i.e.
```
ln -s ~/.idris2/bin/idris2 ~/bin/idris2
```

3) Clean and rebuild as in step 4
```
make clean
make
make install
make install-api
```

edwinb:
I don't think this is necessary, as others have said, and I'd rather not include a step that isn't necessary, without explaining it, or without understanding what's actually going on. If there is something going wrong that we're not understanding, maybe it'd be better to open an issue and we'll try to work it out?

Issue #1507, opened 2021-06-05
[Proposal] nix env script + docker OR remove nix from the CI


ohad:
nix is an impressive package/configuration manager, but if you don't have it installed and the nix continuous-integration tests fail, it is a pain to fix the error. The purpose of this proposal is to collect opinions on proposals to avoid that.

# Proposal 1: a nix container + scripts
Create some container + script that sets up the environment used by the nix CI.
This way, if one's PR failed, one can easily fire up the environment and debug the problem.

# Proposal 2: remove nix from the CI

The learning curve for nix is sufficiently steep that perhaps we should move it to a separate repository, where nix-specialists could troubleshoot such issues.

stefan-hoeck:
Unless we find someone who is willing to keep fixing nix related issues in this repository on a regular basis, we should opt for 2). Same goes - in my opinion - for other backends / operating systems that are hard to maintain.

abailly:
I fail to understand why nix is part of the _Continous Integration_ process of Idris2. The purpose of CI is to ensure that code submitted can be _integrated_ in the existing codebase. How does this impact nix or the distribution of Idris2 package through nix?

fabianhjr:
@abailly, the contributed nix scripts are mostly for setting up a dev environment (Like Asdf `.tool-versions`) and not for packaging.

ohad:
I agree with @fabianhjr: we have nix scripts in the repo, so we should be testing them during CI.
If we go with proposal 2, we should also remove the nix scripts from the repo so that everything on the repo gets tested.

abailly:
That's exactly what I am saying @ohad, thanks for making it clearer.

abailly:
I don't want to play the troll role here. I have started using nix at work a couple months ago and my experience with nix so far is less than optimal, to say the least. Given that a few smart people out there, even some I know personally, are nix enthusiasts and actively promote its use, I assume this is a limitation on my part. 
[...]

PR #1509, opened 2021-06-05
Simplify Nix Expressions

fabianhjr:
This should simplify the usage of nix to be strictly about declaring dependencies and building a nix-shell for development of Idris2, removes flakes, emacs, and withPackages script.

Related to #1507

Pending: Documentation on a Docker Image and modifying dependencies


guilhermehas:
Maybe, it is better to pin the <nixpkgs> channel to a specific commit. Because there is a small probability that some update of <nixpkgs> channel breaks the build.

fabianhjr:
> Maybe, it is better to pin the Nixpkgs channel to a specific commit. Because there is a small probability that some update of the Nixpkgs channel breaks the build.

The dependencies are very stable, (gmp, clang, chez, makeWrapper) OR CI-here (idris2 itself has a "build from previous version" CI check)

The main complain has been that flakes has been flaky (experienced this myself and other contributors) and flakes is an experimental unstable feature so so far flakes has been causing more failed CI builds and many people don't know what is going on when the Nix CI fails.

The nix channel could be changed to a release channel to ensure more stability

claymager:
I don't see how moving to a channel would provide more stability than a pinned flake.

Simplifying the nix and docker documentation is a great idea.

fabianhjr:
Hi @claymager, thanks for your approval your idris2-nix with flakes repo is amazing. <3

Would it be ok if I do a further merge request documenting how to use idris2 on nix/nixos and send people your way?

claymager:
A PR like this should update INSTALLING.md

fabianhjr:
Ah that is right, I should also add some notes to contributing.

PR #1522, opened 2021-06-09
Cleanup src libraries (Remove `Libraries.Data.Bool.Extra` and half `Libraries.Data.List1`) since they have been included since v0.3.0

fabianhjr:


andrevidela (approved):
Looks good to me

Issue #1244, opened 2021-03-28
"idris2.so: no such file or directory" error when building with MSYS2 on Win 10


Ryan1729:
I've tried to build on Windows 10, using MSYS2, but after some fiddling around with dependencies and environment variables, I cannot get it to progress past errors about `idris2.so` not being created.

## Installation steps I performed
* Installed MSYS2 fresh, using the `msys2-x86_64-20210228.exe` linked on https://www.msys2.org/
* Installed Chez Scheme 9.5.4 using the exe on [the releases page](https://github.com/cisco/ChezScheme/releases/tag/v9.5.4)
* Added the following to the MSYS2 `.bashrc`:
```bash
PATH=$PATH:$HOME/.idris2/bin
LD_LIBRARY_PATH=$HOME/.idris2/lib
PATH="$PATH:/c/Program Files/Chez Scheme 9.5.4/bin/ta6nt/"
```
( I understand `ta6nt` to refer to the threaded amd64 version)
* Unzipped the Idris2 source code into a folder in the MSYS2 home folder (I tried [0.3.0](https://github.com/idris-lang/Idris2/releases/tag/v0.3.0) and https://github.com/idris-lang/Idris2/commit/ec77ad21abfcc335971c8170e3dbe94372f86431 with similar results.
* Since after adding Chez Scheme to my path, running `scheme` runs that interpreter, I ran `make bootstrap SCHEME=scheme` in the source folder. After the few attempts, I settled on running `rm -rf ~/.idris2/ && make clean && make distclean && make bootstrap SCHEME=scheme` just in case there was something left over from previous attempts. 

## Error messages when building
0.3.0
```
162/162: Building Idris.Main (src\Idris\Main.idr)
make[2]: Entering directory '/home/Ryan/Idris2-0.3.0/tests'
/home/Ryan/Idris2-0.3.0/bootstrap/bin/idris2 --build tests.ipkg
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
1/2: Building Lib (Lib.idr)
2/2: Building Main (Main.idr)
make[2]: Leaving directory '/home/Ryan/Idris2-0.3.0/tests'
[...]

PR #1525, opened 2021-06-09
Put chez executable in quotes on windows, part of #1244

TDecki:
A proper fix will need to update the bootstrap files, but the main issue should be otherwise resolved.

### DIFF
--- a/bootstrap-stage1-chez.sh
+++ b/bootstrap-stage1-chez.sh
@@ -14,8 +14,8 @@ echo "Bootstrapping SCHEME=$SCHEME IDRIS2_VERSION=$IDRIS2_VERSION"
 # Compile the bootstrap scheme
 # TODO: Move boot-build to Makefile in bootstrap/Makefile
 cd bootstrap
-echo "Building idris2-boot from idris2-boot.ss"
-${SCHEME} --script compile.ss
+echo "Building idris2-boot.so from idris2-boot.ss"
+"${SCHEME}" --script compile.ss
 
 # Put the result in the usual place where the target goes
 mkdir -p ../build/exec
--- a/bootstrap-stage1-racket.sh
+++ b/bootstrap-stage1-racket.sh
@@ -11,7 +11,7 @@ echo "Bootstrapping IDRIS2_VERSION=$IDRIS2_VERSION"
 # Compile the bootstrap scheme
 cd bootstrap
 echo "Building idris2-boot from idris2-boot.rkt"
-raco exe idris2_app/idris2-boot.rkt
+"${RACKET_RACO:-raco}" exe idris2_app/idris2-boot.rkt
 
 # Put the result in the usual place where the target goes
 mkdir -p ../build/exec
--- a/bootstrap/idris2-boot.sh
+++ b/bootstrap/idris2-boot.sh
@@ -7,14 +7,15 @@ if [ -z "$SCHEME" ]; then
     exit 1
 fi
 
-if [ "$(uname)" = Darwin ]; then
+if [ "$OS" = windows ] || [ "$OS" = Windows_NT ]; then
+    DIR=$(dirname "$(readlink -f -- "$0" || cygpath -a -- "$0")")
+    PATH=$DIR/idris2_app:$PATH
+elif [ "$(uname)" = Darwin ]; then
     DIR=$(zsh -c 'printf %s "$0:A:h"' "$0")
 else
     DIR=$(dirname "$(readlink -f -- "$0")")
 fi
 
-LD_LIBRARY_PATH="$DIR/idris2_app":$LD_LIBRARY_PATH
-PATH="$DIR/idris2_app":$PATH
-export LD_LIBRARY_PATH PATH
+export LD_LIBRARY_PATH="$DIR/idris2_app:$LD_LIBRARY_PATH"
 
-${SCHEME} --script "$DIR/idris2_app/idris2-boot.so" "$@"
+"${SCHEME}" --script "$DIR/idris2_app/idris2-boot.so" "$@"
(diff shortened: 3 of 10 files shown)


benhormann:
Looks like `raco` and `gsc` also need quoted. I'd inline `quoted_node` (remove the comment) from [src/Compiler/ES/Node.idr](https://github.com/idris-lang/Idris2/blob/abd5432885331390cfde74e944091e677dd4b128/src/Compiler/ES/Node.idr#L52), to match your change.

Wrapping every system command in quotes is not really a good idea?
Say you type `:sh "echo test.cmd"` at the REPL; this should mean `"echo" "test.cmd"`, but you're changing it to mean `"echo test.cmd"` (which appears to do different things depending on the existence of file "echo test.cmd").

TDecki:
The remaing compilers are now quoted too.

As for the quotes inside the `system` function, I'm doing it because of [this](https://superuser.com/questions/238810/problem-with-quotes-around-file-names-in-windows-command-shell).

I'd gladly leave `system` untouched, it's just that at the time of writing, I did not have a idea.

benhormann:
Windows... So do we add `systemRaw` that doesn't wrap (for [src/Idris/REPL.idr#L919](https://github.com/idris-lang/Idris2/blob/ec1e7c0db6b9ddace9b70bdaded87078163b9048/src/Idris/REPL.idr#L919))?

There's more quoting to be done for racket:
- See `findRacoExe`, the `exe` arg needs moved (then rename to `findRaco`).
- Script templates [src/Compiler/Scheme/Racket.idr#L441-L446](https://github.com/idris-lang/Idris2/blob/ec1e7c0db6b9ddace9b70bdaded87078163b9048/src/Compiler/Scheme/Racket.idr#L441-L446)

Haven't tested Gambit, but wouldn't /usr/bin/env need removed?

---

Thinking aloud, is anyone using `CHEZ="\"some path\" some args"`?
(Like how there's `IDRIS2_BOOT='idris2 --codegen racket'` in [INSTALL.md](https://github.com/idris-lang/Idris2/blob/master/INSTALL.md#4-optional-self-hosting-step)).

benhormann:
> [cmd | Microsoft Docs](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd#remarks)
> If the first character is an opening quotation mark, it is stripped along with the closing quotation mark. Any text following the closing quotation marks is preserved.

So we should wrap if the command starts with a quote.
It's even less likely to trip anyone up. It's more like CMD's prompt too.

Thinking longer term, an alternative is to consider switching these commands to `_spawnvp` / `posix_spawnp`.

benhormann:
@TDecki Could prepend `;` instead of adding quotes.
```idris
system cmd = primIO $ prim__system $ if isWindows then ";" ++ cmd else cmd
```

Unlike sh, `;` is a parameter separator. Could choose `,` or `=` instead, but we are using it like `:;` in sh.
It's basically a space that doesn't get trimmed. Ref: https://ss64.com/nt/syntax-esc.html#delimiters

---

> Thinking longer term, an alternative is to consider switching these commands to _spawnvp / posix_spawnp.

Turns out _spawn isn't much help: [Docs | Everyone quotes command line arguments the wrong way
](https://docs.microsoft.com/en-gb/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way#the-c-runtime-library-is-useless)

TDecki:
Thanks for the info.

As a replacement for `system`, do the scheme implementations allow calls to `execve`/`CreateProcessW` ?

benhormann:
Calls to `system` go through FFI (to C or scheme). It could be added as a new primitive to support/c/idris_support.c and base lib. Racket has `system*` but the Chez guide says the system module has "[various undocumented system bindings](https://cisco.github.io/ChezScheme/csug9.5/syntax.html#./syntax:s35)".

> **execve()** does not return on success, and the text, data, bss, and stack of the calling process are overwritten by that of the program loaded.

So it'd probably be `posix_spawnp`. There are all sorts of win32 calls in libidris2_support, so should be OK to add CreateProcess.
Could be good for performance too, e.g. [e0e8403](https://github.com/idris-lang/Idris2/commit/e0e8403317a16f818bbfd15ba415a1f02863c8ca) calls system a lot.

benhormann:
```
++<<<<<<< HEAD
 +         path <- pathLookup ["chez", "chezscheme9.5", "scheme"]
 +         pure $ fromMaybe "scheme" path
++=======
+          path <- pathLookup ["chez", "chezscheme", "chezscheme9.5", "scheme"]
+          pure $ fromMaybe "/usr/bin/env scheme" path
++>>>>>>> upstream/master
```
Can drop my change to remove this conflict, makes no real difference.

Although it would probably be better to abort or emit a warning if pathLookup fails. @gallais ?
It should also do a path lookup if `CHEZ` is only a name. An absolute path is required for the bang in `schHeader`.

I missed `compileChezLibraries`, `compileChezLibrary` & `compileChezProgram` in ChezSep. @TDecki It might be better to wait until #1567 is merged, then rebase.

edwinb:
This looks like it's gone stale, and I'm not sure what it's fixing - also there's some conflicts introduced, as well as a do-not-merge flag. So, I'm going to close it for now, but if there is an issue that it's solving please feel free to reopen (as long as it's a proper fix!)

melted:
This should be reopened. There are trouble with spaces in paths. I think I would prefer to do the quoting in pathLookup, once and for all. As long as were using system it has to be done.

Well, it looks like it can't be reopened due to the branch renaming.

benhormann:
There will be a lot of conflicts now.
_I added the windows fixes here, but ended up merging them via another PR instead._

Better to start a new one with the good bits.
Note that `/usr/bin/env` is a bad fallback (e.g. breaks Chez's shebang). It ought to be fatal when compiler lookup fails.

Could even simplify things by not adding a shebang...
It isn't really needed, since Chez's path can be added to the launcher script. Also easier to edit later.
Maybe there's a reason / preference to keep it?

PR #1526, opened 2021-06-09
[ cleanup ] Test.Golden

gallais:


libs/test/Test/Golden.idr
```diff
@@ -246,15 +247,15 @@ runTest opts testPath = forkIO $ do
         Nothing => putStr $ unlines
           [ "Golden value missing. I computed the following result:"
           , out
-          , "Accept new golden value? [yn]"
+          , "Accept new golden value? [y/N]"
```
fabianhjr, line 250: If the prompt is changed to uppercase it might be good to change L234 to `case toLower str of` to avoid casing confusion.

libs/test/Test/Golden.idr
```diff
@@ -227,7 +234,8 @@ runTest opts testPath = forkIO $ do
       case str of
```
fabianhjr, line 234:
```suggestion
      case toLower str of
```

If the prompt would be changed to `[y/N]` it might help avoid some casing confusion.

gallais: I think it's standard to have upper case used to indicate the default.

fabianhjr: Yeah, but it is more of a "ignore casing just in case".

PR #1527, opened 2021-06-09
Add color to tests

fabianhjr:
image

### DIFF
--- a/libs/test/Test/Golden.idr
+++ b/libs/test/Test/Golden.idr
@@ -64,6 +64,8 @@
 
 module Test.Golden
 
+import Control.ANSI
+
 import Data.Either
 import Data.Maybe
 import Data.List
@@ -92,20 +94,23 @@ record Options where
   onlyNames    : List String
   ||| Should we run the test suite interactively?
   interactive  : Bool
+  ||| Should we use colors?
+  color        : Bool
   ||| Should we time and display the tests
   timing       : Bool
   ||| How many threads should we use?
   threads      : Nat
   ||| Should we write the list of failing cases from a file?
-  failureFile     : Maybe String
+  failureFile  : Maybe String
 
 export
-initOptions : String -> Options
-initOptions exe
+initOptions : String -> Bool -> Options
+initOptions exe color
   = MkOptions exe
               Nothing
               []
               False
+              color
               False
               1
               Nothing
@@ -117,6 +122,7 @@ usage exe = unwords
   , "runtests <path>"
   , "[--timing]"
   , "[--interactive]"
+  , "[--[no-]color, --[no-]colour]"
   , "[--cg CODEGEN]"
   , "[--threads N]"
   , "[--failure-file PATH]"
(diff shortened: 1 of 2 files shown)


libs/test/Test/Golden.idr
```diff
@@ -257,6 +269,10 @@ runTest opts testPath = forkIO $ do
     printTiming True  clock msg = putStrLn (unwords [msg, show clock])
     printTiming False _     msg = putStrLn msg
 
+    wrapColor : Int -> Bool -> String -> String
+    wrapColor color bold s = "\x1b[" ++ (if bold then "1;" else "0;") ++
+      show color ++ "m" ++ s ++ "\x1b[0m"
+
```
andrevidela, line 275: Can this (and its callers) be replaced by using `Libraries.Control.ANSI`?

andrevidela (commented):
Looks good, just one nitpick about colors, good job!

libs/test/Test/Golden.idr
fabianhjr: Hi @andrevidela, I think that since this is a very lightweight use ( only color / bold or not bold) this 3 line function would be better than importing something more feature-rich. If more color/effects/modes are desired in the feature it could be better to use Control.ANSI

andrevidela: I agree, however, contrib is already a dependency of the `test` framework (https://github.com/idris-lang/Idris2/blob/f7f944a2367ebb3c0c882ad30d4f9ea687ea051c/libs/test/test.ipkg) and adding `import Control.ANSI` doesn't seem to have as much of a cost as re-inventing the wheel here.

There is also a benefit in terms of readability `show $ bolden $ colored Green "success"` reads much better than `wrapColor 32 False "success"`

fabianhjr: I am having issues mixing up bolden and colored together since

```
bolden : String -> DecoratedString
colored : Color -> String -> DecoratedString
```

And the only `DecoratedString -> other` is a field accesor or `show = escapeSGR dstr.sgr ++ dstr.str ++ escapeSGR [Reset]`

Will use Control.ANSI without bold for the time being; pushed the changes.

andrevidela: I didn't realise that, maybe that's also the opportunity to revise the `ANSI` api. I think @ShinKage was the one to write it. What do you think @ShinKage ? should we open an issue to update the ANSI API to allow colored and bolded text?

fabianhjr: Well, not gonna bikeshed the use of bold so this could be merged with only the color changes.

andrevidela: Yes

ShinKage: We can open a PR to improve the ANSI API and add the change there.
On the matter of the API, you can already have bolded and colored text, but you have to construct or unwrap `DecoratedString` manually. We can probably add a function that merges the decoration annotations of multiple helpers like bolden or colored, just for ergonomics.

fabianhjr: Merginig would be great, yeah I could have done that wraping/unwrapping but then the 2 line wrapColor helper would have ended shorter than using Control.ANSI. unu

libs/test/Test/Golden.idr
```diff
@@ -154,7 +164,8 @@ options args = case args of
 
     mkOptions : String -> List String -> IO (Maybe Options)
     mkOptions exe rest
-      = do let Just (mfp, opts) = go rest Nothing (initOptions exe)
+      = do color <- (Just "DUMB" /=) <$> getEnv "TERM"
```
gallais, line 167: We should probably add a function to ANSI to check whether a terminal
supports highlighting. We already have at least 2 occurrences of this
recipe in the codebase and if we need to update it at some point, chances
are we'll only update one version.

PR #1528, opened 2021-06-09
Add Vect drop function and fix bug with signal C code

mattpolzin:
Two unrelated commits -- I can drop (heh) the `drop` function from this PR if folks don't want it in base.

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -88,10 +88,15 @@ take 0 xs = Nil
 take (S k) (x :: xs) = x :: take k xs
 
 ||| Drop the first `n` elements of a Vect.
-drop : (n : Nat) -> Vect l elem -> Vect (l `minus` n) elem
-drop 0 xs = rewrite minusZeroRight l in xs
-drop (S k) [] = rewrite minusZeroLeft (S k) in []
+drop : (n : Nat) -> Vect (n + m) elem -> Vect m elem
+drop 0 xs = xs
 drop (S k) (x :: xs) = drop k xs
+
+||| Drop up to the first `n` elements of a Vect.
+drop' : (n : Nat) -> Vect l elem -> Vect (l `minus` n) elem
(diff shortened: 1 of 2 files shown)


libs/base/Data/Vect.idr
```diff
@@ -87,6 +87,12 @@ take : (n  : Nat)
 take 0 xs = Nil
 take (S k) (x :: xs) = x :: take k xs
 
+||| Drop the first `n` elements of a Vect.
+drop : (n : Nat) -> Vect l elem -> Vect (l `minus` n) elem
```
gallais, line 91: I'd rather have `drop : (n : Nat) -> Vect (n + m) elem -> Vect m elem`
to have it follow the same convention as `take`.

buzden: Originally suggested definition with `minus` has a property that `drop 4 []` is correctly `[]` and is an incorrect call for what you suggest. I don't know whether is it important or not, but is a visible change between the signatures.

mattpolzin: I think both are useful, personally. I had thought of both, maybe as `drop` and `drop'` (not sure which would be which), but I only ended up writing the one that I actually needed in the project I was working on. I had a type like 
```idris
Fin (totalSize `minus` visibleSize)
```
So having a drop definition that similarly cared less about bottoming out at zero and more about offsetting the Vect as much as possible made sense (i.e. fit the types perfectly).

mattpolzin: I'd be happy to add both here, or change to the `n + m` definition.

mattpolzin: @gallais I added the `(n + m)` definition in addition to renaming the "drop up to `n`" definition `drop'`.

I'm glad you pushed back, I do agree that the `(n + m)` definition feels more at home with `Vect` so I like it as `drop` with any alternative definition having a different name.

As I mentioned above, if you still feel the `minus` definition is out of place (at least in base) I am happy to push another commit to remove it entirely.

PR #1535, opened 2021-06-11
[ new ] detect changes using `sha256` rather than timestamps

### LINKED ISSUE #23: REPL 'last change' time-stamp should be taken from start of reload
<a href="https://github.com/ohad"></img></a> **Issue by [ohad](https://github.com/ohad)**
_Friday Oct 18, 2019 at 12:17 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/127_

----

# Steps to Reproduce
Take a file that takes idris2 a while to type-check, say `Foo.idr`.
While idris2 is type-checking, make a change in the file and save before idris2 finishes.
Reload the file.

# Expected Behavior
```
Main> :r
1/1 Building Main (Foo.idr)
Main> 
```
# Observed Behavior
```
Main> :r
Main> 
```

I think this is because the timestamp on the last change to this file is taken after type-checking finishes, and it should be taken before type-checking starts.

### LINKED ISSUE #1519: Change `needsBuilding` from `ttcTime` to a `ttcHash`
While experimenting with caching in #1517 there is an issue where the Idris2 compiler will use modification dates rather than content hashes so:

- When a source file has not been modified but has been checked out by git more recently than its last modification it will have a more recent date than the ttcTime so Idris2 will rebuild unnecessarily. (Eg, `make`, `git checkout v0.2.0`, `git checkout -`, `make`)
- If the modification time is set to the commit time by a script (or someone resets the file to a previous version and somehow ends with a previous modification time) then Idris2 would not rebuild source files previous dates than those on the ttcFile even if they have changed and need rebuilding.

A challenge would be that there is currently no hash function in Idris2 so I am opening this as discussion and for tracking.

https://github.com/idris-lang/Idris2/blob/7aee7c9b7c50d16461e8d96d22d4ced2f19a52ef/src/Idris/ModTree.idr#L177-L180

fabianhjr:
Closes #1519
Fixes #23 (modification time is not stored but read from the filesystem)

### DIFF
--- a/tests/idris2/import001/run
+++ b/tests/idris2/import001/run
@@ -2,5 +2,8 @@ $1 --no-color --console-width 0 --no-banner --no-prelude Test.idr < input
 sleep 1
 touch Mult.idr
 $1 --no-color --console-width 0 --no-banner --no-prelude Test.idr < input
+sleep 1
+touch Mult.idr
+$1 --no-color --console-width 0 --no-banner --no-prelude -Xcheck-hashes Test.idr < input
 
 rm -rf build
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -17,8 +17,6 @@ import Libraries.Utils.Path
 import Data.List
 import Data.List1
 import Data.Maybe
-import Libraries.Data.NameMap
-import Libraries.Data.Version
 import Data.Strings
 import Data.Vect
 
@@ -28,6 +26,10 @@ import System
 import System.Directory
 import System.File
 import System.Info
+
+import Libraries.Data.NameMap
+import Libraries.Data.Version
+import Libraries.Utils.String
 
 %default covering
 
@@ -80,15 +82,6 @@ findLibs ds
              then Just (trim (substr 3 (length d) d))
              else Nothing
 
-export
-escapeString : String -> String
-escapeString s = pack $ foldr escape [] $ unpack s
-  where
-    escape : Char -> List Char -> List Char
-    escape '"' cs = '\\' :: '\"' :: cs
-    escape '\\' cs = '\\' :: '\\' :: cs
-    escape c   cs = c :: cs
-
 schHeader : String -> List String -> String
 schHeader chez libs
   = (if os /= "windows" then "#!" ++ chez ++ " --script\n\n" else "") ++
@@ -101,7 +94,7 @@ schHeader chez libs
     "  [(i3nt ti3nt a6nt ta6nt) (load-shared-object \"msvcrt.dll\")" ++
     "                           (load-shared-object \"ws2_32.dll\")]\n" ++
     "  [else (load-shared-object \"libc.so\")])\n\n" ++
-    showSep "\n" (map (\x => "(load-shared-object \"" ++ escapeString x ++ "\")") libs) ++ "\n\n" ++
+    showSep "\n" (map (\x => "(load-shared-object \"" ++ escapeStringChez x ++ "\")") libs) ++ "\n\n" ++
     "(let ()\n"
 
 schFooter : Bool -> String
@@ -246,7 +239,7 @@ cCall appdir fc cfn clib args ret collectSafe
                            copyLib (appdir </> fname, fullname)
                            put Loaded (clib :: loaded)
                            pure $ "(load-shared-object \""
-                                    ++ escapeString fname
+                                    ++ escapeStringChez fname
                                     ++ "\")\n"
          argTypes <- traverse (\a => cftySpec fc (snd a)) args
          retType <- cftySpec fc ret
--- a/src/Compiler/Scheme/ChezSep.idr
+++ b/src/Compiler/Scheme/ChezSep.idr
@@ -21,8 +21,6 @@ import Libraries.Utils.Path
 import Data.List
 import Data.List1
 import Data.Maybe
-import Libraries.Data.NameMap
-import Libraries.Data.Version
 import Data.Strings
 import Data.Vect
 
@@ -32,6 +30,10 @@ import System
 import System.Directory
 import System.File
 import System.Info
+
+import Libraries.Data.NameMap
+import Libraries.Data.Version
+import Libraries.Utils.String
 
 %default covering
 
@@ -46,7 +48,7 @@ schHeader libs compilationUnits = unlines
   , "  [(i3nt ti3nt a6nt ta6nt) (load-shared-object \"msvcrt.dll\")"
   , "                           (load-shared-object \"ws2_32.dll\")]"
   , "  [else (load-shared-object \"libc.so\")]"
-  , unlines ["  (load-shared-object \"" ++ escapeString lib ++ "\")" | lib <- libs]
+  , unlines ["  (load-shared-object \"" ++ escapeStringChez lib ++ "\")" | lib <- libs]
   , ")"
   ]
 
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -1,3 +1,6 @@
+||| Reading and writing 'Defs' from/to a binary file. In order to be saved, a
+||| name must have been flagged using 'toSave'. (Otherwise we'd save out
+||| everything, not just the things in the current file).
 module Core.Binary
 
 import Core.CaseTree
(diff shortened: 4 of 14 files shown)


fabianhjr:
I am blocked by Windows, if someone is a Windows dev and could help with the sha256 function it would be appreciated.

edwinb:
Something I often do while working on Idris is `touch` a file to force it to rebuild. Or, just resave it. This probably isn't a common thing to want to do, but even so, what should I do instead now? I guess remove the relevant ttc?

fabianhjr:
Hi @edwinb, an easier option than hunting down the idris ttc file to might be to add a blank line or space in the file you are working on. I did some testing on my fork and got this time differences:

image
image
image
image

Comparing this runs:
 - https://github.com/idris-lang/Idris2/actions/runs/925170679
 - https://github.com/LibreCybernetics/Idris2/actions/runs/927413234
 
 Since its intention is to reduce friction to contributing by speeding up the CI (Particularly time to first error and completion time) if it causes more inconvenience on the coding stage it would be counterproductive to the intention.

andorp:
What if we hide one behind an arg?

fabianhjr:
Yes, an arg is also a possibility.

I have no preference as a default so to keep backwards compatibility the sha256 based one could be the one that needed a toggle. Any preference for a flag? (`-Xcheck-hash`)

benhormann:
Quoting and escaping is different on windows, thanks to cmd.exe. But if the paths are relative you don't need to.

Isn't SHA2 overkill? Would've thought MD5 or even CRC is good for this job.
Only `cksum` is standard. Even macOS and BSD have different names for SHA. Could upgrade to a FFI call later.

fabianhjr:
> Isn't SHA2 overkill? Would've thought MD5 or even CRC is good for this job.

MD5, SHA-1, and SHA-2 (256 to 512) are in the same magnitude of performance: https://automationrhapsody.com/md5-sha-1-sha-256-sha-512-speed-performance/ (In the comments there is even some pointers to get SHA-256 to be faster than MD5 and SHA-1)

I don't want to risk collisions (either accidental or as an intentional attack) so CRC is out.

> Only `cksum` is standard.

`sha256sum` is on the same standard (corutils) as `cksum` ( https://www.gnu.org/software/coreutils/manual/coreutils.html#sha2-utilities ) and comes preinstalled on most UNIX based systems. (BSDs, macOS, Linux, etc)

> Quoting and escaping is different on windows, thanks to cmd.exe. But if the paths are relative you don't need to.

Oh ok, then it should be easy to fix and even windows has `sha256sum` preinstalled so it seems like this could get to a mergeable state :3

> Could upgrade to a FFI call later.

Yes that would probably help with the slight performance penalty of launching trough a shell and piping console IO. (And the path escaping or not)

benhormann:
> ... comes preinstalled on most UNIX based systems. (BSDs, macOS, Linux, etc)

But it doesn't, only Linux and MSYS. Neither is gmp.
Turns out `shasum` is implemented in Perl on macOS.
I'll pipe down now.

fabianhjr:
Ok, this new behavior is now behind an experimental flag (`-Xcheck-hashes`)

There is still a need for a windows Idris-dev since I don't have a windows machine and removing the escape didn't solve the issue. (I could leave the windows version unimplemented and leave a constant string in place but that would be less than ideal. UnU)

benhormann:
TL;DR  `popen "certutil -hashfile \"" ++ fileName ++ "\""`

Example output:
```
SHA1 hash of file C:\Windows\System32\cmd.exe:
0f 3c 4f f2 8f 35 4a ed e2 02 d5 4e 9d 1c 55 29 a3 bf 87 d8
CertUtil: -hashfile command completed successfully.
```
Assumes no `%` in fileName. Can append " SHA256" to upgrade. [certutil -hashfile](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/certutil#-hashfile)

---

Or if not quoting, here are my best guesses (untested):

/bin/sh (ref: https://unix.stackexchange.com/a/270979)
```idr
escapeUnquotedSh s = pack $ foldr escape [] $ unpack s
  where
    special : List Char
    special = unpack ";<=>?@[\]^`{|}~"
    escape : Char -> List Char -> List Char
    escape c cs = if c < '+' || isInfixOf [c] special
     then '\\' :: c :: cs 
     else c :: cs
```
/bin/sh
```idr
escapeQuotedSh s = pack $ foldr escape [] $ unpack s
  where
    escape : Char -> List Char -> List Char
    escape '\"' cs = '\\' :: '\"' :: cs
    escape '\\' cs = '\\' :: '\\' :: cs
    escape '\$' cs = '\\' :: '\$' :: cs
    escape '\`' cs = '\\' :: '\`' :: cs
    escape c    cs = c :: cs
```
C:\Windows\System32\cmd.exe (ref: https://ss64.com/nt/syntax-esc.html)
```idr
escapeUnquotedCmd s = pack $ foldr escape [] $ unpack s
  where
    escape : Char -> List Char -> List Char
    escape '^' cs = '^' :: '^' :: cs
    escape '%' cs = '^' :: '%' :: cs
    escape '!' cs = '^' :: '!' :: cs
    escape '*' cs = '^' :: '*' :: cs
    escape '?' cs = '^' :: '?' :: cs
    escape ';' cs = '^' :: ';' :: cs
    escape '&' cs = '^' :: '&' :: cs -- illegal in paths
    escape '|' cs = '^' :: '|' :: cs -- illegal in paths
    escape '<' cs = '^' :: '<' :: cs -- illegal in paths
    escape '>' cs = '^' :: '>' :: cs -- illegal in paths
    escape ' ' cs = '^' :: ' ' :: cs
    escape '\n' cs = '^' :: '\n' :: cs
    escape '\t' cs = '^' :: '\t' :: cs
    escape c cs = c :: cs
```
For paths, could simply drop the illegal chars.

/usr/local/bin/idris2
```idr
escapeString s = pack $ foldr escape [] $ unpack s
  where
    escape : Char -> List Char -> List Char
    escape '\"' cs = '\\' :: '\"' :: cs
    escape '\\' cs = '\\' :: '\\' :: cs
    escape c  cs = c :: cs
```
Or, for Multi-line / Single-line literals:
```idr
escapeString ml s = pack $ foldr escape ml [] $ unpack s
  where
    escape : Bool -> Char -> List Char -> List Char
    escape False '\n' cs = '\\' :: 'n'  :: cs
    escape False '\r' cs = '\\' :: 'r'  :: cs
    escape _     '\"' cs = '\\' :: '\"' :: cs
    escape _     '\\' cs = '\\' :: '\\' :: cs
    escape _     c    cs = c :: cs
```
Do we actually need `escapeString`?

fabianhjr:
Thanks @benhormann will use what you documented to rewrite the escapes and select the more complete ones for Windows Unquoted and POSIX/*NIX Unquoted.

benhormann:
@fabianhjr
`C:\> vim ab^ cd ef` results in three files to edit...
`C:\> vim "ab cd" ef` for two files.

Escaping doesn't work for args, but does for > redirection. Quoting make more sense for file paths anyway.
Even for sh, not many people are going to use a path with <code>$\\"`</code>. It'd break the makefiles too.

benhormann:
@fabianhjr Escaping doesn't work inside quotes. Remove `escapeStringWindows`.

tests/idris2/import001/run
```diff
@@ -2,5 +2,8 @@ $1 --no-color --console-width 0 --no-banner --no-prelude Test.idr < input
 sleep 1
 touch Mult.idr
 $1 --no-color --console-width 0 --no-banner --no-prelude Test.idr < input
+sleep 1
+touch Mult.idr
+$1 --no-color --console-width 0 --no-banner --no-prelude -Xcheck-hashes Test.idr < input
```
fabianhjr, line 8: Added a test case of -Xcheck-hashes

fabianhjr:
> @fabianhjr Escaping doesn't work inside quotes. Remove escapeStringWindows.

Removed, thanks for the help. `osEscape` now sends Windows to id and other to escapeUnix (`\` and `"` with `\` as escape char)

fabianhjr:
Hey @edwinb, @gallais, I would appreciate comment on the high level design ( usage of -X flag or any change that might be desired)

benhormann:
Any chance of `-Xcheck-hashes` optionally taking a command name? E.g. 
- `idris2 -Xcheck-hashes --build ...`
- `idris2 -Xcheck-hashes gsha256sum --build ...`

fabianhjr:
Don't think so but on Linux I would use a symbolic link.

I have no way to validate the windows behaviour other than the CI. :C

On the other hand that would be a great PR/contribution to improve the windows experience.

fabianhjr:
Oh, :o great to have this merged. @benhormann I can open a new PR for the windows certutil option or if you prefer I can help you with that PR. (Both as an explicit option or maybe as a fallback for windows)

benhormann:
> Don't think so but on Linux I would use a symbolic link.

Same for macOS or BSD. Outside of MSYS, you'd find an exe to put on your PATH.

> but I don't feel confident increasing the complexity of this PR and keeping it mergable.

Fair call.

---

R.I.P. ⚰️ `67 /196: Building IdrisPaths (src/IdrisPaths.idr)` as a progress indicator. Good job 👍

PR #1536, opened 2021-06-11
Fix deadlocks [Rebased, Squashed]

fabianhjr:
This is a rebase + squashes of @Russoul and @abailly work to fix deadlocks.

Supersedes #1298

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -197,19 +197,25 @@ cftySpec fc (CFStruct n t) = pure $ "(* " ++ n ++ ")"
 cftySpec fc t = throw (GenericMsg fc ("Can't pass argument of type " ++ show t ++
                          " to foreign function"))
 
-cCall : {auto c : Ref Ctxt Defs} ->
-        {auto l : Ref Loaded (List String)} ->
-        String -> FC -> (cfn : String) -> (clib : String) ->
-        List (Name, CFType) -> CFType -> Core (String, String)
-cCall appdir fc cfn clib args (CFIORes CFGCPtr)
+cCall : {auto c : Ref Ctxt Defs}
+     -> {auto l : Ref Loaded (List String)}
+     -> String
+     -> FC
+     -> (cfn : String)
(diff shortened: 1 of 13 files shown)


src/Compiler/Scheme/Chez.idr
```diff
@@ -282,19 +312,25 @@ schemeCall fc sfn argns ret
...
            Just ("scheme", [sfn]) =>
                do body <- schemeCall fc sfn (map fst args) ret
                   pure ("", body)
-           Just ("C", [cfn, clib]) => cCall appdir fc cfn clib args ret
-           Just ("C", [cfn, clib, chdr]) => cCall appdir fc cfn clib args ret
-           _ => throw (NoForeignCC fc)
+           Just ("C__collect_safe", (cfn :: clib :: _)) => do
+             when (unsupportedCallingConvention version) $
+               throw (GenericMsg fc $ "Chez version " ++ show version ++  " is undefined or does not support '__collect_safe' calling convention")
+             cCall appdir fc cfn clib args ret True
```
fabianhjr, line 328: Here is the a check of support by @andorp

gallais, line 328: Wasn't the plan to keep generating code without the collect safe calling convention if
the available scheme wasn't recent enough?

fabianhjr: I think I could create an extra commit on top of Russoul's and Abailly's work to change the detection from an error if unsupported to switch to the safe on collect if supported.

fabianhjr: Changed to:

```
             if unsupportedCallingConvention version
               then cCall appdir fc cfn clib args ret False
               else cCall appdir fc cfn clib args ret True
```

fabianhjr: Ended up amending rather than creating an extra commit.

abailly-iohk:
Thanks @fabianhjr for following up on this!

fabianhjr:
You are welcome, thanks a lot for your work! UwU

PR #1538, opened 2021-06-11
Split RefC/Scheme buffers

madman-bob:
Fully separate the RefC and Scheme representations of `Buffer`s.

I've added the `RefC` language specifier, to distinguish RefC FFIs from common FFIs usable by all backends using C.

Further, I've removed the `freeBuffer` function, making each backend fully responsible for its own memory management. In practice it wasn't used anyway, as RefC was the only backend requiring it, causing memory leaks if you happened to compile with RefC. Instead, RefC now GCs `Buffer`s in the same way it GCs everything else.


src/Compiler/RefC/RefC.idr
```diff
@@ -887,7 +887,7 @@ createCFunctions n (MkACon tag arity nt) = do
 
 
 createCFunctions n (MkAForeign ccs fargs ret) = do
-  case parseCC ["C"] ccs of
+  case parseCC ["RefC", "C"] ccs of
```
gallais, line 890: Given that `RefC` will come first but `C` will act as a fallback, is it really necessary
to make all the foreign bindings in `System.Clock`, `String.Iterator`, etc. `RefC`-only?

madman-bob: Given that they can't actually be used by other backends, yes.

That said, with some rearrangement, they could be made to work for other backends. I based my tagging on the current state of the code.

PR #1540, opened 2021-06-11
Return Maybe from strengthen

nickdrozd:
If the bound can't be strengthened, the current bound is returned
as-is in Left. But the current bound is already available, so the
returned bound isn't needed. Just return Nothing in that case.

### DIFF
--- a/src/Libraries/Data/Fin.idr
+++ b/src/Libraries/Data/Fin.idr
@@ -1,0 +1,16 @@
+module Libraries.Data.Fin
+
+import public Data.Fin
+
+%default total
+
+-- TODO: Remove this, once Data.Fin.strengthen from base is available
+--       to the compiler
+
+export
+strengthen : {n : _} -> Fin (S n) -> Maybe (Fin n)
+strengthen {n = S _} FZ = Just FZ
+strengthen {n = S _} (FS p) with (Libraries.Data.Fin.strengthen p)
+  strengthen (FS _) | Nothing = Nothing
+  strengthen (FS _) | Just q  = Just $ FS q
+strengthen _ = Nothing
--- a/src/TTImp/ProcessBuiltin.idr
+++ b/src/TTImp/ProcessBuiltin.idr
@@ -80,7 +80,7 @@ getNEIndex : (arity : Nat) -> Term vars -> Maybe (Fin arity)
 getNEIndex ar (Bind _ x b tm) = case b of
     Let _ _ val _ => getNEIndex ar $ subst {x} val tm
     Pi _ mul _ arg => if isErased mul
-        then getNEIndex ar tm >>= \k => Libs.strengthen (FS k)
+        then getNEIndex ar tm >>= Libs.strengthen . FS
         else natToFin 0 ar
     _ => Nothing
 getNEIndex _ _ = Nothing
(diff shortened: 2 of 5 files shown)


nickdrozd:
This PR changes a function in `libs` as well as a call site in `src`. I think that's causing the `self-host-previous-version` test to fail. How should that be handled?

src/TTImp/ProcessBuiltin.idr
```diff
@@ -3,7 +3,7 @@
 module TTImp.ProcessBuiltin
 
 import Libraries.Data.Bool.Extra
-import Data.Fin
+import Libraries.Data.Fin
```
fabianhjr, line 6: It might help to do

```suggestion
import Libraries.Data.Fin as Libs
```

src/TTImp/ProcessBuiltin.idr
```diff
@@ -80,10 +80,7 @@ getNEIndex : (arity : Nat) -> Term vars -> Maybe (Fin arity)
 getNEIndex ar (Bind _ x b tm) = case b of
     Let _ _ val _ => getNEIndex ar $ subst {x} val tm
     Pi _ mul _ arg => if isErased mul
-        then getNEIndex ar tm >>=
-            \k => case strengthen (FS k) of
-                Left _ => Nothing
-                Right k' => Just k'
+        then getNEIndex ar tm >>= \k => strengthen (FS k)
```
fabianhjr, line 83: And use it here as

```suggestion
        then getNEIndex ar tm >>= \k => Libs.strengthen (FS k)
```

src/Libraries/Data/Fin.idr
```diff
@@ -0,0 +1,16 @@
...
+-- TODO: Remove this, once Data.Fin.strengthen from base is available
+--       to the compiler
+
+public export
+strengthen : {n : _} -> Fin (S n) -> Either (Fin (S n)) (Fin n)
+strengthen {n = S _} FZ = Right FZ
+strengthen {n = S _} (FS i) with (Libraries.Data.Fin.strengthen i)
+  strengthen (FS _) | Left x  = Left $ FS x
+  strengthen (FS _) | Right x = Right $ FS x
+strengthen f = Left f
```
fabianhjr, line 16: Use your new function here rather than the previous one.

nickdrozd:
Thanks @fabianhjr, it looks like that worked.

fabianhjr:
You are welcome @nickdrozd, glad I could help

src/TTImp/ProcessBuiltin.idr
```diff
@@ -80,10 +80,7 @@ getNEIndex : (arity : Nat) -> Term vars -> Maybe (Fin arity)
 getNEIndex ar (Bind _ x b tm) = case b of
     Let _ _ val _ => getNEIndex ar $ subst {x} val tm
     Pi _ mul _ arg => if isErased mul
-        then getNEIndex ar tm >>=
-            \k => case strengthen (FS k) of
-                Left _ => Nothing
-                Right k' => Just k'
+        then getNEIndex ar tm >>= \k => Libs.strengthen (FS k)
```
fabianhjr, line 83:
```suggestion
        then getNEIndex ar tm >>= Libs.strengthen . FS
```

Not really a requirement but checkout out your branch and just realized it is easy to turn it dot-free.

PR #1548, opened 2021-06-13
Use map for some maybes

nickdrozd:


src/Compiler/ES/TailRec.idr
```diff
@@ -318,11 +318,10 @@ makeFusionBranch fusionName functionsIdx (i, _, args, body) =
...
         rep (IEApp (IEVar n) arg_vals) =
-            case lookup n functionsIdx of
-                Nothing => Nothing
-                Just i => Just $ IEApp
-                          (IEVar fusionName)
-                          [IEConstructor (Left $ cast i) arg_vals]
+          map
+            (\i => IEApp (IEVar fusionName)
+                         [IEConstructor (Left $ cast i) arg_vals])
+            (lookup n functionsIdx)
```
gallais, line 324: I don't think we should sacrifice readability just for the sake of using `map`.

src/Core/Unify.idr
```diff
@@ -341,12 +341,11 @@ patternEnv {vars} env args
...
-              Nothing => pure Nothing
-              Just vs =>
-                 let (newvars ** svs) = toSubVars _ vs in
-                     pure (Just (newvars **
-                                     (updateVars vs svs, svs)))
+         pure $ map
+           (\vs =>
+             let (newvars ** svs) = toSubVars _ vs in
+               (newvars ** (updateVars vs svs, svs)))
+           (getVars [] args')
```
gallais, line 348: Same comment here

src/Core/Unify.idr
```diff
@@ -376,12 +375,11 @@ patternEnvTm : {auto c : Ref Ctxt Defs} ->
...
-              Nothing => pure Nothing
-              Just vs =>
-                 let (newvars ** svs) = toSubVars _ vs in
-                     pure (Just (newvars **
-                                     (updateVars vs svs, svs)))
+         pure $ map
+           (\vs =>
+             let (newvars ** svs) = toSubVars _ vs in
+               (newvars ** (updateVars vs svs, svs)))
+           (getVarsTm [] args)
```
gallais, line 382: And here

PR #1550, opened 2021-06-14
[ base ] add List functions

alissa-tung:


libs/base/Data/List.idr
```diff
@@ -111,12 +111,49 @@ filter p (x :: xs)
...
+
+public export
+filterWith' : (a -> Bool) -> List a -> List (a, Nat)
+filterWith' p = h 0 where
+  h : Nat -> List a -> List (a, Nat)
+  h _ [] = []
+  h lvl (x :: xs) = if p x
+    then (x, lvl) :: h (S lvl) xs
+    else             h (S lvl) xs
+
```
gallais, line 131: These seem like very ad-hoc function whose name is not very enlightening.
Other occurrences of `xWith` in the libs seem to be for functions of the shape:

```idris
xWith : Constraint b => (a -> b) -> f a -> g a
```

I think the name should be different & this should be in (at most) contrib.

alissa-tung: thx

libs/base/Data/List.idr
```diff
@@ -117,6 +119,25 @@ find : (p : a -> Bool) -> (xs : List a) -> Maybe a
...
+||| Find indices of all elements that satisfy the given test.
+public export
+findIndices : (a -> Bool) -> List a -> List Nat
+findIndices p = h 0 where
+  h : Nat -> List a -> List Nat
+  h _         []  = []
+  h lvl (x :: xs) = if p x
+    then lvl :: h (S lvl) xs
+    else        h (S lvl) xs
+
```
Russoul, line 140: I think these better return an informative proof that the natural number is actually an index into the list,
like `InBounds`:
https://github.com/idris-lang/Idris2/blob/0209cee41a2487f42dc81dc9755cab4b87419a10/libs/base/Data/List.idr#L40

alissa-tung: thx

alissa-tung: InBounds for snoc too

alissa-tung:
Should we refactor `InBounds` a little bit?

libs/base/Data/SnocList.idr
```diff
@@ -96,6 +95,35 @@ public export
 Monoid (SnocList a) where
   neutral = Lin
 
+public export
+Foldable SnocList where
```
gallais, line 99: This instance is inconsistent with the `Traversable` one.
* this insists that `SnocList` is just a normal list with a flipped `(::)`
* but traversable insists that `SnocList` is a reversed list

libs/base/Data/List.idr
```diff
@@ -117,6 +119,27 @@ find : (p : a -> Bool) -> (xs : List a) -> Maybe a
 find p [] = Nothing
 find p (x::xs) = if p x then Just x else find p xs
 
+||| Find the index and proof of InBounds of the first element (if exists) of a
+||| list that satisfies the given test, else `Nothing`.
+public export
+findIndex : (a -> Bool) -> (xs : List a) -> Maybe (n : Nat ** InBounds n xs)
```
buzden, line 125: Why not to use `Subset Nat (\n => InBounds n xs)` insead of `(n : Nat ** InBounds n xs)` in the return type, just to not to have an `InBounds` at runtime? As far as I remember, `InBounds` argument has always `0` quantity in functions that use it (like `index`), so there is no need to have it at runtime.

Another similar question, why not to go more radical and to have a `Fin (length xs)` instead of `(n : Nat ** InBounds n xs)` or `Subset Nat (\n => InBounds n xs)`. As far as we have a runtime optimisation for `Fin`s to be fast integers, it feels plausible. (Moreover, I personally have a finding function returning `Fin (length xs)` in my codebase and feel it better than `InBounds` version). However, `Fin`-based vs. `InBounds`-based approaches can be a holy war.

Russoul: I didn't mention `Subset` only because we don't have sugar for it, therefore I personally consider it painful to use...

Russoul: In case of `Fin (length xs)`, `Fin` is indexed by a computation. I don't have a feel for particular circumstances when it can turn out to be a nuisance. But I do know that sometimes in makes computations stuck which is to be avoided...

gallais:
Ugh, another PR from an org which means I can't push to it. :(

alissa-tung:
> Ugh, another PR from an org which means I can't push to it. :(
> I've opened a PR there: [alissa-tung#1](https://github.com/alissa-tung/Idris2/pull/1)

Merged. thx QAQ sorry i am suffering my exam week these days
and transfered to alissa-tung/Idris2

PR #1553, opened 2021-06-14
Improve case-splitting formatting to not introduce new whitespace; add a bit of comments+docs.

CodingCellist:
```idris2
f : {n : Nat} -> Nat
f {n = a} = ?f_rhs
```

# Expected behaviour
Case-splitting on `a` produces
```idris
f {n = Z} = ?f_rhs_1
f {n = (S k)} = ?f_rhs_2
```

# Observed behaviour
Case-splitting on `a` produces
```idris
f {  n = Z} = ?f_rhs_1
f {  n = (S k)} = ?f_rhs_2
```
introducing indentation where there previously was none.

This was due to the case-splitting moving all whitespace to the start of the new expression (so the two spaces in `{n = a}` would get moved to the front and then re-introduced). This new implementation correctly preserves the indentation, even when whitespace is present in the expression along with whitespace at the end of the expression. Additional tests are added to check this, and the code has been commented a bit for readability.


src/Idris/IDEMode/CaseSplit.idr
```diff
@@ -65,23 +65,36 @@ doUpdates : {auto u : Ref UPD (List String)} ->
...
 doUpdates defs ups (LBrace :: xs)
-    = let (ws, nws) = spanSpace xs in map (LBrace :: ws ++) $
+    -- the cases we care about are easy to detect w/o whitespace, so separate it
+    = let (ws, nws) = span isWhitespace xs in
          case nws of
            Name n :: RBrace :: rest =>
+                -- expand something in braces into its cases, e.g.
+                -- { x} where x : Nat, becomes { x = Z} (and later { x = (S k)})
+                map (LBrace :: ws ++) $
                 pure (Name n ::
```
gallais, line 75:
```suggestion
                pure (LBrace :: ws ++ Name n ::
```

src/Idris/IDEMode/CaseSplit.idr
```diff
@@ -65,23 +65,36 @@ doUpdates : {auto u : Ref UPD (List String)} ->
...
                       Whitespace " " :: Equal :: Whitespace " " ::
                       !(doUpdates defs ups (Name n :: RBrace :: rest)))
            Name n :: Equal :: rest =>
+                -- update the RHS of the Equal, e.g. `rest` in {x = rest}
+                map (LBrace :: ws ++) $
                 pure (Name n ::
```
gallais, line 81:
```suggestion
                pure (LBrace :: ws ++ Name n ::
```

src/Idris/IDEMode/CaseSplit.idr
```diff
@@ -65,23 +65,36 @@ doUpdates : {auto u : Ref UPD (List String)} ->
...
                       Whitespace " " :: Equal :: Whitespace " " ::
                       !(doUpdates defs ups rest))
-           _ => doUpdates defs ups xs
+           Name n :: Whitespace s :: RBrace :: rest =>
```
gallais, line 84: I suppose we could also have some whitespace before `Equal`.
It may be better to restructure the whole block as:

```idris
    = let (ws, nws) = span isWhitespace xs in
      case nws of
        Name n :: rest => 
          let (ws2, nws2) = span isWhitespace rest in
          case nws2 of
             Equal ... =>
			 RBrace ... => 
        _ => map (LBrace ::) $ doUpdates defs ups xs
```

CodingCellist: Aah yes, I had been thinking of that but wasn't quite sure how to proceed. I'll try this asap : )

src/Idris/IDEMode/CaseSplit.idr
CodingCellist: Done : )

src/Idris/IDEMode/CaseSplit.idr
CodingCellist: Done : )

src/Idris/IDEMode/CaseSplit.idr
CodingCellist: Done : )

src/Idris/IDEMode/CaseSplit.idr
```diff
@@ -60,37 +60,69 @@ toStrUpdate _ = pure [] -- can't replace non user names
...
+        case nws of
+          -- brace is immediately closed, so generate a new pattern-match on the
+          -- values the name can have, e.g. { x} where x : Nat would become
+          -- { x = Z}  (and later { x = (S k)})
+          Name n :: RBrace :: rest =>
+             pure (LBrace :: ws ++   -- preserve whitespace
+                   Name n ::
+                   Whitespace " " :: Equal :: Whitespace " " ::
+                   !(doUpdates defs ups (Name n :: RBrace :: rest))
+                   )
```
gallais, line 83: Isn't this case handled by the similar one after line 87's `span isWhitespace rest`?

CodingCellist: Ah, so it is. Thanks for spotting that ^^;

PR #1554, opened 2021-06-14
[ refactor ] Right-align padded numbers.

CodingCellist:
Just a little pet peeve of mine, so feel free to ignore if you disagree, but this looks nicer imo (and it's also the usual way of formatting progress trackers AFAIK) ^^


gallais:
booooooo :-1:

PR #1555, opened 2021-06-14
[test] Add preformance test of Fin/Vect

fabianhjr:
Closes idris-lang/Idris2#16


fabianhjr:
Ah messed up  and missed the expected output

fabianhjr:
Empty test (very small N) takes close to 1.2 seconds on my machine. Increased n to a bigger value (500) which takes roughly 2 seconds on my machine.

tests/idris2/perf008/run
```diff
@@ -0,0 +1,3 @@
+$1 --check FinPerf.idr
+
```
gallais, line 2: Don't you want to run the program too? Otherwise what's the point of the `main`?

fabianhjr: I thought the issue was during typechecking, will change the code regardless.

fabianhjr: Changed to `--exec main` instead of `--check`. Same performance (roughly 2 seconds for N=500)

fabianhjr:
Fun note, N=1000 goes to 10seconds so there is still performance increase opportunities. :3

Also, the last example on the issue page now behaves better than the first example. uwu

PR #1556, opened 2021-06-14
Hoist some pures

nickdrozd:


src/Core/Core.idr
```diff
@@ -449,9 +449,9 @@ export
 wrapError : (Error -> Error) -> Core a -> Core a
 wrapError fe (MkCore prog)
     = MkCore (prog >>=
-                 (\x => case x of
-                             Left err => pure (Left (fe err))
-                             Right val => pure (Right val)))
+                 (\x => pure $ case x of
+                             Left err => Left (fe err)
+                             Right val => Right val))
```
buzden, line 454: You can go more radical:
```idris
wrapError fe (MkCore prog)      
    = MkCore $ prog <&>      
                 \case                  
                   Left err  => Left (fe err)                      
                   Right val => Right val
```
or better even more:
```idris
wrapError fe (MkCore prog)      
    = MkCore $ prog <&> mapFst fe
```

fabianhjr: :+1: nice

nickdrozd: What does that mean? I don't understand the `<&>` notation (or `<$>`).

fabianhjr: Flipped version of `<$>`: https://github.com/idris-lang/Idris2/blob/a28bc655440e963591aaa7cfc47adefc4e785452/libs/prelude/Prelude/Interfaces.idr#L87-L93

fabianhjr: And `f <$> traversable` is `map f traversable`, so `traversable <&> f` is also `map f traversable`.

nickdrozd: I see. I'll take care of that in another PR.

PR #1557, opened 2021-06-14
[ fix ] print postfix projections... postfix

gallais:


src/Idris/Pretty.idr
```diff
@@ -266,7 +266,16 @@ mutual
           let_ <++> braces (angles (angles "definitions")) <+> line <+> in_ <++> go startPrec sc
       go d (PUpdate _ fs) =
         parenthesise (d > appPrec) $ group $ record_ <++> braces (vsep $ punctuate comma (prettyUpdate <$> fs))
-      go d (PApp _ f a) = parenthesise (d > appPrec) $ group $ go leftAppPrec f <++> go appPrec a
+      go d (PApp _ f a) =
+        let catchall : Lazy (Doc IdrisSyntax)
+               := go leftAppPrec f <++> go appPrec a
+
+        in parenthesise (d > appPrec) $ group $ case f of
+          f'@(PRef _ n) =>
```
mattpolzin, line 274: Why rebind f as f' here?

gallais: Ah! Because I used to call a function on `f` to remove potential `PBracketed`
constructors but, as it turns out, it does not seem to be an issue.

PR #1562, opened 2021-06-16
Use truncated case notation

nickdrozd:


src/Core/Core.idr
```diff
@@ -449,10 +449,10 @@ coreFail e = MkCore (pure (Left e))
...
 wrapError : (Error -> Error) -> Core a -> Core a
 wrapError fe (MkCore prog)
-    = MkCore (prog >>=
-                 (\x => pure $ case x of
-                             Left err => Left (fe err)
-                             Right val => Right val))
+    = MkCore $ prog >>=
+                    pure . \case
+                      Left err => Left $ fe err
+                      Right val => Right val
```
gallais, line 455: As suggested earlier, this is a good opportunity to use `mapFst` instead.

src/Core/Unify.idr
```diff
@@ -1437,18 +1437,18 @@ retryGuess mode smode (hid, (loc, hname))
...
-                                       pure False -- progress not made yet!
-                                _ => do logTermNF "unify.retry" 5 ("Search failed at " ++ show rig ++ " for " ++ show hname)
-                                                  [] (type def)
-                                        case smode of
-                                             LastChance =>
-                                                 throw !(normaliseErr err)
-                                             _ => pure False) -- Postpone again
+                     \case
+                       DeterminingArg _ n i _ _ =>
+                         let msg = "Failed (det " ++ show hname ++ " " ++ show n ++ ")" in
```
gallais, line 1442: I wouldn't be surprised if you need to add `Lazy String` annotations here to keep the
same semantics. Same for the other branch.

src/Parser/Rule/Package.idr
```diff
@@ -20,117 +20,119 @@ EmptyRule = Grammar () Token False
 export
 equals : Rule ()
 equals = terminal "Expected equals"
```
gallais, line 22: For all of these, I am tempted to suggest the addition of a `Eq Token` implementation
in `Parser.Lexer.Package` and then turn them all into `guard (TOK ==)`.

src/Core/Core.idr
nickdrozd: added

src/Core/Unify.idr
nickdrozd: I put the messages back in the function calls.

src/Parser/Rule/Package.idr
nickdrozd: Certainly the existing arrangement feels a little less elegant than it could be.

src/Core/Core.idr
```diff
@@ -448,7 +448,7 @@ coreFail e = MkCore (pure (Left e))
 
 export
 wrapError : (Error -> Error) -> Core a -> Core a
-wrapError fe (MkCore prog) = MkCore $ prog <&> mapFst fe
+wrapError fe (MkCore prog) = MkCore $ prog >>= pure . mapFst fe
```
fabianhjr, line 451: This is fancier though, `<&>` for `>>= pure .`.

gallais: Fancier in the sense it's not shipped by the previous version.

fabianhjr: Oh right

buzden: If `<&>` is unavailable, then use `<$>`, it's definitely available. I.e.
```idris
wrapError fe (MkCore prog) = MkCore $ mapFst fe <$> prog
```
IMO, still better than `>>= pure .`

PR #1567, opened 2021-06-17
Reduce noise in chez-sep codegen

fabianhjr:
Currently does this when creating the *.so files:

image

/cc @ziman


benhormann:
@fabianhjr
`, ">", if isWindows then "nul" else "/dev/null"`

https://ss64.com/nt/nul.html / https://ss64.com/nt/syntax-redirection.html

edwinb:
Now that we have incremental code generation, the chez-sep codegen will probably be removed (although large parts of it are likely to be useful in the API). So since this is still a draft, I'm going to close it. If it can be completed, please do reopen as merging certainly won't do any harm.

PR #1568, opened 2021-06-17
Add ability to extend RefC backend to create further backends

madman-bob:
Add the ability add aditional FFI targets to the RefC backend. This allows backends that extend RefC to add their own FFIs (for example, Python FFIs in the Python backend).

I've also added some documentation on how to extend RefC, both on creating basic backends, and adding FFI support to those backends.


gallais:
And well documented too! :partying_face:

PR #1570, opened 2021-06-17
Move System.Future from contrib to base

fabianhjr:
This has been in use for a while in `test` and has compiler primitives to work. This concurrency primitive will help others do some parallelism without having to worry about shared memory/channels/mutexes/syncronization/etc from the get go.

Related to #753 (Merged at 2020Q4, currently in v0.3.0)

> The main thing is that, since this is clearly a new experiment, it shouldn't be in base, which is for things that have become stable. I'd prefer it to be in contrib for now @edwinb


buzden:
> This concurrency primitive will help others do some parallelism without having to worry about ...

But having it in the `contrib` library does not prevent people from using it. So why to have it in `base`? Moreover (correct me it I'm wrong), this module for now works only with Chez backend -- a thing not so desired for something in `base`.

CodingCellist:
Hi Fabian, I have some doubts and questions regarding Futures and this suggestion:

1. My immediate worry is that it sets a precedent for merging something from `contrib` into `base` which, afaik, has not been done previously. This creates the question of "When is something in `contrib` considered 'stable enough' to be moved to `base`?", which is a very difficult question to answer. Given that one of the major goals is to eventually have a package manager, would it not be better to keep the core language slimmer and then eventually be able to add packages to projects as need be using the package manager?
2. IIUC, thread creation, external parallel and/or concurrent work, etc. live in `IO` or `HasIO` since they have to interact with the OS, which is an IO action. Futures seem to hide this to an extent, and I'm not convinced it should be  
 `await : Future a -> a`  
 but rather  
 `await : Future a -> IO a` (or a `HasIO io` version).  
 This could just be a design decision though and I'd really like to hear your thoughts on this  : )
3. Idris's concurrency model has been that of threads/processes and channels. Abstractions and tools can be built on top of these, or alternative concurrency models can be created, but I would argue that they are _a_ way of doing concurrency (since nobody has really figured out if there is a "right/best" way to do concurrency), so having them all live in `contrib` as alternative ways of viewing, abstracting over, implementing, and facilitating concurrency seems like the more appropriate option to me.
4. I've been concerned about `test` using futures for a bit, since I believe the core bits of Idris should avoid using things from `contrib`. Libraries in `contrib` could be experimental features, so we shouldn't rely on them for the core; or rather, modifying things in `contrib` should not require also changing parts of Idris. It is possible to do the parallel test execution using the existing primitives : )

Again, I think futures are a neat way to help people do concurrent programming. They are simple to reason about, and they provide a quick+easy way to make things happen in parallel. I just also think that moving them to `base` would mean there were suddenly two "official" ways Idris does concurrency, which I'm not sure would be good (having two `fork` functions really rubs me the wrong way). If futures were implemented in Idris, that would be a different story, since we could then just let people have access to the underlying, more granular stuff if they want, whilst also having an abstraction built on top of those things. Of course, I'm sure there are problems with this approach as well : )

CodingCellist:
@buzden
> this module for now works only with Chez backend

I think the library works in Racket as well. They're a [primitive](https://docs.racket-lang.org/reference/futures.html) in Racket and we have [bindings](https://github.com/idris-lang/idris2/blob/master/support/racket/support.rkt#L425-L426) for the Racket backend  : )

fabianhjr:
> Moreover (correct me it I'm wrong), this module for now works only with Chez backend 

Also racket! (Though, had to be implemented in two steps) And any backend that wants to self-host would need to implement it to run the test suite.

> This creates the question of "When is something in contrib considered 'stable enough' to be moved to base?", which is a very difficult question to answer. Given that one of the major goals is to eventually have a package manager, would it not be better to keep the core language slimmer and then eventually be able to add packages to projects as need be using the package manager?

I would agree however in this case this is mostly a wrapper around some compiler primitives since there is no way to marshall a lazy / delayed through an FFI `%foreign` call so the `prim__makeFuture` already ship with the compiler.

> IIUC, thread creation, external parallel and/or concurrent work, etc. live in IO or HasIO since they have to interact with the OS, which is an IO action. Futures seem to hide this to an extent, and I'm not convinced it should be

There are non-IO futures and IO futures are in a IO context on creation (forkIO), they could be considered "completed" on creation but delayed; await is for the synchronization of results and not an evaluation/computation per se.

>  so having them all live in contrib as alternative ways of viewing, abstracting over, implementing, and facilitating concurrency seems like the more appropriate option to me.

I disagree since I have been doing experiments with using futures to add concurrency to the idris compiler via Futures and while most have been really buggy due to shared memory as other concurrency primitives like MVars/TVars get implemented Futures might end up being used in the compiler itself.

Which goes in hand with your next point, I also believe the following:

> I've been concerned about test using futures for a bit, since I believe the core bits of Idris should avoid using things from contrib. Libraries in contrib could be experimental features, so we shouldn't rely on them for the core; or rather, modifying things in contrib should not require also changing parts of Idris.

Though, in particular:

> It is possible to do the parallel test execution using the existing primitives : )

Yes, it would be possible but would also require a lot more code. Since tests are self-contained it was a straightforward usage of futures.

gallais:
> It is possible to do the parallel test execution using the existing primitives

We're using Futures because this is what the person who wrote the PR used (and
at the time `Test.Golden` was part of `contrib` so having a local dependency was
not a big deal). I'd be happy to merge a refactoring removing this dependency.

edwinb:
I think there's too many open questions about concurrency to commit to maintaining this in `base` for now.

I can understand the desire to use these nice new features in Idris itself, but we should resist the temptation until we've had enough experience elsewhere.

I suspect we probably *will* move this to base. But I'd like to decide what the criteria for that are first, rather than doing it because it seems like it's probably a good idea. So I'm going to close this for now, sorry. We will revisit this.

Issue #1572, opened 2021-06-18
Seems impossible to turn on profiling with the Chez compiler


redfish64:
The Chez compiler will silently ignore it if you turn the profile flag on and run a command using ":exec" within the repl or ide.
Also, if you specify "--profile" as a command line argument, along with "--build", it responds with:
```
Not all command line options can be used to override package options.
```
and there doesn't seem to be a option for setting "profile" within the ipkg

redfish64:
Added a pull request: https://github.com/idris-lang/Idris2/pull/1573

CodingCellist:
Closed by #1576?  : )

PR #1573, opened 2021-06-18
Issue #1572

redfish64:
Added "Profile" as a flag that can be used when building a project.  Since, if you try to use :exec within a repl, the profile flag is ignored by the Chez backend, without this flag, and there is no "profile" setting available within the "ipkg" file, there is not way to turn profiling on with the Chez backend.

### DIFF
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -711,6 +711,7 @@ partitionOpts opts = foldr pOptUpdate (MkPFR [] [] False) opts
     optType (OutputDir f)          = POpt
     optType WarningsAsErrors       = POpt
     optType HashesInsteadOfModTime = POpt
+    optType Profile          = POpt
     optType (ConsoleWidth n)       = PIgnore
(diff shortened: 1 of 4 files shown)


src/Idris/Package.idr
```diff
@@ -711,6 +711,7 @@ partitionOpts opts = foldr pOptUpdate (MkPFR [] [] False) opts
     optType (OutputDir f)          = POpt
     optType WarningsAsErrors       = POpt
     optType HashesInsteadOfModTime = POpt
+    optType Profile          = POpt
```
fabianhjr, line 714: It would be good to keep the `=` alignment.

redfish64: Sorry, I wasn't able to figure out how to change this pull request, and I accidentaly created a new one, #1576 , which has the spacing fixed, so I'm closing this.

fabianhjr: No worries uwu

PR #1574, opened 2021-06-18
Add clarifying note re: PREFIX in config.mk

I-B-3:
Found this to be true in the process of trying to install Idris 2 into its own `git clone`d folder.

I initially tried `./.idris2`, but this led to the library files not being installed in the right places.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -1,6 +1,6 @@
 ##### Options which a user might set before building go here #####
 
-# Where to install idris2 binaries and libraries (must not be a relative path)
+# Where to install idris2 binaries and libraries (must be an absolute path)
 PREFIX ?= $(HOME)/.idris2
 
 # For Windows targets. Set to 1 to support Windows 7.
--- a/config.mk
+++ b/config.mk
@@ -1,6 +1,6 @@
 ##### Options which a user might set before building go here #####
 
-# Where to install idris2 binaries and libraries
+# Where to install idris2 binaries and libraries (must not be a relative path)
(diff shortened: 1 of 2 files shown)


config.mk
```diff
@@ -1,6 +1,6 @@
 ##### Options which a user might set before building go here #####
 
-# Where to install idris2 binaries and libraries
+# Where to install idris2 binaries and libraries (must not be a relative path)
```
CodingCellist, line 3: Potentially change to "(must be an absolute path)"? I can't really decide if it's clearer or not  ^^;

CodingCellist (commented):
Good spot! We should definitely have specified that... Could you possibly also please update `INSTALL.md` (around line 37) to have the same note? : )

config.mk
I-B-3: I suppose the positive view would be more constructive than a [Not](https://github.com/idris-lang/Idris2/blob/master/libs/prelude/Prelude/Basics.idr#L11)!

I-B-3: Corrected: 722bab81100e3272971cc5f0141e8a2ef8aa451d

I-B-3:
> Good spot! We should definitely have specified that... Could you possibly also please update `INSTALL.md` (around line 37) to have the same note? : )

Et voilá da21f0477aa534d76d9613a45a44c5d12674e40d :)

edwinb:
Ah yes, that makes it clearer, thanks.

PR #1575, opened 2021-06-18
[refactor:Data.Vect] Get length from type `O(1)` instead of `O(n)` calculation

fabianhjr:


gallais:
This means that the length cannot be erased anymore. It was a conscious choice
not to implement it that way: users can already require the nat to be available
at runtime in their local definition by changing its type. They don't need an
auxiliary function for that.

PR #1578, opened 2021-06-18
[ Fix #1577 ] Actually use natMinus hack

### LINKED ISSUE #1577: Nat is not optimized for minus
Compiling and running the below program shows that `minus` is slow. However in `CompileExpr.idr` in the `natHack` function, a "minus" is supposed to be replaced with `natMinus`, so it looks like it should be optimized.

```
0: Pass duration: 0s 245ns
1: Pass duration: 0s 137ns
2: Pass duration: 0s 127588978ns
3: Pass duration: 0s 241ns
4: Pass duration: 0s 193ns
```

```
module NatSpeedBug

import System
import System.Clock
import Data.List
import Data.Nat
import Data.String

timeit : Lazy x -> IO (x,Clock Duration)
timeit x =
  do 
    start <- clockTime Monotonic
    let x' = force x
    end <- clockTime Monotonic
    pure (x', timeDifference end start)

v1 : Nat
v1 = 100000000

v2 : Nat
v2 = 100000000

v3 : Nat
v3 = 100000001

test : List (Lazy Bool)
test =
 [
   --fast ops                        
   delay $ (natToInteger $ v1 + v2) == 200000000
  ,delay $ (natToInteger $ 2 * v1) ==  200000000
  ,delay $ (v1 `minus` v2) == 0    --should be fast, but is slow
  ,delay $ (natToInteger v1) ==      100000000
  ,delay $ const True (integerToNat  100000000)
 ]

toNano : Clock type -> Integer
toNano (MkClock seconds nanoseconds) =
  let scale = 1000000000
   in scale * seconds + nanoseconds
   
main : IO ()
main =  
  do
    cutOff <- getCutoffArg
    ignore $ traverse (prnIt cutOff) (zip [0..length test] test)
    pure ()
  where
    getCutoffArg : IO (Maybe Integer)
    getCutoffArg =
      do 
        args <- getArgs
        let (_ :: cutoffStr :: _) = args
               | x => pure Nothing
        let cutOff = cast $ stringToNatOrZ cutoffStr
        pure $ if cutOff == 0 then Nothing else Just cutOff
   
    showit : Bool -> String -> String -> String
    showit v a b = if v then a else b     
    prnIt : (Maybe Integer) -> (Nat, Lazy Bool) -> IO ()
    prnIt mCutOff (ind, lb) = 
        do 
          (b, diff) <- timeit lb
          putStrLn $ (show ind) ++ ": " ++ showit b "Pass" "Fail" ++ " " ++
             maybe 
               (show diff)
               (\co => showit (toNano diff < co) "Fast" "Slow")
               mCutOff
```

redfish64:
This should fix #1577, also I added a regression test speed check for optimizations. Let me know if it's in the wrong place.

### DIFF
--- a/libs/prelude/Prelude/Types.idr
+++ b/libs/prelude/Prelude/Types.idr
@@ -33,7 +33,7 @@ prim__integerToNat i
 
 public export
 integerToNat : Integer -> Nat
-integerToNat 0 = Z -- Force evaluation and hencing caching of x at compile time
+integerToNat 0 = Z -- Force evaluation and hence caching of x at compile time
 integerToNat x
   = if intToBool (prim__lte_Integer x 0)
        then Z
--- a/src/Compiler/CompileExpr.idr
+++ b/src/Compiler/CompileExpr.idr
@@ -176,21 +176,32 @@ magic ms e = go ms e where
     Nothing => go ms e
     Just e' => e'
 
-natMinus : FC -> FC -> forall vars. Vect 2 (CExp vars) -> CExp vars
-natMinus fc fc' [m,n] = CApp fc (CRef fc' (UN "prim__sub_Integer")) [m, n]
+%inline
+magic__integerToNat : FC -> FC -> forall vars. Vect 1 (CExp vars) -> CExp vars
+magic__integerToNat fc fc' [k]
+  = CApp fc (CRef fc' (NS typesNS (UN "prim__integerToNat"))) [k]
+
+magic__natMinus : FC -> FC -> forall vars. Vect 2 (CExp vars) -> CExp vars
+magic__natMinus fc fc' [m,n]
+  = magic__integerToNat fc fc'
+    [CApp fc (CRef fc' (UN "prim__sub_Integer")) [m, n]]
+
+-- We don't reuse natMinus here because we assume that unsuc will only be called
+-- on S-headed numbers so we do not need the truncating integerToNat call!
+magic__natUnsuc : FC -> FC -> forall vars. Vect 1 (CExp vars) -> CExp vars
+magic__natUnsuc fc fc' [m]
+  = CApp fc (CRef fc' (UN "prim__sub_Integer")) [m, CPrimVal fc (BI 1)]
 
 -- TODO: next release remove this and use %builtin pragma
 natHack : List Magic
 natHack =
-    [ MagicCRef (NS typesNS (UN "natToInteger")) 1
-         (\ _, _, [k] => k)
-    , MagicCRef (NS typesNS (UN "integerToNat")) 1
-         (\ fc, fc', [k] => CApp fc (CRef fc' (NS typesNS (UN "prim__integerToNat"))) [k])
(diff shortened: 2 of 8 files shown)


gallais:
Looks like the `natHack` is buggy. :rofl:

edwinb:
Ah yes, I spotted this recently (among other related things). Thanks for the update. I think for the test, though, we really need to check the output (`:di` in the REPL is handy for this) not the run time, because we can't predict how the VM running CI will perform. I just had a failure where it wasn't `minus` that was the problem.

This particular optimisation could be done more cleanly, anyway... not that I know exactly how yet.

edwinb:
Oh, we should also make sure that any tests that use chez or some other back end are run after we've checked that chez is available. We've had a few creep in now. which I'll tidy up before the imminent release if I can find them all. I'm not sure if there's a way to spot automatically whether they creep in, unfortunately. It's an easy thing to do by accident.

redfish64:
> Ah yes, I spotted this recently (among other related things). Thanks for the update. I think for the test, though, we really need to check the output (`:di` in the REPL is handy for this) not the run time, because we can't predict how the VM running CI will perform. I just had a failure where it wasn't `minus` that was the problem.
> 
> This particular optimisation could be done more cleanly, anyway... not that I know exactly how yet.

I wasn't able to get ":di" to show any optimization. For example, running `:di t` on the following:
```
import Data.Nat

v1 : Nat
v1 = 100000000

v2 : Nat
v2 = 100000000

t : Nat
t = v1 + v2
```

produces:

```
NatSpeedBug.t ==> [];
Compile time tree: [0] (Prelude.Num.+ Prelude.Types.Nat Prelude.Types.Num implementation at Prelude.Types:78:1--83:33 NatSpeedBug.v1 NatSpeedBug.v2)
Run time tree: Error: ""
RigW
Erasable args: []
Detaggable arg types: []
Specialise args: []
Inferrable args: []
Compiled: []: (Prelude.Types.+ [(NatSpeedBug.v1 []), (NatSpeedBug.v2 [])])
Refers to: [NatSpeedBug.v2, NatSpeedBug.v1, Prelude.Types.Num implementation at Prelude.Types:78:1--83:33, Prelude.Types.Nat, Prelude.Num.+]
Refers to (runtime): [NatSpeedBug.v2, NatSpeedBug.v1, Prelude.Types.+]
Flags: [covering]
Size change: Prelude.Num.+: [Nothing, Nothing, Nothing, Nothing], Prelude.Types.Num implementation at Prelude.Types:78:1--83:33: [], NatSpeedBug.v1: [], NatSpeedBug.v2: []
```
which doesn't seem to indicate that any optimization took place. So I ended up changing the test to parse the ".ss" file directly.

> Oh, we should also make sure that any tests that ue chez or some other back end are run after we've checked that chez is available. We've had a few creep in now. which I'll tidy up before the imminent release if I can find them all. I'm not sure if there's a way to spot automatically whether they creep in, unfortunately. It's an easy thing to do by accident.

Moved it to "chez/perf002"

I created a separate pull request for this, #1588

PR #1580, opened 2021-06-18
[ fix #1579 ] Nat hack for comparison operators

### LINKED ISSUE #1579: Nat should be optimized for (==) and compare
Nat is optimized for (+), minus, (*), natToInteger and integerToNat.

I think it should also be optimized for (==) and compare, since (==) and compare are IMO more primitive than (+), minus, and especially (*). Having some primitives optimized and some not caused confusion for myself, and is hard to debug without profiling and/or looking at the Idris2 source code.

redfish64:
The optimization hack for compare probably isn't ideal, but I'm unsure of a better way.


fabianhjr:
Hey, you have the commits of PR #1576 in this branch too.

libs/prelude/Prelude/EqOrd.idr
```diff
@@ -136,9 +136,13 @@ Ord Int where
   (>) x y = intToBool (prim__gt_Int x y)
   (>=) x y = intToBool (prim__gte_Int x y)
 
+public export
+compareInteger : (x : Integer) -> (y : Integer) -> Ordering
+compareInteger x y = if (intToBool $ prim__lt_Integer x y) then LT else if (intToBool $ prim__eq_Integer x y) then EQ else GT
```
gallais, line 141: Why are you not reusing `if x < y then LT else if x == y then EQ else GT`?

libs/prelude/Prelude/Types.idr
```diff
@@ -71,16 +84,11 @@ Num Nat where
 
 public export
 Eq Nat where
-  Z == Z = True
-  S j == S k = j == k
-  _ == _ = False
+  (==) = equal
```
gallais, line 87: As mentioned in #1182, can't we use a `transform` rule instead?

%transform "nateq" (==) {a = Nat} m n = natToInteger n == natToInteger m

libs/prelude/Prelude/EqOrd.idr
redfish64: Because `compareInteger` is defined before `OrdInteger`.

If I try to use `if x < y then LT else if x == y then EQ else GT`, I get:
```
 `-- Prelude.EqOrd line 141 col 24:
     While processing right hand side of compareInteger. Can't find an implementation for Ord Integer.
     
     Prelude.EqOrd:141:25--141:30
      137 |   (>=) x y = intToBool (prim__gte_Int x y)
      138 | 
      139 | public export
      140 | compareInteger : (x : Integer) -> (y : Integer) -> Ordering
      141 | compareInteger x y = if x < y then LT else if x == y then EQ else GT
                                    ^^^^^
```

gallais: You could insert it after the `Ord` declaration and have `compareInteger = compare` though

libs/prelude/Prelude/Types.idr
redfish64: I was testing this out, `%transform` doesn't seem to work all the time. 

Here is an example. As you can see, `fee5` calls `fee` directly whereas `fee3` calls `fee2` which in turn calls `fee`. For the `fee5` call, the result is transformed at runtime, for the `fee3` call, it is not.

```
interface Foo x where
  fee : x -> String
  
data Bar = Bee | Boot

Foo Bar where
  fee Bee = "Bee"
  fee Boot = "Boot"

%transform "feechange" fee {x=Bar} x = "No matter"

fee2 : Foo x => x -> String
fee2 v = fee v

fee3 : Bar -> String
fee3 v = fee2 v

fee5 : Bar -> String
fee5 v = fee v

test : IO ()
test = do 
         --using :exec test 
         --at runtime, prints:
         -- fee3 : Bee
         -- fee5 : No matter

         putStrLn $ "fee3 : " ++ fee3 Bee
         putStrLn $ "fee5 : " ++ fee5 Bee
```

Z-snails: If `%transform` is unreliable shouldn't we try and fix that rather than introduce more compiler magic that the end user can't extend in their own code?

redfish64: I was just thinking that it seemed that the hacks general purpose was to speed up Nat computation, which it wasn't doing. So, fixing the hack wouldn't introduce significant more dirtiness into the code that was already there, and was easy to do.

IMO, getting something working quickly is better since it attracts more users, but I don't know if the idris developers are looking for more users right now, so maybe this isn't a goal of theirs. 

However, the best would probably be to rip out the hack completely and use %transform for everything, including minus,+, and * for which the hack already works.

PR #1583, opened 2021-06-19
[fix] Loading libidris2_support.dll with Racket

benhormann:
### Summary:
Makes the problem of loading ws2_32.dll with Racket disappear.
File path issues remain. With some legwork, it's possible to bootstrap.

### Details:
Fixes the racket case for `os` on macOS - unblocks self-hosting.

Fixes the build flags for OLD_WIN, and automatically enables it on Windows 7.

Adds a define for reducing the headers included by windows.h; the intention is to prevent windows.h from including winsock.h (winsock2.h should be first).

Adds support functions to call POSIX interfaces, and associated tidy-up. The win32 api has some deprecated aliases, which racket doesn't like. Note the lack of procedure name in the error message:
> ffi-lib: couldn't open "libidris2_support.dll" (The specified procedure could not be found.; errid=127)

Moves `listen` into idris_net.c to be with its buddies (reverts [6054e86](https://github.com/idris-lang/Idris2/commit/6054e86b55f89a40586aa37ee852afb27ce193bb)).

Once the bootstrap is updated, [win_hack.c](https://github.com/idris-lang/Idris2/blob/v0.3.0/support/c/windows/win_hack.c) can be removed.

---

Adding a racket step to the windows CI could help prevent regressions, e.g. `make idris2-exec IDRIS2_CG=racket && ./build/exec/idris2 --version`.

#### Related:
win_hack.c makes chmod is a no-op. It would be good to note why, presumably because ntfs doesn't use the unix model.
Would it be better to check the parameter for absence of write permissions, then make the file read-only. E.g. `_chmod(path, mode & 0222 ? _S_IWRITE : _S_IREAD)`? This is close, but not quite the same as `_chmod(path, mode)`. It'll return an error if the file doesn't exist, which is probably better than the current approach.

`Network.Socket` checks for == -1, while `Idris.IDEMode.REPL` checks /= 0 instead. Winsock2 docs don't specify -1 as POSIX does, just 0 for success.

There's a GCC warning for `idris_signal.c`, the switch in `lock` is missing a default case.


gallais:
The self hosted binary built using racket on macOS seems to yield some errors:

```log
ffi-lib: could not load foreign library
  path: libidris2_support.dylib
  system error: dlopen(libidris2_support.dylib, 6): image not found
  context...:
   /Applications/Racket v8.1/collects/ffi/unsafe.rkt:131:0: get-ffi-lib
   body of "/Users/runner/work/Idris2/Idris2/tests/idris2/basic054/build/exec/_tmpracket_app/_tmpracket.rkt"
ffi-lib: could not load foreign library
  path: libidris2_support.dylib
  system error: dlopen(libidris2_support.dylib, 6): image not found
  context...:
   /Applications/Racket v8.1/collects/ffi/unsafe.rkt:131:0: get-ffi-lib
   body of "/Users/runner/work/Idris2/Idris2/tests/idris2/basic053/build/exec/_tmpracket_app/_tmpracket.rkt"
ffi-lib: could not load foreign library
  path: libidris2_support.dylib
  system error: dlopen(libidris2_support.dylib, 6): image not found
  context...:
   /Applications/Racket v8.1/collects/ffi/unsafe.rkt:131:0: get-ffi-lib
   body of "/Users/runner/work/Idris2/Idris2/tests/idris2/basic055/build/exec/_tmpracket_app/_tmpracket.rkt"
```

benhormann:
My impression was *all* the racket cg tests fail. I wasn't able to diagnose it quickly. Putting a copy of the dylib in `~/lib/` didn't seem to help.

Just speculating here, using `raco make` instead of `raco exe` might help (assuming it's a macOS security issue). This would be analogous to how Chez produces bytecode. It'd be executing `racket` instead of a brand new binary each time.

> Fixes the racket case for `os` on macOS - unblocks self-hosting.

"Unblocks" being the operative word. The issue with runtests is independent, so far as I could deduce.

gallais:
Alright, I'll turn off the self-host testing for racket for now.
I'll try to add the one for windows.

gallais:
I got a build using `make bootstrap-racket` started and it seems to be able to build
the .rkt to get a basic Idris2 that can be used to actually compile the current HEAD.

However [the build](https://github.com/gallais/Idris2/runs/2946547881) currently fails with:

```
/d/a/Idris2/Idris2/build/exec/idris2 --build prelude.ipkg
ffi-obj: could not find export from foreign library
  name: fdopen
  library: [all opened]
  system error: The specified procedure could not be found.; win_err=127
  context...:
   .../ffi/unsafe.rkt:255:20
   .../ffi/unsafe.rkt:249:2: get-ffi-obj*
   body of '#%mzc:idris2-boot
```

benhormann:
That will work after the bootstrap rkt is updated with this fix. Until then, start from Chez bootstrap.

gallais:
I came back to this but I still can't get the build to work. On the chez bootstrap
phase `make install` puts the the content into the non-standard `bootstrap` directory
instead of `.idris2`.

[This is the build](https://github.com/gallais/Idris2/actions/runs/1027063559)

and on the successful chez part we get a lot of:

> Installing D:\a\Idris2\Idris2\libs\X\Y\Z.ttc to D:\a\Idris2\Idris2\bootstrap\idris2-0.4.0\X-0.4.0

And unsurprisingly saving `D:\a\Idris2\Idris2\.idris2` for the next step leads to a
build failure with

> Uncaught error: Can't find package network (any)

Any idea why it's picking this non-standard installation target?

benhormann:
That's because the chez build has generated "src/IdrisPaths.idr" with the bootstrap path. It's not obvious why though... The support files and bin are installed to the right prefix.

In the first half of bootstrapping it uses the prefix from the scheme source, which is patched before compiling.
In the second half it overrides the path with env vars, so `IdrisPaths` should only be generated with the actual prefix.

I haven't done a windows build lately, I'll have a go. Should I rebase this PR?

benhormann:
@gallais It'll be this:
> `PREFIX=$PWD/bootstrap`
> [bootstrap-stage2.sh#L5](https://github.com/idris-lang/Idris2/blob/v0.4.0/bootstrap-stage2.sh#L5)

If "config.mk" exported `PREFIX` (perhaps it should), other CI builds would have the same problem.
Quick fix: `sed -i.bak -E 's/(^|\$)PREFIX/\1prefix/' bootstrap-stage2.sh`?

edwinb:
This is next in my queue for "merge or close" decisions but it looks like it's still active and close to a fix so I'll leave it alone! It'd be great to have this working.

benhormann:
@gallais Is it OK if I rebase, dropping your commits? Should make it simpler to add your CI jobs.

Alternative workaround for windows CI job: `echo 'PREFIX := <path>' > custom.mk`?
(I don't want to touch bootstrap-stage2.sh, every time I look at it I end up rewriting makefiles instead.)

benhormann:
@edwinb @gallais I vote we merge this, regenerate the bootstrap, and _then_ add the CI jobs proper.

gallais:
> It'll be [$PREFIX]

You're correct. I am surprised this local definition leaks out of this script.
Naïvely I would expect the alias to only be local. But I should know better given
the number of times naming a local var `PATH` has bitten me...

Changing this name does lead the CI to install the executable & libs in the right
place. I'm now hitting a different problem when building the self-host based on racket:
`make` builds Idris2 successfully but then the executable cannot be found to build
the libraries.

https://github.com/gallais/Idris2/runs/3103831688

```
echo "-- @""generated" > src/IdrisPaths.idr
echo 'module IdrisPaths' >> src/IdrisPaths.idr
echo 'export idrisVersion : ((Nat,Nat,Nat), String); idrisVersion = ((0,4,0), "0ec40a8b3")' >> src/IdrisPaths.idr
echo 'export yprefix : String; yprefix="D:/a/Idris2/Idris2/.idris2"' >> src/IdrisPaths.idr

(...)

make -C libs/prelude IDRIS2=/d/a/Idris2/Idris2/build/exec/idris2 IDRIS2_INC_CGS=racket IDRIS2_PATH="D:/a/Idris2/Idris2/libs/prelude/build/ttc;D:/a/Idris2/Idris2/libs/base/build/ttc;D:/a/Idris2/Idris2/libs/contrib/build/ttc;D:/a/Idris2/Idris2/libs/network/build/ttc;D:/a/Idris2/Idris2/libs/test/build/ttc"
make[1]: Entering directory '/d/a/Idris2/Idris2/libs/prelude'
/d/a/Idris2/Idris2/build/exec/idris2 --build prelude.ipkg
make[1]: /d/a/Idris2/Idris2/build/exec/idris2: No such file or directory
```

Are all of the directory separators the wrong way around?!

benhormann:
> I am surprised this local definition leaks out of this script.

It's not local, all these variables are global. Just because the script doesn't export it, doesn't mean it isn't already exported.

---

Most likely the file is in fact missing. (I'd put `ls -lR build/exec` instead of find, to make sure other files are in the right place).

Where is `raco` installed?
We've seen this sort of error in other issue reports... @gallais **Add `RACKET_RACO=raco`**

benhormann:
> ls: cannot access 'build/exec': No such file or directory

@gallais Why not use bash for these scripts? https://www.msys2.org/docs/ci/
```yml
  <job-name>:
    runs-on: windows-latest
    defaults:
      run:
        shell: msys2 {0}
```

gallais:
> Why not use bash for these scripts?

I did not write the original script and I don't know enough about msys to even
think it was possible. That would be a nice cleanup PR once we get this running
(the latest build seems promising!)

benhormann:
Unix doesn't buffer stdout but windows requires flushing ([nodejs docs](https://nodejs.org/api/process.html#process_a_note_on_process_i_o) - seems relevant). It's quite noticeable with the build progress lines, also the interleaving of output in the tests log. Add it to the fix-up list.

> ✗ racket not found

This one seems odd, but it's probably called 'Racket.exe'. Try adding `RACKET=racket`?

Could just pick some minimal set of tests (maybe `only=ideMode`?), or even skip testing. Building prelude proves `libidris2_support.dll` loads; building base proves prelude loads.

gallais:
I think we can skip testing for now and we can open another PR to deal with the
rest of the CI issues. My main concern was to get the CI to build the self-hosted
version using racket on windows given that it is this PR's contribution.

benhormann:
@gallais Want me to re-rebase, or will it come out in the wash?

benhormann:
Hurrah!

gallais:
Sorry it took so long! Each CI experiment was taking ~45 minutes so I kept context
switching away from the PR, forgetting about it and sometimes only checking the
result of the run multiple days later.

PR #1587, opened 2021-06-20
Some evaluator tidying, CBV evaluation.

edwinb:
There were some internal details preventing CBV evaluation working. This was fine, since we didn't do CBV :). But CBV evaluation might be a better default at the REPL since it saves duplication and it's closer to what the runtime does. So, I've tidied up the evaluator to fix the internals, and switched to CBV at the REPL.


edwinb:
I get a different result locally than the failing test. I haven't looked into what the test is actually doing, but if it's some performance threshold, I'd worry that we're relying on the performance of the VM that's running the test suite.

edwinb:
I just restarted the job and it worked fine this time... I think I'll change the test to show compiler output for reliability but not until tomorrow.

edwinb:
Well this is annoying, I'm doing two things at once and so of course I've got my branches confused locally...

edwinb:
I've moved the tests around so that any one that uses a code generator is now behind the check that the specific code generator is available. I think this removes the need for the `IDRIS2_TEST_CG` variable. I'm also (separately) setting things up so that code generation is disabled for the language/typechecking batch of tests.

One reason I don't like the tests being parameterised by code generator is that code generators might behave differently. So if we want to test that code generators all behave the same, and correctly, we should have separate tests for each.

gallais:
> One reason I don't like the tests being parameterised by code generator is that code generators might behave differently. So if we want to test that code generators all behave the same, and correctly, we should have separate tests for each.

One nice thing about the previous setup was that the CI was at least running the
racket & chez backends and these tests were sharing the same `expected` file. So
there was no risk of updating a golden file to """fix""" a build thus introducing
a disparity between the results of the different backends.

edwinb:
Hmm, maybe we can put those tests in their own category instead then? Either that, or we should probably change the test suite so that we check for an available CG right at the start.

It used to be that the tests would happily run with no CG available. That makes less sense now, since having no CG available means Idris wouldn't build at all.

I only did the shuffling to get out of the mess I just got myself into. Perhaps I should undo that and get out of it a different way.

edwinb:
Unrelated to this PR, which I've reopened elsewhere, but perhaps instead of (most of) the `chez` tests we should have a collection of code generation tests which are expected to run the same on *all* back ends.

gallais:
> we should have a collection of code generation tests which are expected to run the same on all back ends.

I have a design in mind, I'll open a PR.

PR #1588, opened 2021-06-21
Moved reg042 to chez/perf002 and now examines ".ss" file directly

redfish64:
Moved reg042 to chez test group, as chez/perf002 since it relies on chez and is probably closer to a performance test.

Reworked the test to examine the ".ss" file output to determine whether
performance improvements were made, rather than relying on clock on the
wall timing


gallais:
I think we can close this given that Edwin has changed the test to use `:di`
and look at the generated code instead (e35930a2aab1d073eda00a65cae6f5b90fd38b8c).

Issue #666, opened 2020-09-13
scheme binary not found despite specifying `SCHEME`


CodingCellist:
On my system, Chez Scheme gets installed as the binary `chez-scheme`. When trying to run `make bootstrap SCHEME=chez-scheme` the build initially seems to proceed without issues, but then errors with the following messages:
```
[...]
158/158: Building Idris.Main (src/Idris/Main.idr)
/usr/bin/env: ‘scheme’: No such file or directory
make[2]: Entering directory '$HOME/Projects/Idris2-git/tests'
$HOME/Projects/Idris2-git/bootstrap/bin/idris2 --build tests.ipkg
1/1: Building Main (Main.idr)
/usr/bin/env: ‘scheme’: No such file or directory
make[2]: Leaving directory '$HOME/Projects/Idris2-git/tests'
make -C libs/prelude IDRIS2=../../build/exec/idris2 IDRIS2_PATH="$HOME/Projects/Idris2-git/libs/prelude/build/ttc:$HOME/Projects/Idris2-git/libs/base/build/ttc:$HOME/Projects/Idris2-git/libs/contrib/build/ttc:$HOME/Projects/Idris2-git/libs/network/build/ttc"
make[2]: Entering directory '$HOME/Projects/Idris2-git/libs/prelude'
../../build/exec/idris2 --build prelude.ipkg
../../build/exec/idris2: line 14: $HOME/Projects/Idris2-git/build/exec/idris2_app/idris2.so: No such file or directory
make[2]: *** [Makefile:2: all] Error 127
make[2]: Leaving directory '$HOME/Projects/Idris2-git/libs/prelude'
make[1]: *** [Makefile:69: prelude] Error 2
make[1]: Leaving directory '$HOME/Projects/Idris2-git'
make: *** [Makefile:149: bootstrap-build] Error 2
```

It seems that despite specifying `SCHEME=chez-scheme`, some internals in the bootstrap tests have Chez Scheme hardcoded to `scheme`?

### Setup:
- Linux 5.8.7
- Chez Scheme 9.5.4
- `$HOME/.idris2/bin` was on my `PATH` and `$HOME/.idris2/lib` was on `LD_LIBRARY_PATH` as per the instructions in INSTALL.md

melted:
[...]

PR #1589, opened 2021-06-21
Add chezscheme binary to search path

00-matt:
The Gentoo Chez package and some versions of the Ubuntu package install Chez scheme as `/usr/bin/chezscheme` which requires manually setting the CHEZ environment variable.  Adding it to the search path makes it easier for these users.

Maybe there is a better solution but I think this is an okay temporary fix.

Closes: https://github.com/idris-lang/Idris2/issues/666


gallais:
Thanks!

gallais:
You're right, @benhormann. Re-opened #666. Thanks!

PR #1591, opened 2021-06-21
[ refactor ] to allow testpools to specify a backend

gallais:


buzden:
I'm sure [`test` lib's readme](https://github.com/idris-lang/Idris2/blob/master/libs/test/README.md) should be updated since it describes `MkTestPool` and has an example of its use, but it's changed.

PR #1593, opened 2021-06-22
fix #1514 (Broken doc in the theorem proving part of Idris2)

### LINKED ISSUE #1514: Broken doc in the theorem proving part of Idris2
The display for this grid on the [theorem proving introduction](https://github.com/idris-lang/Idris2/blob/master/docs/source/proofs/definitional.rst)  is broken: 
image

However the [corresponding one for Idris1](http://docs.idris-lang.org/en/latest/proofs/definitional.html) seems good.

zseri:
fixes #1514


edwinb:
Thanks!

PR #1594, opened 2021-06-22
[ contrib ] Add Queues and Buffered Channels for inter-process communication and/or sharing

CodingCellist:
The existing `Channel` implementation in `libs/base/System/Concurrency` can be used for inter-process communication, however, it blocks on both `put` and `get`, forcing these to match up. This is desirable in some scenarios, but sometimes it is nicer to be able to buffer a bunch of things and have the receiver(s) receive them at their own pace.

This PR adds shared, mt-safe, queues (queues being based on the description in Chris Osaki's "Purely Functional Datastructures") and builds buffered channels on top of these. There is also a test-suite to confirm that both behave as expected.

Queues support the expected operations of:
- `enqueue`
- `dequeue`
- `peek`

Buffered Channels operate as follows:
1. A new, shared channel is created through `makeBufferedChannel`
2. Each endpoint can then obtain the ability to send or receive by calling `becomeSender` or `becomeReceiver` respectively. These functions return a dependent pair of a channel that can be sent on and the `send` or `receive` function respectively.
   - When becoming a receiver, it must be decided whether to be a `Blocking` or `NonBlocking` receiver. The former blocks if the `BufferedChannel` is empty, but returns a thing of the BC's type; the latter doesn't block, but returns `Nothing` if the BC is empty.
   - When sending, since there can be threads blocked on the channel, the sender must decide how to wake these threads: `Signal` wakes one thread blocked on the channel, whereas `Broadcast` wakes all the threads blocked on the channel. It is up to the user programming the sender to make sure that there are enough things for all the blocked threads before a `Broadcast` is issued. Unfortunately, due to the shared nature of the channel, it is not possible (afaik) to check/prove in the types that `Broadcast` is safe.
3. To help with the `Broadcast` problem, and address a common use-case, there is a `becomeBuffer` function. It behaves similarly to `becomeSender` except that the input to the `buffer` function is a list of items to send on the BC (think `map send`). As with `send`, either `Signal` or `Broadcast` must be specified, but contrary to simply doing `map send Signal/Broadcast`, this "notification" only happens after _all_ the elements in the list have been placed on the BC, rather than after each element's placement.


CodingCellist:
Given that we're slowly but surely trying out [Sirdi](https://github.com/eayus/sirdi) as the official package manager, I've included Sirdi support in my [buffered channels repo](https://github.com/CodingCellist/idris2-buffered-channels) and recommend people try installing from there instead, if they're interested in using this  : )

Issue #1552, opened 2021-06-14
`System.Concurrency.Channel` non-deterministic behaviour in Chez-Scheme


CodingCellist:
When using a channel to communicate with multiple processes, things sent on the channel can get lost. Small example:
```idris
import System.Concurrency

child : Channel Nat -> IO ()
child c =
  do (S k) <- channelGet c
       | Z => putStrLn "Child thread done."
     child c

main : IO ()
main =
  do c <- makeChannel
     tids <- for [1..4] $ \_ => fork (child c)
     ignore $ for [1..8] $ \_ => channelPut c 1
     ignore $ for [1..4] $ \_ => channelPut c Z
     ignore $ traverse (\t => threadWait t) tids
     putStrLn "All threads + main done."
```

(a more flexible example can be found [here](https://gist.github.com/CodingCellist/618ccf6562ac8133bda4a58207e4a128))

# Steps to Reproduce
Execute `main` in the code above (or from the gist), first with `--cg chez` then with `--cg racket`.

# Expected Behavior
No messages are lost and all threads terminate correctly, followed by the main thread terminating.

# Observed Behavior
On the Chez-Scheme backend, the behaviour is non-deterministic: Sometimes all the child threads terminate, but more often they don't and so the program never finishes because at least one of the threads in `threadWait` never receives the `Z` that was sent.

By contrast, on the Racket backend, the program seems to always behave as expected, suggesting this is a problem with the `Channel` implementation in `support/chez/support.ss`.

PR #1595, opened 2021-06-22
[ base,tests ] Reduce dependency on `contrib`

CodingCellist:
In an attempt to minimise the number of dependencies on `contrib` in the core of the language, tests can be run in parallel using the lower-level `fork` and `Channel`. On the one lucky run where everything worked, this also seems to be faster, possibly due to re-using threads. **However** this implementation led me to discover #1552 and as such, this should not be merged until #1552 is fixed.


CodingCellist:
Just a quick note (since there's a PR review spree going on) that this is still being worked on. Currently ironing out merging with the latest `master`  : )

gallais:
@CodingCellist This seems to be working on my machine, should we consider this ready?

gallais:
I'm planning to merge this by the end of the week. Make yourself known if you don't think it's a good idea. :)

CodingCellist:
@gallais Hi! Yes! Sorry! I meant to reply to this a week ago and then got distracted ^^;;

It seems to work on my machine as well, which is really neat, so I'm happy for this to be merged! I'm somewhat confused as to why it's fixed itself, but I suspect it could've been something to do with #1596...
From a bit of benchmarking on my machine, using the `chez` backend with 8 threads and `/usr/bin/time` to measure things, these changes still seems to make things slightly faster (~7 seconds faster for `make test`) and more resource efficient (I get ~480% CPU usage with `Future` vs a consistent 600% CPU usage using this branch), so I think we can just merge it  : )

gallais:
:ship:

benhormann:
@CodingCellist `test` still depends on `contrib` for `Control.ANSI`.

@gallais The change to test.ipkg broke my bootstrap script in #1992. Shall I add `-p contrib` or `depends = contrib`?

gallais:
I see... We can't rely on the builds to check that these dependencies are correct because we're passing

> IDRIS2_PATH="/home/runner/work/Idris2/Idris2/libs/prelude/build/ttc:/home/runner/work/Idris2/Idris2/libs/base/build/ttc:/home/runner/work/Idris2/Idris2/libs/contrib/build/ttc:/home/runner/work/Idris2/Idris2/libs/network/build/ttc:/home/runner/work/Idris2/Idris2/libs/test/build/ttc:/home/runner/work/Idris2/Idris2/libs/linear/build/ttc"

:tired_face: 

I'd say put back the `depends = contrib` because that is the correct thing to do.

benhormann:
> I see... We can't rely on the builds to check that these dependencies are correct because we're passing

My bootstrap script does away with that nonsense, so you'll be able to rest easy 😉 

> I'd say put back the `depends = contrib` because that is the correct thing to do.

Done.

PR #1596, opened 2021-06-22
[ fix ] Chez channels

### LINKED ISSUE #1552: `System.Concurrency.Channel` non-deterministic behaviour in Chez-Scheme

CodingCellist:
These are my attempts at fixing #1552. And actually, this currently does fix #1552. However, it also has this ~~lovely~~ horifying little function:
```idris
spin : HasIO io => IORef Bool -> io ()
spin p =
  do True <- readIORef p
      | False => spin p
     pure ()
```

It is hacky, but I found it necessary to keep the functionality/semantics of `Channel`s:
- `channelGet` retrieves a value from the channel, blocking until a value is available.
- `channelPut` puts a value on the channel **blocking** until it has been received.

As detailed by #1552, the current implementation _does_ achieve this, but is also non-deterministic when using a single sender and multiple receivers on the same channel. Interestingly, it seems to work fine and reliably with multiple senders and a single receiver.

I initially thought the problem could be fixed by implementing channels natively in Idris:
<details>
<summary>Initial, non-hacky Idris implementation</summary>
<p>

```idris
module System.Concurrency.Channel

import Data.IORef
import System.Concurrency

%hide Channel
%hide channelGet
%hide channelPut
%hide makeChannel

data Msg : Type -> Type where
  Blank : Msg _
  TheVal : (val : a) -> Msg a

data Channel : Type -> Type where
  MkChannel : {0 a : Type}
            -> (putCtrl : (Mutex, Condition))
            -> (getCtrl : (Mutex, Condition))
            -> (mRef : IORef (Msg a))
            -> Channel a

||| Create a new, empty Channel.
export
makeChannel : HasIO io => {0 a : Type} -> io (Channel a)
makeChannel =
  do putMutex <- makeMutex
     putCV <- makeCondition
     getMutex <- makeMutex
     getCV <- makeCondition
     mRef <- newIORef Blank
     pure (MkChannel (putMutex, putCV) (getMutex, getCV) mRef)

||| Puts the @val@ on the Channel and blocks until it has been received.
export
channelPut : HasIO io => Channel a -> (val : a) -> io ()
channelPut (MkChannel (_, putCV) (getM, getCV) mRef) val =
  do mutexAcquire getM
     writeIORef mRef (TheVal val)
     conditionSignal putCV
     conditionWait getCV getM   -- atomically releases getM and blocks on getCV
     mutexRelease getM

||| Blocks until `val` is available on the Channel, at which point it receives
||| it.
export
channelGet : HasIO io => Channel a -> io a
channelGet (MkChannel (putM, putCV) (_, getCV) mRef) =
  do mutexAcquire putM
     conditionWait putCV putM   -- atomically releases putM and blocks on putCV
     (TheVal val) <- readIORef mRef
        | Blank => assert_total $ idris_crash "Somehow read a blank channel."
     conditionSignal getCV
     mutexRelease putM    -- FIXME: correct?
     pure val


testGet : IO ()
testGet =
  do c <- makeChannel {a=Nat}
     v <- channelGet c
     putStrLn "Shouldn't get here."

testPut : IO ()
testPut =
  do c <- makeChannel
     channelPut c 1
     putStrLn "Shouldn't get here."

testPutGet : IO ()
testPutGet =
  do c <- makeChannel
     t <- fork $ do putStrLn "Sending Z"
                    channelPut c Z
                    putStrLn "Sent"
     v <- channelGet c
     threadWait t
     if v == Z
        then putStrLn "Success!"
        else putStrLn "Fail."

smallTest : IO ()
smallTest =
  do c <- makeChannel
     t <- fork $ ignore (for [0..2] $ \n => channelPut c n)
     v0 <- channelGet c
     v1 <- channelGet c
     v2 <- channelGet c
     if v0 == 0 && v1 == 1 && v2 == 2
        then putStrLn "Success!"
        else putStrLn "Fail."

child : Channel Nat -> IO ()
child c =
  do (S k) <- channelGet c
      | Z => putStrLn "Thread done."
     child c

bigTest : IO ()
bigTest =
  do c <- makeChannel
     tids <- for [1..4] $ \_ => fork (child c)
     ignore $ for [1..8] $ \_ => channelPut c 1
     ignore $ for [1..4] $ \_ => channelPut c Z
     ignore $ traverse (\t => threadWait t) tids
     putStrLn "All threads + main done."
```

</p>
</details>

Unfortunately, this messed up the behaviour in Racket, probably due to Racket threads being green threads.

I then tried implementing it in Chez-Scheme using Condition Variables:

<details>
<summary>Chez-Scheme implementation</summary>
<p>

```scheme
;; Semaphore

(define-record semaphore (box mutex condition))

(define (blodwen-make-semaphore init)
  (make-semaphore (box init) (make-mutex) (make-condition)))

(define (blodwen-semaphore-post sema)
  (with-mutex (semaphore-mutex sema)
    (let [(sema-box (semaphore-box sema))]
      (set-box! sema-box (+ (unbox sema-box) 1))
      (condition-signal (semaphore-condition sema))
    )))

(define (blodwen-semaphore-wait sema)
  (with-mutex (semaphore-mutex sema)
    (let [(sema-box (semaphore-box sema))]
      (when (= (unbox sema-box) 0)
        (condition-wait (semaphore-condition sema) (semaphore-mutex sema)))
      (set-box! sema-box (- (unbox sema-box) 1))
      )))

;; Threads

(define-record thread-handle (semaphore))

(define (blodwen-thread proc)
  (let [(sema (blodwen-make-semaphore 0))]
    (fork-thread (lambda () (proc (vector 0)) (blodwen-semaphore-post sema)))
    (make-thread-handle sema)
    ))

(define (blodwen-thread-wait handle)
  (blodwen-semaphore-wait (thread-handle-semaphore handle)))

;; Channels

(define-record channel
               (p-mutex g-mutex box put-mutex put-condition get-mutex get-condition))

(define (blodwen-make-channel ty)
  (display "Made channel\n")
  (make-channel
    (make-mutex)
    (make-mutex)
    (box '())         ; a ref to the value(s)
    (make-mutex)      ; 'put' mutex
    (make-condition)  ; 'put' condition variable
    (make-mutex)      ; 'get' mutex
    (make-condition)  ; 'get' condition variable
    ))

(define (blodwen-channel-put ty chan val)
  (display "Put: Acquiring mutex\n")
  (with-mutex (channel-p-mutex chan)
    (display "Put: Acquired mutex\n")
    (let ([get-m (channel-get-mutex chan)])
      (with-mutex get-m   ; automatic acquire+release
        (let ([chan-box (channel-box chan)]
              [put-cv (channel-put-condition chan)]
              [get-cv (channel-get-condition chan)]
              )
          (display "Put: Inside let-block\n")
          ;;(set-box! chan-box val)   ; write the value to the box/IORef
          (set-box! chan-box (append (unbox chan-box) (list val)))
          (display "Wrote the new value\n")
          (condition-signal put-cv)
          (display "Put: Signalled put-cv\n")
          (condition-wait get-cv get-m)
          (display "Put: Passed get-cv\n")
          )))))

(define (blodwen-channel-get ty chan)
  (display "Get: Acquiring mutex\n")
  (with-mutex (channel-g-mutex chan)
    (display "Get: Acquired mutex\n")
    (let ([put-m (channel-put-mutex chan)])
      (with-mutex put-m
        (let ([chan-box (channel-box chan)]
              [put-cv (channel-put-condition chan)]
              [get-cv (channel-get-condition chan)]
              )
          (condition-wait put-cv put-m)
          (display "Get: Passed put-cv\n")
          (let* ([chan-val-list (unbox chan-box)]   ; let, but in sequence
                 ;; [chan-val (car chan-val-list)]     ; x
                 [chan-vals (cdr chan-val-list)])   ;   (::) xs
            (set-box! chan-box chan-vals)           ; change box to contain xs
            (display "Get: Set the box\n")
            (condition-signal get-cv)
            (display "Get: Signalled get-cv\n")
            (car chan-val-list)
            ))))))


;; TEST
(define (testGet)
  (display "-- testGet --\n")
  (let ([c (blodwen-make-channel '())])
    (display "testGet: Inside let-block\n")
    (blodwen-channel-get '() c)
    ))

(define (testPutChild c)
  (blodwen-channel-put '() c 0))

(define (testPut)
  (display "-- testPut --\n")
  (let ([c (blodwen-make-channel '())])
    (fork-thread (lambda () (sleep (make-time 'time-duration 0 2)) (testPutChild c)))
    (display "Forked thread\n")
    (let ([v (blodwen-channel-get '() c)])
      (eq? v 0))))

(define (smallTest)
  (display "-- testPut --\n")
  (let* ([c (blodwen-make-channel '())]
         [t (fork-thread (lambda ()
;;         [t (blodwen-thread (lambda ()
                               (blodwen-channel-put '() c 0)
                               (blodwen-channel-put '() c 1)))])
    (display "Forked thread\n")
    (let* ([v0 (blodwen-channel-get '() c)]
           [v1 (blodwen-channel-get '() c)]
           )
      (display (eq? v0 0))
      (eq? v1 1)
      )))
```

</p>
</details>

However, when running the test, this repeated the same non-deterministic behaviour. This led me to realise that there might be something more fundamentally wrong and so I wrote an implementation in C to check whether the behaviour was just local to Chez-Scheme, or if something else might be amiss.

<details>
<summary>C implementation</summary>
<p>

```c
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <assert.h>
#include <error.h>

typedef struct {
    pthread_mutex_t* getM;
    pthread_cond_t*  getCV;
    pthread_mutex_t* putM;
    pthread_cond_t*  putCV;
    int* val;
} channel;

channel* makeChannel()
{
    channel* c = calloc(1, sizeof(channel));
    if (!c) perror("Couldn't calloc channel");

    c->getM = calloc(1, sizeof(pthread_mutex_t));
    if (!(c->getM)) perror("Couldn't calloc getM");
    int e = pthread_mutex_init(c->getM, NULL);
    if (e) perror("Couldn't create getM");

    c->getCV = calloc(1, sizeof(pthread_cond_t));
    if (!(c->getCV)) perror("Couldn't calloc getCV");
    e = pthread_cond_init(c->getCV, NULL);
    if (e) perror("Couldn't create getCV");

    c->putM = calloc(1, sizeof(pthread_mutex_t));
    if (!(c->putM)) perror("Couldn't calloc putM");
    e = pthread_mutex_init(c->putM, NULL);
    if (e) perror("Couldn't create putM");

    c->putCV = calloc(1, sizeof(pthread_cond_t));
    if (!(c->putCV)) perror("Couldn't calloc putCV");
    e = pthread_cond_init(c->putCV, NULL);
    if (e) perror("Couldn't create putCV");

    c->val = calloc(1, sizeof(int));
    if (!(c->val)) perror("Couldn't calloc val");

    return c;
}

void destroyChannel(channel* c)
{
    int e;
    e = pthread_mutex_destroy(c->getM);
    assert(!e);
    e = pthread_cond_destroy(c->getCV);
    assert(!e);
    e = pthread_mutex_destroy(c->putM);
    assert(!e);
    e = pthread_cond_destroy(c->getCV);
    assert(!e);
    free(c->val);
    free(c);
}


void channelPut(channel* c, int val)
{
    int e = pthread_mutex_lock(c->getM);
    assert(!e);
    *(c->val) = val;
    e = pthread_cond_signal(c->putCV);
    assert(!e);
    e = pthread_cond_wait(c->getCV, c->getM);
    assert(!e);
    e = pthread_mutex_unlock(c->getM);
}

int channelGet(channel* c)
{
    int e = pthread_mutex_lock(c->putM);
    assert(!e);
    e = pthread_cond_wait(c->putCV, c->putM);
    assert(!e);
    int* vPtr = c->val;
    assert(vPtr != NULL);
    e = pthread_cond_signal(c->getCV);
    assert(!e);
    e = pthread_mutex_unlock(c->putM);
    assert(!e);
    return *vPtr;
}
    
void* child(void* arg)
{
    channel* c = (channel*) arg;
    int v;
    do {
        v = channelGet(c);
        //printf("Get: %d\n", v);
    } while (v);
    //printf("Thread done.\n");
    pthread_exit(0);
}


int main(void)
{
    channel* c = makeChannel();
    assert(c);
    pthread_t** tids = calloc(4, sizeof(pthread_t*));
    assert(tids);
    for (int i = 0; i < 4; i++)
    {
        tids[i] = calloc(1, sizeof(pthread_t));
        assert(tids[i]);
    }
    int e;
    for (int i = 0; i < 4; i++)
    {
        e = pthread_create(tids[i], NULL, child, (void*) c);
        assert(!e);
    }
    for (int i = 0; i < 4; i++) channelPut(c, 1);
    for (int i = 0; i < 4; i++) channelPut(c, 0);
    int retval = 1;
    for (int i = 0; i < 4; i++)
    {
        e = pthread_join(*(tids[i]), ((void**) &retval));
        assert(!e);
        assert(!retval);
    }
    for (int i = 0; i < 4; i++)
    {
        free(tids[i]);
    }
    free(tids);
    destroyChannel(c);
    printf("Success!\n");    
}
```

</p>
</details>

Unfortunately, this is also non-deterministic.

What I believe all this suggests is that there is either something weird about using Condition Variables, and so they should be avoided (which led to my hacky solution) or I've messed up the logic/control-flow somewhere. But the problem is clearly more complex than I anticipated and so I would be extremely grateful if anyone with more experience or knowledge (or both!) about these sorts of things could help narrow things down and possibly even fix them.

One of my first further thoughts was to do with race-conditions with multiple threads trying to acquire mutexes, but Chez's mutexes [are recursive](https://cisco.github.io/ChezScheme/csug9.5/threads.html#./threads:h2) which means that that shouldn't be the problem...

Once again, any help and/or insight would be much appreciated!  : )


zseri:
I don't have experience with green threads, but the C implementation seems a bit weird...
```c

void channelPut(channel* c, int val)
{
    int e = pthread_mutex_lock(c->getM);
    assert(!e);
    /* c->val should also be protected by c->putM, right? */
    *(c->val) = val;
    /* and putM should be unlocked before checking c->putCV */
    e = pthread_cond_signal(c->putCV);
    assert(!e);
    e = pthread_cond_wait(c->getCV, c->getM);
    assert(!e);
    e = pthread_mutex_unlock(c->getM);
}

int channelGet(channel* c)
{
    int e = pthread_mutex_lock(c->putM);
    assert(!e);
    e = pthread_cond_wait(c->putCV, c->putM);
    assert(!e);
    /* race condition here: c->val is only protected by c->getM, but we only locked c->putM */
    int* vPtr = c->val;
    /* the following assert may fail even if the above code is correct,
      because condvars may have spurious wake-ups */
    assert(vPtr != NULL);
    /* we transferred ownership at this point, it would make sense to set c->val to NULL,
      to be able to check for spurious wake-ups after pthread_cond_wait(c->getCV, c->getM)*/
    e = pthread_cond_signal(c->getCV);
    assert(!e);
    e = pthread_mutex_unlock(c->putM);
    assert(!e);
    return *vPtr;
}
```

zseri:
I tried to fix the C implementation, but it still deadlocks...
related: https://stackoverflow.com/questions/22939896/pthread-cond-wait-lost-signal-from-pthread-cond-signal

zseri:
<details>
<summary>fixed C code</summary>

```c
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <assert.h>
#include <error.h>

/* linux-specific debugging */
#include <sys/types.h>
#include <unistd.h>
#include <sys/syscall.h>

/* I omitted the heap allocations, if it is a thread sync problem,
  not doing heap allocations shouldn't fix it */
typedef struct {
    pthread_mutex_t M;
    pthread_cond_t CVg, CVsb, CVd;
    int val;
    bool isset, isdone;
} channel;

channel* makeChannel()
{
    channel* c = calloc(1, sizeof(channel));
    if(!c) perror("Couldn't calloc channel");

    int e = pthread_mutex_init(&c->M, NULL);
    if(e) perror("Couldn't create M");

    e = pthread_cond_init(&c->CVg, NULL);
    if(e) perror("Couldn't create CVg");

    e = pthread_cond_init(&c->CVsb, NULL);
    if(e) perror("Couldn't create CVsb");

    e = pthread_cond_init(&c->CVd, NULL);
    if(e) perror("Couldn't create CVd");

    c->val = 0;
    c->isset = false;
    c->isdone = false;

    return c;
}

void destroyChannel(channel* c)
{
    int e;
    e = pthread_mutex_destroy(&c->M);
    assert(!e);
    e = pthread_cond_destroy(&c->CVg);
    assert(!e);
    e = pthread_cond_destroy(&c->CVsb);
    assert(!e);
    e = pthread_cond_destroy(&c->CVd);
    assert(!e);
    free(c);
}

pid_t curtid()
{
    return syscall(__NR_gettid);
}

void channelPut(channel* c, int val)
{
    pid_t mytid = curtid();
    printf("[%d] P with %d\n", mytid, val);
    printf("[%d] P 0 M L\n", mytid);
    int e = pthread_mutex_lock(&c->M);
    assert(!e);
    while(c->isset) {
      printf("[%d] P 0 CVsb W\n", mytid);
      e = pthread_cond_wait(&c->CVsb, &c->M);
      assert(!e);
    }
    c->val = val;
    c->isset = true;
    printf("[%d] P 0 M U\n", mytid);
    e = pthread_mutex_unlock(&c->M);
    assert(!e);
    printf("[%d] P 1 CVg S\n", mytid);
    e = pthread_cond_signal(&c->CVg);
    assert(!e);
    printf("[%d] P 2 M L\n", mytid);
    e = pthread_mutex_lock(&c->M);
    assert(!e);
    while(!c->isdone) {
      printf("[%d] P 2 CVd W\n", mytid);
      e = pthread_cond_wait(&c->CVd, &c->M);
      assert(!e);
    }
    c->isdone = false;
    printf("[%d] P 2 M U\n", mytid);
    e = pthread_mutex_unlock(&c->M);
    assert(!e);
    printf("[%d] P 3 CVsb S\n", mytid);
    e = pthread_cond_signal(&c->CVsb);
    assert(!e);
    printf("[%d] P with %d DONE\n", mytid, val);
}

int channelGet(channel* c)
{
    pid_t mytid = curtid();
    printf("[%d] G 0 M L\n", mytid);
    int e = pthread_mutex_lock(&c->M);
    assert(!e);
    while(!(c->isset)) {
      printf("[%d] G 0 CVg W\n", mytid);
      e = pthread_cond_wait(&c->CVg, &c->M);
      assert(!e);
    }
    assert(c->isset);
    int ret = c->val;
    /* transferred ownership */
    c->isset = false;
    c->isdone = true;
    printf("[%d] G 0 Mv U\n", mytid);
    e = pthread_mutex_unlock(&c->M);
    assert(!e);
    printf("[%d] G 1 CVd S\n", mytid);
    e = pthread_cond_signal(&c->CVd);
    assert(!e);
    printf("[%d] G with %d DONE\n", mytid, ret);
    return ret;
}

void* child(void* arg)
{
    channel* c = (channel*) arg;
    int v;
    do {
        v = channelGet(c);
        printf("[%d] Get: %d\n", curtid(), v);
    } while (v);
    printf("[%d] Thread done.\n", curtid());
    pthread_exit(0);
}

int main(void)
{
    channel* c = makeChannel();
    assert(c);
    pthread_t** tids = calloc(4, sizeof(pthread_t*));
    assert(tids);
    for (int i = 0; i < 4; i++)
    {
        tids[i] = calloc(1, sizeof(pthread_t));
        assert(tids[i]);
    }
    int e;
    for (int i = 0; i < 4; i++)
    {
        e = pthread_create(tids[i], NULL, child, (void*) c);
        assert(!e);
    }
    for (int i = 0; i < 4; i++) {
        printf("[%d] Put: %d\n", curtid(), 1);
        channelPut(c, 1);
    }
    for (int i = 0; i < 4; i++) {
        printf("[%d] Put: %d\n", curtid(), 0);
        channelPut(c, 0);
    }
    int retval = 1;
    for (int i = 0; i < 4; i++)
    {
        e = pthread_join(*(tids[i]), ((void**) &retval));
        assert(!e);
        assert(!retval);
    }
    for (int i = 0; i < 4; i++)
    {
        free(tids[i]);
    }
    free(tids);
    destroyChannel(c);
    printf("Success!\n");
    return 0;
}
```

</details>

CodingCellist:
@zseri thank you so much for the detailed comments, resources, and fixed version!  : D

> c->val should also be protected by c->putM, right? (missing synchronization here)

> and putM should be unlocked before checking c->putCV
>      needed to avoid unnecessary context switches

Seems I misunderstood what my code was doing/how CVs work: I assumed that CVs kept track of signalling, so if `get` got called before `put` there wouldn't be an issue since the "handshaking" would protect the variable _and_ ensure that progress was made.

Having seen the stuff about having some sort of predicate and your working version, things make a lot more sense. I might try to implement your version in Chez or Idris, with full credit of course, and see if that still has problems, if you're okay with me doing that?  : )

zseri:
> with full credit of course, and see if that still has problems, if you're okay with me doing that? : )

no problem.
please credit myself as "Alain Zscheile" if the change is big enough to justify an add of myself to `CONTRIBUTORS`.

CodingCellist:
@zseri, after a bit more testing, it seems we're unfortunately not quite in the clear yet:
- `put/get` with one producer and one consumer works fine
- one producer and multiple consumers also seems to work fine
- **but** one consumer and multiple producers still non-deterministically deadlocks...  : /

<details>
<summary>Multiple producers version of your C code</summary>
<p>

```c
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <assert.h>
#include <error.h>

/* linux-specific debugging */
#include <sys/types.h>
#include <unistd.h>
#include <sys/syscall.h>

/* I omitted the heap allocations, if it is a thread sync problem,
  not doing heap allocations shouldn't fix it */
typedef struct {
    pthread_mutex_t M;
    pthread_cond_t CVg, CVsb, CVd;
    int val;
    bool isset, isdone;
} channel;

channel* makeChannel()
{
    channel* c = calloc(1, sizeof(channel));
    if(!c) perror("Couldn't calloc channel");

    int e = pthread_mutex_init(&c->M, NULL);
    if(e) perror("Couldn't create M");

    e = pthread_cond_init(&c->CVg, NULL);
    if(e) perror("Couldn't create CVg");

    e = pthread_cond_init(&c->CVsb, NULL);
    if(e) perror("Couldn't create CVsb");

    e = pthread_cond_init(&c->CVd, NULL);
    if(e) perror("Couldn't create CVd");

    c->val = 0;
    c->isset = false;
    c->isdone = false;

    return c;
}

void destroyChannel(channel* c)
{
    int e;
    e = pthread_mutex_destroy(&c->M);
    assert(!e);
    e = pthread_cond_destroy(&c->CVg);
    assert(!e);
    e = pthread_cond_destroy(&c->CVsb);
    assert(!e);
    e = pthread_cond_destroy(&c->CVd);
    assert(!e);
    free(c);
}

pid_t curtid()
{
    return syscall(__NR_gettid);
}

void channelPut(channel* c, int val)
{
    pid_t mytid = curtid();
    printf("[%d] P with %d\n", mytid, val);
    printf("[%d] P 0 M L\n", mytid);
    int e = pthread_mutex_lock(&c->M);
    assert(!e);
    while(c->isset) {
      printf("[%d] P 0 CVsb W\n", mytid);
      e = pthread_cond_wait(&c->CVsb, &c->M);
      assert(!e);
    }
    c->val = val;
    c->isset = true;
    printf("[%d] P 0 M U\n", mytid);
    e = pthread_mutex_unlock(&c->M);
    assert(!e);
    printf("[%d] P 1 CVg S\n", mytid);
    e = pthread_cond_signal(&c->CVg);
    assert(!e);
    printf("[%d] P 2 M L\n", mytid);
    e = pthread_mutex_lock(&c->M);
    assert(!e);
    while(!c->isdone) {
      printf("[%d] P 2 CVd W\n", mytid);
      e = pthread_cond_wait(&c->CVd, &c->M);
      assert(!e);
    }
    c->isdone = false;
    printf("[%d] P 2 M U\n", mytid);
    e = pthread_mutex_unlock(&c->M);
    assert(!e);
    printf("[%d] P 3 CVsb S\n", mytid);
    e = pthread_cond_signal(&c->CVsb);
    assert(!e);
    printf("[%d] P with %d DONE\n", mytid, val);
}

int channelGet(channel* c)
{
    pid_t mytid = curtid();
    printf("[%d] G 0 M L\n", mytid);
    int e = pthread_mutex_lock(&c->M);
    assert(!e);
    while(!(c->isset)) {
      printf("[%d] G 0 CVg W\n", mytid);
      e = pthread_cond_wait(&c->CVg, &c->M);
      assert(!e);
    }
    assert(c->isset);
    int ret = c->val;
    /* transferred ownership */
    c->isset = false;
    c->isdone = true;
    printf("[%d] G 0 Mv U\n", mytid);
    e = pthread_mutex_unlock(&c->M);
    assert(!e);
    printf("[%d] G 1 CVd S\n", mytid);
    e = pthread_cond_signal(&c->CVd);
    assert(!e);
    printf("[%d] G with %d DONE\n", mytid, ret);
    return ret;
}

void* producer(void* arg)
{
    channel* c = (channel*) arg;
    int v;
    for (int i = 1; i <= 100; i++)
    {
        channelPut(c, i);
    }
    printf("[%d] Producer done.\n", curtid());
    pthread_exit(0);
}

#define NTHREADS 4

int main(void)
{
    channel* c = makeChannel();
    assert(c);
    pthread_t** tids = calloc(NTHREADS, sizeof(pthread_t*));
    assert(tids);
    for (int i = 0; i < NTHREADS; i++)
    {
        tids[i] = calloc(1, sizeof(pthread_t));
        assert(tids[i]);
    }
    int e;
    for (int i = 0; i < NTHREADS; i++)
    {
        e = pthread_create(tids[i], NULL, producer, (void*) c);
        assert(!e);
    }
    int v;
    int sum = 0;
    // each thread is sending 100 messages
    for (int i = 0; i < NTHREADS * 100; i++) {
        v = channelGet(c);
        sum += v;
        printf("[%d] Got: %d\n", curtid(), v);
    }
    int retval = 1;
    for (int i = 0; i < NTHREADS; i++)
    {
        e = pthread_join(*(tids[i]), ((void**) &retval));
        assert(!e);
        assert(!retval);
    }
    printf("[%d] Sum: %d\n", curtid(), sum);
    for (int i = 0; i < NTHREADS; i++)
    {
        free(tids[i]);
    }
    free(tids);
    destroyChannel(c);
    printf("Success!\n");
    return 0;
}
```

</p>
</details>

It seems to be a tiny bit more stable in `valgrind`, but not entirely. Any idea what might be happening?  : )

zseri:
ok, I'm trying to investigate it. first problem: `P 2 M L` may randomly block longer than expected, with another producer highjacking the channel.

zseri:
In general, I would suggest introducing a `pipe` primitive which just doesn't wait for the other thread to receive the data, instead it waits if data is already in the channel until it is taken out. That's way easier to implement, and on top of that it would be easy implementing this as a wrapper. My suggestion for the `P2ML` problem would be sending `(Value, (Channel ()))` inside the channel, and then blocking on the subchannel (although cleaning that up is a bit more complex in the C version, because we don't have a garbage collector there).

CodingCellist:
Thanks for investigating things! Interesting that that's where the problem lies  : )
> In general, I would suggest introducing a `pipe` primitive which just doesn't wait for the other thread to receive the data, instead it waits if data is already in the channel until it is taken out.

Yeah, that's probably what I'll end up doing. I believe the initial semantics were introduced to facilitate building session-types on top of `Channel`, but having `put` not block unless data is already present makes things _much_ simpler; I actually think I have an old implementation of that in native Idris2 that I abandoned because it didn't wait on `get`. I'll see if I still have it lying around on my drive.

zseri:
a problem with the current fix (e.g. just waiting if there is still data in there) is that if the value isn't marked with an origin thread id (so that the current thread blocks if the value currently there was placed by it(thread)self), a thread could place data in the channel, then try to receive it, but instantly getting back the data that was just placed by itself in there....

PR #1597, opened 2021-06-23
A couple of expression search improvements

edwinb:
Firstly, add a configurable timeout (via `%search_timeout`), defaulting to 1 second (or rather, 1000 milliseconds), after which it will stop searching for more solutions. We search for a batch then sort to find the best of the batch, but in some cases it can start exploring too much, in which case we need the timeout. Typically if it hasn't found a solution in a second, it's not going to find a good one anyway.

Secondly, if a hole already has a solution by unification, we should just give that. We nearly did, but it was rendered incorrectly.

This adds a simple timeout mechanism, in which we can start a timer and check whether it's gone over some threshold. It could be useful elsewhere, e.g. in checking whether ambiguity resolution is struggling (the current check for depth is clearly not what we need, unfortunately.)

### DIFF
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -1115,9 +1115,9 @@ record Defs where
      -- again
   timings : StringMap (Bool, Integer)
      -- ^ record of timings from logTimeRecord
-  timer : Maybe Integer
-     -- ^ for timing and checking timeouts; the time in nanoseconds when
-     -- the timer was initialised
+  timer : Maybe (Integer, String)
+     -- ^ for timing and checking timeouts; the maximum time after which a
+     -- timeout should be thrown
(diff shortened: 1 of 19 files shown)


buzden:
In the context of recent discussion of _stable_ metrics (e.g. working in the same way on fast local machines and on slow CIs), have you considered some other metric rather than time? Those that would be more stable and repeatable; like, I don't know, count of some steps of search or whatever.

edwinb:
It's hard to count anything here because there's lots going on (search steps, unification, reductions). I wondered about the same thing here, but from a user's point of view time is probably most meaningful. If it does turn out to be a problem in CI, I think I'd start by cranking up the threshold for the test in question.

PR #1598, opened 2021-06-23
[ fix ] missing Show implementations in libs

gallais:


edwinb:
Shall I just go ahead and merge this? They look like good additions to me. After this, I'm going to get on with the release.

PR #1601, opened 2021-06-23
Enable cache of TTC files, update flake.lock

fabianhjr:
image

Building once takes about 3 minutes, self hosting around another 3 minutes and between 4 and 20 minutes to run the test-suite on self hosting (racket still being quite slow)


.github/workflows/ci-ubuntu-combined.yml
```diff
@@ -8,23 +8,23 @@ on:
       - '*'
   pull_request:
     branches:
-      - master
-      - main
+      - '*'
 
 env:
-  IDRIS2_VERSION: 0.3.0 # For previous-version build
+  IDRIS2_VERSION: 0.4.0 # For previous-version build
```
stefan-hoeck, line 14: Sorry, but why is this bumped to 0.4.0? This should be the _previous version_, no? Otherwise, the comment is misleading and should be removed.

fabianhjr: Though a better name might be `latest-version`

stefan-hoeck: Ah, my bad. Thanks for the clarification.

fabianhjr: No worries, though not sure if I should rename the jobs too to make that comment clearer.

fabianhjr:
Notice: Chez/Racket bootstraps are the codegen output of v0.4.0 so testing from that version not via bootstrap should be compatible/interchangebale; furthermore both macOS/Windows would still self-host from the chez bootstrap so bootstrapping is still being tested.

.github/workflows/ci-macos-combined.yml
```diff
@@ -7,7 +7,7 @@ on:
       - '*'
   pull_request:
     branches:
-      - master
+      - '*'
```
gallais, line 10: You can use `$default-branch` instead if you want support for `main`/`master`

.github/workflows/ci-macos-combined.yml
```diff
@@ -54,8 +57,26 @@ jobs:
...
+          path: |
+            build/
+            libs/base/build/
+            libs/prelude/build/
+            libs/contrib/build/
+            libs/network/build/
+            libs/test/build/
+          key: ${{ runner.os }}-self-host-chez-${{ github.ref }}-${{ env.TTC_VERSION }}-${{ env.GITHUB_SHA }}
+          restore-keys: |
+            ${{ runner.os }}-self-host-chez-refs/heads/main-${{ env.TTC_VERSION }}
```
gallais, line 75: Same comment here.

.github/workflows/ci-macos-combined.yml
fabianhjr: I want to support any branch in case of any release branch or collaborative branch with many people doing PRs to said branch

fabianhjr: Due to the quickcheck change conflict going to leave this change out.

.github/workflows/ci-quick-check.yml
```diff
@@ -8,8 +8,7 @@ on:
       - '*'
   pull_request:
     branches:
-      - master
-      - main
+      - '*'
```
fabianhjr, line 11: Here is the same change as before, the intention is to be able to have branches that work on bigger changes and can receive PRs that are checked by the CI. (Not currently used)

fabianhjr:
Reverted #1643 due to disassociated duplicate runs:
image
image

And the CI checks should run on the PRs to avoid issues on the main branch:
image

fabianhjr:
Not sure how this diff happened! :o

image

PR #1602, opened 2021-06-24
[ fix #1581 ] public export Range functions and implementations

### LINKED ISSUE #1581: NonEmpty doesn't work with list comprehensions
# Steps to Reproduce
Write a function which uses a provably non-empty list, then use a list-comprehension with the function e.g.

```idris
import Data.List

foo : (l : List a) -> {auto 0 ok : NonEmpty l} -> a

bar : Integer
bar = foo [1..1000]
```

# Expected Behavior
This type-checks since the list `[1..1000]` is non-empty.

# Observed Behavior
```idris
Error: While processing right hand side of bar. Can't find an implementation for NonEmpty (rangeFromTo (fromInteger 1) (fromInteger 1000)).

LComp:6:7--6:20
 2 |
 3 | foo : (l : List a) -> {auto 0 ok : NonEmpty l} -> a
 4 |
 5 | bar : Integer
 6 | bar = foo [1..1000]
           ^^^^^^^^^^^^^
```

fabianhjr:
Fixes #1581

There is however a performance issue:

```
 ~/D/O/Idris2   fix-1581 $+…  time idris2 --check test.idr                                            Wed 23 Jun 2021 11:02:59 PM CDT
1/1: Building test (test.idr)

Executed in    8.45 secs    fish           external
   usr time    8.32 secs    0.00 micros    8.32 secs
   sys time    0.12 secs  335.00 micros    0.11 secs
```

### DIFF
--- a/libs/base/Data/Stream.idr
+++ b/libs/base/Data/Stream.idr
@@ -91,23 +91,6 @@ cycle (x :: xs) = x :: cycle' xs
         cycle' []        = x :: cycle' xs
         cycle' (y :: ys) = y :: cycle' ys
 
-public export
-partial
-takeUntil : (n -> Bool) -> Stream n -> List n
-takeUntil p (x :: xs)
-    = if p x
-         then [x]
-         else x :: takeUntil p xs
-
-public export
-partial
-takeBefore : (n -> Bool) -> Stream n -> List n
-takeBefore p (x :: xs)
-    = if p x
-         then []
-         else x :: takeBefore p xs
-
-
 --------------------------------------------------------------------------------
 -- Interleavings
 --------------------------------------------------------------------------------
--- a/libs/prelude/Prelude/Types.idr
+++ b/libs/prelude/Prelude/Types.idr
@@ -771,21 +771,21 @@ ceiling x = prim__doubleCeiling x
 
 -- These functions are here to support the range syntax:
 -- range expressions like `[a..b]` are desugared to `rangeFromXXX` calls.
--- They are not exported, but similar functions are exported from
--- `Data.Stream` instead.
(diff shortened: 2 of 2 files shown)


libs/prelude/Prelude/Types.idr
```diff
@@ -771,20 +771,20 @@ ceiling x = prim__doubleCeiling x
 
 -- These functions are here to support the range syntax:
 -- range expressions like `[a..b]` are desugared to `rangeFromXXX` calls.
--- They are not exported, but similar functions are exported from
--- `Data.Stream` instead.
 
-total
```
fabianhjr, line 777: Notice: Module has `%default total` so this is redundant.

libs/base/Data/Stream.idr
```diff
@@ -97,15 +97,15 @@ takeUntil : (n -> Bool) -> Stream n -> List n
...
+         else x :: Stream.takeUntil p xs
 
 public export
 partial
 takeBefore : (n -> Bool) -> Stream n -> List n
 takeBefore p (x :: xs)
     = if p x
          then []
-         else x :: takeBefore p xs
+         else x :: Stream.takeBefore p xs
```
fabianhjr, line 108: Data.Stream `takeUntil` and `takeBefore` should probably be deduplicated since it was defined in Prelude and needs to be `public export` to get some `auto` searches of `Range`s to work.

gallais: Why don't you remove them? `Prelude` is automatically loaded so these will always be clashing.

fabianhjr: Amended

PR #1603, opened 2021-06-24
Remove all mentions of `realpath` from docs

ska80:
`realpath` is no longer needed to run executables.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,11 +9,10 @@ The easiest way to install is via the existing generated Scheme code.
 The requirements are:
 
 - A Scheme compiler; either Chez Scheme (default), or Racket.
-- `bash`, with `realpath`. On Linux, you probably already have this.
-  On a macOS, you can install this with `brew install coreutils`.
-  On FreeBSD, OpenBSD and NetBSD, you can install `realpath` and `GNU make`
-  using a package manager.  For instance, on OpenBSD you can install all of them
-  with `pkg_add coreutils gmake` command.
+- `bash`, `GNU make` and `sha256sum`.  On Linux, you probably already have
+  this.  On macOS and major BSD flavours, you can install them using a package
+  manager: for instance, on macOS, you can install this with the `brew install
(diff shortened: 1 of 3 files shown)


benhormann:
`nproc` isn't required either. Only the benchmark script uses `bash`?

GMP and its headers are required for support/refc, some systems don't have it installed (even after installing make & gcc). To make things awkward, the package name varies: `gmp`, `gmp-dev`, `gmp-devel`, `libgmp`, `libgmp-dev`, `libgmp-devel`... For macOS it's `brew install gmp`.

ska80:
> `nproc` isn't required either. Only the benchmark script uses `bash`?

Does `sysctl -n hw.ncpu` work in other *BSD?

benhormann:
> Does sysctl -n hw.ncpu work in other *BSD?

Yup. At least the popular ones.

benhormann:
Ah, thought `sha256sum` was behind a flag. Hopefully #1584 will remove the dependency by switching to a C implementation.

mattpolzin:
Aside from mentioning gmp as a requirement (at least as of right now), I think this is ready to go, right?

ska80:
> Aside from mentioning gmp as a requirement (at least as of right now), I think this is ready to go, right?

Yep, it is ready. @gallais

mattpolzin:
@ska80 sorry, my comment was not very clear. I meant to suggest that we do add GMP to the list of requirements in this PR because it is indeed needed to successfully build Idris 2 right now. 

Although, even without adding a mention of GMP in this PR, this PR does represent a nice update to the docs!

ska80:
> @ska80 sorry, my comment was not very clear. I meant to suggest that we do add GMP to the list of requirements in this PR because it is indeed needed to successfully build Idris 2 right now.
>
> Although, even without adding a mention of GMP in this PR, this PR does represent a nice update to the docs!

@mattpolzin I will update this PR by mentioning GMP. Thanks for the clarification.

mattpolzin (approved):
Looks good. Thanks!

PR #1608, opened 2021-06-24
Improvements to VMCode

Z-snails:
- Include Foreign definitions in `VMDef`
- Assign a unique tag to each constructor that doesn't alreay have a tag (Types)
- Only project arguments to constructors when they are used
- If a case statement has exactly one alternative and no default, then skip `CASE` instruction.

Overall this reduces the number of instructions significantly.
With the following idris program:

```idris
first : (a, b) -> a
first (x, _) = x

switch : (a, b) -> (b, a)
switch (x, y) = (y, x)

headMaybe : List a -> Maybe a
headMaybe [] = Nothing
headMaybe (x :: _) = Just x

main : IO ()
main = do
    printLn $ first (10, 11)
    printLn $ switch (8, 9)
    printLn $ headMaybe [1 .. 11]
```

- `DECLARE` instructions (number of registers) decrease by 15%
- `CASE` instructions decrease by 40%
- `PROJECT` instructions decrease by 65%
- overall instruction count decrease by 15%


src/Compiler/VMCode.idr
```diff
@@ -101,59 +106,113 @@ Show VMInst where
...
-    = [APPLY res (toReg f) (toReg a)]
+    = pure [APPLY res (toReg f) (toReg a)]
 toVM t res (ALet fc var val body)
-    = toVM False (Loc var) val ++ toVM t res body
+    = [| toVM False (Loc var) val ++ toVM t res body |]
 toVM t res (ACon fc n ci (Just tag) args)
-    = [MKCON res (Left tag) (map toReg args)]
+    = pure [MKCON res tag (map toReg args)]
 toVM t res (ACon fc n ci Nothing args)
-    = [MKCON res (Right n) (map toReg args)]
```
gallais, line 125: Wouldn't it be better to this work on a more high level IR so that the `ACon`
already arrives with an `Int` tag no matter whether the constructor is a type
or data one? This way all of the backends could benefit from this optimisation!

Z-snails: Yep, I'll do that!

Z-snails:
I broke every backend again...
pinging all the backend authors
@ziman @karroffel @bamboo @chrrasmussen @Russoul @mmhelloworld @cypheon @andorp 
This shouldn't be hard to fix.

gallais:
The further tests following the "Quick check" one did not get triggered during
the previous build. I suspect it may be because the updated .yml files are not
on this branch hence the merge. :crossed_fingers:

Z-snails:
I've just realised this possibly won't work with multiple modules thanks to [this](https://github.com/idris-lang/Idris2/blob/master/src/Idris/ModTree.idr#L179). I will have to check that, so don't merge for now

Z-snails:
How I currently have it relies on whole program compilation (in fact how it currently is just doesn't work with more than one module) so a different approach is needed.
Possibly a utility function that ensures all tags are `Just` that the whole program backends can use.
It might also be possible to number every imported module and set that as the top 16 bits of that tag and the type's number inside it's source module to the bottom 16 bits, although that is much more complicated.

Z-snails:
I'm going to split this PR into 2 parts, one with improvements to VMCode (#1662) and one with assigning tags to types.

Z-snails:
I'm going to close this in favour of #1716 .

PR #1610, opened 2021-06-24
[ fix ] Various Test.Golden fixes

gallais:
* Do not pass a codegen choice when the test pool specifies it does
  not require a codegen to be specified.

* padding shouldn't be affected by ANSI escape codes

### DIFF
--- a/libs/test/Test/Golden.idr
+++ b/libs/test/Test/Golden.idr
@@ -228,10 +228,10 @@ runTest opts testPath = forkIO $ do
   let time = timeDifference end start
 
   if result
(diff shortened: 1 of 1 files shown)


jfdm:
I see no reason for this *not* to be merged now.

Thanks for all the hard work!

gallais:
@jfdm I thought you wanted to add a test case to the PR so that's why
I left it open. Sorry about the bug.

jfdm:
Yes, I have a PR waiting. I hadn't thought about contributing to *this* one, and have done it separately.

PR #1611, opened 2021-06-24
Nix: make nixpkgs options overridable

zseri:
This allows for cross-flake testing of content-addressible nix derivations without affecting other users of the flake (which would be the cases if CA-drvs would be hard-coded upstream) and without losing the ability to quickly update to the latest idris2 version (which would happen if a fork of the repo would be used).

See also: https://discourse.nixos.org/t/content-addressed-nix-call-for-testers/12881

The flake can still be used as usual, but it will gain an output `overrideOptions`, which can be used like e.g.
```nix
let
  idrisx = idris.overrideOptions {
    config.contentAddressedByDefault = true;
  };
in { ... }
```

Additonally, this tries to improve `buildIdris` a bit.

### DIFF
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -1,0 +1,12 @@
+rec {
+  # source: https://nixos.org/guides/nix-pills/override-design-pattern.html
+  # but with the ability to change the name of the override method
+  makeOverridable = kind: f:
+    let inner = oldArgs: (f oldArgs) // { ${kind} = newArgs: inner (oldArgs // newArgs); };
+    in  inner;
+
+  # create a flake with overridable options. useful because we can't pass
+  # fine-grained overrides to flakes otherwise, we can only change inputs
+  # (therefore, this can't change inputs per default, and that should also be unnecessary)
+  mkOvrOptsFlake = flakefn: makeOverridable "overrideOptions" flakefn { };
+}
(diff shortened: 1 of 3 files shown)


nix/lib.nix
```diff
@@ -0,0 +1,14 @@
...
+  # source: https://nixos.org/guides/nix-pills/override-design-pattern.html
+  # but with the ability to change the name of the override method
+  makeOverridable = kind: f:
+    let inner = oldArgs: (f oldArgs) // { ${kind} = newArgs: inner (oldArgs // newArgs); };
+    in  inner;
+
+  # create a flake with overridable options. useful because we can't pass
+  # fine-grained overrides to flakes otherwise, we can only change inputs
+  # (therefore, this can't change inputs per default, and that should also be unnecessary)
+  mkOvrOptsFlake = flakefn: makeOverridable "overrideOptions" flakefn { };
```
zseri, line 13: point for discussion: We nay call it `overrideOptions`, but simply `override` may be also appropriate.

nix/lib.nix
```diff
@@ -0,0 +1,14 @@
+{ }:
```
zseri, line 1: point for discussion: this empty argument list may be removed.

nix/lib.nix
claymager: I think I prefer it as is. `override` implies more control over the flake than this provides.

claymager:
Sorry for the merge conflict. @gallais , other than that, ready for merge

PR #1613, opened 2021-06-24
[ fix ] test Test.Golden with non-idris2 projects.

jfdm:
Links with  #1610 and provides a simple test to catch the issue. The test should probably be more exhaustive but that can come later.

This will need to be updated once #1610 is merged so that the expected file is correct.

ping @gallais


jfdm:
this should be ready now!, works locally.

gallais:
Would this have caught the issue? The executable is ignoring the arguments
passed to it so I think this would have been green even with the bug that
we found today. I think something as simple as adding `echo "$1"` to the
`run` file that only contains `$1` at the moment should at least give us a
warning if we start passing extra arguments without notice.

jfdm:
> Would this have caught the issue?

Running the tests before I rebased onto main, showed that the error was caught. The tests that discovered the bug, only had a single argument. That being said I agree that we should explicitly force the issue.

gallais:
Could be a fluke on my end but I was trying to see what the error would be so I tried
and it wrongly worked:

```shell
:~/idris2$ git log -n 1
commit f9eafa812c5726a8824800726504024d5333b255 (HEAD -> yo-dawg)
Author: Guillaume ALLAIS <guillaume.allais@ens-lyon.org>
Date:   Fri Jun 25 09:22:50 2021 +0100

    Revert "[ fix ] Only pass --cg to test exec if requested"
    
    This reverts commit c2179115737c16e6d7c251e8adc350141f32f8e5.
:~/idris2$ make test only=gold
make[1]: Entering directory '/home/gxa1/projects/idris2'
make[1]: '/home/gxa1/projects/idris2/build/env/idris2-0.4.0' is up to date.
make[1]: Leaving directory '/home/gxa1/projects/idris2'
make[1]: Entering directory '/home/gxa1/projects/idris2/tests'
/home/gxa1/projects/idris2/build/exec/idris2 --build tests.ipkg
make[1]: Leaving directory '/home/gxa1/projects/idris2/tests'

NOTE: `make test` does not rebuild Idris or the libraries packaged with it; to do that run `make`

make[1]: Entering directory '/home/gxa1/projects/idris2/tests'
./build/exec/runtests /home/gxa1/projects/idris2/build/exec/idris2 --interactive --timing --failure-file failures --threads 8 --only gold

------------------------------------------------------------------------
Misc
------------------------------------------------------------------------

idris2/golden001: success                                        01.694s
1/1 tests successful
make[1]: Leaving directory '/home/gxa1/projects/idris2/tests'
```

I'll add the `echo "$1"` because that does fail with:

```shell
Expected:
../build/exec/hello
Hello World

Given:
../build/exec/hello --cg chez
Hello World
```

jfdm:
test suites in test suites are hard! But yeah, i've added the update.

gallais:
The issue with self host looks like a fluke: `sh` was found just fine during the
last build. I've restarted this one.

jfdm:
Shall we merge this? Every thing looks fine...

jfdm:
thanks!

PR #1621, opened 2021-06-27
Support for incremental compilation with Chez

edwinb:
This adds a new flag '--inc <backend>' which, if set, compiles all modules incrementally, and for executables, links the incrementally compiled modules rather than building the whole program. Also, adds an environment variable IDRIS2_INC_CGS for providing a comma separated list of backends to use for incremental builds.

Also, adds '--whole-program', which overrides incremental builds for an executable.

Libraries are now built with IDRIS2_INC_CGS set, which means they'll still build okay on older versions, but now also allow Idris itself to be built incrementally. This massively reduces turnaround time during development (I know because I've been using this feature to implement this feature :)).

Incremental builds are much faster if there's nothing to recompile, but for the Chez backend, generate code which runs at about half the speed.

Currently only works for Chez - other backends ignore the flag.

Also, incremental building of an executable will only work if *all* required modules have been built incrementally for the backend in use. I've added a page to the documentation to explain all this, and updated the CHANGELOG and INSTALL with brief details.

We may want to consider adding a CI job for an incremental build too.

### DIFF
--- a/.github/linters/.markdown-lint.yml
+++ b/.github/linters/.markdown-lint.yml
@@ -1,0 +1,37 @@
+---
+###########################
+###########################
+## Markdown Linter rules ##
+###########################
+###########################
+
+# Linter rules doc:
+# - https://github.com/DavidAnson/markdownlint
+#
+# Note:
+# To comment out a single error:
+#   <!-- markdownlint-disable -->
+#   any violations you want
+#   <!-- markdownlint-restore -->
+#
+
+###############
+# Rules by id #
+###############
+MD004: false                  # Unordered list style
+MD007:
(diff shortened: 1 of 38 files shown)


edwinb:
By the way, I'm leaving the `chez-sep` backend for the moment, but I think this will remove the need for it so if this works out perhaps we should consider dropping it?

edwinb:
I knew there was no way this was going to work smoothly...

edwinb:
Also interestingly, the incremental build of the libraries has caught some places where methods are missing from an interface (they show up as warnings). I had forgotten that we don't check that properly yet, but I'll leave that for a separate patch.

PR #1624, opened 2021-06-27
Add Split and SplitRec views for Vect

technic:
* ported from the Idris1 base library
* the counter by list deconstruction is replaced with Nat counter

Hi,
I adopted the splitRec View from Idris1 [defined here](https://github.com/idris-lang/Idris-dev/blob/a13caeb4e50d0c096d34506f2ebf6b9d140a07aa/libs/base/Data/Vect/Views.idr#L25), which is referenced in the exercise after Chapter 10.2 in tydd book.
The complexity O(...) comment is left from the Idris1 implementation, however please check that it is still valid.
Some of the arguments I have to make explicit.

Cheers,


libs/contrib/Data/Vect/Views/Extra.idr
```diff
@@ -0,0 +1,66 @@
...
+total
+splitHelp : {n: Nat} -> (head : a) -> (zs : Vect n a) ->
+            Nat -> Split (head :: zs)
+splitHelp head [] _ = SplitOne
+splitHelp head (z :: zs) 0 = SplitPair {xs = []} {ys = zs}
+splitHelp head (z :: zs) 1 = SplitPair {xs = []} {ys = zs}
+splitHelp head (z :: zs) (S (S k))
+  = case splitHelp head zs k of
+         SplitOne => SplitPair {xs = []} {ys = []}
+         SplitPair {xs} {ys} => SplitPair {xs = z :: xs} {ys = ys}
```
gallais, line 28: If you're going to always have to pass `xs` and `ys` explicitly
then they should probably be explicit arguments of the `SplitPair`
constructor.

libs/contrib/Data/Vect/Views/Extra.idr
```diff
@@ -0,0 +1,66 @@
...
+
+||| View for splitting a vector in half, recursively
+|||
+||| This allows us to define recursive functions which repeatedly split vectors
+||| in half, with base cases for the empty and singleton lists.
+public export
+data SplitRec : Vect k a -> Type where
+  SplitRecNil : SplitRec []
+  SplitRecOne : SplitRec [x]
+  SplitRecPair : {n : Nat} -> {m : Nat} -> {xs : Vect n a} -> {ys : Vect m a} ->
```
gallais, line 45: Shouldn't we have `{xs : Vect (S n) a} -> {ys : Vect (S m) a} ->` here?
Otherwise the split is not necessarily unique: `SplitRecPair SplitNil SplitNil`
and `SplitNil` are 2 valid splits of the same empty vector.

libs/contrib/Data/Vect/Views/Extra.idr
buzden: It's not necessary to pass `ys`'s (you can see all typechecking here: https://github.com/buzden/Idris2/commit/c2037d8450e78f5c335ce737124269a714832767)

technic: @buzden thanks! Can you add your commit to this PR?

technic: ... or actually I can merge your changes myself.

PS sorry for messed up white-spaces :-/

libs/contrib/Data/Vect/Views/Extra.idr
```diff
@@ -0,0 +1,66 @@
...
+
+||| Covering function for the `SplitRec` view
+||| Constructs the view in O(n lg n)
+public export total
+splitRec : {k : Nat} -> (xs : Vect k a) -> SplitRec xs
+splitRec input with (sizeAccessible k)
+  splitRec input | acc with (split input)
+    splitRec {k = 0} []  | acc | SplitNil = SplitRecNil
+    splitRec {k = 1} [x] | acc | SplitOne = SplitRecOne
+    splitRec {k = S nl + S nr} (x :: xs ++ y :: ys) | Access rec | SplitPair {x} {xs} {y} {ys} =
```
gallais, line 63: The fact that we still need to mention `xs` explicitly in the definition
of `splitHelp` and that the use site explicitly binds the various bits of
the split still suggests to me that all of these arguments ought to be
explicit.

libs/contrib/Data/Vect/Views/Extra.idr
```diff
@@ -0,0 +1,66 @@
...
+      SplitRecPair
+        (splitRec {k = S nl} (x :: xs) | rec _ $ smallerPlusL nl nr)
```
gallais, line 65: Do we really need to specify `k` explicitly? I expect it should be inferred
from the type of `x :: xs`.

technic: Yep, the compiler was complaining about it, e.g.

Error: While processing left hand side of with block in with block in splitRec. When unifying Split [] and Split [].
Pattern variable k unifies with: 0.

gallais: I mean in the recursive calls passed to `SplitRecPair`.

technic: I didn't manage to make it work otherwise, but this automatic deduction is kind of magic to me, sorry.

libs/contrib/Data/Vect/Views/Extra.idr
```diff
@@ -0,0 +1,66 @@
...
+                Split (x :: xs ++ y :: ys)
+
+total
+splitHelp : {n : Nat} -> (head : a) -> (zs : Vect n a) ->
+            Nat -> Split (head :: zs)
+splitHelp head [] _ = SplitOne
+splitHelp head (z :: zs) 0 = SplitPair {xs = []}
+splitHelp head (z :: zs) 1 = SplitPair {xs = []}
+splitHelp head (z :: zs) (S (S k))
+  = case splitHelp head zs k of
```
gallais, line 26: This recursive call should probably be `splitHelp z zs k`.
This becomes evident when `SplitPair`'s `x, y, xs, ys` are all explicit arguments.

technic: Actually, you are right, but the code surprisingly worked correctly.

PR #1629, opened 2021-06-28
[ new ] Missing integer type interfaces

stefan-hoeck:


libs/prelude/Prelude/EqOrd.idr
```diff
@@ -181,6 +197,42 @@ Ord Bits64 where
   (>) x y = intToBool (prim__gt_Bits64 x y)
   (>=) x y = intToBool (prim__gte_Bits64 x y)
 
+public export
+Ord Int8 where
+  compare x y = if x < y then LT else if x == y then EQ else GT
```
gallais, line 202: In [this commit](https://github.com/idris-lang/Idris2/pull/1580/commits/724db0dc18bdecd95db2395d0d6dfc506d1e8da0#diff-91a27794ac7278a19f8524c32fe13371b87913969e6530cf1e70766e6e8e0f2eR91-R92) I add a default definition of `compare` to avoid having to repeat this everywhere.

stefan-hoeck: So, should we wait for that PR then, and afterwards I'll rebase and remove the explicit `compare` implementations?

gallais: You could probably add that line right now & it shouldn't create a merge conflict
given it's the exact same code. :crossed_fingers:

gallais: Nevermind: it breaks a test case so it probably easier to wait for the PR to be merged
and then rebase on top of it.

PR #1642, opened 2021-06-29
[ new ] Support reflecting on new intX types

stefan-hoeck:


gallais (commented):
Should these be sized i.e. `| I8 Int8`?

stefan-hoeck:
> Should these be sized i.e. `| I8 Int8`?

I can't think of a reason why they shouldn't. I'll give it a go tonight to see, whether tests start failing.

Right now they reflect the types used in `Constant` in the compiler. Once the dust settles, I think there we should as well use the proper primitives.

stefan-hoeck:
No problems locally with the suggested change. Let's see what CI makes out of it. :crossed_fingers:

PR #1643, opened 2021-06-29
[ actions ] add quick check step

gallais:
Rather than launching all of the builds immediately, we first try
to make sure that the project builds.


gallais:
The additional checks are only run if they're part of the default branch so it's not
surprising this PR is only running the 4 basic checks. Cf. https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run

PR #1649, opened 2021-06-30
[ docs ] Record docs fixes

cypheon:
This PR fixes a few issues with docs for records:
 * docstrings for field projections were missing when listing the docs for a record type and when looking up docs with the `:doc (.fieldName)` syntax
 * in the HTML docs, all field projections were shown in one line, now they're correctly formatted
 * in the HTML docs, all field names were wrapped in parentheses, now that happens only if their name is an operator

### DIFF
--- a/src/Idris/Doc/String.idr
+++ b/src/Idris/Doc/String.idr
@@ -244,11 +244,10 @@ getDocsForName fc n
                 | Nothing => pure Empty
            ty <- resugar [] =<< normaliseHoles defs [] (type def)
            let prettyName = pretty (nameRoot nm)
-           let projDecl = hsep [ fun nm prettyName, colon, prettyTerm ty ]
+           let projDecl = annotate (Decl nm) $ hsep [ fun nm prettyName, colon, prettyTerm ty ]
            let [(_, str)] = lookupName nm (docstrings syn)
                   | _ => pure projDecl
-           pure $ annotate (Decl nm)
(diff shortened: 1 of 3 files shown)


src/Idris/Doc/String.idr
```diff
@@ -244,13 +244,14 @@ getDocsForName fc n
...
-           let [(_, str)] = lookupName nm (docstrings syn)
-                  | _ => pure projDecl
-           pure $ annotate (Decl nm)
-                $ vcat [ projDecl
-                       , annotate DocStringBody $ vcat (reflowDoc str)
-                       ]
+           let projDecl = annotate (Decl nm) $ hsep [ fun nm prettyName, colon, prettyTerm ty ]
+           case lookupName nm (docstrings syn) of
+                [(_, "")] => pure projDecl
+                [(_, str)] =>
```
cypheon, line 250: It seems we always get a docstring back (just that it's sometimes the empty string). I wasn't sure how to best handle missing docstrings. E.g. for interface methods a missing docstring currently result in an empty line in the REPL output. IMHO this way it looks nicer (saves space) but it complicates the code a bit.

gallais: It looks like we're systematically using `option "" documentation` in the parser.
We should probably refactor this to use `optional documentation` instead and only
bother adding a doc string if there is actually one.

But this will do nicely in the meantime!

Issue #1325, opened 2021-04-22
Redesign Functor interface hierarchy ala Semigroupoids


stefan-hoeck:
Today, I was in need of `Alternative` for `Either e` (well, only the `<|>` part, but that's the whole point). Yet, of course, there is no such thing, as we can't implement `empty`. The "alt" operator `<|>`, however, could be implemented and would be very useful. In a similar way, we could have bind (`>>=`) for map-like structures like `Data.SortedMap`, but not `pure`, so there is neither `Applicative` nor `Monad` for those.

The Haskell library [semigroupoids](https://hackage.haskell.org/package/semigroupoids) fixes this by introducing several intermediary typeclasses. The following diagram was shamelessly copied from there:

```
Foldable ----> Traversable <--- Functor ------> Alt ---------> Plus           Semigroupoid
     |               |            |                              |                  |
     v               v            v                              v                  v
Foldable1 ---> Traversable1     Apply --------> Applicative -> Alternative      Category
                                  |               |              |                  |
                                  v               v              v                  v
                                Bind ---------> Monad -------> MonadPlus          Arrow
```

So here's my suggestion: As Idris2 is still at a very early state, we could still redesign the interface hierarchy in the Prelude in a similar manner as what's shown above (minus `Plus` and `MonadPlus`: We won't need those). The main pain point would of course be, how much such a change would affect bootstrapping, and in what order things would have to be adjusted where (I'll forever struggle with that one :-) ).

Here are two use cases this would allow us to abstract over:

- Select the first computation not failing with an error from a non-empty list of choices (nees `Alt` and `Foldable1`)
- Use `WriterT` to accumulate values with a `Semigroup` but no `Monoid` instance (`List1`, for instance, but also `First` and `Last` (both not yet in Idris2), which keep the first or last occurrence of a value).

buzden:
I'd like to mention that transition from `Category` to `Arrow` is also too far. I ran into it when using arrow-like classes for invertible computations -- stuff like `***` and `+++` is useful, but `Arrow` also requires `arr` which contradicts reversibility.

An example of gradual `Category` to `Arrow` transition can be seen there: https://hackage.haskell.org/package/semi-iso-1.0.0.0/docs/Control-Category-Structures.html

ysangkok:
[...]

PR #1661, opened 2021-07-05
[ refactor, breaking ] Redesign Functor interface hierarchy as suggested in #1325

stefan-hoeck:
This was much less painful than I thought it would be (though, I should probably wait until CI had its say). Luckily, there was only one `Monad` in the compiler libs (Data.List.Lazy) and the interface was actually never used so I just removed the implementation. Some instances of `Apply`, `Alt`, or `Bind` might still be missing, and I'll need to add some additional docstrings.

@buzden: You mentioned that `Control.Arrow` should also be split up. I didn't do this so far, since there is also talk on discord about adding `Profunctor` and related interfaces to contrib. This will allow us to move some stuff from `Arrow` to `Profunctor`-related interfaces anyway.

### DIFF
--- a/libs/prelude/Prelude/Interfaces.idr
+++ b/libs/prelude/Prelude/Interfaces.idr
@@ -150,41 +150,57 @@ public export
 mapHom : Bifunctor f => (a -> b) -> f a a -> f b b
 mapHom f = bimap f f
 
+||| An applicative functor without `pure`.
 public export
-interface Functor f => Applicative f where
-  constructor MkApplicative
-  pure : a -> f a
+interface Functor f => Apply f where
+  constructor MkApply
   (<*>) : f (a -> b) -> f a -> f b
 
 public export
-(<*) : Applicative f => f a -> f b -> f a
+(<*) : Apply f => f a -> f b -> f a
 a <* b = map const a <*> b
 
 public export
-(*>) : Applicative f => f a -> f b -> f b
+(*>) : Apply f => f a -> f b -> f b
 a *> b = map (const id) a <*> b
 
-%allow_overloads pure
 %allow_overloads (<*)
 %allow_overloads (*>)
+
+namespace Apply
+  ||| Composition of applicative functors is an applicative functor.
+  public export
+  [Compose] (Apply f, Apply g) => Apply (f . g)
+    using Functor.Compose where
+      fun <*> x = (<*>) <$> fun <*> x
+
+public export
+interface Apply f => Applicative f where
+  constructor MkApplicative
+  pure : a -> f a
+
+%allow_overloads pure
 
 namespace Applicative
   ||| Composition of applicative functors is an applicative functor.
   public export
   [Compose] (Applicative f, Applicative g) => Applicative (f . g)
-    using Functor.Compose where
+    using Apply.Compose where
       pure = pure . pure
-      fun <*> x = [| fun <*> x |]
 
 public export
-interface Applicative f => Alternative f where
-  constructor MkAlternative
-  empty : f a
+interface Apply f => Alt f where
+  constructor MkAlt
   (<|>) : f a -> Lazy (f a) -> f a
 
 public export
-interface Applicative m => Monad m where
-  constructor MkMonad
+interface Applicative f => Alt f => Alternative f where
+  constructor MkAlternative
+  empty : f a
+
+public export
+interface Apply m => Bind m where
+  constructor MkBind
   ||| Also called `bind`.
   (>>=) : m a -> (a -> m b) -> m b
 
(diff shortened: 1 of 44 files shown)


stefan-hoeck:
Does anybody have an idea why the chez/futures001 test fails on Windows? I don't have a Windows machine, so I probably won't be able to figure out what's going on there.

stefan-hoeck:
> Does anybody have an idea why the chez/futures001 test fails on Windows? I don't have a Windows machine, so I probably won't be able to figure out what's going on there.

Now it completes without failing. Seems like a flaky bit of testing...

libs/prelude/Prelude/Interfaces.idr
```diff
@@ -150,41 +150,69 @@ public export
...
 public export
-interface Applicative f => Alternative f where
+interface Apply f => Alt f where
+  constructor MkAlt
+  (<|>) : f a -> Lazy (f a) -> f a
+
+public export
+interface Applicative f => Alt f => Alternative f where
   constructor MkAlternative
   empty : f a
```
gallais, line 211: I guess this design decision echoes with #1472: Why should alternative be an `Alt Applicative` with a point rather than a `Pointed Applicative` (`interface Pointed a where point : a`) with
a `Sum`?

Should we flatten the hierarchy entirely?

stefan-hoeck: I'm afraid I can't follow you here. Could you please quickly explain how this is related to #1472 and how you'd like to flatten the hierarchy?

stefan-hoeck: So, I had a look at how Purescript does it and added an intermediate interface `Plus` plus some doc strings (which I more or less copied from Purescript as well). Reference: [purescript-control](https://pursuit.purescript.org/packages/purescript-control/5.0.0).

stefan-hoeck: In addition, I somewhat reorganized `Prelude.Interfaces` to more cleanly group interfaces in the same hierarchy.

gallais: This is the relevant quote:

> Preorder is transitive and reflexive, and Equivalence is a symmetric preorder. But how does a "tolerance relation", which is reflexive and transitive, fit into this scheme? Why does an equivalence relation have to be defined as a symmetric preorder instead of a transitive tolerance relation? Why bring a preorder into the definition of equivalence at all?

Why do we define an applicative as an apply with pure rather than lift (`interface Lift f where pure : a -> f a`) with an apply?

In #1472 we decided to have separate constraints for:
* reflexive
* symmetric
* transitive
* etc.
and define common notion such as equivalence as aliases to the set of constraints they represent.

```idris
interface (Reflexive r, Symmetric r, Transitive r) => Equivalence r where
  -- nothing
```

rather than

```idris
interface Preorder r => Equivalence r where
  symmetric : r a b -> r b a
```


I'm asking whether we should be doing the same here:

```idris
interface Functor f where
  map : (a -> b) -> f a -> f b

interface Lift f where
  pure : a -> f a

interface Apply f where
  (<*>) : f (a -> b) -> f a -> f b

interface Bind f where
  (>>=) : f a -> (a -> f b) -> f b

interface (Lift f, Apply f) => Applicative f where
interface (Applicative f, Bind f) => Monad f where
```

stefan-hoeck: I'll give this a try and will report back here if some unexpected problems occurred. One advantage I see already that for Monads, `(<*>)` can be implemented in terms of `(>>=)`, which is tedious if we have a strict hierarchy. With your suggestion, we could just implement `Functor` and `Bind` first.

One disadvantage would be, that we generate some intermediary interfaces without a proper set of laws. Not sure, what the purists think of this...

stefan-hoeck: So, here's what I found: We need `Functor` => `Bind`, otherwise we will not be able to choose whether to implement `join` or `>>=` or both, which I consider to be quite a drawback. Therefore, if we (almost) completely flatten the `Monad` hierarchy, here's what I suggest:

* `Functor f` : `map : (a -> b) -> f a -> f b`
* `Apply f` : `(<*>) : f (a -> b) -> f a -> f b`
* `Lift f` : `pure : a -> f a`
* `Functor f => Bind f` : `(>>=) : f a -> (a -> f b) -> f b`
* `Functor f => Apply f => Semiapplicative f`
* `Semiapplicative f => Lift f -> Applicative f`
* `Bind f => Apply f => Semimonad f`
* `Semimonad f => Lift f => Monad f`

The names `Semiapplicative` and `Semimonad` are taken from [this issue in the semigroupoids library](https://github.com/ekmett/semigroupoids/issues/26).

In addition, we'll have

* `Alt f` : `(<|>) : f a -> f a -> f a`
* `Alt f => Plus f` : `empty : f a`
* `Alt f => Applicative f => Alternative f`

I haven't inlcuded this new version here yet, since this is quite a bit of work and I first would like to here your thoughts.

stefan-hoeck: After these experiments, just to give my personal opinion on this: I'd like to stick with the hierarchy suggested in #1325. This is, how it is done in Purescript and in haskell's semigroupoids library, so people familiar with one of these will be comfortable immediately. If we flatten the hierarchy, people will have to write (dummy) implementations for three more interfaces: `Lifted`, `Semiapplicative`, and `Semimonad`. We could give default definitions for the latter two, but some of these (for instance the IO ones) are marked as to be inlined, and I'm not sure how this will be affected by default definitions.

Therefore, I think the hierarchy as it is right now (in this PR) is the way to go.

edwinb:
This looks like it's quite a big breaking change. So, while I have nothing particular to contribute to the discussion at the moment, because I haven't thought it through, I just want to say: please don't rush into merging this, because there seems to be quite a lot of possible design choices to make, and a lot of tradeoffs to consider. Also, please make sure all the documentation (mainly, the tutorial) is consistent before merging.

As a general point, I'm not that keen on complex hierarchies with lots of constraints, I prefer it to be reasonably flat so I don't have to think about too many things at once. On the other hand, I don't want to lose the ability to define `Monad` via either `>>=` or `join`.

PR #1662, opened 2021-07-06
[ improvement ] VMCode

Z-snails:
See #1608 for context.

This PR includes the improvements to `VMCode`, namely:
- Include Foreign definitions in `VMDef`
- Only project arguments to constructors when they are used
- If a case statement has exactly one alternative and no default, then skip CASE instruction (speeds up record projections).

### DIFF
--- a/src/Compiler/Interpreter/VMCode.idr
+++ b/src/Compiler/Interpreter/VMCode.idr
@@ -1,0 +1,282 @@
+module Compiler.Interpreter.VMCode
+
+
+import Core.Core
+import Core.Context
+import Core.Context.Log
+import Core.Primitives
+import Core.Value
+import Compiler.Common
+import Compiler.VMCode
+import Data.IOArray
+import Libraries.Data.NameMap
+import Data.Nat
+import Data.Vect
+
+public export
+data Object : Type where
+    Closure : (predMissing : Nat) -> (args : List Object) -> Name -> Object
+    Constructor : (tag : Either Int Name) -> (args : List Object) -> Object
+    Const : Constant -> Object
+    Null : Object
+
+showType : Object -> String
+showType (Closure _ _ _) = "Closure"
+showType (Constructor _ _) = "Constructor"
+showType (Const _) = "Constant"
+showType Null = "Null"
+
+mutual
+    showSep : Nat -> List Object -> String
+    showSep k [] = ""
+    showSep k [o] = showDepth k o
+    showSep k (o :: os) = showDepth k o ++ ", " ++ showSep k os
+
+    showDepth : Nat -> Object -> String
+    showDepth (S k) (Closure mis args fn) = show fn ++ "-" ++ show mis ++ "(" ++ showSep k args ++ ")"
+    showDepth (S k) (Constructor (Left t) args) = "tag" ++ show t ++ "(" ++ showSep k args ++ ")"
+    showDepth (S k) (Const c) = show c
+    showDepth _ obj = showType obj
+
+Show Object where
+    show = showDepth 5
+
+data State : Type where
+record InterpState where
+    constructor MkInterpState
+    defs : NameMap VMDef
+    locals : IOArray Object
+    return : Maybe Object
+
+initInterpState : List (Name, VMDef) -> Core InterpState
+initInterpState defsList = do
+    let defs = fromList defsList
+    locals <- coreLift $ newArray 0
+    let return = Nothing
+    pure $ MkInterpState defs locals return
+
+0
+Stack : Type
+Stack = List Name
+
+interpError : Ref State InterpState => Stack -> String -> Core a
+interpError stk msg = do
+    MkInterpState _ ls ret <- get State
+    lsList <- coreLift $ toList ls
+    throw $ InternalError $ "Interpreter Error in " ++ show (take 10 stk) ++ ":\n" ++ msg
+        ++ "\n\nlocals:\n" ++ showWithIndex lsList
+        ++ "\nreturn:\n  " ++ show ret
+  where
+    showWithIndex : forall a. {default 0 idx : Nat} -> Show a => List a -> String
+    showWithIndex {idx} [] = ""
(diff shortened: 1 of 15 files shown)


gallais:
:+1: 

Do we have an official backend using the VM IR i.e. are these changed being tested by the CI?

Z-snails:
There's no official backend, I'll test it with @cypheon's LLVM backend.
For testing purposes we could write interpreters for all the backends, and test them that way, but that should probably be a seperate PR.

src/Compiler/VMCode.idr
```diff
@@ -127,19 +165,18 @@ toVM t res (AOp fc _ op args)
     = [OP res op (map toReg args)]
 toVM t res (AExtPrim fc _ p args)
     = [EXTPRIM res p (map toReg args)]
+toVM t res (AConCase fc (ALocal scr) [MkAConAlt n ci mt args code] Nothing) -- exactly one alternative, so skip matching
+    = let body = toVM t res code
+          used = foldMap collectUsed body
+       in projectArgs scr 0 used args ++ body
```
stefan-hoeck, line 171: Such an optimization is also possible for cases with zero alternatives and just a single default value as well as for matches on constants. Or are all of these already optimized in some earlier IR (in which case this should probably be reflected by the types)? I'm asking this because the refactored JS backend in #1609 performs the same optimizations.

Z-snails: I don't know if that happens, but I've only optimised this because of how common record projections are. I don't expect it to be very common to have no alternatives and just a default, but that can be easily optimised at the `CExp` level

Z-snails:
All the tests that haven't been broken by changes to idris2 pass.

Z-snails:
I've written a vmcode interpreter and added it as a new backend.
It actually picked up on an error I made.

gallais:
Oh wow, that's fantastic! :star_struck:

src/Compiler/Interpreter/VMCode.idr
```diff
@@ -0,0 +1,281 @@
...
+        setReg stk target $ Closure (pred missing) argObjs fn
+    step stk (MKCONSTANT target c) = setReg stk target $ Const c
+    step stk (APPLY target fn arg) = do
+        fnObj <- getReg stk fn
+        argObj <- getReg stk arg
+        case fnObj of
+            Closure Z args fn => do
+                res <- callFunc stk fn (args ++ [argObj])
+                setReg stk target res
+            Closure (S k) args fn => setReg stk target $ Closure k (args ++ [argObj]) fn
```
gallais, line 190: Should `Closure` store its list of args using a snoc list? You'd have:

```idris
callFunc stk fn (args <>> [argObj])
setReg stk target $ Closure k (args :< argObj)
```

`(<>>)` is still linear in `args` but `(:<)` is constant time!

src/Compiler/Interpreter/VMCode.idr
```diff
@@ -0,0 +1,281 @@
...
+                Just arg => setReg stk target arg
+            _ => interpError stk $ "PROJECT: Expected Constructor, found " ++ showType valObj
+    step stk (NULL reg) = setReg stk reg Null
+    step stk (ERROR msg) = interpError stk $ "ERROR: " ++ msg
+
+    callFunc : Ref State InterpState => Stack -> Name -> List Object -> Core Object
+    callFunc stk fn args = saveLocals $ do
+        let ind = pack $ '|' <$ stk
+        logCallStack <- logging "compiler.interpreter" 25
+        when logCallStack $ coreLift $ putStrLn $ ind ++ "Calling " ++ show fn ++ " with args: " ++ show args
```
gallais, line 250: AFAIU this is just

```suggestion
        log "compiler.interpreter" 25 $ ind ++ "Calling " ++ show fn ++ " with args: " ++ show args
```

src/Compiler/Interpreter/VMCode.idr
```diff
@@ -0,0 +1,281 @@
...
+
+Show Object where
+    show = showDepth 5
+
+data State : Type where
+record InterpState where
+    constructor MkInterpState
+    defs : NameMap VMDef
+    locals : IOArray Object
+    return : Maybe Object
```
gallais, line 49: I think `return` is a bit confusing here. Later down there's something that looks like:
```
case return <$> get State of
  Nothing => branch1
  Just obj => branch2
```
and my first reaction was to suggest regactoring it as
```
obj <- get State
branch2
```
because it looked like `return <$>` was just going under the `Core` layer and adding
a `Just` constructor! It took me quite some time to figure out that the code was fine
as is.

src/Compiler/Interpreter/VMCode.idr
```diff
@@ -0,0 +1,281 @@
...
+    callFunc stk fn args = saveLocals $ do
+        let ind = pack $ '|' <$ stk
+        logCallStack <- logging "compiler.interpreter" 25
+        when logCallStack $ coreLift $ putStrLn $ ind ++ "Calling " ++ show fn ++ " with args: " ++ show args
+        let stk' = fn :: stk
+        defs <- defs <$> get State
+        res <- case lookup fn defs of
+            Nothing => interpError stk $ "Undefined function: " ++ show fn
+            Just (MkVMFun as is) => do
+                is' <- beginFunction (zip as args) is (foldl max (-1) as)
```
gallais, line 256: Do we want to check here that `length as == length args`?

src/Compiler/Interpreter/VMCode.idr
Z-snails: I think `log` would add more noise at the start of the line.

src/Compiler/Interpreter/VMCode.idr
Z-snails: Ah, ok. I never ever used `return` instead of `pure` when I used Haskell, but I'll change it to avoid confusion.

src/Compiler/Interpreter/VMCode.idr
Z-snails: Yep, I'll do that.

PR #1668, opened 2021-07-07
[fix] refc backend broken with msys2

rujialiu:
I was getting "The system cannot find the drive specified" when compiling any code with "--cg refc", with msys2, no additional environment variables set.
Tried two different PCs both with Windows 10 + fresh msys2 install.

The problem is that operator `</>` is infix 5 so we're regarding the concatenation of everything (including `cc`) on the left side of `</>` a path, which is not valid (when CC is not set to an absolute path), so we end up with an invalid path.

### DIFF
--- a/src/Compiler/RefC/CC.idr
+++ b/src/Compiler/RefC/CC.idr
@@ -60,7 +60,7 @@ compileCFile {asShared} objectFile outFile =
 
      let runcc = cc ++ " -Werror " ++ sharedFlag ++ objectFile ++
                        " -o " ++ outFile ++ " " ++
-                       fullprefix_dir dirs "lib" </> "libidris2_support.a" ++ " " ++
+                       fullprefix_dir dirs ("lib" </> "libidris2_support.a") ++ " " ++
                        "-lidris2_refc " ++
                        "-L" ++ fullprefix_dir dirs "refc " ++
                        clibdirs (lib_dirs dirs) ++
--- a/src/Compiler/RefC/CC.idr
+++ b/src/Compiler/RefC/CC.idr
@@ -60,7 +60,7 @@ compileCFile {asShared} objectFile outFile =
 
      let runcc = cc ++ " -Werror " ++ sharedFlag ++ objectFile ++
                        " -o " ++ outFile ++ " " ++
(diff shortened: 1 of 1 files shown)


ohad:
Tagging @madman-bob

madman-bob:
That's a good point about operator precedence. I think this should be

```idris
(fullprefix_dir dirs "lib" </> "libidris2_support.a")
```

That is, we find the directory, and then concatenate the file; rather than searching for a directory that's actually a file.

rujialiu:
> That is, we find the directory, and then concatenate the file; rather than searching for a directory that's actually a file.

Yeah, that was my initial thought too. Then I realized that there are lots of calls to `fullprefix_dir` without paranthesis. Changing them all seems overkill for this fix, but adding a paranthesis around the whole call seems to be "inconsistent" with other calls, so at the end I chose to minimize the "affected code"

Anyway, I definitely agree with you. If you think it's better to change them all, just let me know. I'm studying RefC backend (and also your idris2-python on top of that) recently, so I'd be happy to contribute more.

madman-bob:
That seems to be the only use of `fullprefix_dir` in combination with a `</>`, so I'd say only that occurence should be changed.

May I ask what you're using these backends for? @ohad is always eager to hear of the impact of this project.

rujialiu:
> May I ask what you're using these backends for? @ohad is always eager to hear of the impact of this project.
I'm experimenting how to utilize functional languages with dependent type in areas where refactoring is frequent, but correctness is very important. But I'm very new to dependent types, so it will take some time before I have any concrete plan. But I hope I can use it in mobile app, browser and server side, so I'm most interested in RefC and JS backend.

As I said in discord (Industry), I will start by implementing some small plugins in Idris. But I need to integrate libBF to replace GMP before using it in our commercial product first, though I don't know how long will it take.

madman-bob:
Ah, I hadn't noticed you were the same person.

Interesting stuff. If you need help with RefC, I can probably point you towards what needs changing.

rujialiu:
> Interesting stuff. If you need help with RefC, I can probably point you towards what needs changing.

Thank you :) Anything else I need to do before it could be merged?

PR #1676, opened 2021-07-10
Need --script for incrementally compiled apps

edwinb:
Otherwise it doesn't load the compiled modules and can't find the compiled definitions!


edwinb:
The output here suggests to me that incremental compilation doesn't work on Windows, which perhaps wouldn't be surprising since I haven't tested it on Windows until now. The error message suggests it can't find the relevant .so, which should be in the app's subdirectory.

PR #1689, opened 2021-07-12
Use gsha256sum when sha256sum doesn't exist [ Re: #1584 ]

fabianhjr:


src/Core/Options.idr
```diff
@@ -229,11 +231,21 @@ defaultElab : ElabDirectives
...
-                     defaultElab Nothing Nothing
-                     (MkPrimNs Nothing Nothing Nothing Nothing) []
-                     []
+defaultHashFn : Core String
+defaultHashFn
+    = do Nothing <- coreLift $ pathLookup ["sha256sum", "gsha256sum", "sha256"]
+           | Just p => pure p
+         Nothing <- coreLift $ pathLookup ["openssl"]
+           | Just p => pure "openssl sha256 -r"
+         pure "sha256sum"
```
gallais, line 240: Should this be an exception? We've already attempted pathLookup on `"sha256sum"`
so a priori it does not exist and returning it has no chance of success?

fabianhjr: Yeah, it should be an exception.

fabianhjr: Amending

benhormann:
@fabianhjr BSD's `sha256` requires `-r` (except NetBSD has no such option - [man page](https://man.netbsd.org/sha256.1)).

Why not opt for BSD style output instead?
Use coreutil's `--tag` and drop `-r` for openssl. Change [src/Libraries/Utils/Binary.idr#L484](https://github.com/idris-lang/Idris2/blob/bc9ff537248500b143d2b7da479f1e51a0f2d500/src/Libraries/Utils/Binary.idr#L484) to grab the last word.

src/Core/Options.idr
fabianhjr: Amended for an error explaining the attempts.

fabianhjr:
There was a runner issue on macOS CI, rerunning should clear the issue.

fabianhjr:
Rebased for the runner re-run @gallais.

PR #1694, opened 2021-07-13
Enable incremental compilation on Windows.

melted:
Update Windows launch script.

Add back the current directory to source-directories (not needed for
this, but it's the default and it's better to keep load working as usual,
there's no risk as the new dir has priority).

Copy libs instead of linking when testing (linking has bad interactions
with admin privileges).


melted:
This is related to #1676

melted:
Guess it would be better to cons the directory to the old `source-directories`, in case it's different from the default or the default changes.

PR #1700, opened 2021-07-14
Apply "qualified do" notation to bangs and comprehensions

Russoul:
Example in the test file is kind of silly and meaningless though.

Comprehensions might not be the best fit for qualified do, because they target `Applicative`'s and `Alternative`'s.
However, it probably won't hurt to include them too.
As a possible future extension, we could also qualify user-written `pure` and `guard` calls as well.


edwinb:
Good idea, thanks. Please could you update the tests to resolve the conflicts, though? Then I'll merge - this might be a victim of my recent merge frenzy, sorry about that...

Russoul:
@edwinb All done

src/Idris/Desugar.idr
```diff
@@ -392,15 +393,21 @@ mutual
...
+  desugarB side ps (PComprehension fc ret conds) = do
+        let ns = mbNamespace !(get Bang)
+        desugarB side ps (PDoBlock fc ns (map (guard ns) conds ++ [toPure ns ret]))
     where
-      guard : PDo -> PDo
-      guard (DoExp fc tm) = DoExp fc (PApp fc (PRef fc (UN "guard")) tm)
-      guard d = d
+      mbNS : Maybe Namespace -> Name -> Name
+      mbNS Nothing n = n
+      mbNS (Just ns) n = NS ns n
```
gallais, line 402: Can we please lift this to `Core.Name`?

stefan-hoeck:
Just a quick question: Does this also work for idiom brackets (`[| a + b |]`). Would be great to have that as well.

Russoul:
@stefan-hoeck I haven't tested. They are desugared into `pure`, among other things. `pure` is not covered by the PR, that's for sure

edwinb:
Idiom brackets are a different bit of desugaring that don't necessarily come under do blocks, so it'd need a different approach unfortunately. The qualified do notation was really just about solving the occasional problem of having to be explicit about which `>>=` you need if you have some fancy DSL.

Russoul:
Personally, I use qualified do for handling effects (as composition of monads), by specialising (>>=) to an instance of a wanted composition in a dedicated namespace.

Issue #1498, opened 2021-06-03
Add more help topics to the command line


gallais:
We already have `idris2 --help logging` to get a list of the logging options.

We may want to have more of these topics for things that are annoying to look up in the
documentation e.g. the format of package files.

CodingCellist:
A brief overview of how to go about adding these:

1. Find the `src/Idris/CommandLine.idr` file and add a constructor for your help topic to the `HelpTopic` data-type.
2. In the same file, add a case for your topic string to the `recogniseHelpTopic`
3. Update the relevant source file to contain a string/string-generating function, ideally using the pretty-printer (see `src/Core/Options/Log.idr:helpTopics` and `src/Idris/Syntax.idr:pragmaTopics` for examples).
4. Update `src/Idris/Driver.idr:quitOpts` to handle your help topic, i.e. display the new string and quit.

PR #1705, opened 2021-07-14
[ doc ] Add command-line help/topic for the various pragmas.

CodingCellist:
Similar to how you can run `idris2 --help logging` to get information about the different logging topics, this adds the ability to do `idris2 --help pragma` to get information about the various pragmas that can be used. Part of improvements suggested in #1498.


src/Idris/Syntax.idr
```diff
@@ -269,6 +270,56 @@ mutual
...
+  isPragma (CGAction _ _) = False
+  isPragma _              = True
+
+  export
+  pragmaTopics : String
+  pragmaTopics =
+    show $ vsep $ map (((<++>) "+") . pretty {ann=()} . showDirective) $ filter isPragma directiveList
+    where
+      showDirective : Directive -> String
+      showDirective (Hide _)             = "%hide term"
```
Z-snails, line 296: This should be name I think.

CodingCellist: Yep, that also matches what the parser wants it to be. Thanks for spotting it, I'll fix that  : )

edwinb:
Good idea! Maybe `--help` should also list the topics too, and each topic could have a brief introduction to say what it is. I guess there's always more you can add for this sort of thing :).

Issue #1703, opened 2021-07-14
Improve error messages when parsing a return type with a name


andrevidela:
Currently when parsing a function signature, the parser fails with a very unhelpful message. In particular the two functions:

```
func : (n : Bool)

func : String -> (n : Bool)
```

fail with the messages

```
Error: Expected 'case', 'if', 'do', application or operator expression.

Example:3:8--3:9
 1 | module Main
 2 |
 3 | func : (lol : Bool)
            ^
```

and 

```
Error: Couldn't parse declaration.

Example:3:1--3:2
 1 | module Main
 2 |
 3 | func : String -> (lol : Bool)
```

Neither of which point to the correct location nor explain how to solve the problem

## Proposed solution

It would be nice if the error messages looked more like this:

```
Error: Cannot return a named argument

Example:3:1--3:2
 1 | module Main
 2 |
 3 | func : String -> (lol : Bool)
                      ^^^^^^^^^^^^
```

gallais:
Hmm I bother generating good ranges for the error and then this kicks in
and ruins everything (in `Parser.Support`):

```idris
fromParsingError origin (Error msg (Just t))
    = let l = t.startLine
          c = t.startCol in
          ParseFail (MkFC origin (l, c) (l, c + 1)) (msg +> '.')
```

Not sure what the justification for this is.
Why not just preserve the whole range?

gallais:
I give up for today: `binder` (the parser for pi types) is part of `simplerExpr` which,
in turn, is used to parse the arguments of an application. This is why this weird beast
is valid syntax:

```idris
ID : Type -> Type
ID a = a
 
test : ID (a : Type) -> a -> a -- parsed as `ID ((a : Type) -> a -> a)`
test = \ a, x => x
```

This means that when we parse `interface Show (0 a : Type) where` (because we first attempt
to parse constraints i.e. applications followed by `=>`) we first try to parse `Show` as a
function name and `(0 a : Type)` as the beginning of a pi type that is its argument.
We fail before backtracking & realising the list of constraints is empty and we can now parse
a name followed by a list of interface parameters.

So we can't actually hard fail when parsing a pi type upon not finding a `->/=>`: we explicitly
rely on the backtracking behaviour here!

I have tried moving `binder` out of `simplerExpr` so that it needs to be bracketed to be used
as an argument to a function but I'm now getting parse failures on `public export` so I clearly
got something wrong.

The only bit of code worth salvaging is:

```idris
[...]

PR #1711, opened 2021-07-15
[ breaking ] remove parsing of dangling binders

gallais:
It used to be the case that

```
ID : Type -> Type
ID a = a

test : ID (a : Type) -> a -> a
test = \ a, x => x
```

and

```
head : List $ a -> Maybe a
head [] = Nothing
head (x :: _) = Just x
```

were accepted but these are now rejected because:

* `ID (a : Type) -> a -> a` is parsed as `(ID (a : Type)) -> a -> a`
* `List $ a -> Maybe a` is parsed as `List (a -> Maybe a)`

Similarly if you want to use a lambda / rewrite / let expression as
part of the last argument of an application, the use of `$` or parens
is now mandatory.

This should hopefully allow us to make progress on #1703


melted:
There are a lot of weird files added in contib, like cases-25.txt

gallais:
Given that merging master into this branch revealed two recently
introduced dangling lambdas, I should maybe hold off until the big
mergefest is over before I push this breaking change.

ohad:
Is it possible to generate a warning first for a few weeks/1 release?

gallais:
Shoot, I've mixed the new feature (a `Warning` constructor for the `Grammar`) with the merge.
Oh well, it'll all get squashed anyway.

PR #1716, opened 2021-07-15
[ backends ] always assign tag to constructors

Z-snails:
This is a cleaned-up second approach to #1608.
This uses a polynomial rolling hash to get tags for types, which should minimise chance of collisions
The refc backend is broken by this, I'd appreciate some help.


Z-snails:
I can't think of a good way to deal with collisions, so, for now, I will close this.

PR #1718, opened 2021-07-15
Postpone when instantiating if type is unknown

edwinb:
This is an update for #949 to bring it up to date with some changes to the unifier. Thanks to @Russoul!


Russoul:
Great, thank you @edwinb

PR #1344, opened 2021-04-27
[ fix #759 ] Deal with non-ascii chars in Show Char implementation

glmxndr:

PR #1719, opened 2021-07-15
Updated: Deal with non-ascii chars in Show Char implementation

edwinb:
Updating #1344 for latest changes. Let's see what happens...


edwinb:
@madman-bob - do you know what's up with the C implementation here? I haven't looked closely but it's a bit surprising to me.

madman-bob:
This change affects all backends, not just RefC. RefC just happened to be the only one to test this particular corner case.

This changes `show '\159'` to result in `"'�'"`, rather than `"'\\159'"`. Whether this is desirable is another matter entirely.

edwinb:
Oh right, of course. I misread the diff! I'm now not so sure about this. So following the rules I was following when ploughing through PRs, I'm going to close this. If someone can justify why it's desirable (there were no comments or details on the original PR) they can open a new PR with explanation and we'll take it then.

PR #1730, opened 2021-07-16
[ fix ] `Name` in `Language.Reflection.TT`

Z-snails:
This was different to `Name` in `Core.Name`
Specifically `CaseBlock` and `WithBlock` had an `Int` instead of a `String`


edwinb:
Oops! That's probably something I overlooked in yesterday's mergefest. I think this was in an old PR that took so long to merge that things changed...

PR #1739, opened 2021-07-17
[ Test ] allow 'N' as an acceptable response

alexhumphreys:
Situation I just ran into that I found confusing:
```
Golden value differs from actual value.
Accept actual value as new golden value? [y/N]
N
Invalid answer.
```

### DIFF
--- a/libs/test/Test/Golden.idr
+++ b/libs/test/Test/Golden.idr
@@ -246,6 +246,7 @@ runTest opts testPath = forkIO $ do
       case str of
         "y" => pure True
         "n" => pure False
+        "N" => pure False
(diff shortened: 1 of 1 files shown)


edwinb:
Oh yes, I've tripped over that one too. I assume the capital indicates a default. I wonder if "Y" should be allowed too for consistency. Anyway, thanks for this.

alexhumphreys:
You know I had `Y` in this pull request too but wasn't sure so removed it before opening. Given that a lot of "are you sure?" terminal prompts take any of yes/no/y/n in any case, it probably wouldn't hurt to add `Y`...

PR #1747, opened 2021-07-17
Delayed elaborator tweaks

edwinb:
This is for the sake of error messages (changing the ordering) and removing a restriction on how often we'd retry them. That restriction was for the sake of performance but given that the evidence is that it isn't the source of any performance problems, we now retry until we stop making progress. Let's see what effect it has.

I accidentally included an FAQ update here too. Sorry about that. I'm not going to try disentangling it since I'm just going to merge this if CI confirms it's okay...


edwinb:
Well that's an intriguing error given that it's not related to the change and not using the new version yet... I'll investigate later.

edwinb:
Turns out it is something I broke. I'll close this until I work it out.

Issue #1755, opened 2021-07-18
Remove buggy `fromString` insertion with string interpolation


andrevidela:
When using string interpolation the compiler inserts `fromString` calls at the wrong places. The details in #555 indicate that attempting to support `fromString` in interpolated string leads to inconsistent and unpredictable behaviour. A more moral argument is that interpolated strings are meant to concatenate strings specifically.

# Steps to Reproduce

Use

```
"hello \{"world"} !"
```

# Expected Behavior

The tree should return

```
"hello " ++ "world" ++ " !"
```

for the reason highlighted in #555 

# Observed Behavior

The tree is instead

```
fromString "hello" ++ fromString "world" ++ " !"
```

andrevidela:
It turns out we cannot remove the `fromString` from the `"world"` call because we cannot guess what the type of its expression should be. In the example I assumed it was possible to assume each string literal in an interpolated slice evaluates to a string but that is not the case:

```haskell
"Take this path: \{ show ("path" / "to" / "file") } for example"
```
Here the quotes within an interpolated slice _must_ have the `fromString` call otherwise this expression wouldn't typecheck

Issue #1132, opened 2021-02-26
Reporting ill-indented multi-line strings


andrevidela:
#1097 introduces multiline strings but the error reporting does not point to the correct areas

```
hello : String
hello = """
      hello
        """
```

should point to the excess characters on the left, the error should be reported as:

```
Error: While processing multi-line string. Line is less indented than the closing delimiter
      hello
      ^^"""
```

PR #1756, opened 2021-07-18
Document string literals

andrevidela:
#555, #1027, #1029 have been implemented, but the documentation and the changelog weren't updated at the same time. Documenting those feature is the last item in their todo list.

The features seem to work well except for #1755 and #1132 which are fairly innocuous.

### DIFF
--- a/docs/source/reference/strings.rst
+++ b/docs/source/reference/strings.rst
@@ -1,49 +1,64 @@
 String literals in Idris
-===============
+========================
 
-Strings are ubiquitous in programming, they allow the programmer to ingest data, display it and communicate information in a format that's legible to humans.
-
-To facilitate the use of string in daily programming, idris provides feature to make them easier to read and easier to understand. Those features are multi-line strings, raw strings and interpolated strings. Let us start with plain string literals.
+To facilitate the use of string literals, idris provides three features
+in addition to plain string literals: multiline strings, raw strings and interpolated
+strings.
 
 Plain string literals
 ---------------------
 
-String literals behave the way you expect from other programming language. Use quotation marks `"` around the piece of text that you want to use as a string:
+String literals behave the way you expect from other programming language. Use quotation marks
+``"`` around the piece of text that you want to use as a string:
 
 ``"hello world"``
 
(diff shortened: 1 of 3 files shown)


docs/source/reference/strings.rst
```diff
@@ -0,0 +1,137 @@
...
+    myRegex = parseRegex #"\\{"#
+
+If you need to escape characters you still can by using a ``\\`` followed by the same number of
+``#`` that you used for your string delimiters. In the following example we are using two
+``#`` characters as our escape sequence and want to print a line return:
+
+.. code-block::
+
+    markdownExample : String
+    markdownExample = ##"markdown titles look like this: \##n"# Title \##n body""##
```
gallais, line 91: I'm confused: why do you have `\##n` instead of `\n`?
According to the next example using a multiline string you *do* want the newline?

docs/source/reference/strings.rst
```diff
@@ -0,0 +1,137 @@
...
+    markdownExample = ##"""
+        markdown titles look like this:
+        "# Title
+        body"
+        """##
+
+Interpolated strings
+--------------------
+
+Concatenating string literals with runtime values happens all the time, but sprinking our code
```
gallais, line 107:
```suggestion
Concatenating string literals with runtime values happens all the time, but sprinkling our code
```

docs/source/reference/strings.rst
andrevidela: Because the escape sequence is `\##` in that string, if you were to write `\n` the _text_ `\n` would appear instead of a new line.

gallais: I see. I always understood `\n` as a single character, not as `n` being escaped to mean newline. :thinking:

andrevidela: the character `n` preceeded with the escape sequence renders a single character, just like `\t`, `\r`, `\0`. Raw strings allow the user to type those character without having to escape the `\` character. But you can recover their meaning by using the `\#` escape sequence inside a raw string (or however many `#` you're using for the raw string)

Issue #1630, opened 2021-06-28
Missing type specifier kills compiler


nickdrozd:
Here is a proof from `libs/contrib/Control/Algebra/Laws.idr`:

```idris
||| x0 = 0
public export
multNeutralAbsorbingR : Ring ty => (l : ty) ->
  l <.> neutral {ty} = neutral {ty}
multNeutralAbsorbingR {ty} l =
  rewrite sym $ monoidNeutralIsNeutralL $ l <.> neutral in
    rewrite sym $ groupInverseIsInverseL $ l <.> neutral in
  rewrite semigroupOpIsAssociative (l <.> ((l <.> neutral) <+> (inverse $ l <.> neutral))) (l <.> neutral) (inverse $ l <.> neutral) in
    rewrite groupInverseIsInverseL $ l <.> neutral in
  rewrite sym $ ringOpIsDistributiveL l neutral neutral in
    rewrite monoidNeutralIsNeutralL $ the ty neutral in
  groupInverseIsInverseL $ l <.> neutral
```

It compiles fine. But consider this apparently minor change:

```diff
-    rewrite monoidNeutralIsNeutralL $ the ty neutral in
+    rewrite monoidNeutralIsNeutralL neutral in
```

This doesn't merely fail to typecheck; it actually kills the compiler:

```
120/122: Building Control.Algebra.Laws (Control/Algebra/Laws.idr)
Killed
make[1]: *** [Makefile:2: all] Error 137
```

Possibly related to https://github.com/idris-lang/Idris2/issues/72, or maybe it's something else entirely. I can't reproduce it now, but at some point the Idris process message was `RenaKilled`, which maybe relates to renaming?

ohad:
Not an explanation for the crash (and associated bug), but I don't expect this rewrite to work without the type annotation.
The reason is that I expect `monoidNeutralIsNeutralL` has the type `Ring ty0 =>(x : ty0) -> ...`.
If you're calling it with `neutral`, which I expect to be similarly constrained, the type-checker shouldn't be able to work out what `ty0` is. You'll need to disambiguate it with a type annotation.

Alternatively, you could use equational reasoning, which will allow you to elide a lot of the information you must spell out for rewriting, and perhaps make the proof more readable because you'll explicate the hidden proof-state.

nickdrozd:
> Not an explanation for the crash (and associated bug), but I don't expect this rewrite to work without the type annotation.
[...]

PR #1761, opened 2021-07-19
Define and implement Injective interface

nickdrozd:
The idea here is similar to https://github.com/idris-lang/Idris2/pull/1472. Instead of claiming that functions are injective on an ad-hoc basis, make one Injective interface and then implement it for each function.

Unlike with the relation interfaces, there is no need to specify any implicit arguments (with one notable exception).

There are some existing ad-hoc injectivity proofs that I couldn't convert to the new interface, like `finToNatInjective`. I couldn't understand the error messages.

### DIFF
--- a/libs/base/base.ipkg
+++ b/libs/base/base.ipkg
@@ -29,6 +29,7 @@ modules = Control.App,
           Control.Monad.Writer.CPS,
           Control.Monad.Writer.Interface,
           Control.WellFounded,
+          Control.Function,
           Control.Relation,
           Control.Order,
 
--- a/libs/contrib/Data/Nat/Factor.idr
+++ b/libs/contrib/Data/Nat/Factor.idr
@@ -9,7 +9,6 @@ import Syntax.PreorderReasoning
 
 %default total
 
-
 ||| Factor n p is a witness that p is indeed a factor of n,
 ||| i.e. there exists a q such that p * q = n.
 public export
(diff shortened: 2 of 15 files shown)


libs/contrib/Data/Nat/Factor.idr
```diff
@@ -124,8 +123,7 @@ multOneSoleNeutral (S k) (S (S j)) prf =
         rewrite plusCommutative k j in
         rewrite sym $ plusAssociative j k (k * S j) in
         rewrite sym $ multRightSuccPlus k (S j) in
-        succInjective k (j + (S (S (j + (k * (S (S j))))))) $
-        succInjective (S k) (S (j + (S (S (j + (k * (S (S j)))))))) prf
+        injective {f = S} $ injective prf
```
nickdrozd, line 126: The implicit argument `{f = S}` is is required here. If it is removed, the compiler is killed, as in https://github.com/idris-lang/Idris2/issues/1630.

libs/base/base.ipkg
```diff
@@ -29,6 +29,7 @@ modules = Control.App,
           Control.Monad.Writer.CPS,
           Control.Monad.Writer.Interface,
           Control.WellFounded,
+	  Control.Function,
```
andorp, line 32: Is the indentation right here?

nickdrozd: fixed

Issue #1012, opened 2021-02-01
Defining fromInteger causes unification error


michaelmesser:
# Steps to Reproduce

1:
```idris
f : (x : Nat) -> {auto 0 p : Either (x=2) (x=3)} -> ()
f _ = ()

g : ()
g = f 3
```

2:
```idris
fromInteger : Integer -> String

f : (x : Nat) -> {auto 0 p : Either (x=2) (x=3)} -> ()
f _ = ()

g : ()
g = f 3
```

3:
```
fromInteger : Integer -> String

fw : (x : Nat) -> {auto 0 _ : Either (x=2) (x=3)} -> ()
fw _ = ()

gw : ()
gw = fw (the Nat 3)
```

# Expected Behavior

All three programs work correctly.

# Observed Behavior
The 1 and 3 works but 2 fails with:
```idris
Error: While processing right hand side of g. When unifying 2 and 3.
Mismatch between: 0 and 1.
    |
    | g = f 3
    |     ^^^
```

gallais:
Here is my understanding of the issue:

* Without the annotation we consider that the ambiguous `fromInteger 3` could benefit from being delayed
* Auto picks an arbitrary solution `Left Refl`
* We come back to the delayed problem
* Elaborate 3 to `S S S Z`
* But `S S S Z` is not equal to `S S Z`!
* :(

The fix should be a one-liner.

gallais:
> I hope this won't break other cases though.

It does break other cases. :(

PR #1764, opened 2021-07-20
A bit of refactoring of argument elaboration order

### LINKED ISSUE #1743: Irrefutable pattern in `let` binding in a type declaration stops patterns from accessing implicit arguments in its scope
# Steps to Reproduce
```idris
works : (n : (Nat, Nat)) -> let mk = n in {y : Bool} -> (z : Bool) -> String
works (m, k) {y} z = ?h1

fails : (n : (Nat, Nat)) -> let (m, k) = n in {y : Bool} -> (z : Bool) -> String
fails (m, k) {y} z = ?h2
```
# Expected Behavior
Both should type-check fine.

# Observed Behavior
Only the first definition type-checks, the second one fails with:
```
 line 5 col 0:
     While processing left hand side of fails. y is not a valid argument in fails ?n ?z.
     
     Foo:5:1--5:19
      1 | works : (n : (Nat, Nat)) -> {y : Bool} -> (z : Bool) -> String
      2 | works (m, k) {y} z = ?h1
      3 | 
      4 | fails : (n : (Nat, Nat)) -> let (m, k) = n in {y : Bool} -> (z : Bool) -> String
      5 | fails (m, k) {y} z = ?h2
          ^^^^^^^^^^^^^^^^^^
```

edwinb:
In theory argument elaboration order doesn't matter, but in practice we
sometimes make choices for performance reasons, like helping with
disambiguation by knowing the target type.

This was kind of messy, now we can more clearly see what's going on.
Also, more importantly, it gives us a bit more control which we
sometimes need. For example, if we go choose to go right to left for
some performance heuristic it might turn out we don't have enough
information yet, in which case we need to delay and try again later.

Fixes #1743

### DIFF
--- a/src/TTImp/Elab/Ambiguity.idr
+++ b/src/TTImp/Elab/Ambiguity.idr
@@ -353,7 +353,7 @@ checkAlternative rig elabinfo nest env fc (UniqueDefault def) alts mexpected
          let solvemode = case elabMode elabinfo of
                               InLHS c => inLHS
                               _ => inTerm
-         delayOnFailure fc rig env expected ambiguous Ambiguity $
+         delayOnFailure fc rig env (Just expected) ambiguous Ambiguity $
              \delayed =>
                do solveConstraints solvemode Normal
                   defs <- get Ctxt
@@ -406,7 +406,7 @@ checkAlternative rig elabinfo nest env fc uniq alts mexpected
                 let solvemode = case elabMode elabinfo of
                                       InLHS c => inLHS
                                       _ => inTerm
-                delayOnFailure fc rig env expected ambiguous Ambiguity $
+                delayOnFailure fc rig env (Just expected) ambiguous Ambiguity $
                      \delayed =>
                        do defs <- get Ctxt
                           exp <- getTerm expected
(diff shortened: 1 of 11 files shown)


gallais:
If you have the executable lying around, what's the impact of this PR on #1012?

edwinb:
> If you have the executable lying around, what's the impact of this PR on #1012?

It didn't make any difference, but I'll see if the changes I've made make it any easier to find out what's going on with it.

Not sure what's up with CI here. I'll investigate.

edwinb:
Ah, right, it wouldn't help #1012 because it's only a particular case where it's willing to delay at the moment. Maybe there's a way to get a bit more control though. I'll take a look while I'm still thinking about it.

gallais:
> Not sure what's up with CI here. I'll investigate.

It's not specific to your PR: ubuntu-self-host-racket has been hanging since
[this unrelated commit](https://github.com/idris-lang/Idris2/commit/ada33b90a1b026cfef8e924456a507f030a3446a) where github killed the job after 6h
stuck on VMCode.

edwinb:
Well that is mystifying. I'm going to hope it's a temporary glitch, then, but I suppose I could check if it's reproducible locally just in case.

PR #1765, opened 2021-07-21
[Not for merging] CI timing experiment

edwinb:
Added --timing flags just to see if it sheds any light on what's getting stuck in CI.


edwinb:
Well that was sadly not very enlightening. Mostly we've learned that it proceeds happily then just stops at some point, but not the same point every time.

PR #1766, opened 2021-07-21
[ ci ] controlling builds via commit messages

gallais:
Checking commit messages for instructions:

* `[ci: skip]` to skip all idris2 builds
* `[ci: ubuntu]` to build the ubuntu-based ones (same for `macos`, `windows`, `nix`)
* `[ci: chez]` to only build the chez-based ones (same for `racket`)

If `[ci: skip]` is not present then the resulting action is the *union* of all the tags so
a message containing both `[ci: windows]` and `[ci: racket]` will run all of the windows
tests as well as the racket ones on ubuntu & macos.

### DIFF
--- a/.github/workflows/ci-idris2.yml
+++ b/.github/workflows/ci-idris2.yml
@@ -46,11 +46,11 @@ jobs:
       - name: Get commit message
         id: get_commit_message
         run: |
-          echo ${{ github.event_name }}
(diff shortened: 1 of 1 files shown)


gallais:
It's infuriating: the check works on [the branch](https://github.com/gallais/Idris2/commits/actions) but not on the PR.

PR #1768, opened 2021-07-21
Update bootstrap code

edwinb:
We shouldn't strictly need this, but it doesn't do any harm, and it also
means we can use --inc chez in the build if we do. Let's see if it helps
or hinders the CI problem...


edwinb:
Sigh. Looks like exactly the same problem. Any ideas anyone?

PR #1770, opened 2021-07-21
Fix #1767

### LINKED ISSUE #1767: if-expression requires parentheses within string interpolation
When using an `if`-expression within string interpolation an error gets reported. But the error mentions that the token 'if' belongs to the expected tokens.

# Steps to Reproduce
Evaluate the following string:
`"\{ if True then "a" else "b" }"`

# Expected Behavior
I would expect that it evaluates to `"a"` 
or a parse error that says that `if` is not allowed in this position.

# Observed Behavior
```
Main> "\{ if True then "a" else "b" }"
Couldn't parse any alternatives:
1: Expected 'case', 'if', 'do', application or operator expression.
(Interactive):1:5--1:7
 1 | "\{ if True then "a" else "b" }"
         ^^
```

It works as expected when the `if`-expression is surrounded with parentheses.
```
Main> "\{ (if True then "a" else "b") }"
"a"
```

# Idris2 Version
Version 0.4.0-5f3480120

andrevidela:
The `if then else` syntax expects a block for the `then` and `else`
parts. Before this patch, the token `InterpEnd` was not a valid
follow up token to end a block. This adds `InterpEnd` as a closing
token for blocks, allowing `if then else` in interpolation slices
without additional parens.


andylokandy:
Would you like to also include the fix for token printing?

andrevidela:
Ah that's a good idea, let's do that quickly, which commit should I cherry-pick?

andylokandy:
> Ah that's a good idea, let's do that quickly, which commit should I cherry-pick?

hmm, i didn't start to write the code. But it should be just swap the "multiple string begin" and "string begin" I refered in the issue.

andrevidela:
I just fixed it now, commit inc

PR #1774, opened 2021-07-22
Document --init

eayus:
Reference the --init option for creating package files in the tutorial documentation.


gallais:
Thank you!

Issue #1771, opened 2021-07-22
Programing being checked as %default total when it probably should not be total.


donovancrichton:
A program checked with %default total is type checking when it probably shouldn't.

# Steps to Reproduce

The [gist](https://gist.github.com/donovancrichton/9c2db6156949979985c7d010454b9e22).

write this code:
```idris
%default total

data Fix : (Type -> Type) -> Type where
  MkFix : f (Fix f) -> Fix f

F : Type -> Type
F x = x -> x

fix : Fix F -> Fix F
fix (MkFix f) = f (MkFix f)

use : Fix F
use = fix (MkFix {f=F} fix)
```
# Expected Behavior
An error or warning because `use` will never terminate.
# Observed Behavior
No errors or warnings and then non-termination when I go to evaluate `use`.

andrevidela:
Is this due to `Type : Type`?

gallais:
`Fix` is not strictly positive (because we can't guarantee that `f` is) so it should be rejected.

gallais:
Boom:

```idris
%default total

data Fix : (Type -> Type) -> Type where
  MkFix : f (Fix f) -> Fix f

F : Type
F = Fix (\ x => x -> Void)

notF : Not F
yesF : Not F -> F

notF (MkFix f) = f (yesF f)
yesF nf = MkFix nf

argh : Void
argh = notF (yesF notF)
```

edwinb:
I'm a bit surprised this gets past the positivity checker. I suppose we'd better fix that...

At some point we should also get on with fixing `Type : Type`. Mostly that involves writing a mini constraint solver for levels, which I'd really like to be much faster than the one we had in Idris 1, so I'll be looking for help from our local constraint experts for that!

gallais:
Here's my understanding:

When we check the type of `MkFix`, the first thing we encounter is a Pi-type binding
`{0 f : Type -> Type}`. This argument is fine so we inspect the body of the pi type.
To do so we feed the closure `Erased` as the value for `f`. Which means that the body
computes to `[__] -> Fix [__]` and we've now lost the reference to `Fix`.

gallais:
A variant that's not relying on `Erased`: we drop the arguments to a type constructor
that are in parameter positions by considering these are safe positive positions.
That's what allows us to define e.g. `data Rose a = Node (List (Rose a))`.

The issue is that we do not actually check that we are using the name in a positive
manner in this safe positions! So we can use the parameter bypass to hide a negative
[...]

PR #1776, opened 2021-07-23
Fixing 3 out of the 4 proofs of Void in #1771

gallais:
The last one will require a bit more work. So let's get these in first.


libs/contrib/Data/InductionRecursion/DybjerSetzer.idr
```diff
@@ -4,8 +4,6 @@
 
 module Data.InductionRecursion.DybjerSetzer
 
-%default total
-
```
andrevidela, line 8: Is that deletion intentional?

gallais: Yes unfortunately that module does not check as total anymore.

PR #1777, opened 2021-07-23
Quote chez exe path

melted:
Update the windows documentations with gotchas. Which will still be
there until the fix is in the bootstrap.

Remove unused function


melted:
Found that the findExists there wasn't used after pathLookup got moved. I didn't add the quoting to pathLookup to keep down the scope of the fix.

This is small part of #1525.

melted:
Ok, that was too naive, don't have time to rework it right now, so I will make a new PR with the doc fix only.

PR #1781, opened 2021-07-24
[ fix ] bash TAB completion for files

stefan-hoeck:
As mentioned on [discord](https://discord.com/channels/827106007712661524/834611018775789568/868145988508975195), TAB completion for Idris2 does not work for files but only for directories. This PR should fix the issue.


gallais:
Thanks!

PR #1785, opened 2021-07-25
Add ability to manipulate scheme objects

edwinb:
This is step 0 in a plan to use the scheme evaluator to evaluate Idris
expressions at compile time. As a proof of concept, I've got this
working for a toy language here: https://github.com/edwinb/SchemeEval

We won't be able to do anything interesting with this in Idris itself
until the next release because it involves updating the bootstrap code
and adding the ability to pass 'Integer' to foreign calls, which really
should have been allowed anyway since it's for a backend to decide what
it can cope with, not Idris itself.


buzden:
Will it be possible to use the scheme evaluator during the elaboration script calculation? (Or, is my question silly?)

PR #1789, opened 2021-07-26
[ fix #1782 ] remove the case-specific code

### LINKED ISSUE #1782: A partial function blindly assumed to be total
Tested on `51f59b95308b015e9b596a8e071cbb5b77b64e7c` (today's `master` branch).

# Steps to Reproduce
The following code compiles just fine. Note that althought the function contains a recursive call in one of its branches, no attempt whatsoever is made to prove its totality. Yet, it is marked as such and the totality checker seems to blindly accept that.

```
import Data.Fin
import Data.Nat
import Data.Nat.Factor

||| Given a natural number n, this function returns a pair of numbers
||| (e, m) such that m is odd and (power 2 e) * m = n. It is almost
||| total, except for 0, which cannot be expressed as such a product.
||| Indeed, given 0 the function loops indefinitely, but the compiler
||| does not seem to mind
export
total
factorOutTwos : (n : Nat) -> (Nat, Nat)
factorOutTwos n with (decFactor n 2)
  factorOutTwos n | ItIsNotFactor (ProperRemExists q FZ prf) =
    (0, n)
  factorOutTwos n | ItIsFactor (CofactorExists q prf) =
    let (e, m) = factorOutTwos q in
    (S e, m)
```

# Expected Behavior
This function should be rejected by the totality checker, but it is not.

# Observed Behavior
The above module  compiles just fine and can be run in the REPL. The function seems to return correct results for most inputs, but given 0 it loops forever, because 2 is indeed considered a factor of 0, it's cofactor (`q`) being also 0, which makes the function step into the recursive call with the same argument over and over again.

gallais:
I can't make sense of this code, it seems to try to convert the
case function corresponding to `let (a, b) = f n in ...` into a
different case function where `f n` and `(a, b)` have been unified.
But if `f n` is a bona fide stuck computation, there's no chance of
this happening.

Just turning this off solves the #1782 and only breaks one function
in the whole of the idris2 repo (I would have expected our current
termination oracle to be too weak to detect it as valid anyway!)
and one in frex (which, again, should not have been seen as terminating).

### DIFF
--- a/libs/base/Data/List1.idr
+++ b/libs/base/Data/List1.idr
@@ -94,10 +94,12 @@ snoc xs x = xs ++ (singleton x)
 
 public export
 unsnoc : (xs : List1 a) -> (List a, a)
-unsnoc (h ::: Nil)       = (Nil, h)
-unsnoc (h ::: (x :: xs)) =
-  let (ini,lst) = unsnoc (x ::: xs)
-   in (h :: ini, lst)
+unsnoc (x ::: xs) = go x xs where
+
+  go : (x : a) -> (xs : List a) -> (List a, a)
+  go x [] = ([], x)
+  go x (y :: ys) = let (ini,lst) = go y ys
+                   in (x :: ini, lst)
 
 ------------------------------------------------------------------------
 -- Reverse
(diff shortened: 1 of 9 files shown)


gallais:
I guess I was a bit too eager to remove `Libraries.Data.List1`. :grimacing:

PR #1794, opened 2021-07-27
Fix bad range behaviour for Nat

Sventimir:
Lists produced by `Range` instance of `Nat` differ from those produced for `Integer`. This creates a confusing inconsistency:
```
Main> rangeFromThenTo (the Nat 2) 3 2
[]
Main> rangeFromThenTo (the Integer 2) 3 2
[2]
```
The issue is fixed in favour of behaviour for `Integer`, which seems more natural. Additionally some tests for Range instance for `Nat` were added.

### DIFF
--- a/tests/prelude/nat001/nats.idr
+++ b/tests/prelude/nat001/nats.idr
@@ -1,0 +1,36 @@
+nats : List Nat -> List Nat
+nats = the (List Nat)
+
+singletonRange : nats [1..1] = nats [1]
+singletonRange = Refl
+
+basicIncreasingRange : nats [1..3] = nats [1, 2 , 3]
+basicIncreasingRange = Refl
+
+basicDecreasingRange : nats [3..1] = nats [3, 2, 1]
+basicDecreasingRange = Refl
+
+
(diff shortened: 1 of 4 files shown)


tests/prelude/nat001/nats.idr
```diff
@@ -0,0 +1,36 @@
...
+singletonRangeWithStep : nats [3, 4..3] = nats [3]
+singletonRangeWithStep = Refl
+
+zeroStepEmptyList : nats [3, 3..5] = nats []
+zeroStepEmptyList = Refl
+
+zeroStepWhenBoundEqual : nats [1, 1..1] = nats [1]
+zeroStepWhenBoundEqual = Refl
+
+decreasingRanngeWithStep : nats [11, 8..1] = nats [11, 8, 5, 2]
```
Z-snails, line 29: Awesome work, just a small typo you might as well fix before it gets merged (rannge -> range).

Sventimir: Fixed.

ohad:
I agree with @Z-snails : very nicely done!

PR #1796, opened 2021-07-28
Add `break` in each case alt in js backend

### LINKED ISSUE #1795: Error with generated javascript

# Steps to Reproduce
```idris
-- Main.idr

module Main

import Data.String.Parser

main : IO ()
main = do
    let x = parse (satisfy isDigit) "100"
    printLn x
```
```
$ idris2 -o test Main.idr --cg node
$ node ./build/exec/test
```
# Expected Behavior
```
Right ('1', 1)
```
# Observed Behavior
```
Left "Parse failed at position 0: satisfy"
```

Here is the bad JS code (prettified):
```js
function Data_String_Parser_satisfy($0, $1, $2) {
    let $7;
    switch (Prelude_EqOrd_x3c_Ord_Int($2.a2, $2.a3)) {
        case 1: {
            const $e = Data_String_strIndex($2.a1, $2.a2);
            switch ($1($e)) {
                case 1: {
                    $7 = {
                        h: 1,
                        a1: $e,
                        a2: { a1: $2.a1, a2: _add32s($2.a2, 1), a3: $2.a3 },
                    };
                    break;
                }
                case 0: {
                    $7 = { h: 0, a1: $2.a2, a2: "satisfy" };
                    break;
                }
            }
        }
        case 0: {
            $7 = { h: 0, a1: $2.a2, a2: "satisfy" };
            break;
        }
    }
    return $0.a2(undefined)($7);
}
```
There is a missing `break` statement in the first case, should be:
```js
function Data_String_Parser_satisfy($0, $1, $2) {
    let $7;
    switch (Prelude_EqOrd_x3c_Ord_Int($2.a2, $2.a3)) {
        case 1: {
            const $e = Data_String_strIndex($2.a1, $2.a2);
            switch ($1($e)) {
                case 1: {
                    $7 = {
                        h: 1,
                        a1: $e,
                        a2: { a1: $2.a1, a2: _add32s($2.a2, 1), a3: $2.a3 },
                    };
                    break;
                }
                case 0: {
                    $7 = { h: 0, a1: $2.a2, a2: "satisfy" };
                    break;
                }
            }
            break; // HERE
        }
        case 0: {
            $7 = { h: 0, a1: $2.a2, a2: "satisfy" };
            break;
        }
    }
    return $0.a2(undefined)($7);
}
```

pinging @stefan-hoeck

Z-snails:
Fixes #1795
Currently this isn't very clever, and inserts an uneeded `break` after every `return`.

### DIFF
--- a/src/Compiler/ES/Codegen.idr
+++ b/src/Compiler/ES/Codegen.idr
@@ -616,7 +616,7 @@ mutual
     as <- traverse alt alts
     d  <- traverseOpt stmt def
     pure $  switch (minimal sc <+> ".h") as d
-    where 
+    where
         alt' : {r : _} -> EConAlt r -> Core (Doc,Doc)
         alt' (MkEConAlt _ RECORD b)  = ("undefined",) <$> stmt b
         alt' (MkEConAlt _ NIL b)     = ("0",) <$> stmt b
--- a/src/Compiler/ES/Codegen.idr
+++ b/src/Compiler/ES/Codegen.idr
@@ -570,7 +570,7 @@ switch sc alts def =
 
   where anyCase : Doc -> Doc -> Doc
         anyCase s d =
-          let b = if isMultiline d then block d else d
+          let b = block $ vcat [d, "break;"]
            in s <+> softColon <+> b
 
         alt : (Doc,Doc) -> Doc
@@ -610,7 +610,7 @@ mutual
   stmt (Declare v s) =
     (\d => vcat ["let" <++> var v <+> ";",d]) <$> stmt s
   stmt (Assign v x) =
-    (\d => vcat [hcat [var v,softEq,d,";"], "break;"]) <$> exp x
+    (\d => vcat [hcat [var v,softEq,d,";"]]) <$> exp x
 
   stmt (ConSwitch r sc alts def) = do
     as <- traverse alt alts
--- a/src/Compiler/ES/Codegen.idr
+++ b/src/Compiler/ES/Codegen.idr
@@ -614,7 +614,7 @@ mutual
   stmt (Declare v s) =
     (\d => vcat ["let" <++> var v <+> ";",d]) <$> stmt s
   stmt (Assign v x) =
-    (\d => vcat [hcat [var v,softEq,d,";"]]) <$> exp x
+    (\d => hcat [var v,softEq,d,";"]) <$> exp x
 
   stmt (ConSwitch r sc alts def) = do
     as <- traverse (map (insertBreak r) . alt) alts
--- a/src/Compiler/ES/Codegen.idr
+++ b/src/Compiler/ES/Codegen.idr
@@ -583,6 +583,10 @@ lambdaArgs : List Var -> Doc
 lambdaArgs [] = "()" <+> lambdaArrow
 lambdaArgs xs = hcat $ (<+> lambdaArrow) . var <$> xs
 
+insertBreak : (r : Effect) -> (Doc, Doc) -> (Doc, Doc)
+insertBreak Returns x = x
+insertBreak (ErrorWithout _) (pat, exp) = (pat, vcat [exp, "break;"])
+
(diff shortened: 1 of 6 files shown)


stefan-hoeck:
As we distinguish between value assignments and return statements in the types (see function `stmt` in the Codegen), it should be possible to not add unnecessary breaks after returns.

stefan-hoeck:
I came up with the following (sorry, don't know how to push to a PR) (note that the `break` in the `Assign` case
of `stmt` was removed:

```idris

breakAfterAssignment : Effect -> (Doc,Doc) -> (Doc,Doc)
breakAfterAssignment Returns p = p
breakAfterAssignment (ErrorWithout _) (a,d) = (a,vcat [d, "break;"])

  stmt : {e : _} -> {auto c : Ref ESs ESSt} -> Stmt e -> Core Doc
  stmt (Return y) = (\e => "return" <++> e <+> ";") <$> exp y
  stmt (Const v x) = constant (var v) <$> exp x
  stmt (Declare v s) =
    (\d => vcat ["let" <++> var v <+> ";",d]) <$> stmt s
  stmt (Assign v x) =
    (\d => hcat [var v,softEq,d,";"]) <$> exp x

  stmt (ConSwitch r sc alts def) = do
    as <- traverse alt alts
    d  <- traverseOpt stmt def
    pure $  switch (minimal sc <+> ".h") (breakAfterAssignment r <$> as) d
    where alt : EConAlt r -> Core (Doc,Doc)
          alt (MkEConAlt _ RECORD b)  = ("undefined",) <$> stmt b
          alt (MkEConAlt _ NIL b)     = ("0",) <$> stmt b
          alt (MkEConAlt _ CONS b)    = ("undefined",) <$> stmt b
          alt (MkEConAlt _ NOTHING b) = ("0",) <$> stmt b
          alt (MkEConAlt _ JUST b)    = ("undefined",) <$> stmt b
          alt (MkEConAlt t _ b)       = (tag2es t,) <$> stmt b

  stmt (ConstSwitch r sc alts def) = do
    as <- traverse alt alts
    d  <- traverseOpt stmt def
    ex <- exp sc
    pure $ switch ex (breakAfterAssignment r <$> as) d
    where alt : EConstAlt r -> Core (Doc,Doc)
          alt (MkEConstAlt c b) = do d    <- stmt b
                                     pure (Text $ jsConstant c, d)

  stmt (Error x)   = pure $ jsCrashExp (jsStringDoc x) <+> ";"
  stmt (Block ss s) = do
    docs <- traverse stmt $ forget ss
    doc  <- stmt s
    pure $ vcat (docs ++ [doc])

```

Z-snails:
That's basically what I did, but yours is neater, so I'll copy that.

stefan-hoeck:
Sorry, looks like we were coding at the same time. :-)

stefan-hoeck:
Would be great if we had a test for this to make sure it works.

src/Compiler/ES/Codegen.idr
```diff
@@ -610,28 +614,31 @@ mutual
   stmt (Declare v s) =
     (\d => vcat ["let" <++> var v <+> ";",d]) <$> stmt s
   stmt (Assign v x) =
-    (\d => vcat [hcat [var v,softEq,d,";"], "break;"]) <$> exp x
+    (\d => vcat [hcat [var v,softEq,d,";"]]) <$> exp x
```
stefan-hoeck, line 617: No need for the outer `vcat` here.

Z-snails: Good spot!

Z-snails:
If you're happy with this @stefan-hoeck , I'll merge.

stefan-hoeck:
Yeah, please go ahead. And sorry for the delay. Thanks a lot for the fix!

Z-snails:
Thanks @stefan-hoeck for your help!

PR #1797, opened 2021-07-28
Add copyFile to System.File

madman-bob:
This function is currently defined in `Idris.Package`, but is more generally useful.

If my understanding is correct, I can't remove the version in `Idris.Package` yet, due to how the dependancies work.

Adding this function requires an updated version of #1666, to avoid a circular dependancy between `System.File` and `System.Buffer`. @edwinb I know you weren't too keen on that PR, but hopefully you're happy with a more practical reason for that flip.

I also took the liberty of splitting `System.File` into multiple files, as it was beginning to get unwieldly large in a single file. This also gave me some ideas about name resolution, but that's a discussion for another time.

I wasn't sure where to put the test for `copyFile`. Semantically, `tests/base` seems the most appropriate place, but those tests are intended for tests that are run on all backends. As `copyFile` is not an immediate wrapper for FFI functions, it is redundant to test it on multiple backends. Personally, I would consider those tests more appropriate in `tests/allbackends`. Nonetheless, that still seems the most appropriate place for it.


gallais:
Checking this builds with `main` now that all of the tests have been reactivated.

PR #1808, opened 2021-07-31
Update contrib Text.Parser to match Library.Text.Parser

alexhumphreys:
This copies the latest changes from Libraries.Text to Contrib.

The main differences were from the introduction of WithBounds in #527,
and the changes for semantic highlighting (#1335).

The changes to `libs/contrib/Text/*` are mostly a straight `cp -r src/Library/Text/*`,  I just removed the `Libraries.` from the module names.

The other changes are to:
- libs/contrib/Data/String/Extra.idr
- libs/contrib/Language/JSON.idr
- libs/contrib/System/Path.idr

I think I preserved the behaviour, but I'm not too familiar with these modules (especially the `Data.String.Extra`). There's also few/no tests for these. I ran a few commands in the repl and they seem to still work, but someone who's more familiar with them should probably make sure.

The motivation for this PR was I wanted to use those `WithBounds` changes to get position information while parsing.


libs/contrib/Data/String/Extra.idr
```diff
@@ -5,8 +5,12 @@ import Data.List1
 import Data.Nat
 import Data.String
 
-%default total
+%hide Data.String.lines
```
mattpolzin, line 8: What necessitated hiding these and redefining them? The comment below that says "should be removed once v.0.4.0 is released" makes the changes to this file particularly suspect :)

libs/contrib/Language/JSON.idr
```diff
@@ -6,6 +6,8 @@ import Language.JSON.Parser
 
 import public Language.JSON.Data
 
+import public Text.Bounded
```
mattpolzin, line 9: This module does not live in prelude, base, or contrib, so it will break the use of contrib.'s Language.JSON -- curiously, this does not cause contrib to fail to build, but that must be a side effect of it being built at the same time as the compiler and somehow the compiler's access to a `Text.Bounded` module leaking into the contrib compilation.

mattpolzin: Correction: You did copy this module into contrib, but the public import will fail until it is added to the exposed modules in the contrib.ipkg file (I believe this is what is going on with the failing Frex compatibility test).

alexhumphreys: Good catch! I added the modules to contrib.ipkg as they appear in idris2api.ipkg. With the exception of `Text.Distance.Levenshtein`, as this would also require pulling `Data.IOMatrix` into contrib. Can also do that of course, but maybe at a later stage if someone needs it.

libs/contrib/Data/String/Extra.idr
alexhumphreys: Aye, I saw that too. I'd like to keep this as similar to the files in `src/Libraries` as possible, but this does seem like an awkward moment to copy that file. I'll try fix the duplication in `src/Libraries/Data/String/Extra.idr` to reduce the diff between it and its contrib counterpart. Though `libs/contrib/Data/String/Extra.idr` also has the functions `fastReplicate`, `justifyLeft`, `justifyRight`, which are absent from the `Libraries` version...

libs/contrib/Data/String/Extra.idr
```diff
@@ -89,10 +93,64 @@ index n str with (unpack str)
   index Z str | (x :: xs) = Just x
   index (S n) str | (x :: xs) = index n str | xs
 
+||| Produce a string by repeating the character `c` `n` times.
+public export
+replicate : (n : Nat) -> (c : Char) -> String
+replicate n c = pack $ replicate n c
+
+-- lines and unlines are cloned from Data.String for compatible reason, should be removed
```
Z-snails, line 101: I don't think that applies here, so you can use `Data.String.lines` rather than cloning it.

Z-snails: I should have read matt polzin's review, sorry

libs/contrib/Data/String/Extra.idr
stefan-hoeck: This happens, when stuff is added to the standard libs that is also needed in the compiler. In these cases, the function in question is also added to an `Extra` module in `Libraries`, because it won't otherwise be available during bootstrapping. After a minor release, these temporary additions can be removed from the compiler libraries and the version from the standard libraries can be used.

Sorry for stating the obvious, if this was already known to you.

mattpolzin: Right, stuff in the contrib version but not in the compiler version of the Extra module are fine; that just means the functions have never been needed by the compiler. 

As Stefan noted, the compiler will duplicate functions prior to their availability in the previous release of the base libraries, but that consideration doesn't need to be made for the contrib package. In short, we should not need to hide and then re-export base library functions in contrib. 

The difference, IIUC, is in the order of compilation. The compiler is built using the prelude/base from the previous release and then _after_ that the new compiler is used to build the new version of prelude/base/contrib, etc.

stefan-hoeck: You are right, @mattpolzin . In this case, the problem with new stuff in the standard libs is actually not the bootstrap compiler (this will build the current version of the standard libs first) but the previous release, which will use the previous release version of the standard libs.

alexhumphreys: https://github.com/idris-lang/Idris2/pull/1808/commits/e91a322d4184f53ed77d482673d1b8a69a373914 i pushed this commit as an attempt to remove the duplication from `Data.String.Extra`, it's currently failing on `Text.Literate` here: https://github.com/alexhumphreys/Idris2/blob/feat/sync-parsers/src/Libraries/Text/Literate.idr#L84 (and the same error in `contrib`). Changing `:::` to `::` causes a "not covering" error, and the function is kinda confusing so I'm not sure which cases or missing, or how the should be handled.

Thought I'd push this anyway, just to see if I'm on the right track with the de-duplication effort.

stefan-hoeck: Could this be related to #1585, where the semantics (and types) of `lines` was changed?

alexhumphreys: Yep I think so. Now it probably has to handle an empty case which couldn't exist before.

alexhumphreys: Or to be more precise:
```
Missing cases:
    with block in reduce _ [] _ _ _ _ _ _
```

mattpolzin: I suggest focusing on getting the contrib version working without redefining lines/unlines (using the new semantics called out by Stefan) and leave the task of getting the compiler version fixed fro a separate PR.

Because #1585 was merged _after_ version 0.4.0 was cut, removing the compiler redefinitions of lines/unlines will actually be a two-step process for all the same reasons discussed above:
1. actually update the redefined versions to align with the definitions in #1585 but leave them in plac
2. remove the redefinitions after the _next_ version of Idris is cut (0.5.0 or whatever it will be).

As noted above, we do not need to go through the same song and dance in contrib, we just need to use the semantics of the lines/unlines as defined in the current commit. Fixing remaining errors in contrib will likely still be the process @alexhumphreys is describing above of adding pattern clauses for the empty list case that is introduced by the new semantics.

alexhumphreys: Aye, glad you said that as I was about to suggest the same. Even after the above error there's a bunch of changes to `src/Core` to be made. I just pushed a new commit to remove the duplication in contrib, but I duplicated the `lines` function in `Text.Literal` to preserve the old behaviour and not have to fiddle with that `reduce` function. Couldn't use a `where` clause for some reason, kept getting parse errors, so had to make a private namespace so the names didn't leak out.
Bit hacky, but I don't have much more time today so just wanted to get something reasonable up :D

gallais:
Hmmm the frex test is getting in the way. Not sure how to proceed here. It's a bit
annoying to have to update frex, update the PR & merge it, and update frex again.
But I guess that's the cost we have to pay for being part of the test suite and
making sure no performance regressions land in Idris2.

alexhumphreys:
Oh nice one, thanks for finding/fixing that!

gallais (approved):

libs/contrib/Data/String/Extra.idr
alexhumphreys: Thanks to gallais the tests should now be passing. I added the commit https://github.com/idris-lang/Idris2/pull/1808/commits/7fc48582160640dd873de3681f4da8f7e9f3c995 to explain that random `namespace` I added in `Text.Literal` and give it a better name. I was hoping to hide those functions in a `where` block but I don't think that's possible with `with`, and I've opened #1813 to track that.
Is there anything else stopping this from getting merged? If that `namespace` I added is too hacky, the `reduce` function could changed to use the new `lines`, but that'd probably need someone who understands it. I tried but wasn't able to grok it enough to be confident in changing it.

CodingCellist:
Is this fine to merge? @gallais, you mentioned something about `TokenData` vs `WithBounds` when helping with my parser example?  : )

gallais:
@CodingCellist Yeah sorry it'll break your project now that the lib has caught up with
the version in the compiler's internals. :(

jfdm:
and break other people's projects...let's hope the reviewer's won't complain about something not working against HEAD...

alexhumphreys:
Thanks for merging! 

And aye, in hindsight I should have put "breaking" in the PR title :/ 

At least with the Grammar changes, updating the JSON parser to the new version was pretty straightforward, mostly just adding an extra unused arg to the Type signatures. The diff to any of the files in the JSON dir should show what kind of changes are needed.

PR #1810, opened 2021-08-01
Added "lookupBetween" to SortedMap

redfish64:
This allows for looking up the entries between a given a key quickly.

Not sure if this belongs here, but since SortedMap uses private data structures, there is no way to do this without adding the function directly into Data.SortedMap, and I just thought it might be useful.

### DIFF
--- a/libs/contrib/Data/Validated.idr
+++ b/libs/contrib/Data/Validated.idr
@@ -98,6 +98,28 @@ Traversable (Validated e) where
   traverse _ $ Invalid e = pure $ Invalid e
 
 public export
+Semigroup e => Zippable (Validated e) where
+  zipWith f (Valid l)   (Valid r)   = Valid $ f l r
+  zipWith _ (Valid _)   (Invalid r) = Invalid r
+  zipWith _ (Invalid l) (Valid _)   = Invalid l
+  zipWith _ (Invalid l) (Invalid r) = Invalid $ l <+> r
+
+  zipWith3 f (Valid x)   (Valid y)   (Valid z)   = Valid $ f x y z
+  zipWith3 _ (Valid _)   (Valid _)   (Invalid z) = Invalid z
+  zipWith3 _ (Valid _)   (Invalid y) (Valid _)   = Invalid y
+  zipWith3 _ (Valid _)   (Invalid y) (Invalid z) = Invalid $ y <+> z
+  zipWith3 _ (Invalid x) (Valid _)   (Valid _)   = Invalid x
+  zipWith3 _ (Invalid x) (Valid _)   (Invalid z) = Invalid $ x <+> z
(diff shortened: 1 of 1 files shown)


Z-snails:
There's a way around it being a private structure.
You can make a module with a name that's a child of `Data.SortedMap`
(eg `Data.SortedMap.Extra`) and then you can import `Data.SortedMap` and then it's public imported in that module.

We (the idris community) are currently considering refactoring contrib and splitting off parts into other libraries/repos but I don't see why we shouldn't add stuff in the meantime.

libs/contrib/Data/SortedMap.idr
```diff
@@ -319,6 +319,44 @@ export
...
+treeLookupBetween : Ord k => k -> Tree n k v o -> (Maybe (k,v),Maybe (k,v))
+treeLookupBetween k (Leaf k' v) with (k < k')
+  treeLookupBetween k (Leaf k' v) | True = (Nothing, Just (k',v))
+  treeLookupBetween k (Leaf k' v) | False = (Just (k',v), Nothing)
+treeLookupBetween k (Branch2 t1 k' t2) with (k < k')
+  treeLookupBetween k (Branch2 t1 k' t2) | True =
+    let t1r = treeLookupBetween k t1
+    in
+      case t1r of
+        (x,Nothing) => (x,snd $ treeLookupBetween k t2)
```
gallais, line 331: I feel like the logic would be clearer if we had separate `leftMost` and `rightMost`
functions instead of `snd` and `fst` of tree lookups for a key that's clearly too
small or big to be found.

redfish64: Ok, that makes sense. I modified the code and I also exported `leftMost` and `rightMost` since they have easily understood functionality and seem like they might come in handy for users. But if you would rather me to keep them private, let me know.

libs/contrib/Data/SortedMap.idr
```diff
@@ -319,6 +319,68 @@ export
...
+treeLookupBetween k (Leaf k' v) with (k < k')
+  treeLookupBetween k (Leaf k' v) | True = (Nothing, Just (k',v))
+  treeLookupBetween k (Leaf k' v) | False = (Just (k',v), Nothing)
+treeLookupBetween k (Branch2 t1 k' t2) with (k < k')
+  treeLookupBetween k (Branch2 t1 k' t2) | True =
+    let t1r = treeLookupBetween k t1
+    in
+      case t1r of
+        (x,Nothing) => (x,Just $ treeLeftMost t2)
+        _ => t1r
```
gallais, line 342: I'd probably rewrite this the following way (and similarly for the symmetric case):
```suggestion
    let (lower, upper) = treeLookupBetween k t1 in
    (lower, upper <|> pure (treeLeftMost t2))
```

PR #1812, opened 2021-08-01
Return a Vect from Stream take

joelberkeley:
I've added a new version of `Stream.take` which returns a `Vect`


joelberkeley:
I have just noticed `Data.Vect` isn't imported in this file, and I'm guessing it can't be because `Prelude` is higher up the tree. Should `take` be in `Data.Vect` perhaps?

libs/base/Data/Vect.idr
```diff
@@ -87,6 +87,14 @@ take : (n  : Nat)
 take 0 xs = Nil
 take (S k) (x :: xs) = x :: take k xs
 
+namespace Stream
+  ||| Take precisely n elements from the stream.
+  ||| @ n how many elements to take
+  ||| @ xs the stream
+  public export
+  take : (n : Nat) -> (xs : Stream a) -> Vect n a
+  take = fromList . take
```
joelberkeley, line 96: i'm not clear on the implications of using `fromList` instead of copy-pasting from the `List` implementation. Anyone know?

libs/base/Data/Vect.idr
```diff
@@ -87,6 +87,14 @@ take : (n  : Nat)
 take 0 xs = Nil
 take (S k) (x :: xs) = x :: take k xs
 
+namespace Stream
+  ||| Take precisely n elements from the stream.
+  ||| @ n how many elements to take
+  ||| @ xs the stream
+  public export
+  take : (n : Nat) -> (xs : Stream a) -> Vect n a
+  take n xs = fromList $ take n xs
```
joelberkeley, line 96: i'm not clear on the implications of using `fromList` instead of copy-pasting from the `List` implementation. Anyone know?

gallais: In scheme `fromList` should essentially be the identity function given that both
`List` and `Vect` will be represented using scheme lists. I don't know whether either
the backend of Chez detects that and optimises it away though.

joelberkeley: `fromList` is defined later so I can't use it here. Since it's not clear which is better, I just copied the implementation from the `List` version

PR #1814, opened 2021-08-05
Reuse ide socket.

redfish64:
Reuse ide socket.

This sets SO_REUSEADDR which allows the socket to be immediately 
reopened on restart if the process didn't close the socket on the 
last exit.

Otherwise, trying to immediately restart idris with --ide-mode-socket
after the process was killed or exited uncleanly will cause idris to
refuse to start.


gallais:
Anyone understands why this seems to be breaking unix sockets?
It would be nice to be able to merge this fix.

Issue #1216, opened 2021-03-21
Function hole leads to internal error


ysangkok:
# Steps to Reproduce
```
main : IO ()
main = let
  x : String
  (?h x) = Just "hi!"
  in
  print x
```

# Expected Behavior
Should report that the hole `?h` is a constructor of type `String -> Maybe String`.

# Observed Behavior
`Error: INTERNAL ERROR: ?h is not a function application`

andrevidela:
It took me a moment to understand what was going on here. I think Idris is parsing this as a clause with `?h x` as the left hand side and `Just "hi!"` as the right hand side, the left side should be a pattern and indeed something like `Just x` would be a valid pattern. However, holes cannot appear in patterns and so Idris aborts with an ugly error message because it's been given something that shouldn't have been possible in the first place.

gallais:
> holes cannot appear in patterns

That was my first reaction too but then it got me wondering: is there
a deep reason for this restriction?

gallais:
The fix should be fairly similar to that of #1035 so I'll mark this as a good first issue.

gallais:
This was fixed in #1816.

PR #1816, opened 2021-08-05
Fix error message for incorrect function patterns

### LINKED ISSUE #1441: Wrong Code Similar to Pattern Matching on a Pair Raises an Internal Error
# Steps to Reproduce

```idris
module Main

(a, b) = (1, 2)
```

# Expected Behavior

Idris gives an error message about where and why it's wrong.

# Observed Behavior

```
Error: INTERNAL ERROR: (|((Builtin.Pair a) b),((Builtin.MkPair a) b)|) is not a function application
```

eayus:
Produces a proper error message when the leftmost term in a function clause's pattern is not a function name. Fixes #1441

### DIFF
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -694,14 +694,15 @@ mutual
            pure $ MkImpTy fc nameFC n !(bindTypeNames fc (usingImpl syn)
                                                ps !(desugar AnyExpr ps ty))
 
-  -- Attempt to get function name from function pattern. For example, given the
-  -- pattern 'f x y', getClauseFn would return 'f'.
+  -- Attempt to get the function name from a function pattern. For example,
+  --   - given the pattern 'f x y', getClauseFn would return 'f'.
+  --   - given the pattern 'x == y', getClausefn would return '=='.
   getClauseFn : RawImp -> Core Name
   getClauseFn (IVar _ n) = pure n
   getClauseFn (IApp _ f _) = getClauseFn f
   getClauseFn (IAutoApp _ f _) = getClauseFn f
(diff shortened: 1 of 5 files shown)


eayus:
Note this also changes the error message for #1216 .

buzden:
Why not to add a test case for that?

eayus:
Ok I'll have a go at adding a test now

src/Idris/Desugar.idr
```diff
@@ -694,14 +694,14 @@ mutual
...
+  -- pattern 'f x y', getClauseFn would return 'f'.
   getClauseFn : RawImp -> Core Name
   getClauseFn (IVar _ n) = pure n
   getClauseFn (IApp _ f _) = getClauseFn f
   getClauseFn (IAutoApp _ f _) = getClauseFn f
   getClauseFn (INamedApp _ f _ _) = getClauseFn f
-  getClauseFn tm = throw $ case tm of
-    Implicit fc _ => GenericMsg fc "Invalid name for a declaration"
-    _ => InternalError (show tm ++ " is not a function application")
+  getClauseFn tm = throw $ GenericMsg (getFC tm) "Leftmost term in pattern must be a function name"
```
gallais, line 704: It's not the leftmost one but the head one. Patterns on the LHS can you use infix
names, e.g. `(x :: xs) !! Z = x`

eayus: Ah right, I'll change that now

eayus: Ok I've fixed the wording. I also added a comment above the `getClauseFn` to try and help other new contributors, however it's probably worth verifying the accuracy of what I've written.

gallais:
Thank you!

PR #1817, opened 2021-08-05
[ breaking? ] Some optimisations mainly involving Nat and Fin

Z-snails:
- Constant folding, which also moves constants to the right wherever possible (+ and *)
- An inlining heuristic.
  - this inlines very small functions (at most one function application on variables), which should remove most aliases
- 'Identity' optimisation for functions which are the identity function at runtime
  - thanks to inlining, in most cases this completely removes these functions.

These optimisations combined allow idris to automatically optimise casts between natural types and integers to be `O(1)`
Although they have to be written in a fairly specific way (see `Data.Fin.natToFinLT` and `Data.Fin.fromInteger`)

To make the new definition of `fromInteger` actually usable, `LTE` has to be in scope, so I `import public`ed `Data.Nat`
Which might cause some ambiguity issues, but, at least in the compiler, almost all modules that import `Data.Fin` also import `Data.Nat`


src/Compiler/Opts/ConstantFold.idr
```diff
@@ -0,0 +1,70 @@
...
+-- constant folding of primitive functions
+-- if a primitive function is applied to only constant
+-- then replace with the result
+-- if there's only 1 constant argument to a commutative function
+-- move the constant to the right. This simplifies Compiler.Opts.Identity
+constFold : {vars : _} -> CExp vars -> CExp vars
+constFold e@(CLocal _ _) = e
+constFold e@(CRef fc x) = e
+constFold (CLam fc x y) = CLam fc x $ constFold y
+constFold (CLet fc x inlineOK y z) = CLet fc x inlineOK (constFold y) (constFold z)
```
gallais, line 19: Should we inline the `let` if `constFold y` yields a constant?

src/Compiler/Opts/ConstantFold.idr
```diff
@@ -0,0 +1,70 @@
...
+    fromNF (NPrimVal fc c) = Just $ CPrimVal fc c
+    fromNF _ = Nothing
+    constRight : FC -> PrimFn ar -> Vect ar (CExp vars) -> CExp vars
+    constRight fc (Add ty) [x@(CPrimVal _ _), y] = COp fc (Add ty) [constFold y, x]
+    constRight fc (Mul ty) [x@(CPrimVal _ _), y] = COp fc (Mul ty) [constFold y, x]
+    constRight _ _ _ = e
+constFold (CExtPrim fc p xs) = CExtPrim fc p $ constFold <$> xs
+constFold (CForce fc x y) = CForce fc x $ constFold y
+constFold (CDelay fc x y) = CDelay fc x $ constFold y
+constFold (CConCase fc sc xs x) = CConCase fc (constFold sc) (foldAlt <$> xs) (constFold <$> x)
```
gallais, line 40: Same question here: is there any chance `constFold sc` will yield a constant?

src/Compiler/Opts/ConstantFold.idr
Z-snails: I think that would be a better fit for `Compiler.Inline` (there's some other times where you can inline a let without duplicating work). I also have no idea where the `inlineOK` comes from, so I don'y know if I should change that

gallais: My point being that `constFold` of `z` where `x` has been replaced by the
constant could fold further.

Z-snails:
This is causing some issues, because overflow depends on the backend, but we don't know what backend it is until later, so I may have to disable constant folding, or at least check values are much too small to even have a chance of overflowing

Also idris2/perf008 is failing, because idris isn't normalising enough, any ideas on how to fix?

src/Compiler/Opts/ConstantFold.idr
Z-snails: Good point!

gallais:
> because overflow depends on the backend

That's only true for underspecified types like `Int`, right? So you could do constant folding
for all but this one type.

Z-snails:
I don't know why, but it seems the old definition of `Fin.fromInteger` seems to reduce much better than any of my new definitions. If anyone has any ideas of how to improve this, I'll do that, otherwise I'll just revert the change.

Z-snails:
I've fixed all the casts, the only failing test now is the ubuntu-bootstrap-racket which is timing out again.

src/Compiler/Opts/ConstantFold.idr
```diff
@@ -0,0 +1,153 @@
...
+     in if foldableOp fn
+        then case the (Maybe (Vect arity (NF vars))) $ Prelude.traverse toNF xs' of
+                Nothing => e
+                Just nfs => case getOp fn nfs of
+                    Just nf => fromMaybe e $ fromNF nf
+                    Nothing => e
+        else e
+  where
+    toNF : CExp vars -> Maybe (NF vars)
+    toNF (CPrimVal fc (I _)) = Nothing -- don't fold `Int` because it has varying widths
```
gallais, line 101: Should we also leave floats out because of varying widths?

src/Compiler/Opts/ConstantFold.idr
```diff
@@ -0,0 +1,153 @@
...
+    toNF (CPrimVal fc c) = Just $ NPrimVal fc c
+    toNF _ = Nothing
+
+    fromNF : NF vars -> Maybe (CExp vars)
+    fromNF (NPrimVal fc c) = Just $ CPrimVal fc c
+    fromNF _ = Nothing
+
+    constRight : {ar : _} -> FC -> PrimFn ar -> Vect ar (CExp vars) -> CExp vars
+    constRight fc (Add ty) [x@(CPrimVal _ _), y] = COp fc (Add ty) [constFold y, x]
```
gallais, line 110: Same comment here: I don't think we can guarantee that addition on floats is commutative.

src/Compiler/Opts/ConstantFold.idr
```diff
@@ -0,0 +1,164 @@
...
+import Data.Vect
+
+find : FC -> SizeOf outer -> CExp inner -> Var (outer ++ [x] ++ inner) -> CExp (outer ++ inner)
+find fc out val var = case sizedView out of
+    Z => case var of
+        MkVar {i=Z} First => val
+        MkVar {i=S _} (Later p) => CLocal fc p
+    S out' => case var of
+        MkVar {i=Z} First => CLocal fc First
+        MkVar {i=S _} (Later p) => weaken $ find fc out' val (MkVar p)
```
gallais, line 18: This function should probably be reorganised to call `weakenN` once instead of
calling `weaken` many times (`SizeOf outer`). Something like:

```
aux : SizeOf outer -> CExp inner -> Var (outer ++ [x] ++ inner) ->
	  Either (Var (outer ++ inner)) (CExp inner)
```

should be doable.

src/Compiler/Opts/ConstantFold.idr
```diff
@@ -0,0 +1,164 @@
...
+    smaller (Signed x) (Unsigned y) = x < P y
+    smaller (Unsigned x) (Signed y) = P x < y
+foldableOp _ = True
+
+-- constant folding of primitive functions
+-- if a primitive function is applied to only constant
+-- then replace with the result
+-- if there's only 1 constant argument to a commutative function
+-- move the constant to the right. This simplifies Compiler.Opts.Identity
+constFold : {vars : _} -> CExp vars -> CExp vars
```
gallais, line 74: I wonder whether this could have a more general type so that you're not
forced to call `inlineLet` before re-traversing the term (and therefore
the inlined term too!).

gallais (commented):
I had to kill the build because it was taking hours to self-host.
There seems to be a performance issue in the code. I've suggested
a few things but I don't know if that's where the issue is.

PR #1821, opened 2021-08-05
Cache CI

fabianhjr:
- Refactor Makefiles to use and export `IDRIS2_FLAGS`
- Add TTC cache for Quickcheck and Racket Self-hosting with `-Xcheck-hashes`

Benefits:
- Halves Quickcheck from 6m to 3m on the best case scenario (rerun with cache or no `.idr` changes with cache)
- Reduces ubuntu-racket-selfhost time from ~32m to ~26m on the best case scenario

### DIFF
--- a/.github/workflows/ci-idris2.yml
+++ b/.github/workflows/ci-idris2.yml
@@ -70,6 +70,7 @@ jobs:
     env:
       IDRIS2_CG: chez
       SCHEME: scheme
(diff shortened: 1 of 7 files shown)


gallais:
I'm not convinced this is a good idea as part of the reason for CI is to test the compiler, even if that means
rebuilding stuff we assume should work.

PR #1826, opened 2021-08-10
[ base ] Relax requirements of `Alternative` of `ReaderT` + lazy `<|>` for transformers + cleanup

buzden:
There are three more-or-less independent changes, but all about transformers. All of them are in different commits and all can be considered independently if needed:

- the current implementation of `Alternative` for `ReaderT` requires `Monad` but it may not;
- some (hopefully) readability cleanup;
- the current signature of `Alternative` has its second argument lazy; however, the current implementations of `Alternative` for transformers do not treat it nicely, they force its second argument despite the underlying alternative/monad may not require its second argument; the fix is proposed.

### DIFF
--- a/libs/base/Control/Monad/Reader/Reader.idr
+++ b/libs/base/Control/Monad/Reader/Reader.idr
@@ -42,35 +42,33 @@ runReader s = runIdentity . runReaderT s
 
 public export
 implementation Functor f => Functor (ReaderT stateType f) where
-  map f (MkReaderT g) = MkReaderT (\st => map f (g st))
+  map f (MkReaderT g) = MkReaderT $ \st => f <$> g st
 
 public export
 implementation Applicative f => Applicative (ReaderT stateType f) where
-  pure x = MkReaderT (\st => pure x)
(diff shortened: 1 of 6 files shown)


libs/base/Control/Monad/Reader/Reader.idr
```diff
@@ -42,35 +42,33 @@ runReader s = runIdentity . runReaderT s
 
 public export
 implementation Functor f => Functor (ReaderT stateType f) where
-  map f (MkReaderT g) = MkReaderT (\st => map f (g st))
+  map f (MkReaderT g) = MkReaderT \st => f <$> g st
```
gallais, line 45: Trailing binders (and so lambdas too) are going to be removed (cf. 1711) so that
all the arguments are treated uniformly & we can generate better error messages
in type signatures.

gallais (changes requested):

libs/base/Control/Monad/Reader/Reader.idr
buzden: I've rebased upon just merged 1711.

PR #1830, opened 2021-08-10
Implement Cast in place of ad hoc conversions

nickdrozd:
This PR doesn't introduce any new interfaces. It just implements an existing interface in places where I feel it ought to be used. The net result is a significant reduction in horrible names like `fromList`.

As usual there are still a few places where inference struggles, requiring explicit type annotations and other hacks.

### DIFF
--- a/libs/base/Data/Colist.idr
+++ b/libs/base/Data/Colist.idr
@@ -19,14 +19,14 @@ data Colist : (a : Type) -> Type where
 
 ||| Convert a list to a `Colist`.
 public export
-fromList : List a -> Colist a
-fromList []        = Nil
-fromList (x :: xs) = x :: fromList xs
+Cast (List a) (Colist a) where
+  cast []        = Nil
+  cast (x :: xs) = x :: cast xs
 
 ||| Convert a stream to a `Colist`.
 public export
-fromStream : Stream a -> Colist a
-fromStream (x :: xs) = x :: fromStream xs
+Cast (Stream a) (Colist a) where
+  cast (x :: xs) = x :: cast xs
 
 ||| Create a `Colist` of only a single element.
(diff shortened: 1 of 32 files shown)


mattpolzin:
I wonder if `Cast` has the right semantics here given that it is explicitly for potentially lossy conversations (so my mental model is to take heed when using it).
```
  ||| Perform a (potentially lossy!) cast operation.
  ||| @ orig The original type
  cast : ...
```

buzden:
I've always wondered, maybe it is a good idea to have `Cast` for definitely lossy conversions and some another interface with semantics of lossless conversions.

stefan-hoeck:
This is a breaking change to the standard libs. It should therefore first be discussed in the issue tracker as described [here](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md#things-that-should-be-discussed-via-the-issue-tracker-first).

I'm also not sure this is actually an improvement. Additional passing of dictionaries is introduced and type inference for `cast` might be worse that for `fromList` or for `fromStream`. Finally, this seems like a mere cosmetic change. Is it worth to potentially break a lot of code for this?

nickdrozd:
My interpretation of the `Cast` interface is that it should be used for conversions that are "stupid" -- no interesting calculations are done, and the conversion is implemented exactly as anyone would think to implement it. There are a lot of conversions like this in the codebase. For instance, there are several variations of:

```idris
fromList : List a -> Colist a
fromList []        = Nil
fromList (x :: xs) = x :: fromList xs
```

To me, that kind of code is boring and doesn't deserve to be named, and cutting back on named functions is a simplification of the library API.

But I don't feel too strong about it, so I can drop it if it has no support.

PR #1833, opened 2021-08-11
[ elab ] Add an ability to fail elaboration script with a custom `FC`

buzden:
I sometimes find myself wanting to complain about a particular point of a user code inside my elaboration scripts. This PR adds such an ability while keeping the original behaviour unchanged.

### DIFF
--- a/src/TTImp/Elab/RunElab.idr
+++ b/src/TTImp/Elab/RunElab.idr
@@ -63,9 +63,12 @@ elabScript fc nest env script@(NDCon nfc nm t ar args) exp
                               !(sc defs (toClosure withAll env
                                               !(quote defs env act'))) exp
                   x => failWith defs $ "non-function RHS of a Bind: " ++ show x
-    elabCon defs "Fail" [_,msg]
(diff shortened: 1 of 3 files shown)


src/TTImp/Elab/RunElab.idr
```diff
@@ -63,9 +63,14 @@ elabScript fc nest env script@(NDCon nfc nm t ar args) exp
...
                   x => failWith defs $ "non-function RHS of a Bind: " ++ show x
-    elabCon defs "Fail" [_,msg]
+    elabCon defs "Fail" [_, mbfc, msg]
         = do msg' <- evalClosure defs msg
-             throw (GenericMsg fc ("Error during reflection: " ++
+             customFC <- reify defs !(evalClosure defs mbfc)
+             let customFC' = case the (Maybe FC) customFC of
+                               Nothing      => fc
+                               Just EmptyFC => fc
+                               Just x       => x
```
gallais, line 72: Should we just take an `FC` then given that `Nothing` and `Just EmptyFC` are treated the same way?

PR #1837, opened 2021-08-12
[ doc ] Fix a complain of not being in a toctree

buzden:
I've figured out the genesis of the warning introduced at #1775. Sphinx (and, thus Github) requires every `rst` file to be either included using `.. include::` or included through table of contents (`toctree`) or explicitly orphaned. `windows.rst` does not conform to any of these.

What we can do is to include it as hidden into the `starting.rst` document, an addition to which `windows.rst` actually is. Unfortunately, you cannot add a `toctree` item between existing sections, only above them. So, if this PR is applied, `toctree` (table of contents on the left) starts looking like this:

x

i.e., prerequisites for windows are listed before the general prerequisites.

If you add a `toctree` directive to the "Prerequisites" section (not to the upper "Installing from Source", as it is in this PR), then you don't have "Prerequisites for Windows" listed in the table of contents at all because it starts to be too deep inside the table of contents. However, this still removes the warning.

I'm not sure whether this table of contents' look is acceptable, but I think it is still better than having warnings in each PR and during the docs building.

### DIFF
--- a/docs/source/tutorial/starting.rst
+++ b/docs/source/tutorial/starting.rst
@@ -7,6 +7,11 @@ Getting Started
 Installing from Source
 ======================
 
+.. toctree::
+   :hidden:
+
+   windows
+
 Prerequisites
 -------------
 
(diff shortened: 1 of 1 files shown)


gallais:
I guess the principled solution would be to have toc for installation that
directs you to the right advice based on your OS?

buzden:
> I guess the principled solution would be to have toc for installation that
> directs you to the right advice based on your OS?

Yes, kind of

gallais:
I'd still rather have the principled solution but let's silence the annoying warning.

Issue #1727, opened 2021-07-16
Library organisation


edwinb:
I'm opening this, with some trepidation, because I'd like to do something about the state of `contrib`, and it'd be interesting to hear people's opinions. Essentially, the problem is that it's become - or maybe it always was - something of a dumping ground for experimental libraries, but it also includes some libraries that are being actively used (syntax/proof related things), and libraries that are good to have in a "batteries included" language implementation (e.g. JSON, parsing, pretty printing). But, everything that's in this repository implies a committment on our part (and, ultimately, on my part) to maintain it. There's only a certain amount of that I can do, realistically, as many of you have probably noticed lately.

So, what I want to do is organise them better. Some should be carved out and put in a library that's listed on https://github.com/idris-lang/Idris2/wiki/Libraries. Some should be moved into more meaningfully named packages that are shipped with Idris itself. Anything left over, maybe it's a sign that it's not that useful after all.

Please let me know:

* Any parts of the library you're willing to claim and maintain. In this case, I think the best thing to do would be to start a new repo with those parts, and add them to https://github.com/idris-lang/Idris2/wiki/Libraries
* Opinions - I don't promise to respond to them or necessarily agree, but I will at least read them. Including:
  - Any parts of the library you really can't live without
  - Suggestions for good names for subsets of `contrib` that you think

I think `base` is mostly okay - these are things that are effectively part of the language. I don't think `Control.App` belongs there, because it's too experimental, although I'd still like that to be available because I know people are using it.

I'd still like to be a bit careful about dropping things, because without package management we could end up with awkward dependency problems.

I'm also willing to consider moving all libraries that aren't essential to the Idris 2 build into a new repository. In this case, we'd need a volunteer to maintain that new repository. I would also update the release scripts to package Idris 2 and the libraries together in a single distribution if we did this.

mattpolzin:
To make things more complicated (but recognizing that at the end of the day as long as I can get this stuff from somewhere I am happy to pull in other additional libraries), some of the `Extra` modules are a mix of really useful and maybe less often needed things even within the same module; the name "Extra" also doesn't give much guidance on what to put there or what to expect to find there. 

So, on the topic of reorganization, maybe it's still worth _some_ things in contrib being considered for base and maybe this reorganization will motivate a closer look and ultimatum for those `Extra` modules.

edwinb:
[...]

PR #1842, opened 2021-08-15
The math package

Sventimir:
Following discussion under #1727 I have extracted math-related stuff from `contrib` into its own package called `math`. My criteria for including a module in `math` were twofold:

1. It's related to a numeric type defined in the standard library (i.e not built into the language) or defines such types.
2. It contains (ideally exclusively) definitions or proofs of properties of these types.

This is meant to be mostly a library of proofs, so it makes little sense to include modules related to built-in types like `Integer`. It's purposes are threefold:
1. Demonstrate the ability of Idris as a proof-assistant;
2. Provide a place for people interested in theorem proving in Idris to put their proofs in;
3. In the long term, provide an extensive library of proofs of properties of basic numeric types that can be used to describe and prove properties of practical programs more easily.

I'm not entirely sure if I tracked down all the modules in `contrib` that fulfill my criteria. If there are some that I missed, I'm happy to include them as well.


nickdrozd:
Seems like a reasonable compromise between leaving the math stuff strewn about contrib and cutting it entirely.

andrevidela:
It looks like this has diverged a little bit. I suspect it's because none of us are really willing to touch the libraries until we've found a real consensus. @Sventimir would you be ok postponing this PR until the developper meetup in december? I encourage you to talk about it with everyone else there

Sventimir:
Yep, I', fine with that. However, I have no idea if I'll be able to attend the meeting at this point. I'm having other occupations than Idris recently. Closing the PR.

Issue #1843, opened 2021-08-16
void-returning foreign functions given unit type


DrBearhands:
Came across what is probably a niche problem. In the documentation, foreign imperative functions with no return value are declared as returning a unit. E.g. this example:

```Idris
%foreign "browser:lambda:x=>{document.body.innerHTML = x}"
prim__setBodyInnerHTML : String -> PrimIO ()
```

taken from: https://idris2.readthedocs.io/en/latest/backends/javascript.html

In the javascript backend at least, this function does not actually return a unit, but `undefined`. This can result in runtime exceptions e.g. if the fake unit is pattern-matched. On the other hand, while incorrect, pretending the foreign function returns unit (without adding the overhead of generating a real unit value) seems useful in the majority of situations.

I suspect this problem will resolve itself with performance improvements to the backend, as a switch on unit seems redundant, but it might be a good idea to mention the problem in the docs.

stefan-hoeck:
Interesting. It would be pretty straight forward (in the JS backend at least) to identify Idris2 `()` with JS `undefined` in pattern matches and constructor applications. Do you think @DrBearhands this would resolve this issue or can you see other problems with such a "fix"?

stefan-hoeck:
So, here is a pathological example. We expect the following program to print
```
a test
foo
a test
```
but it actually prints
```
a test
foo
oopsie
```

```idris
test : () -> IO ()
test () = putStrLn "a test"
test _  = putStrLn "oopsie"

main : IO ()
main = do test ()
          putStrLn "foo" >>= test
```

The pattern match on `()` is converted to a switch on the integer constant `0`, so it might not be trivial to figure out that this actually came for `()` in the first place. If, however, the catch all case is removed from `test`, the JS backend optimizes away the single case switch statement. So, at least in the JS statement, this seems not to be a problem unless one writes nonsensical pattern matches as shown above.

I wonder, however, why Idris does not complain about an unreachable case in the code above...

DrBearhands:
> Do you think @DrBearhands this would resolve this issue or can you see other problems with such a "fix"?

I suspect it would work, but frankly I don't know Idris well enough to make that call. I'd be concerned about weird behavior of undefined and runtime type information. E.g. for the former `undefined >= undefined` is false, but `undefined === undefined` is true. 

Concerning your pathological example, I didn't get that optimization, so I thought it didn't exist. Will have to investigate and return with a minimal example.

Z-snails:
[...]

PR #1844, opened 2021-08-16
Optimise away case statements on unit-y types

Z-snails:
This fixes the example in #1843, however, as @DrBearhands notes, it might not fix all cases.

I couldn't immediately think of how to fix the root cause of this issue, as the js ffi can have arbitrary expressions in.
Maybe @stefan-hoeck can think of something (feel free to make your own PR if you want)

### DIFF
--- a/src/Compiler/CompileExpr.idr
+++ b/src/Compiler/CompileExpr.idr
@@ -270,6 +270,14 @@ enumTree (CConCase fc sc alts def)
     toEnum _ = Nothing
 enumTree t = t
 
+-- remove pattern matches on unit
+unitTree : CExp vars -> CExp vars
+unitTree exp@(CConCase fc sc alts def) = fromMaybe exp
+    $ do let [MkConAlt _ UNIT _ [] e] = alts
+             | _ => Nothing
+         pure e
+unitTree t = t
+
 -- See if the constructor is a special constructor type, e.g a nil or cons
 -- shaped thing.
 dconFlag : {auto c : Ref Ctxt Defs} ->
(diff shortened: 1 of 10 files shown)


Z-snails:
Don't merge this yet I can think of an easier way, which doesn't involve a ttc version change.

stefan-hoeck:
Am I right in assuming that a `()` value is still converted to a 0 constant with this change, or will it actually be a `CCon ... ... UNIT (Just 1) []`? In the latter case, we could indeed convert unit values to `undefined` in the JS backends.

Z-snails:
@stefan-hoeck no it won't be comnverted to an int, however it will be now available for backends as a new ConInfo flag (so for example, the js backend could convert it into undefined)
It wouldn't be hard to convert it to an int, but then backends wouldn't get a choice on how it's represented.

stefan-hoeck:
So this is the way forward then: After merging this pull request, backend implementors are able to choose an appropriate representation for `()` values in order to make sure that they agree with the types of FFI function calls as described in #1843. For the JS backends this means we should represent `()` as `undefined`. I'll write a PR once this is merged.

Z-snails:
I've thought of another issue with this PR (once inlining happens more) in that it might remove effectful functions when removing the case statement.
So I'll change the js backend while I'm at it.

src/Compiler/ES/Codegen.idr
```diff
@@ -145,6 +145,7 @@ applyCon NOTHING _ [] = "{h" <+> softColon <+> "0}"
 applyCon CONS    _ as = applyObj (conTags as)
 applyCon JUST    _ as = applyObj (conTags as)
 applyCon RECORD  _ as = applyObj (conTags as)
+applyCon UNIT    _ [] = "undefined"
```
stefan-hoeck, line 148: For consistency, we should do the same in pattern matches (switch statements), even if we hopefully will never need it.

stefan-hoeck (changes requested):

PR #1847, opened 2021-08-19
[ elab ] Erase `check` and `quote`'s main argument

buzden:
This is effectively an equivalent change, but allows not have arguments, that are already known to the compiler, to be passed as runtime arguments in macro and reflection-level functions.


berewt:
For `[Cc]heck`, as `expected` is a type, your can totally remove the implicit declaration, it's probably (a bit) cleaner.

buzden:
> For `[Cc]heck`, as `expected` is a type, your can totally remove the implicit declaration, it's probably (a bit) cleaner.

Well, maybe you are right. Let's do so.

PR #1852, opened 2021-08-20
[ refactor ] Put holes in their own namespace

gallais:
1. Put holes in their own namespace so that we don't have to worry about
clashing with user defined functions that come later in the file.

2. Record the hole names we've seen when parsing as part of the syntax info.

This works in conjunction with https://github.com/idris-community/idris2-mode/pull/4 which removes the
bug dropping the last character of the code that is spliced in.

Todo
----

* [x] fix bugs
* [x] add test


Z-snails:
Does the new name want to be part of [reflection](https://github.com/idris-lang/Idris2/blob/main/libs/base/Language/Reflection/TT.idr)? (I don't know the answer, just checking you haven't forgotten)

gallais:
> just checking you haven't forgotten

I 100% have. :D

There are lots of places where we have catch-all cases which means we don't get a warning
when we add a new constructor. I wonder whether we should have a disjunction pattern so that
we could express stuff like

```idris
isUserName : Name -> Bool
isUserName (NS _ n) = isUserName
isUserName (UN _) = True
isUserName (RF _) = True
isUserName (MN{} | CaseBlock{} | WithBlock{} | ...) = False
```

so that we get a coverage error if we add a new constructor.

mattpolzin:
> so that we get a coverage error if we add a new constructor.

I think this is an opportunity to get really creative. Not that your idea isn’t, but what if we think really outside the box. 

What if an editor action plopped a hash value representing the non-order-dependent full list of current data constructors for the type and then if that changed it was a coverage error until updated?

```idris
foo : AlphabetChar -> …
foo A = …
foo B = …
foo _<#A20C4> = …
```

I realize this conversation doesn’t belong on this ticket. Just thought the problem of how to get something akin to coverage checking on functions with default patterns sounded like a fun one to dream about a bit.

gallais:
Together with a patch to idris2-mode I got a few things fixed.
Proof search for recursive functions still seems broken. :(

gallais:
This is utterly puzzling. Generating a definition by case-splitting is somehow broken by this patch. :confused:

gallais:
A little progress thanks to splitting off the name generation from the refactoring! :eyes:
This is, if there ever was, a poster child for small self-contained PRs.

src/Idris/Syntax.idr
```diff
@@ -574,12 +574,16 @@ data EditCmd : Type where
...
      ExprSearchNext : EditCmd
      GenerateDef : Bool -> Int -> Name -> Nat -> EditCmd
      GenerateDefNext : EditCmd
-     MakeLemma : Bool -> Int -> Name -> EditCmd
-     MakeCase : Bool -> Int -> Name -> EditCmd
+     MakeLemma : Bool -> Int ->
+                 String -> -- hole name, without the question mark
+                 EditCmd
+     MakeCase : Bool -> Int ->
+                String -> --hole name, without the question mark
```
mattpolzin, line 585: It feels like we lost something in the way of self documenting types with the changes from Name to String in a number of places. What was the motivation for this change?

src/Yaffle/REPL.idr
```diff
@@ -146,13 +154,14 @@ processCatch cmd
...
-       {auto u : Ref UST UState} ->
        {auto s : Ref Syn SyntaxInfo} ->
+       {auto u : Ref UST UState} ->
        Core ()
 repl
     = do coreLift_ (putStr "Yaffle> ")
          inp <- coreLift getLine
          case runParser (Virtual Interactive) Nothing inp command of
-              Left err => do coreLift_ (printLn err)
+              Left err => do coreLift_ $ putStrLn "Argh"
```
mattpolzin, line 164: 😀. Feels about right.

mattpolzin (commented):
Such a useful change. I made it through all the changed code and really just had one question about whether we needed to give up some type information (see inline comment).

andrevidela:
Is this still going on? What's the status @gallais @mattpolzin ?

gallais:
The goal was to merge it right after the next release because the idris-modes will need to be changed to
properly send hole-names over the wire and i'd like to have ample testing time to make sure we got things
right before shipping it as part of a release.

CodingCellist:
Given that we're slowly but surely trying to get a post-IDM release going, do we "just" need to fix the merge conflicts for this to go in?  : )

gallais:
>  do we "just" need to fix the merge conflicts for this to go in? : )

We also need to patch the emacs mode. It may be worth waiting for idris2-mode to be deprecated first so that we only need to make the change in one place. :D

gallais:
I'll keep the branch around in case someone is interested into porting the work.

PR #1855, opened 2021-08-23
Support tarm64osx (AArch64, Apple M1) target

loovjo:
On Racket, the Mac M1 target (AArch64) is known as `tarm64osx`. Currently, Idris2 does not account for this when looking for the `libc.(so/dylib)` shared object file. This does not cause the build to fail, and it actually takes a while for the error to surface (for me it only appeared when trying to run `idris2-lsp`, getting the error `Exception: (while loading libc.so) dlopen(libc.so, 2): image not found`).

This pull request simply addr `tarm64osx` to the `case`-statements selecting macOS-targets, overriding the default behaviour of defaulting to `libc.so`.


loovjo:
Thinking about it, maybe a better solution would be to do some simple pattern matching on the strings instead of having a lookup-list. It seems like the suffix of every machine name is enough to determine what libc it uses, and doing prefix-matching would be more future proof when new targets get added. Thoughts?

PR #1857, opened 2021-08-24
[ fix ] fix #1839

### LINKED ISSUE #1839: JS: ReferenceError: Can't find variable: require due to sysos
# Steps to Reproduce

Compile with `--cg javascript`. Run the program in a browser.

# Expected Behavior

Program works.

# Observed Behavior

https://github.com/idris-lang/Idris2/blob/523c0a6d7823d2b9a614d4a30efb52da015f9367/support/js/support.js#L48-L49
```
ReferenceError: Can't find variable: require
```

stefan-hoeck:
Fixes #1839.

This fix removes the implementation of `prim__os` from the static Javascript preamble. After this, `prim__os` is only supported in the `node` backend but not the `javascript` backend. If people can show me a reliable and useful way to determine the OS in the browser, I'll happily add that as the new implementation of `prim__os` for the `javascript` backend.

### DIFF
--- a/tests/node/fix1839/run
+++ b/tests/node/fix1839/run
@@ -1,4 +1,4 @@
 rm -rf build
 
 $1 --cg node --no-banner --no-color --console-width 0 OS.idr < input
-$1 --cg javascript --no-color -o os.js OS.idr
+$1 --cg javascript --no-banner --no-color --console-width 0 -o os.js OS.idr
--- a/tests/node/fix1839/run
+++ b/tests/node/fix1839/run
@@ -1,0 +1,4 @@
+rm -rf build
+
+$1 --cg node --no-banner --no-color --console-width 0 OS.idr < input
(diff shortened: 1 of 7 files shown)


stefan-hoeck:
Since one version of the new test fails on Window (the error message contains a line break), I'll drop that one unless someone has an idea, why this behaves differently on Windows than on Ubuntu.

tests/node/fix1839/run
```diff
@@ -0,0 +1,4 @@
+rm -rf build
+
+$1 --cg node --no-banner --no-color --console-width 0 OS.idr < input
+$1 --cg javascript --no-color -o os.js OS.idr
```
Z-snails, line 4: You're missing `--no-banner` and `--console-width 0` here, so maybe it's failing due to the console on windows being a different width.

stefan-hoeck: Ah, yes, that makes sense. Thanks for spotting this!

PR #1860, opened 2021-08-24
[ refactor ] Elaboration script functions were abstracted to interface

buzden:
When writing complex elaboration scripts using mtl-style-like typeclasses, it becomes inconvenient that elaboration interface is nailed to the particular monad. I suggest to abstract it a little bit to make custom monads or transformers stacks be able to fulfill the elaboration interface.

Possibly a strange thing may seem in the `lambda` function interface, since `Elab` monad is present in the interface itself. But I think it is the most correct way, since it requires so to say *the least powerful elaboration-supporting monad* as the return type of the argument function. Any additional effect in such functions cannot be managed in the general case, and it seems that it would be unnecessary anyway.

This is an equivalent change and the code that was working in the `Elab` monad should continue to work unchanged.


buzden:
Rebased upon recently merged~~ #1847

PR #1867, opened 2021-08-26
[ contrib ] Make sorted map be able to store dependently typed values

buzden:
This is done by minimal changes inside the internal data structure: practically it was ready to hold dependently typed values. All was split into two modules because of backward-compatibility: all the code that was working with existing non-dependent `SortedMap` data type should work unchanged.

During the review, probably it's better to look at diff between the original `SortedMap` module and the new `SortedMap.Dependent`, in this case there are not so many differences, e.g. using
```
> git diff HEAD^:libs/contrib/Data/SortedMap.idr HEAD:libs/contrib/Data/SortedMap/Dependent.idr
```

### DIFF
--- a/libs/base/Data/DPair.idr
+++ b/libs/base/Data/DPair.idr
@@ -1,5 +1,7 @@
 module Data.DPair
 
+import Decidable.Equality
+
 %default total
 
 namespace DPair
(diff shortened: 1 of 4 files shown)


gallais:
Believe it or not, I just happen to need this! :heart:
Reviewing now.

PR #1877, opened 2021-08-30
Make the (++) operator for lists tail-recursive. [breaks things; not necessarily a good idea]

falsifian:
This PR does three things:
* Makes `(++)` tail-recursive. (More precisely, it turns it into two calls to `reverseOnto`, which is tail-recursive.)
* Moves reverse and reverseOnto into prelude, since (++) uses them.
* Updates a bunch of proofs to work with the new (++) (and adds a new theorem to `Data.List`).

Here are two reasons to make `(++)` tail-recursive:
* It should be faster. A preliminary benchmark confirms this; see below.
* It should help prevent stack overflows. See [my recent thread](https://groups.google.com/g/idris-lang/c/zD751y1AUOI/m/jcrw0Tw-AQAJ) about the javascript backend. Even if stack overflows aren't an issue, it seems better to avoid allocating unnecessary stack frames.

Unfortunately it looks like a lot of things depend on the implementation being what it is. This PR fixes the libraries, but presumably external libraries will break.

I would have preferred to hide the implementation (i.e. use `export` instead of `public export`) and provide lemmas instead, so that future implementation changes won't break everything again. However I [ran into trouble](https://groups.google.com/g/idris-lang/c/_W_6fQnD4KU/m/JxhrXVCEAQAJ) trying to move one of the necessary lemmas into the prelude.

On the whole, I don't know if this change is a good idea. Personally my vote would be for it, because it's more efficient and I don't think it's good to depend on how functions are implemented. But from fixing all the proofs, I do see that it's mighty convenient in proofs to define `(++)` the old way, and expose the implementation. In the worst case, I've learned a bunch about how to prove things in Idris2 :)

## Benchmark

Here's a silly benchmark. I haven't tried anything real, like building Idris.

(EDIT to add: this is with the racket backend, because chez isn't currently ported to OpenBSD.)

The program:

```
import Data.List

%default total

slowReverse : List a -> List a
slowReverse [] = []
slowReverse (x :: xs) = slowReverse xs ++ [x]

main : IO ()
main = putStrLn $ show $ length $ slowReverse $ replicate 100000 ()
```

Before the change (at commit 4a9f0007):
```
falsifian moth tmp $ idris2 -o f f.idr && echo running... && time build/exec/f 
running...
100000
    1m15.56s real     1m15.25s user     0m00.26s system
falsifian moth tmp $ 
I-search: 
falsifian moth tmp $ time build/exec/f                                         
100000
    1m15.17s real     1m14.74s user     0m00.38s system
falsifian moth tmp $ time build/exec/f 
100000
    1m19.07s real     1m18.69s user     0m00.36s system
```

After the change (at this commit):
```
falsifian moth tmp $ idris2 -o f f.idr && echo running... && time build/exec/f 
running...
100000
    0m37.06s real     0m36.75s user     0m00.25s system
falsifian moth tmp $ time build/exec/f
100000
    0m37.40s real     0m36.92s user     0m00.43s system
falsifian moth tmp $ time build/exec/f 
100000
    0m37.12s real     0m36.72s user     0m00.35s system
```

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -321,15 +321,6 @@ public export
 replaceOn : Eq a => a -> a -> List a -> List a
 replaceOn a = replaceWhen (== a)
 
-public export
-reverseOnto : List a -> List a -> List a
-reverseOnto acc [] = acc
-reverseOnto acc (x::xs) = reverseOnto (x::acc) xs
-
-public export
-reverse : List a -> List a
-reverse = reverseOnto []
-
 ||| Construct a list with `n` copies of `x`.
 ||| @ n how many copies
 ||| @ x the element to replicate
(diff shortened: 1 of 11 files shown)


falsifian:
BTW some of the proofs could probably be simplified, and in general I didn't do much touching up. I wanted to see if people think the change is worthwhile before investing more time.

falsifian:
Oops, of course this breaks Idris itself.

I just patched up `Core.TT` (b9410d0) but I think I'll wait for feedback before I keep fixing.

stefan-hoeck:
Would it make sense to have a transform rule for this? This would allow us to use the simple non-tailrecursive in proofs and the efficient tail-recursive version at runtime. We'd only need a single proof then (if possible), showing that the two versions are actually equal.

falsifian:
That sounds good! Could you point to documentation for that (or summarize here)? I vaguely remember reading about something like that but wasn't able to find it just now.

I think I do have enough to prove they do the same thing, but I had trouble moving one of my proofs into prelude as described in my recent thread.

stefan-hoeck:
I'm not sure this is properly documented yet. Here is the link to the transform rule from `pack` to `fastPack` in the prelude:
https://github.com/idris-lang/Idris2/blob/1e0027721f11fb51f7bcea86f39e5c27133a72bd/libs/prelude/Prelude/Types.idr#L580.

The rule for `(++)` should be similar. You could add a test the the node tests with a large enough pair of lists to make sure that the function is properly tail-call optimized.

gallais:
I agree that this is best done via a `%transform` directive.

falsifian:
Okay, I opened a new PR so I can keep this branch around for posterity. #1888

PR #1882, opened 2021-08-31
[ new ] ellipses for with patterns

gallais:
Rather than Agda's `...` we use the common symbol for "I don't care": `_`.


mattpolzin (commented):
This is really nice! I wouldn't consider myself adept enough to give a proper approval of the code, but I can roughly follow and the test case is solid.

PR #1888, opened 2021-08-31
Add a transform rule making (++) for List tail-recursive.

falsifian:
Previous discussion: #1877.

I moved some related functions from `Data.List` to `Prelude` since I'm using them anyway, but I don't know if that's the right decision. (I rewrote `reverseInvolutive` so it doesn't depend on the other theorems from `Data.List`.)

One small problem: I had to put the new proofs at the end of `Types.idr` because I get a compiler error if I put them near the new `%transform` rule. Specifically, they have to go after the declaration and/or definition for `reverse : String -> String`. Otherwise, when I run `gmake prelude`, I get "`consReverse is not total`". Full output: https://gist.github.com/falsifian/bc4008830013dce320548998d38bedf8


falsifian:
I re-ran the benchmark program from the old PR. To switch libraries I just checked out the appropriate commit and ran `gmake install-libs`. This is with the racket backend.

Before change (commit 79177a70ea57445be4ae440a9e34fe0d9851ea03):
```
falsifian moth tmp $ idris2 -o f f.idr && time build/exec/f 
100000
    1m15.29s real     1m15.03s user     0m00.23s system
falsifian moth tmp $ time build/exec/f
100000
    1m15.34s real     1m14.98s user     0m00.29s system
falsifian moth tmp $ time build/exec/f 
100000
    1m15.56s real     1m15.14s user     0m00.35s system
```

After (commit bd6ae6ef89e65d1cc8c717a956de199d1fbf531c):
```
falsifian moth tmp $ idris2 -o f f.idr && time build/exec/f 
100000
    0m24.48s real     0m24.22s user     0m00.21s system
falsifian moth tmp $ time build/exec/f
100000
    0m25.05s real     0m24.78s user     0m00.23s system
falsifian moth tmp $ time build/exec/f 
100000
    0m25.49s real     0m25.20s user     0m00.25s system
```

gallais:
Do we need the proofs in the prelude? Can't they just be in `base`?

falsifian:
Sure, I can do that if you prefer.

Should I export `tailRecAppend` from `Prelude`, then, so the proof in base can see it?

It looks like a check is failing because a package called `frex` expects `Data.List.reverse` to exist. I'm not sure if it's better to (a) add an alias `Data.List.reverse = reverse`; (b) make `Prelude`'s `reverse` private (so it gets implemented twice), or (c) update `frex`.

buzden:
> Do we need the proofs in the prelude? Can't they just be in `base`?

I always thought that it's a good idea for such proofs to be as close as possible to the `%trasnform` directive.

falsifian:
I just added a commit changing the `consReverse` proof. I suspect the previous one shouldn't have been accepted by the complier (https://github.com/idris-lang/Idris2/issues/1896). (Let me know if I should squash once everything's ready.)

I'm inclined to agree with @buzden that it's good to have the proof close to the `%transform` but I don't have much experience with practical trade-offs here, so am happy to move the proofs to `Data.List` in base if there's a consensus.

I haven't fixed the issue with `frex` yet. I tried adding an alias `reverse = Prelude.List.reverse` in `Data.List` but that just caused more trouble (`reverse` becomes ambiguous). I guess the remaining options are:
* Hide the definitions of `reverse` and `reverseOnto` in `Prelude`, and keep the copies of the definitions in `Data.List`.
* Fix `frex` (and probably break other packages for similar reasons).

falsifian:
I hid the new definitions in Prelude, and put copies of them back in `Data.List`. Hopefully that will fix `frex` and any other package that looks for `Data.List.reverse`.  46d7221

gallais:
> I hid the new definitions in Prelude

I don't think that's a good idea. We should update the downstream rather than add duplicates.
I was planning to upgrade frex myself but didn't get the time to.

falsifian:
Okay, I removed 46d7221 from this PR.

I think I'll start work on more `%transforms` for functions in `prelude` and `base`, using this PR as a model. (My goal is to get rid of all the stack overflows in a web program I'm working on. At the very least `List.length` is still a problem. `io_bind` is also an issue but that needs more than a `%transform`, I guess.)

gallais:
`prelude` is really meant to be a very small library that's loaded by default.
Especially if you intent to add more `%transform` then I don't think all the
proofs should be included in the prelude.

falsifian:
> `prelude` is really meant to be a very small library that's loaded by default.
> Especially if you intend to add more `%transform` then I don't think all the
> proofs should be included in the prelude.

Is the goal to make Prelude's interface very small, or keep the total amount of code small?

Anyway, I'll defer to you. fa32d26 moves the proof to `base`. I kept my new `reverseInvolutive` proof (but moved it to `base`) because I like the fact that it doesn't depend on as much. I had to make `tailRecAppend` public so that the proof in `base` can see it.

gallais:
Thanks & sorry for the delay: it turned out to be a super easy & even backwards compatible
fix in frex: simply drop `Data` from `Data.List.reverse`!

Issue #1891, opened 2021-09-01
Potential Regression in Coverage Checker


jfdm:
Consider the code [in this gist](https://gist.github.com/jfdm/713eb26592b05a074d4de63f87f07ed7).

This code used to type-check, presenting a simple proof-of-progress on projection on tuples.
Now, however, the coverage checker tells me that the code is not covering.

At somepoint in time Idris has changed its mind. So who is correct here: me or Idris?

# Steps to Reproduce

Compile it.

# Expected Behavior

The in-progress proof-of-progress should pass the coverage checker for the hole to be filled in.
The old equation is presented.

# Observed Behavior

The in-progress proof-of-progress presented complains about missing patterns, AFAIK there are no missing patterns.
Adding in a place-holder pattern satisfies the coverage checker, allows the proof to be given albeit with the same hole that needs filling.

ohad:
Thanks!

Can you see a way to minimise it further and still get the error?

jfdm:
I am trying to too...

gallais:
Can't you have a `Proj` too in that case split?

jfdm:
RIght, I am not saying that the term should be be impossible to construct but I am reducing the terms and when expanding proof that tuple is a value then it should be fully reduced.

gallais:
Then don't you need to enforce that in the type, introducing a notion of normal form?

jfdm:
Yes, you had me confused there for a second. I have defined the value form, the issue is that when case splitting on `Done tuple` it *should* rule out the other reducible patterns, but it is not.

jfdm:
So the problem is that when I last tried this, Idris was a bit more loose in what it would accept. It is now a bit more strict, and the computation in the type of `Proj` (aka the 'green slime') can now confuse the checker. 

The problem can be addressed by moving (fording) the computation from being 'within the type' to being 'along side the type'.
We do this by using a secondary datatype that synthesizes the required type.

So given an 'index' type that captures the indexing computation:

```idris
public export
data Index : (idx : Fin n) -> (Vect n a) -> a -> Type where
    Here  : Index    FZ  (x::xs) x

    There : Index     n      xs  x
         -> Index (FS n) (y::xs) x
````
we can redefine the `Proj` constructor as follows.

```idris
Proj : {types : Vect (S n) Ty}
    -> (tuple : Tupled (TyTuple types))
    -> (idx   : Index i types ty)
             -> Tupled ty
```

[...]

PR #1892, opened 2021-09-01
[ base ] Indexing Vectors.

jfdm:
`Elem` proofs give existential quantification that a given element *is* in the list, but not where in the list it is!
Here we give a predicate that presents proof that a given index of a vector, given by a `Fin` instance, will return a certain element.

Two decidable decision procedures are presented:

1. Check that a given index points to the provided element in the presented vector.
2. Find the element at the given index in the presented vector.

Prompted by #1891.

### DIFF
--- a/libs/base/Data/Vect/AtIndex.idr
+++ b/libs/base/Data/Vect/AtIndex.idr
@@ -11,7 +11,7 @@
 |||    presented vector.
 |||
 ||| 2. Find the element at the given index in the presented vector.
-module Data.Vect.Index
+module Data.Vect.AtIndex
 
 import Data.Vect
 import Decidable.Equality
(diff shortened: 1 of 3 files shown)


jfdm:
I am partial to the name `Index`, but I won't let that block this PR.

But more importantly, it looks like there are a few naming conventions that we might need to write down/decided upon some time some where...

mattpolzin:
I had a similar thought to Guillaume regarding aligning these new definitions to existing ones but my thought was related to `Elem` directly. I think it would be nice if we had `Elem x xs` and `Index idx x xs` rather than the latter swapping the order of its x/xs arguments. Or even `Index x idx xs` would be a bit more similar to the `Elem` structure.

jfdm:
The ordering is purposeful: The most commonly synthesized value in the definition should be the last parameter in the type. This plays nicely with Dependent Pair definitions.

mattpolzin:
> The most commonly synthesized value in the definition should be the last parameter in the type. 

I admit I may have the weaker argument here. It's just unfortunate ergonomics. Users will need to remember that `List.AtIndex` and `Vect.AtIndex` swap the order of their indices around.

I just noticed `List.AtIndex` doesn't do either of the things I suggested above but it does put the index last which means x/xs are in the same order as with `Elem` so it actually does accomplish what I was picturing regarding familiarity for anyone who has used `Elem`.

jfdm:
@mattpolzin To elaborate on the ordering of type-level parameters, which is really about eta reducing the promoted value in a dependent pair.

I wrote `AtIndex` to primarily represent the relation:

> Given a vector `xs` and a valid index `i` in `xs` then there is an `x` such that `index i xs === x`.

Here, `AtIndex` represents the relation: `index i xs === x`. 

Which is what you want to do most of the time. 

With how things go in programming, however, we can reuse `AtIndex` to represent variants of this relation.

1. Given a vector `xs` and a valid index `i` in `xs` then there is an `x` such that `index i xs === x`.
2. Given a vector `xs`, a valid index `i` in `xs`, and a value `x`, then is `x` the value at `index i xs`.
3. Given a vector `xs` and a value `x` then if `x` is in `xs` then it will be at index `i`.

My opinion is that `i` and `xs` are related through the length of the vector (encoded at the type-level) you are more than likely to have these two and then want to get the value at `i`. Thus if we order the parameters correctly we can take advantage of how Dependent Pairs are actually specified and eta reduce the `x` in the type.

That is having:

```idris
index : DecEq type => (idx : Fin n) -> (xs  : Vect n type) -> Dec (DPair type (AtIndex idx xs))
```

instead of

```idris
index : DecEq type => (idx : Fin n) -> (xs  : Vect n type) -> Dec (x ** AtIndex idx xs x)
```

Although another way is to have a custom datatype that negates the need for a dependent-pair.
Regardless of that, everyone is right ordering (nee ergonomics) *does* matter but it depends on what you want to represent primarily with the relation.

mattpolzin:
@jfdm thanks! That explanation was great. I see your point and definitely agree you can’t optimize for all use cases all the time!

Issue #1013, opened 2021-02-02
Memoized Lazy Types


y0105w49:
Currently, lazy types in Idris aren't memoized.  That is, if we pass a value `x : Lazy a` to several different functions, and they all depend on the forced/computed value of type `a`, they will each trigger a separate computation of that value.  The reason for using laziness at all is to avoid unnecessary computations of `x`, but this can sometimes cause the opposite effect.  We can contrast this with GHC's laziness for Haskell, where thunks are computed at most once.

A possible response is to never use laziness when we're concerned about possibly duplicating computations.  However, I think there is a reasonable case for laziness+memoization in Idris.  A few examples for motivation:

- Streams / infinite data structures are often convenient to use, but if computing individual entries is expensive, this becomes incompatible with uses that might require accessing different locations repeatedly.  More concretely, we could have a stream of integer values, each a somewhat expensive-to-compute function (e.g., the hash?) of the previous, and would want to allow access to arbitrary elements.  We would want to compute each entry at most once, and only if required.  Without memoized laziness, this could be done by explicitly storing a list of elements we've computed so far along with a lazy stream of the remaining elements, but this wouldn't allow us to share the stream by passing it to different functions (we would need to return a modified version of it from each function, with possibly some extra terms computed, or wrap everything with some sort of state, to avoid repeated computations).  This also prevents us from passing the same stream object to different data structures (their accesses couldn't share work).

- Implementations of amortized functional data structures becomes possible.  Chris Okasaki's thesis (https://www.cs.cmu.edu/~rwh/theses/okasaki.pdf) and later papers give several functional data structures that require memoization+laziness to be efficient.  Briefly, amortization is a tool that allows us to give guarantees on the total cost of a several operations on a data structure, even if single operations can sometimes be very expensive, by way of a potential argument (the expensive operations are "paid for" by a corresponding decrease in the structure's potential, that guarantees that not too many can occur in a row).  Pure functional data structures seem incompatible with this: if we have a copy of a data structure just prior to applying an expensive operation, we might repeatedly run that operation on the still-available previous state.  However, memoization allows us to share this work, and for the reduction in potential to make subsequent calls cheaper even on other copies of the original state.  We can then do cool things like rolling-back to previous points in history for (basically) free, even with amortization.

Here's a contrived example that shows the performance difference:

```idris
-- definition copied from contrib for convenience:
data LazyList : Type -> Type where
  Nil : LazyList a
[...]

PR #1899, opened 2021-09-04
[ performance ] Memoise toplevel constants

stefan-hoeck:
As discussed in #1869, this PR converts zero argument toplevel lamdas to lazily evaluated, memoised constants on scheme backends. Now that we have `blodwen-lazy` in the scheme support files, we can also easily start experimenting with memoising `Lazy` and `Inf` as requested in #1013. But that's for another PR.


stefan-hoeck:
Any idea, which the racket bootstrap test fails? It says it couldn't find the network package...

buzden:
> this PR converts zero argument toplevel lamdas to lazily evaluated, memoised constants on scheme backends

That's great! Will local constants (i.e. those `let`s and `where`s) also be memoised? I mean situations like
```idris
f : SomeType -> SomeAnotherType
f x = complexExpression arg (f arg) where
  arg : AType
  arg = someExpression
```

stefan-hoeck:
> That's great! Will local constants (i.e. those `let`s and `where`s) also be memoised?

I'll have to check how these constructs are handled at the backends. But I agree that this would be desirable. Depending on the complexity involved with this such a change may have to go into another PR, though.

But first I have to fix that failing racket test (my racket implementation of `blodwen-lazy` is faulty...)

stefan-hoeck:
So, I performed some experiments with memoizing lazy values and values bound in a `let` or `where` block. I'm using the following [branch](https://github.com/stefan-hoeck/Idris2/tree/memo_lazy) for this.

In the following  example, `n1`, `n2` and `n3` in `calc` are lazily evaluated and memoized. The generated Javascrip is listed below (the generated scheme behaves identically but is less pretty :-) )

```idris
import Debug.Trace

calc : Nat -> Nat
calc n0 =
  let n1  = delay $ (trace "foo" $ n0 + n0)
      n2  = delay $ (trace "bar" $ n1 + n1) 
      n3  = delay $ (trace "baz" $ n2 + n2)
   in n2 + n2

main : IO ()
main = do printLn (calc 1)
```

```javascript
function Main_calc($0) {
 const $1 = Debug_Trace_trace('foo', Prelude_Types_x2b_Num_Nat($0, $0));
 const $8 = Debug_Trace_trace('bar', Prelude_Types_x2b_Num_Nat($1, $1));
 return Prelude_Types_x2b_Num_Nat($8, $8);
}
```

As expected, "foo" and "bar" are printed only once when running this on chez or node. "baz" is not printed at all.

However, if I make the following change to the code (which is the same as introducing a `where` block), `n1`, `n2`, and `n3` are introduced as new toplevel functions and are no longer memoized:

```idris
import Debug.Trace

calc : Nat -> Nat
calc n0 =
  let n1 : Nat
      n1  = delay $ (trace "foo" $ n0 + n0)

      n2 : Nat
      n2  = delay $ (trace "bar" $ n1 + n1)

      n3 : Nat
      n3  = delay $ (trace "baz" $ n2 + n2)
   in n2 + n2

main : IO ()
main = do printLn (calc 1)
```

```javascript
function Main_n__2429_385_n2($0) {
 return Debug_Trace_trace('bar', Prelude_Types_x2b_Num_Nat(Main_n__2429_384_n1($0), Main_n__2429_384_n1($0)));
}

function Main_n__2429_384_n1($0) {
 return Debug_Trace_trace('foo', Prelude_Types_x2b_Num_Nat($0, $0));
}

function Main_calc($0) {
 return Prelude_Types_x2b_Num_Nat(Main_n__2429_385_n2($0), Main_n__2429_385_n2($0));
}
```

These are of course no longer memoized, and "foo" is printed four times. So, we need to prevent Idris2 from introducing toplevel functions for zero-argument functions in `where` blocks, and we'll get memoization for free. Whether these constants will then be eagerly evaluated or not depends on whether we wrap them in a `Lazy` or not (which I'm not sure is working properly when I look at the generated Javascript above...)

Z-snails:
I'm happy to merge this now, unless you want to experiment some more with lazily evaluating `let` and `where`?

Although I think that feature would be better as a separate PR.

PR #1901, opened 2021-09-04
nix flake: make it more idiomatic

teto:
While looking at https://github.com/ReplicaTest/REPLica/pull/62, I looked at this flake (I am no idris developer, yet) and think the flake could be more idiomatic/easier to use it if allowed to  pass attributes directly to `buildIdris` instead of the
convoluted `buildIdris (...).build.overrideAttrs(oldAttrs: {})`.


claymager:
Looks good to me!

PR #1902, opened 2021-09-04
Speculative normalisation

edwinb:
In case types and hole types, often normalising them is a good idea, because it makes the types much smaller. However, we can't tell in advance if this is what's going to happen. What we can do, though, is speculatively normalise and give up if it looks like a bad idea. This can be a big win in some cases where there's lots of type level computation needed, and because we give up pretty early, it doesn't do any noticeable harm in other cases.

Also involves a bit of reorganisation of Normalise.idr, which had got very big, so now the evaluation, quotation (readback) and conversion checks are in separate files.

There's a bit more normalisation related stuff I need to do, because we normalise error messages too early and this can be needlessly expensive when removing ambiguity. I'll probably leave that for a separate patch.


edwinb:
Of course it helps if I also add the new files. The size of the diff should have given me a clue...

diakopter:
oops, not on the main branch

Issue #1396, opened 2021-05-10
Slow error propagation in ambiguous elaborations


Russoul:
# Steps to Reproduce
```
namespace A
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace B
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace C
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace D
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace E
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace F
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace G
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace H
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace I
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace J
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat
namespace K
  export
  (>>) : Monad m => m () -> Lazy (m b) -> m Nat

%ambiguity_depth 100
program : IO ()
program = do
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
  putStrLn "hi"
[...]

PR #1907, opened 2021-09-06
Abandon ambiguity resolution on undefined name

edwinb:
This has finally annoyed me enough to do something about it. If we get a "no such variable" error there's no point exploring other branches.


Russoul:
I think it's safe to say that this closes #1396

PR #1908, opened 2021-09-07
Add function that checks whether a file handle points to a TTY device.

mattpolzin:
This feels a bit more robust (and subtly different) than checking whether the `$TERM` environment variable is set to "dumb". It also theoretically opens the door for an app to pipe data to non-`stdout` TTY handles that it opens.

I also swapped out uses of `fileno()` for `_fileno()` when in a Windows environment because the former is a deprecated alias of the latter.

If this PR gets merged, I will submit a (good first issue) follow-up ticket that after the next Idris 2 release we can start using this function either in addition to- or instead of checking for "dumb" terminals when determining if colors are supported in the Idris 2 compiler source code.

### DIFF
--- a/support/c/idris_file.c
+++ b/support/c/idris_file.c
@@ -30,6 +30,14 @@ void idris2_closeFile(FILE* f) {
     IDRIS2_VERIFY(fclose(f) == 0, "fclose failed: %s", strerror(errno));
 }
 
+int idris2_getFileNo(FILE* f) {
+#ifdef _WIN32
+    return win32_getFileNo(f);
+#else
+    return fileno(f);
+#endif
+}
+
 int idris2_fileError(FILE* f) {
     return ferror(f);
 }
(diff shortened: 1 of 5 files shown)


support/c/idris_file.c
```diff
@@ -61,7 +61,11 @@ int idris2_removeFile(const char *filename) {
 }
 
 int idris2_fileSize(FILE* f) {
+#ifdef _WIN32
+    int fd = win32_getFileNo(f);
+#else
     int fd = fileno(f);
+#endif
```
gallais, line 68: Rather than repeating this code snippet multiple times, shouldn't we pull it out into
its own function?

mattpolzin: Yeah. I agree. One of those things that bugged me but not quite enough to fix as I wrote the code. I’ll abstract into a helper function that chooses between the two variants.

mattpolzin: Part of what stopped me initially is the thought that the whole structure of the windows specific functions begs for abstraction where there is currently duplication, but what I’ve done here is a particularly large amount of duplication. Good call out.

PR #1919, opened 2021-09-12
Thread the option of silencing output to stdout

Russoul:
The change helps to avoid the most commonly found dumps of elaboration error messages to stdout when in
quiet mode. That in turn keeps us safe from critical LSP failures due to unexpected writes to stdout.

### DIFF
--- a/src/Idris/REPL/Common.idr
+++ b/src/Idris/REPL/Common.idr
@@ -29,7 +29,10 @@ import System.File
 
 %default covering
 
--- Output informational messages, unless quiet flag is set
+||| Output informational messages, unless suppressed by a flag.
+||| This function should only be called with informational
+||| messages, an unhandled error is an example of what should
+||| *not* end up here.
 export
 iputStrLn : {auto c : Ref Ctxt Defs} ->
             {auto o : Ref ROpts REPLOpts} ->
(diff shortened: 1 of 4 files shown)


src/Idris/REPL/Common.idr
```diff
@@ -50,7 +51,9 @@ printWithStatus : {auto o : Ref ROpts REPLOpts} ->
 printWithStatus render msg
   = do opts <- get ROpts
        case idemode opts of
-         REPL _ => coreLift $ putStrLn !(render msg)
```
mattpolzin, line 53: How about instead of the change you have here deferring to `iputStrLn` (for the `REPL _` case) like you do below in `emitProblem`? That would require an auto implicit ref to Defs here, but maybe it's worth only needing to write `coreLift $ putStrLn !(render msg)` once.

Russoul: I think I actually tried that. But they have different signatures in addition to the `Defs` you mentioned.
Also, the two occurrences of `putStrLn` are far from being the only two in the codebase, unfortunately.
That said, I don't think that's worth the trouble.

Russoul:
A quick update on what's been happening here:
if one runs an idris2 program via the CLI bypassing REPL (that can be done via the `--exec` command-line option),
they don't get any error output at all, because, buried deep in the command-line-option processing, options `--exec` and `--output` entail implicit addition of the `--quiet` option. That's what made the errors be completely silenced in the first commit, breaking tests and all in all resulting in undesirable behaviour.

I removed the addition of `--quiet` to `--exec` and `--output` in the second commit.
That broke some of the tests but the only new line that has to be added to each of them is
`X/X: Building XXX (XXX.idr)`
That line previously was suppressed by the aforementioned implicitly added `--quiet` option.

Hopefully, removing the implicit addition of `--quiet` to `--exec` and `--output` is a sensible solution.

mattpolzin:
I think changing the quietness of --exec will affect or break the non-LSP idris2 mode for vim. I am not sure that's a good reason not to do this, just a consideration. Idris2 mode could be updated, or explicitly deprecated in favor of LSP, but its currently still my daily driver because I have not found the motivation to set the LSP server up yet.

Russoul:
@mattpolzin I guess that needs more testing then, IDE mode and emacs interoperability included.

mattpolzin:
I can give this PR a go with the idris2-vim since that's what I use. At a glance it appears to use --client actually so maybe I was misremembering.

Russoul:
@mattpolzin Thank you! Sure.
I think idris2-vim doesn't use IDE mode indeed.
But emacs surely does

Russoul:
The last commit fully discards everything done previously.
That should be a good solution: editor interoperability stays 100% unchanged. Same applies to tests.

PR #1920, opened 2021-09-13
[ golden ] Allow test directory names to contain spaces

buzden:


mattpolzin (approved):
Nice.

gallais:
I suppose the principled solution will eventually be to use a proper filepath
library. Because I'm guessing that if the path contains a `"`, that's going to
be a problem here.

buzden:
> I suppose the principled solution will eventually be to use a proper filepath
> library. Because I'm guessing that if the path contains a `"`, that's going to
> be a problem here.

I fully agree. And this is not the only place where `"` can make things do wrong: there are some more simple `"`-surroundings already.

PR #1921, opened 2021-09-13
Add chez-scheme to chez search path

shenlebantongying:
According to repology, there are a few systems using `chez-scheme` as chez's name (including Alpine, Guix, Void, ASOC, Freebsd-ports, MacPorts, nix...)

https://repology.org/project/chez-scheme/versions

This small addition can eliminate time-wasting for those users. They will no longer have to change chez name or delegate with corresponding packagers.

### DIFF
--- a/libs/test/Test/Golden.idr
+++ b/libs/test/Test/Golden.idr
@@ -331,7 +331,7 @@ checkRequirement req
   where
     requirement : Requirement -> (String, List String)
     requirement C = ("CC", ["cc"])
-    requirement Chez = ("CHEZ", ["chez", "chezscheme9.5", "chezscheme", "scheme"])
+    requirement Chez = ("CHEZ", ["chez", "chezscheme9.5", "chezscheme", "chez-scheme", "scheme"])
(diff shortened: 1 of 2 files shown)


shenlebantongying:
They use `--installschemename=chez-scheme`, examples:

Void https://github.com/void-linux/void-packages/blob/63f6957acb0b835f296982518ab37773fd5563cc/srcpkgs/chez-scheme/template#L54-L56

Freebsd https://www.freshports.org/lang/chez-scheme
image

Guix https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/chez.scm#n181
image

shenlebantongying:
Related issue https://github.com/idris-lang/Idris2/issues/666

The fix at https://github.com/idris-lang/Idris2/pull/1589 didn't include `chez-scheme` as mentioned by https://github.com/idris-lang/Idris2/pull/1589#issuecomment-866098615

mattpolzin (approved):
I'm sold.

PR #1922, opened 2021-09-14
Made unifying error msg nicer.

redfish64:
Made unifying error msg nicer. ex.:

```
Error: While processing right hand side of foo3. When unifying:
    (Int, (String, (String, (Char, (Int, (String, Char))))))
and:
    (Int, (String, (String, (Char, (Int, (Int, Char))))))
Mismatch between: String and Int.
```
vs:
```
Error: While processing right hand side of foo3. When unifying (Int, (String, (String, (Char, (Int, (String, Char)))))) and (Int, (String, (String, (Char, (Int, (Int, Char))))))
Mismatch between: String and Int.
```

Code below:
```
foo : (Int,String,String,Char,Int,Int,Char) -> ()
foo x = ()


foo2 : (Int,String,String,Char,Int,String,Char)
foo2 = (1,"foo","x",'c',5,"fee",'d')

foo3 : ()
foo3 = foo foo2
```

### DIFF
--- a/src/Idris/Error.idr
+++ b/src/Idris/Error.idr
@@ -231,8 +231,9 @@ perror (CyclicMeta fc env n tm)
 perror (WhenUnifying _ gam env x y err)
     = do defs <- get Ctxt
          setCtxt gam
-         let res = errorDesc (reflow "When unifying" <++> code !(pshow env x) <++> "and"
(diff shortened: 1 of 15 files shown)


gallais:
Thanks!

PR #1929, opened 2021-09-16
Retain the --client mode behaviour (defect introduced in #1919)

Russoul:
I haven't been able to test it on idris2-vim yet...

Edit: tested basic commands: it works


gallais:
Ugh, I had only meant to cancel `ubuntu-bootstrap-racket` as it's diverging atm so why did github
stop all the other ones too? Anyway I'm not restarting the builds.

PR #1933, opened 2021-09-19
Remove __collect_safe from bootstrap chez

edwinb:
We don't need this for bootstrapping, and it prevents building on older Chez, so we can just remove it. Also added a note to the release checklist.

### DIFF
--- a/Release/CHECKLIST
+++ b/Release/CHECKLIST
@@ -5,7 +5,7 @@
 [ ] Change version number in flake.nix
 [ ] Change version number in test pkg010 (TODO: make this step unnecessary!)
 [ ] Update bootstrap chez and racket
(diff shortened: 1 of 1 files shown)


Release/CHECKLIST
```diff
@@ -5,6 +5,7 @@
 [ ] Change version number in flake.nix
 [ ] Change version number in test pkg010 (TODO: make this step unnecessary!)
 [ ] Update bootstrap chez and racket
+[ ] Remove __collect_safe fro generated chez (to avoid need for chez >9.5)
```
memoryruins, line 8:
```suggestion
[ ] Remove __collect_safe from generated chez (to avoid need for chez >9.5)
```

PR #1937, opened 2021-09-19
Abandon auto search on UndefinedName

edwinb:
These might arise from names in other modules that haven't been imported. But it's going to be an error whatever, so give up straight away.


edwinb:
Oh that's exciting, this appears to be more complicated than I expected...

Issue #1935, opened 2021-09-19
idris2-0.5.0 breaks compatibility with busybox sha256sum


nmeum:
Hello, I maintain a idris2 package for [Alpine Linux](https://alpinelinux.org). By default, Alpine uses a Busybox instead of GNU userland. Busybox provides several of the utilities provided by GNU coreutils. However, while it aims for POSIX compatibility it does not support all kinds of GNU extensions. idris2-0.4.0 used to work fine on Alpine, however, since 9b8de95a38c3dc501aa5ad6f46c4aeb7e57a83aa (which is included with idris2-0.5.0) the `--tag` option is used with `sha256sum`. While busybox has a `sha256sum` utility, this utility does not support the `--tag` option thus requiring installation of GNU coreutils which is a bit unfortunate.

It would be possible to add busybox to the following case distinction:

https://github.com/idris-lang/Idris2/blob/257783275ee6b4ee2e8465fffe8daf94a5679300/src/Core/Options.idr#L240-L249

however, this would require checking `busybox sha256sum` before checking for `sha256sum`, i.e. cause the busybox applet to take precedence over `sha256sum` if both are installed. Furthermore, this would require fiddling with the parsing code again as the parser currently always uses the last word from the `sha256sum` output and since busybox does not support `--tag` the checksum is always output as the first word:

https://github.com/idris-lang/Idris2/blob/9865765d1dfc0a065804dcfe9c8d572664c44f5b/src/Core/Binary/Prims.idr#L421-L423

Alternatively, it might make sense to check for openssl first, this would allow making idris depend on openssl on Alpine instead of making idris depend on the entire GNU coreutils toolchain?

edwinb:
Given that this hash function is only used if an experimental option is set, I don't think we should be failing if the executable isn't found, but rather that experimental option shouldn't be available. I'll change its behaviour. I'm about to do a patch release anyway, for a different reason, so this change might as well be included in it.

edwinb:
Oh, it's worse than that isn't it though, because just knowing the utility is there isn't enough, we also need to know how it works and what options it takes. I'm inclined to disable this option completely for the sake of a working portable release - we can come back to it and resolve the details after that.

edwinb:
This should work now, since we no longer run the executable. We'll have to find another way to work out source hashes if we need it.

PR #1940, opened 2021-09-19
Disable -Xcheck-hashes, at least for the moment

edwinb:
If we're going to have this as an option, we need to have a portable way of finding a sha256sum command. At the moment, we might find a command, but different versions accept different options. We should at least allow setting it via an environment variable, and we certainly shouldn't fail if running the command fails.

This solves #1935 in that at least we no longer cause build failures on systems without sha256sum or with a version that runs differently.


edwinb:
I suppose one possible approach would be to try running the command at start up, and only set it if we successfully generate an sha256sum for some test input.

I'm still not entirely convinced of the value of checking hashes of source files, and if it's worth going to the effort of checking another external dependency and more environment variables. If it's about faster file loading in interactive editing, I have a scheme in mind for incremental reloading - although even that seems unrelated since in those situations the file has changed.

In any case, since this is experimental, I'm going to disable it for the 0.5.1 patch release. We can put it back in for the next release when we work out how to do it portably.

Issue #1079, opened 2021-02-17
Unreachability warning is not fired sometimes


buzden:
Consider a pure function that matches its argument and has an unreachable clause:

```idris
g1 : Nat -> Nat
g1 x = 5
g1 _ = 6
```

You expectedly get
```
Warning: Unreachable clause: g1 _

.../UnreachableInDo.idr:3:1--3:4
 1 | g1 : Nat -> Nat
 2 | g1 x = 5
 3 | g1 _ = 6
     ^^^^
```

The same would be if you have some more complicated match:

```idris
g2 : (Nat, Nat) -> Nat
g2 (x, y) = x
g2 (a, b) = 6
```

gives

```
Warning: Unreachable clause: g2 (a, b)

.../UnreachableInDo.idr:7:1--7:10
 3 | g1 _ = 6
 4 |
 5 | g2 : (Nat, Nat) -> Nat
 6 | g2 (x, y) = x
 7 | g2 (a, b) = 6
     ^^^^^^^^^
```

The same in match inside `do`:

```idris
f : Monad m => m (Nat, Nat)

h2 : Monad m => m Nat
h2 = do
  (x, y) <- f
    | (a, b) => pure 5
  pure x
```

gives

```
Warning: Unreachable clause: case (a, b) of
                             {(x, y) => pure x;
                             (a, b) => pure (fromInteger 5)}

.../UnreachableInDo.idr:18:7--18:23
 14 |
 15 | h2 : Monad m => m Nat
 16 | h2 = do
 17 |   (x, y) <- f
 18 |     | (a, b) => pure 5
            ^^^^^^^^^^^^^^^^
```

but...

# Steps to Reproduce

Try unreachably match `_` after non-trivial match, both in pure functions and inside `do`:

```idris
g3 : (Nat, Nat) -> Nat
g3 (x, y) = x
g3 _ = 6
```

and

```idris
h3 : Monad m => m Nat
h3 = do
  (x, y) <- f
    | _ => pure 5
  pure x
```

# Expected Behavior

The warnings of unreachable clauses similar to those above.

# Observed Behavior

No warnings.

gallais:
I believe this is a duplicate for #550

buzden:
Well, this is not obvious, since this issue talks about unreachability warning (i.e. about lack of warning that some case is unreachable) but #550 talks about the case when compiler considers some case to be uncovered when they are.

gallais:
My bad.

Issue #1917, opened 2021-09-10
Replace `ENV` var check being used to disable colors with `TTY` check.


mattpolzin:
https://github.com/idris-lang/Idris2/pull/1908 introduced a `base` library function `System.File.Meta.isTTY/1` that will check whether a file handle (like `stdout`) is a terminal. Although not a perfect proxy, checking for TTY is often used to determine if the receiving file handle will support ANSI control sequences like those used to send colored text to a terminal.

Currently, the Idris 2 compiler checks the `TERM` environment variable to see if it has the value `"dumb"` and if so then it turns color output off: https://github.com/idris-lang/Idris2/blob/9865765d1dfc0a065804dcfe9c8d572664c44f5b/src/Idris/Pretty/Render.idr#L33

Checking the `TERM` environment variable is an established way of determining color support as well, but checking it for equality with `"dumb"` is a bit fragile. The code comments suggest the primary motivator for the check is emacs, which as best I can tell (without being an emacs user myself) is also _not_ going to register as a TTY device. That means that the TTY check will support existing use cases and additional situations as well.

For example, right now Idris 2 is happy to output ANSI color codes to the following Mac OS pasteboard copy command; paste the resulting text into a text editor and you'll get gibberish! A TTY check will catch that case and avoid sending ANSI codes to `pbcopy`:
```shell
idris2 --check Main.idr | pbcopy
```

gallais:
By the way this `TERM` check is also used in the `Test.Golden` file.
(IIRC there's a bug too because we check for `"DUMB"` rather than `"dumb"`).

PR #1942, opened 2021-09-21
Warn about unreachable default clauses

mattpolzin:
Closes https://github.com/idris-lang/Idris2/issues/1079.

This PR adds warnings when a default clause of a function is unreachable due to all possible patterns covered by that default clause being handled by prior clauses.

I did that by borrowing logic from the code that checks for coverage and applying it to the case tree builder. It occurred to me as I worked through this to try to find a way to combine things such that less crawling over the same case tree is needed, but my initial stab at that meant both that the coverage code became more convoluted and at the same time the two places where an "unreachable clause" warning could be generated became very far away from each other in the codebase. The solution in this PR keeps unreachable warnings close together.

Because some of the code previously used only in the coverage code was more broadly useful now, I moved `CaseTree` and `CaseBuilder` into a `Case` module and introduced the `Case.Util` module with some utilities now useful in multiple modules.

~I bumped the previous Idris version in CI to 0.5.1 so I could use the list-difference operator that was added to `base` with version 0.5.1; this also will unblock tickets like https://github.com/idris-lang/Idris2/issues/1917.~

[EDIT] CI was updated by a different now-merged PR.


buzden:
> I bumped the previous Idris version in CI to 0.5.1

As far as I know the convention is that the current Idris must be checked that it's built by the previous minor version, i.e. it is 0.4.0 for the moment. It means that you shouldn't use anything from the 0.5.x until 0.6.x is released.

buzden:
Sad fact: the warnings from this PR are not shown once the file is typechecked (i.e., fresh TTC files are present). It means, there are no warnings on the file reload (if no changes) and, more importantly, there are no warnings even after removing the `build/` dir once you open the file in an editor that uses LSP with idris2-lsp!

mattpolzin:
> As far as I know the convention is that the current Idris must be checked that it's built by the previous minor version, i.e. it is 0.4.0 for the moment. It means that you shouldn't use anything from the 0.5.x until 0.6.x is released.

This is a mostly correct statement, but the previous version is the most recently shipped release. Throughout most of the time after Idris2 v0.4.0 was released and before Idris2 v0.5.0 was released, CI pointed to Idris2 v0.4.0 as the previous version. Now, post v0.5.x release, it can be bumped to v0.5.1. It does not _need_ to get bumped right away, but I believe that not duplicating `base` library functions that became available with v0.5.0 is a good enough reason to bump the CI version. By the time Idris2 v.0.6.0 (hypothetical next version) is released, the requirement to build it will be Idris2 v0.5.x, so using that version in CI covers the most important thing: building from the previously released version is working.

mattpolzin:
@buzden interesting observation about unreachable clause warnings only being given on the first build after changes to a module. On one hand, it is kind of nice that the compiler need not do as much work (because if the case tree isn't changing, the warnings aren't changing either), but I wonder if that's a good enough reason to store information such as these warnings in the TTC itself at some point in the future.

buzden:
> Throughout most of the time after Idris2 v0.4.0 was released and before Idris2 v0.5.0 was released, CI pointed to Idris2 v0.4.0 as the previous version.

Oh, you are right, sorry for the misleading comment. Probably, I was put on the wrong way by the Edwin's comment that v0.5.1 should be still typechecked by v0.4.0, I interpreted it as if all between v0.5.1 and v0.6.0 should be.

mattpolzin:
@madman-bob Would you like me to delete or leave commented-out the unreachable defaults I've currently got commented out in the RefC backend in this PR?

madman-bob:
Please delete any unreachable clauses in RefC. I'm glad to hear they're unreachable - they must have been forgotten about when the final clause was added.

src/Core/Case/CaseBuilder.idr
```diff
@@ -1226,16 +1227,90 @@ simpleCase fc phase fn ty def clauses
          defs <- get Ctxt
          patCompile fc fn phase ty ps def
 
-findReached : CaseTree ns -> List Int
-findReached (Case _ _ _ alts) = concatMap findRAlts alts
+mutual
+  findReachedAlts : CaseAlt ns' -> List Int
```
gallais, line 1231: Should this be producing a set instead? Especially given you seem to `nub` the list in the end.

mattpolzin: This particular function's results are not always nubbed; it's important that reached indices are duplicated where they are used in the list difference below. 

That said, I actually didn't notice we had a set implementation in the compiler libraries (only looked at base), and I _can_ use a sorted set in the implementation of the rep/droprep helpers below so I'll do that. The result of those helpers is indeed always nubbed.

src/Core/Case/Util.idr
```diff
@@ -0,0 +1,62 @@
...
+||| type family, with their type, name, tag, and arity.
+export
+getCons : {auto c : Ref Ctxt Defs} ->
+          {vars : _} ->
+          Defs -> NF vars -> Core (List (NF [], Name, Int, Nat))
+getCons defs (NTCon _ tn _ _ _)
+    = case !(lookupDefExact tn (gamma defs)) of
+           Just (TCon _ _ _ _ _ _ cons _) =>
+                do cs' <- traverse addTy cons
+                   pure (mapMaybe id cs')
```
gallais, line 19:
```suggestion
                   pure (catMaybes cs')
```

src/Core/Case/Util.idr
```diff
@@ -0,0 +1,62 @@
...
+import Core.Env
+import Core.Normalise
+import Core.Value
+
+||| Given a normalised type, get all the possible constructors for that
+||| type family, with their type, name, tag, and arity.
+export
+getCons : {auto c : Ref Ctxt Defs} ->
+          {vars : _} ->
+          Defs -> NF vars -> Core (List (NF [], Name, Int, Nat))
```
gallais, line 14: Can we introduce a `Constructor` record that gives suggestive field names for
these `(NF [], Name, Int, Nat)` values? I'm worried of the over-abundance of
cryptic tuples in the compiler.

src/Core/Case/Util.idr
```diff
@@ -0,0 +1,62 @@
...
+tagIs t (ConCase _ t' _ _) = t == t'
+tagIs t (ConstCase _ _) = False
+tagIs t (DelayCase _ _ _) = False
+tagIs t (DefaultCase _) = True
+
+-- Need this to get a NF from a Term; the names are free in any case
+export
+freeEnv : FC -> (vs : List Name) -> Env Term vs
+freeEnv fc [] = []
+freeEnv fc (n :: ns) = PVar fc top Explicit (Erased fc False) :: freeEnv fc ns
```
gallais, line 61: Is this `Core.Env`'s `mkEnv`?

gallais (commented):
Found a couple of opportunities to tidy the code that is being moved around.

src/Core/Case/Util.idr
mattpolzin: I like this idea a lot. I've also felt encumbered by needing to decipher tuples in parts of the compiler code before.

src/Core/Case/Util.idr
mattpolzin: Identical! I've removed `freeEnv`.

src/Core/Case/Util.idr
mattpolzin: Done. And in the process I discovered that only 3 of the 4 elements of the tuple were needed.

PR #1947, opened 2021-09-22
[ cleanup ] Remove unreachable match clauses

buzden:
#1942 shows that there are some unreachable clauses. Interestingly there was a non-trivial one in the `Unelab` module, but the `_` clause is really unreachable, I checked this by hand.

This must be a backward-compatible change, independent of #1942.


gallais:
Why was this closed?

buzden:
> Why was this closed?

Because before making this PR I didn't check that the original #1942 already contains these changes (perhaps, except for removing unneeded comments in `RefC`).

mattpolzin:
> except for removing unneeded comments in `RefC`

I was unsure whether or not to leave those comments around; I probably should ping Robert instead of playing it safe. My thinking in leaving them was actually that if someone else in the future saw a reason to leave those functions as ostensibly non-covering again then they would know there was precedent for doing so.

PR #1948, opened 2021-09-22
[ fix ] Use qualified type name for record elaboration (#1871)

### LINKED ISSUE #1871: Ambiguity with duplicate record names in different namespaces
If we try to use the same field name and the same record name in two different namespaces, it crashes at compile time when the compiler try to generate the dotted field accessors.

# Steps to Reproduce

Compile: 

```
module Test

namespace A

  public export
  record Dup where

namespace B
  
  public export
  record Dup where
    constructor MkDup
    label : Nat
```

# Expected Behavior

Compilation should pass.

# Observed Behavior

Error:

```
Error: While processing type of .label. Ambiguous elaboration. Possible results:
    Test.B.Dup
    Test.A.Dup
```

madman-bob:
Allow records with the same name in adjacent namespaces. This fixes #1871.

The dotted field getters were using the unelaborated name of the record, potentially resulting in an ambiguous definition. This PR uses the qualified name, removing that ambiguity, and adds a test for the desired behaviour.


gallais (commented):
Looks very sensible to me.

Now that many definitions take 2 `Name`s, I would however like to have at
least descriptive variable names in the type signature too. Comments would
be even nicer.

PR #1949, opened 2021-09-23
[ performance ] CSE in linear space

stefan-hoeck:
The current CSE algorithm stores all closed sub-terms in their unmodified form in a `SortedMap`. This requires O(n^2) space w.r.t. term size and can lead to drastic amounts of memory consumption during CSE as mentioned on [discord](https://discord.com/channels/827106007712661524/827109190266257409/889134639480991764). @Z-snails had the idea to start the CSE analysis at the leaves, in which case it should be possible for the algorithm to consume a linear amount of space w.r.t. term size.

This PR adjusts the algorithm in such a way.

### DIFF
--- a/src/Compiler/Opts/CSE.idr
+++ b/src/Compiler/Opts/CSE.idr
@@ -44,38 +44,17 @@ public export
 UsageMap : Type
 UsageMap = SortedMap (Integer, CExp []) (Name, Integer)
 
---------------------------------------------------------------------------------
---          Size of Expressions
---------------------------------------------------------------------------------
+public export
+data Count = One | C Integer
 
-mutual
-  -- comparison for `Nat` is (or used to be?) O(n),
-  -- therefore, we use `Integer`
-  size : CExp ns -> Integer
-  size (CLocal _ _)         = 1
-  size (CRef _ _)           = 1
-  size (CLam _ _ y)         = size y + 1
-  size (CLet _ _ _ y z)     = size y + size z + 1
-  size (CApp _ _ xs)        = sum (map size xs) + 1
-  size (CCon _ _ _ _ xs)    = sum (map size xs) + 1
-  size (COp _ _ xs)         = sum (map size xs) + 1
-  size (CExtPrim _ _ xs)    = sum (map size xs) + 1
-  size (CForce _ _ y)       = size y + 1
-  size (CDelay _ _ y)       = size y + 1
-  size (CPrimVal _ _)       = 1
-  size (CErased _)          = 1
-  size (CCrash _ _)         = 1
-  size (CConCase _ sc xs x) =
-    size sc + sum (map sizeConAlt xs) + sum (map size x) + 1
-
-  size (CConstCase _ sc xs x) =
-    size sc + sum (map sizeConstAlt xs) + sum (map size x) + 1
-
-  sizeConAlt : CConAlt ns -> Integer
-  sizeConAlt (MkConAlt _ _ _ _ z) = 1 + size z
-
-  sizeConstAlt : CConstAlt ns -> Integer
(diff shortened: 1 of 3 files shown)


src/Compiler/Opts/CSE.idr
```diff
@@ -199,159 +198,291 @@ mutual
...
-analyzeDef : Ref Sts St => CDef -> Core ()
-analyzeDef (MkFun args x)    = analyze x
-analyzeDef (MkCon _ _ _)     = pure ()
-analyzeDef (MkForeign _ _ _) = pure ()
-analyzeDef (MkError _)       = pure ()
-
-analyzeName : Ref Sts St => Ref Ctxt Defs => Name -> Core ()
+  analyze exp = do
+    (sze, exp') <- analyzeSubExp exp
+    case dropEnv {pre = []} exp' of
```
gallais, line 203: Isn't this still exponential? You've already analysed the sub expressions which
means you've already attempted to run `dopEnv` on them, doesn't it?

stefan-hoeck: Yes, that's why I wrote "linear space" not "linear time", but I think it is polymorphic (O(n^2)) not exponential? However, I will give this some more thought, since we might be able to make use of the info from subterm analysis and be able to get rid of the call to `dropEnv` on parent terms.

The problem with quadratic space was that it bites us even in the face of large nested closed terms. The quadratic time above only bites us in the face of large nested open terms, since closed child terms will be replace by machine generated names before the call to `dropEnv`.

Z-snails: I think it could be linear in both if `analyze` returned the subterm with free variables removed as it goes.
This has the disadvantage of having a more complex function, but I think it would be worth it, seeing how much CSE helps with executable size.
eg
```idris
analyse : Ref Sts St => {locals : _} -> CExp (locals :: vars) -> Core (Maybe (CExp locals))
analyse (CLocal fc p) = pure $ CLocal fc <$> dropEnv p
analyse (CApp fc fn as) = do
    fn' <- analyse fn
    as' <- traverse analyse as
    pure $ CApp <$> fn' <*> as'
...
```

stefan-hoeck: > I think it could be linear in both if `analyze` returned the subterm with free variables removed as it goes.
> This has the disadvantage of having a more complex function, but I think it would be worth it, seeing how much CSE helps with executable size.

I experimented with this, an I think it is not as simple as it seems from your example. The hard cases are those, which introduce new variables: Let bindings and lambdas. For instance, what do we do in case of `CLam _ n y`? There are three possibilities to consider:

1. `y` is a closed term (`n` makes no appearance in it), in which case it should be subjected to CSE as should the lambda itself, since it will also be a closed term.
2. `y` has only `n` as a free variable, in which case the lambda is still a closed term and should be subjected to CSE but not so `y`.
3. `y` has additional free variables, in which case neither  `y` nor the lambda should be considered during CSE.

So, in order to stick to the linear algorithm, we need to consider three possibilities in the return type of `analyze`, and these will start to multiply in the face of recursion. If - however - we invoke `analyze` twice with different lists of name prefixes,
we get exponential complexity in the case of nested lambdas (or let bindings).

The same goes for `let` bindings, which also introduce a new variable. Here we could argue that it's unlikely that the new variable is not used in the subexpression, but lambdas ignoring parts of their arguments are pretty common.

A linear version is not lost, since we could always just return the set of free variables (collected in order of appearance) together with every subterm and inspect those sets at runtime to separate closed terms from open ones (actually, this would be O (n * log n) if we use a `SortedSet` to store the names of free variables). However, I could not yet come up with a nice encoding for this, which also behaves well at compile time. Will have to think some more.

gallais:
This is already a big improvement so I'll merge as is but let's keep it in mind.

Issue #1944, opened 2021-09-21
Make declarations optional for aliases


gallais:
We should be able to write definition of the form `f xs = rhs` (where `xs` are variables
only) without having to declare their type first. This would be particularly handy when
moving definitions from a `let` binder to a `where` clause for instance.

This can be elaborated to `f : ?; f = \ xs => rhs`

Note that it may be easier (as a first issue for instance) to start with just `f = rhs`
and then see later on whether we can get the more general one working.

eayus:
Not sure whether you are already aware, but on the latest main branch this feature does not seem to work with visibility modifiers.
E.g.
```idris
public export
N = Nat
```
Causes a parse error:
```
Error: Couldn't parse declaration.

Syntax:4:1--4:7
 1 | module Syntax
 2 |
 3 |
 4 | public export
     ^^^^^^
```
(Adding an explicit type declaration removes the error)

PR #1952, opened 2021-09-23
[ re #1944 ] Allow toplevel aliases

gallais:
Doesn't seem to work in `where` clauses at the moment.


gallais:
Given #2592 and #2593 and degraded error messages I've observed in some cases,
I am planning to revert this hack. People will have to once again declare functions before
defining them.

PR #1953, opened 2021-09-23
Unify IR dumping functions.

danielkroeni:
Inspired by the following comment in `src/Compiler/Common.idr`, I cleaned it up a little.
> -- Hmm, these dump functions are all very similar aren't they...

### DIFF
--- a/src/Compiler/Common.idr
+++ b/src/Compiler/Common.idr
@@ -204,24 +204,8 @@ natHackNames
        UN (Basic "prim__mul_Integer"),
        NS typesNS (UN $ Basic "prim__integerToNat")]
 
--- Hmm, these dump functions are all very similar aren't they...
-dumpCases : String -> List (Name,FC,NamedDef) -> Core ()
-dumpCases fn defs
-    = do Right () <- coreLift $ writeFile fn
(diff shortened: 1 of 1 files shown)


src/Compiler/Common.idr
```diff
@@ -332,19 +289,19 @@ getCompileData doLazyAnnots phase_in tm_in
          defs <- get Ctxt
          whenJust (dumpcases sopts) $ \ f =>
             do coreLift $ putStrLn $ "Dumping case trees to " ++ f
-               dumpCases f namedDefs
+               dumpIR {def=NamedDef} f (map (\(n, _, def) => (n, def)) namedDefs)
```
gallais, line 292: If you're going to need to take `def` explicitly every time, I would just make it an explicit argument:

```
dumpIR : (0 def : Type) -> Show def => String -> List (Name, def) -> Core ()
```

danielkroeni: Thank you for your comment. Actually I don't need to take `def` explicitly. I was thinking (not too much, to be honest), that maybe it makes it more readable. I added a commit which removes the explicit passing of `dev`.

Issue #1946, opened 2021-09-22
`runElab` can't resolve `check` under `lambda`


berewt:
# Steps to Reproduce

Compile:

```
module Test
import Language.Reflection

%language ElabReflection

itPasses : Integer
itPasses = %runElab (do
  x <- check (IPrimVal EmptyFC (BI 3))
  pure x)

itFails : Integer -> Integer
itFails = %runElab (lambda Integer $ \v => do
  x <- check (IPrimVal EmptyFC (BI 3))
  pure v)
```

# Expected Behavior

Compilation passes.

# Observed Behavior

`itFails`fails... Looking at the log, it seems that the compiler isn't able to resolve `check` in the second case.

For `itPasses`:
```
LOG eval.stuck:5: Stuck function: Language.Reflection.check
LOG eval.ref.data:50: Found data constructor: Language.Reflection.Check
```

while `itFails` can't resolve `check` after:
```
LOG eval.stuck:5: Stuck function: Language.Reflection.check
```

berewt:
I tried to pass `Language.Reflection.Elab` to `public export`. Then, if I use `Check` instead of `check`, it works.
My understanding is that the issue lies somewhere in `Core.Normalise.evalRef`, where the compiler get stuck too early, but I'm not sure how to fix it (yet).

berewt:
Indeed, removing the `redok` check in `Core.Normalise.evalRef` solve this particular issue as well (it may introduce endless loop though (?)).

buzden:
I'm not sure the problem is exactly in `check` inside `lambda` since this one typecheks:

```idris
itFails' : Integer -> Integer
itFails' = %runElab lambda Integer $ \v => do
  check (IPrimVal EmptyFC (BI 3))
```

So, maybe there is some problem is `>>=` under `lambda` or so? There is at least one strange behaviour related to `>>=` in elab, namely #1886.

berewt:
Yea, I think bind plays a role there. And it's probably bind with a non empty context, as if the variable introduced by lambda locked the interpretation of the lhs of bind. But it's weird that ˋpureˋ doesn't suffer from the same limitations in a similar use.

Russoul:
I think this issue has disappeared in the recent Idris builds. This works on my machine.

berewt:
I still have a weird elaboration behivour that seems to be improved by setting `Elab` to `public export` IIRC, but I need to test it and provide a minimal failing example (and ideally, grasp why visibility impact it).
[...]

PR #1955, opened 2021-09-24
Visibility workaround for #1946

berewt:
While it doesn't solve #1946 issue per say, it allows at least the issue examples to pass:
if we publicly export `Language.Reflection.check`, the `reducibleInAll` check of `Core.Normalize.Eval.eval` pass to true and `check` is translated into `Check` in all cases.
I still don't get why we don't need this public export in some cases, but we may figure it out later.

### DIFF
--- a/libs/base/Language/Reflection.idr
+++ b/libs/base/Language/Reflection.idr
@@ -61,26 +61,26 @@ Monad Elab where
   (>>=) = Bind
 
 ||| Report an error in elaboration
-export
+public export
 fail : String -> Elab a
 fail = Fail EmptyFC
 
-export
+public export
 failAt : FC -> String -> Elab a
 failAt = Fail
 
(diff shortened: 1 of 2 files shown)


buzden:
Well, I have a still unreported but similarly worked around elab-related problem. For example, when in my elaboration scripts I use something, that's needed to be imported, say `Data.SortedSet`, I have to do `import public Data.SortedSet` in the module where the elaboration script defined, or I get the "stuck function" problem in seemingly arbitrary places.

berewt:
I still have to take some time to investigate  why `check` behaves differently under a lambda. We probably need a more precise condition to decide whether we should try to evaluate a term or not, but it’s hard (with my current knowledge at least) to understand how and when.

gallais:
Abandoned.

PR #1958, opened 2021-09-28
Adding indexed versions of filter,map,fold functions

choltreppe:
I added versions of map, filter and fold functions which keep track of the index.
I personally need at least the mapI quite often. But if its considered not that important, I could put it in contrib instead

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -514,10 +514,11 @@ foldl1 f xs = foldl1By f id xs
 ||| foldrI (\i => \x => (+) (i * x)) 0 [4,1,2,1]
 public export
 foldrI : (Nat -> elem -> acc -> acc) -> acc -> List elem -> acc
-foldrI f z vect = foldrI' Z vect where
-  foldrI' : Nat -> List elem -> acc
-  foldrI' _ [] = z
-  foldrI' i (x :: xs) = f i x (foldrI' (S i) xs)
+foldrI f z list = let (_, z') = foldrI' list in z' where
+  foldrI' : List elem -> (Nat, acc)
+  foldrI' [] = (Z, z)
+  foldrI' (x :: xs) = let (i, z') = foldrI' xs
+                      in (S i, f i x z')
 
 ||| Foldl with index parameter
 |||
@@ -526,7 +527,7 @@ foldrI f z vect = foldrI' Z vect where
 ||| ```
 public export
 foldlI : (acc -> Nat -> elem -> acc) -> acc -> List elem -> acc
-foldlI f z vect = foldlI' Z z vect where
+foldlI f z list = foldlI' Z z list where
   foldlI' : Nat -> acc -> List elem -> acc
   foldlI' _ z [] = z
   foldlI' i z (x :: xs) = foldlI' (S i) (f z i x) xs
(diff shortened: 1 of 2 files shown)


libs/base/Data/List.idr
```diff
@@ -556,6 +595,33 @@ public export
...
+||| ```idris example
+||| mapI (*) [4,1,2,1]
+||| ```
+export
+mapI : (Nat -> a -> b) -> List a -> List b
+mapI f list = mapI' Z list where
+  mapI' : Nat -> List a -> List b
+  mapI' _ [] = []
+  mapI' i (x::xs) = f i x :: mapI' (S i) xs
+
```
ohad, line 609: See `contrib`'s [`Data.List.Extra](https://github.com/idris-lang/Idris2/blob/a9ccf4db4f69677c5287a1278ab167247dac161d/libs/contrib/Data/List/Extra.idr#L11).

ohad:
Should this be in `base`, or maybe in `contrib`?

libs/base/Data/List.idr
choltreppe: oh didnt see that. should I add all the functions in there and in Data.Vect.Extra and do a new request?

libs/base/Data/Vect.idr
```diff
@@ -361,6 +361,34 @@ mapMaybe f (x::xs) =
        Just y  => (S len ** y :: ys)
        Nothing => (  len **      ys)
 
+||| Map with index parameter
+|||
+||| ```idris example
+||| mapI (*) [4,1,2,1]
+||| ```
+export
+mapI : (Nat -> a -> b) -> Vect len a -> Vect len b
```
buzden, line 370: Why using `Nat` index for `Vect`s? (except for the tail recursion, maybe) I'm sure `Fin len` is much more usable for `Vect` (I personally find `Fin`-based `mapI` even for `List`s in my codebase).

choltreppe: yea I already noticed that. (I'm realy new to idris, maybe not best idea to contribute than) I alredy chnaged alll Vect functions, and put them in the contrib library, but now I have a problem being in HEAD detached in git, I'm just looking up the solution for that

choltreppe:
I also changed the names of the existing functions for map, because I think would be better if they have the same name (for List and Vect). But thats probably not so nice for people already using those functions

withoutK:
> I also changed the names of the existing functions for map, because I think would be better if they have the same name (for List and Vect). But thats probably not so nice for people already using those functions

At least for me (since I had put some of them in mess), it's nice to make names neat and mode consistent! This is some how what `contrib` means to be. :)

madman-bob:
To throw another idea into the ring - Python has the `enumerate` function, which we would think of as a function `List a -> List (Nat, a)`. All these `xxxI` functions are then just compositions of the standard forms with `enumerate`, in some manner. By adding an `enumerate` function, we can then remove all the `xxxI` functions, simplifying things.

As we have a stronger type system than Python, we could also strengthen the type of `enumerate` to use `Fin`, or `Elem`, or something even stronger.

choltreppe:
I couldnt figure out how to make indexMapPos in Vect.Properties.Map work with the enum functions because I honestly dont even realy get what it does. Maybe someone else can take a look at it

PR #1960, opened 2021-09-29
[build] fix install-libs target

attila-lendvai:
Without this change `install-libs` will not install the `.so` files, and consequentially the test `chez/chez033` fails with `Error: INTERNAL ERROR: Missing .so:Prelude/Cast.so` when it is run by an installed executable.

Also add a `test-installed` makefile target to run the tests with the installed executable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,6 +138,9 @@ retest: testenv
 	@echo
 	@${MAKE} -C tests retest only=$(only) IDRIS2=${TARGET} IDRIS2_PREFIX=${TEST_PREFIX}
 
+test-installed:
+	@${MAKE} -C tests testbin      IDRIS2=$(IDRIS2_PREFIX)/bin/idris2 IDRIS2_PREFIX=${IDRIS2_PREFIX}
+	@${MAKE} -C tests only=$(only) IDRIS2=$(IDRIS2_PREFIX)/bin/idris2 IDRIS2_PREFIX=${IDRIS2_PREFIX}
 
 support:
 	@${MAKE} -C support/c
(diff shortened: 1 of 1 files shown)


mattpolzin:
I haven't checked myself yet, but are the incrementally compiled libraries compatible with non-incrementally compiled downstream code?

attila-lendvai:
(i can't answer @mattpolzin 's question)

Makefile
```diff
@@ -139,6 +139,9 @@ retest: testenv
 	@${MAKE} -C tests retest only=$(only) IDRIS2=${TARGET} IDRIS2_PREFIX=${TEST_PREFIX}
 
 
+test-installed: testenv
```
mattpolzin, line 142: What's the goal of this target? To my mind you want the compiler test suite to be entirely representative of any changes to the compiler living in the working directory. Is the goal to see how new test cases play out against an older version of the compiler?

mattpolzin:
This change seems to work fine with any combination of incremental and non-incremental I throw at it. I'd personally rather not add the new `test-installed` target without understanding the utility of it, but the incremental flag passed to install-libs seems just fine.

Makefile
attila-lendvai: this target runs the tests with the **installed** version of Idris, i.e. this target detects the faulty installation.

let me know if it's not needed and i'll update the PR.

i have various changes planned for the build system, but i'm not sure this is the right place to discuss that, though. i will open PR's and/or write to the mailing list.

mattpolzin: I'm not totally convinced that testing an installed version of Idris is very useful, but if that's the goal then this target should probably work without relying on build artifacts (so testing is either totally dependent upon the local build of the Idris compiler or entirely dependent upon the installed Idris compiler but not a combination of the two).

Right now, `make clean && make test-installed` does not run because `make-installed` is not fully independent of the local build of the Idris compiler.

attila-lendvai: oh, good point! i have fixed the PR to work that way.

but i'm still open for removing that target altogether, if it's not desirable. note though, that the normal tests didn't catch this issue with the installation.

PR #1961, opened 2021-09-29
[tests] fix idris2/pk010: make it more roboust

attila-lendvai:
`$(pwd)` in the shell means running `pwd` but then there's code next to it that specially handles cygwin.

merge the two into one.


mattpolzin (approved):
Looks pretty reasonable to me!

PR #1964, opened 2021-09-30
[ refactor ] use proper int types for Constant

stefan-hoeck:
Now that the compiler(s) know about all the `IntXY` and `BitsXY` types plus their primops, we should use the proper types in `Core.TT.Constant`. This simplifies the implementations of primops in `Core.Primitives` and makes sure that they stay in sync with the implementations in the scheme backends.

### DIFF
--- a/src/Core/TTC.idr
+++ b/src/Core/TTC.idr
@@ -152,10 +152,10 @@ export
 TTC Constant where
   toBuf b (I x) = do tag 0; toBuf b x
   toBuf b (BI x) = do tag 1; toBuf b x
-  toBuf b (B8 x) = do tag 2; toBuf b x
-  toBuf b (B16 x) = do tag 3; toBuf b x
-  toBuf b (B32 x) = do tag 4; toBuf b x
-  toBuf b (B64 x) = do tag 5; toBuf b x
+  toBuf b (B8 x) = do tag 2; toBuf b $ cast {to = Int} x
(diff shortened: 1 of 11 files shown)


src/Core/TTC.idr
```diff
@@ -152,10 +152,10 @@ export
...
   toBuf b (I x) = do tag 0; toBuf b x
   toBuf b (BI x) = do tag 1; toBuf b x
-  toBuf b (B8 x) = do tag 2; toBuf b x
-  toBuf b (B16 x) = do tag 3; toBuf b x
-  toBuf b (B32 x) = do tag 4; toBuf b x
-  toBuf b (B64 x) = do tag 5; toBuf b x
+  toBuf b (B8 x) = do tag 2; toBuf b $ cast {to = Int} x
+  toBuf b (B16 x) = do tag 3; toBuf b $ cast {to = Int} x
+  toBuf b (B32 x) = do tag 4; toBuf b $ cast {to = Integer} x
+  toBuf b (B64 x) = do tag 5; toBuf b $ cast {to = Integer} x
```
gallais, line 158: Could we get rid of all of these annotations by declaring `TTC BitsN` implementations?

gallais (commented):
A small cleanup suggestion.

stefan-hoeck:
Sorry for not replying to your suggestions. I'm extremely busy with this rhone thing right now. And thanks a lot for merging anyway.

PR #1967, opened 2021-10-01
Implement generic interpolation

andrevidela:
String interpolation can now take values other than strings as long as
they implement the `Interpolation` interface. The `Interpolation`
interface has 1 method `interpolate : a -> String` that converts
a value of the given type into a string for interpolation. By default
only one instance is given for `String` and the method is the identity
function.

Documentation will follow once the design has sthbilized


mattpolzin:
Not that I've wanted to use `:++:` for something else before, but I wonder if we want to expose a prelude operator for something that is almost never written by hand or seen in code. Could we use a named function (possibly even with an infix application in place of the current `++`).

gallais:
Should we instead desugar

`"xxx \{expr} yyy"`

to

`fastConcat [interpolate "xxx ", interpolate expr, interpolate " yyy"]`?
This would ensure we're only allocating small strings for the parts & then
a final big string for the whole.

buzden:
> Should we instead desugar
> 
> `"xxx \{expr} yyy"`
> 
> to
> 
> `fastConcat [interpolate "xxx ", interpolate expr, interpolate " yyy"]`? This would ensure we're only allocating small strings for the parts & then a final big string for the whole.

I see a problem that this form would not reduce, thus blocking such literals to be used in string-angaged proofs.

Z-snails:
> > Should we instead desugar
> > 
> > `"xxx \{expr} yyy"`
> > 
> > to
> > 
> > `fastConcat [interpolate "xxx ", interpolate expr, interpolate " yyy"]`? This would ensure we're only allocating small strings for the parts & then a final big string for the whole.
> 
> I see a problem that this form would not reduce, thus blocking such literals to be used in string-angaged proofs.

We could use regular `concat` then rely on it `%transform`ing into `fastConcat`. Or would normalisation turn that into a sequence of `++`s?

gallais:
`concat` is a bit of a PITA to use with a literal as an argument because
it's polymorphic over `Foldable`. You'd have to write `concat {t = List}`.

andrevidela:
@mattpolzin THat's a really good point, I'll make it a named function rather than operator

@gallais @Z-snails not quite sure what the conclusion is, does `concat` reduce nicely for proofs? I guess I could try it but ideally it would be nice if they were undistinguishable from string concatenation.

libs/prelude/Prelude/Interpolation.idr
```diff
@@ -0,0 +1,27 @@
...
+interface Interpolation a where
+  interpolate : a -> String
+
+||| Concatenation of two things with an interpolation strategy.
+||| Interpolated strings are of the form `"xxxx \{ expr } yyyy"`.
+||| In this example the string `"xxxx "` is concatenated with `expr` and
+||| `" yyyy"`, since `expr` is not necessarily a string, the generated
+||| code looks like this:
+||| ```
+||| interpolate "xxxx " :++: interpolate expr :++: interpolate " yyyy"
```
mattpolzin, line 16: nit: fell out of date with change away from `:++:` operator.

andrevidela: ah yep will fix

andrevidela: I noticed the old documentation was wrong anyways, it said `interpolate "xxx" :++: interpolate expr` instead of `interpolate "xxx" ++ interpolate expr` or `"xxx " :++: expr`

gallais:
> does concat reduce nicely for proofs?

Yes but you probably need to make sure the type of its argument is known to be `List String`.

andylokandy:
Is it possible to give every type a default implementation `Show a => Interpolation a` and to prefer a more specific implementation automatically if exists?

mattpolzin:
@andylokandy there's no existing mechanism for that, but in addition to it being something that could be special-cased and baked into the compiler I can also think of one other way to accomplish that behavior: make the interpolation interface constrained by the show interface and give a default implementation of the interpolate function that equals show.

I'm not entirely sure this is desirable, personally. It saves some boiler plate on one hand, but so would automatic conformance to `Show` (basically bake in what the elaboration-based deriving code does), and yet I think there's some benefit to leaving that to be explicitly derived or implemented. 

I think the same approach here would be nice (`%runElab derive "MyThing" [Eq, Show, Interpolation]`).

andylokandy:
@mattpolzin I hacked a bit and, indeed, Idris doesn't support that. But the interesting thing is Idris tells me it can decide which `Interpolation` implementation for `String` to use: the `Show a => Interpolation a` one or the `Interpolation String` one?

It reminds me what we need is specialization! This could be a general-purpose feature instead of some hacky nasty special-cased workaround only for string interpolation.

buzden:
> But the interesting thing is Idris telling me that it can't decide which `Interpolation` implementation for `String` to use: the `Show a => Interpolation a` one or the `Interpolation String` one?

Yes! And I can tell that it can decide even between harder cases, e.g. between multiple similar instances with the same determining parameter, like [here](https://github.com/buzden/idris2-coop/blob/7b5310c5f762721b048d62b1c61099f696cde65a/src/Control/Monad/State/Pair.idr) (only the second parameter is determining, as [you can see](https://github.com/idris-lang/Idris2/blob/079c032036bec46394e39b83be05e662c07857ee/libs/base/Control/Monad/State/Interface.idr#L15))

andylokandy:
@buzden Is it because of the compiler fallbacks to search on all parameters when multiple feasible implementations on the determining parameter is found?

My case is to allow the compiler to choose one when all parameters are considered but, still, multiple implementations are feasible.

PR #1968, opened 2021-10-01
print location of implicit name shadowing with the warning.

mattpolzin:
I recently got the implicit name warning in a file and actually had a hard time finding where the warning was coming from because the name in question was used commonly in all different contexts.


src/Idris/Error.idr
```diff
@@ -166,9 +166,10 @@ pwarning (UnreachableClause fc env tm)
     = pure $ errorDesc (reflow "Unreachable clause:"
         <++> code !(pshow env tm))
         <+> line <+> !(ploc fc)
-pwarning (ShadowingGlobalDefs _ ns)
+pwarning (ShadowingGlobalDefs fc ns)
     = pure $ vcat
-    $ reflow "We are about to implicitly bind the following lowercase names."
+    $ reflow "We are about to implicitly bind lowercase names that shadow global definitions at the following location."
+   :: !(ploc fc)
    :: reflow "You may be unintentionally shadowing the associated global definitions:"
```
gallais, line 173: I think it'd made more sense to keep the warning as it is and printed the `ploc`
after it. Just like we do in all the other warnings AFAICT.

mattpolzin: Will do!

PR #1978, opened 2021-10-06
Add `buildExpressionParser`, based on Parsec's implementation

gergoerdi:


libs/contrib/contrib.ipkg
```diff
@@ -167,6 +167,7 @@ modules = Control.ANSI,
           Text.Literate,
           Text.Parser,
           Text.Parser.Core,
+          Text.Parser.Expession,
```
gallais, line 170:
```suggestion
          Text.Parser.Expression,
```

Issue #1979, opened 2021-10-06
Type error without source location


gergoerdi:
On https://github.com/gergoerdi/petbasic-parser/tree/4e60da82f1f27dde7e96ed7349dfabcf4fa9679a I get the following type error when trying to compile `Interpreter.idr`:

```
7/8: Building Interpreter (src/Interpreter.idr)
Error: While processing right hand side of exec. While processing right hand
side of exec,next. When unifying:
    List (BASIC r2378 ())
and:
    List1 ?a
Mismatch between: List (BASIC r2378 ()) and List1 ?a.
```

What's weird about it is that there is no source location anywhere pointing at the error.

gallais:
Looks like the `fc` is just ignored in

https://github.com/idris-lang/Idris2/blob/76b90cc816b40b61ffbfbeeddf76a7d9829b5c32/src/Idris/Error.idr#L235

It should be a matter of adding a `<+> line <+> !(ploc fc)` at the end of that errorDesc.

withoutK:
> Looks like the `fc` is just ignored in
> 
> https://github.com/idris-lang/Idris2/blob/76b90cc816b40b61ffbfbeeddf76a7d9829b5c32/src/Idris/Error.idr#L235
> 
> It should be a matter of adding a `<+> line <+> !(ploc fc)` at the end of that errorDesc.

I had done it yesterday and test with the example provide by https://github.com/idris-lang/Idris2/issues/1979#issuecomment-936318371, it turns out even involve the underlined placed fc, Idris cannot report location yet. In addition, if `perr err` is in the situation which contains error locations, we need to elim the duplication of error locations.

PR #1981, opened 2021-10-06
[ chore ] add `Eq` impl for `IdrisAnn`

withoutK:


gallais:
Do you have a use-case for this?

withoutK:
> Do you have a use-case for this?

while trying to fix  #1979 i add them, but the attempt to fix is failed, so not really

PR #1983, opened 2021-10-07
[ build ] Simplify bootstrap

attila-lendvai:
Delete a whole lot of things from `bootstrap-stage2.sh` that doesn't seem to be needed.

Reasoning: `LD_LYBRARY_PATH` is already handled by the wrapper scripts. `IDRIS2_PATH` is not exported and is not propagated anywhere in the `Makefile`.

It causes no visible harm when testing it on Linux using:

`make SCHEME=chez-scheme clean bootstrap test`

----

the CI also seems to be clear (besides some seemingly unrelated issues).

i'm working on a better build system, and i'm trying to understand what's happening in the current setup. in the process i came to the above realization, but i'm still not sure... am i missing here something?

it would be nice if someone reviewed this who understands the bootstrap process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ TEST_PREFIX ?= ${IDRIS2_CURDIR}/build/env
 IDRIS2_BOOT_PREFIX := ${IDRIS2_CURDIR}/bootstrap-build
 
 # These are the library path in the build dir to be used during build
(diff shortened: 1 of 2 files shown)


attila-lendvai:
meanwhile, i have made good progress with refactoring the build process. given my results, this understanding should be valid, but i'll open a new PR when i'm happy with the refactoring.

hence i've converted it as draft, and will close this one when the new PR is ready.

benhormann:
Sounds like you're onto it, it'll be good to get it sorted. I'll try straighten the good bits of mine, maybe there's something you can use.

I'd suggest adding something like IDRIS2_OPTS, so there is less need for taking on more variables.

attila-lendvai:
closing it as this has been grossly superseded by https://github.com/idris-lang/Idris2/pull/1990

@benhormann FYI.

PR #1984, opened 2021-10-08
Add enumerate function for `List` and `Vect`

remimimimimi:


remimimimi:
> This seems very ad-hoc. The index functions of #1958 would be better candidates for inclusion IMO.
Agree with you.

PR #1986, opened 2021-10-08
Re-enable -Xcheck-hashes

benhormann:
This is using some supposedly good enough hash algorithms found on the web: http://www.cse.yorku.ca/~oz/hash.html

I've done a little testing and it seems to work well. I extracted every file from git's db (Idris 1 and 2) to run the function against. About 27k files, zero collisions. I also did `for x in **/*.idr; do echo '' >> $x; done` and all were recompiled them.

I haven't tested much with `Maybe Bit64`, since that is hot of the press. Seems just as good as Int or Integer so far.

One thing to note is that I've assumed the hash won't ever be `0` with text encoding, so zero gets folded in with the equally unlikely `~0`. I imagine the chances of IO errors are low, but it seems better to do it this way regardless. I've tested that part works too, but I'm no C pro.

**TL;DR**
`sed -i.bak ~/.idris2/bin/idris2 's/"\$@"/-Xcheck-hashes &/'`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ idris2-exec: ${TARGET}
 
 ${TARGET}: src/IdrisPaths.idr
 	${IDRIS2_BOOT} --build ${IDRIS2_APP_IPKG}
+	install support/c/${IDRIS2_SUPPORT} ${TARGET}_app
 
 # We use FORCE to always rebuild IdrisPath so that the git SHA1 info is always up to date
 src/IdrisPaths.idr: FORCE
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -33,7 +33,7 @@ import public Libraries.Utils.Binary
 ||| (Increment this when changing anything in the data format)
 export
 ttcVersion : Int
-ttcVersion = 65
+ttcVersion = 66
 
 export
 checkTTCVersion : String -> Int -> Int -> Core ()
@@ -44,7 +44,7 @@ record TTCFile extra where
   constructor MkTTCFile
   version : Int
   totalReq : TotalReq
-  sourceHash : Maybe String
+  sourceHash : Maybe Bits64
   ifaceHash : Int
   importHashes : List (Namespace, Int)
   incData : List (CG, String, List String)
@@ -285,7 +285,7 @@ writeToTTC extradata sourceFileName ttcFileName
          defs <- get Ctxt
          ust <- get UST
          gdefs <- getSaveDefs (currentNS defs) (keys (toSave defs)) [] defs
-         sourceHash <- hashFileWith defs.options.hashFn sourceFileName
+         sourceHash <- coreLift $ hashFile sourceFileName
          totalReq <- getDefaultTotalityOption
          log "ttc.write" 5 $ unwords
            [ "Writing", ttcFileName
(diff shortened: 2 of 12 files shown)


benhormann:
> Tangentially, seeing this implementation in C makes me wonder how far off we are from being able to write the algorithm right in Idris 2 (since Idris would _like_ to be fast enough for that to be practical).

It'd be worth trying. Something like:

```idr
hash : Buffer -> Int -> Bits32 -> Bits32 -> Bits64
hash _ 0 hashA hashB = (cast {to=Bits64} hashA) + (2 exp 32) + hashB
hash buf i hashA hashB = do n <- getBits8 buf i
                            hash buf (i - 1) (calcA hashA n) (calcB hashB n)

export
hashFile : HasIO io => String -> io (Maybe Bits64)
hashFile fileName
  = do Right buf <- createBufferFromFile fileName
           | Left _ => pure Nothing
       size <- rawSize buf
       hash buf size 5381 0
```
Node would have to use a BigInt for the result.

benhormann:
@mattpolzin This would copy the library to `build/exec/` named as `idris2_app` if the directory doesn't exist (which could be possible using a different backend?). Not an immediately fatal error, but could cause surprising ones.

It looks like `make install` would fail if the directory doesn't exist, so if nobody reported that chances are this won't be a problem... Or I can check the directory exists to be safe.

mattpolzin:
@benhormann this is some pretty cool exploration into different approaches!

I'd say anything that is fast enough to make `-Xcheck-hashes` worth using is fast enough to be worth it (since that's the motivation here), and it's actually really nice to have use-cases that motivate various different kinds of optimizations the compiler could aim to adopt in due time (RE your points about there being opportunities to improve `ModTree`/`ProcessIdr`.

It seems to me we still find ourselves in one of two situations, though:
1. We don't need support from an external language (scheme or C) so we can write the new code right into the compiler.
2. We do need support from an external language so we either need to package the _new_ support library with the build (the Makefile workaround you came up with in this PR) or we need to introduce the external language code 1 Idris 2 version prior to when we begin using it in the compiler (my suggestion as what seems like the cleanest approach even though it stretches out the timeline of this improvement).

Does that still seem accurate to you? If so, I vote you pick your favorite approach so far and given that it almost certainly falls into category (2) I also vote you open a separate PR introducing helpers for which you need external language support into the `base` library so we can get that merged before the next version is cut and wrap this existing PR up soon after the next version is cut. I might still be partial to the C version for this purpose even though other versions are conceptually nicer at the cost of some speed, but I stand by my assessment that anything "fast enough" should be acceptable.

mattpolzin:
I went ahead and re-requested my review because I'd say at this point it is more accurate to say you are closing in on a final decision of how to handle this work than it is to say the work is ready to merge as-isgiven my thoughts [here](https://github.com/idris-lang/Idris2/pull/1986#issuecomment-950182355) and the follow-up work you've done on alternative approaches.

benhormann:
I agree, let's close this.

My current thinking is the *has changed?* algorithm needs work.
Core.Hash is the part relevant to this PR. Instead of hashing a limited subset of the AST, let's the hash source.

As far as support goes, it should be fast enough written in Idris. I've created a PR that makes fixnum ops available on Racket and swaps `modulo` for `bitwise-and` in casts &mdash; should get the ball rolling.

Issue #1991, opened 2021-10-11
0.5 & 0.5.1 performance regression (it was ok with 0.4)


csabahruska:
Idris 0.5.x consumes all memory during typechecking my `printf` implementation. It worked fine with Idris 0.4
Idris 0.4 (release version) used only 202MB memory to typecheck, while the 0.5.0 and 0.5.1 and the latest master version run out memory (64GB). However for simplicity I set the memory limit for benchmarks to max 2GB.

# printf source code
```idris
module Printf

import Prelude
import Data.String

data Arg
    = AInt      Arg
    | AOther    Char Arg
    | AEnd

buildArg : List Char -> Arg
buildArg fmt = case fmt of
    '%' :: 'i' :: fmtTail   => AInt (buildArg fmtTail)
    c :: fmtTail            => AOther c (buildArg fmtTail)
    Nil                     => AEnd

argToType : Arg -> Type -> Type
argToType a result = case a of
    AInt fmtTail      => Int -> argToType fmtTail result
    AOther _ fmtTail  => argToType fmtTail result
    AEnd              => result

-- PrintfType "foo" result = result
-- PrintfType "%i\n" result = Int -> result
-- etc
PrintfType : String -> Type -> Type
PrintfType fmt result = argToType (buildArg (unpack fmt)) result

sprintf : (fmt : String) -> PrintfType fmt String
sprintf fmt = go "" (buildArg (unpack fmt)) where
    go : String -> (arg : Arg) -> argToType arg String
    go strTail arg = case arg of
        AInt fmtTail        => \i : Int => go (strTail ++ show i) fmtTail
        AOther c fmtTail    => go (strTail ++ singleton c) fmtTail
        AEnd                => strTail

test : ?result
--test = sprintf "%i%i"
test = sprintf "%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i%i"
--test = sprintf "%i%i%i%i%i%i%i%i%i%i%i%i"
```

# reproduce
Download the files from the gist and run the `./c` bash script.

[...]

PR #1994, opened 2021-10-11
Fix performance regression #1991

edwinb:
This uses a better way of using `nf_metavar_threshold` to check when it's worth normalising a term during unification. Previously we'd do a lot of unnecessary work.


edwinb:
Argh, there is too much unnecessary stuff in this PR. I'll start again...

Issue #1965, opened 2021-10-01
Partial evaluator anomaly


danielkroeni:
The `%spec` annotation leads to surprising effects.

# Steps to Reproduce
Compile and run the given program:
https://gist.github.com/danielkroeni/6d5386f8a8ad7be15713725f4f9bf01a
Observe that the output is `State: 6`.

Then in the same program enable the `%spec ma,mb` annotation on `(>>)`.
Compile and run again. Now the output is `State: 4`.

# Expected Behavior
I would expect that the `%spec` annotation to have no observable effect. (But I may be wrong.)

# Observed Behavior
There is an observable effect, the computed results differ and seem to be wrong when `%spec ma,mb` is enabled.

# Some more anomalies
1. Let `%spec ma,mb` enabled but add an additional (from my beginner point of view) totally unrelated `import Data.List` to the top of the file and it produces again the correct result.
This may have something to do with "things behave differently if name needs to be disambiguated" (e.g. https://github.com/idris-lang/Idris2/issues/1896).

2. And here is an even stranger effect: 
Given the code here in two file: https://gist.github.com/danielkroeni/d082fd3795345f77de740b08902f0e61
Compile and run `problem.idr` and it produces the expected result.
Then change the name of the function `xxxx` to `xxx` in both files and it generates a wrong result.
That's confusing.

# Version
```
> idris2 --version
Idris 2, version 0.5.1-a9ccf4db4
```

ziman:
```
$ diff dumcases_xxxx.idr dumpcases_xxx.idr
5,6c5,6
< Main.comp = [{ext:0}]: (_PE.PE_>>_41f9c421432e45e8 [!{ext:0}])
< _PE.PE_>>_41f9c421432e45e8 = [{ext:0}]: (%con [cons] Builtin.MkPair Just 0 [(+Int [(+Int [(+Int [(+Int [(+Int [(+Int [!{ext:0}, 1]), 1]), 1]), 1]), 1]), 1]), 0])
---
> Main.comp = [{ext:0}]: (_PE.PE_>>_0 [!{ext:0}])
> _PE.PE_>>_0 = [{ext:0}]: (%con [cons] Builtin.MkPair Just 0 [(+Int [(+Int [(+Int [(+Int [!{ext:0}, 1]), 1]), 1]), 1]), 0])
```

ziman:
I find the zero quite suspicious.

danielkroeni:
[...]

PR #1996, opened 2021-10-12
Make the hash function return Bits64.

danielkroeni:
Based on @ziman proposal this PR replaces the result of the `hash` function from `Int` to `Bits64`.
This paves the way for better hash functions such as @Z-snails [SipHash](https://github.com/Z-snails/Idris2-hashable/blob/main/src/Data/Hashable/SipHash.idr) implementation.
This PR ~~fixes https://github.com/idris-lang/Idris2/issues/1965~~ makes the deeper issue concerning hash-collisions less obvious.

### DIFF
--- a/src/Compiler/ES/Codegen.idr
+++ b/src/Compiler/ES/Codegen.idr
@@ -49,7 +49,7 @@ jsString s = "'" ++ (concatMap okchar (unpack s)) ++ "'"
                             '"' => "\\\""
                             '\r' => "\\r"
                             '\n' => "\\n"
-                            other => "\\u{" ++ asHex (cast {to=Int} c) ++ "}"
+                            other => "\\u{" ++ asHex (cast {to=Bits64} c) ++ "}"
 
 ||| Alias for Text . jsString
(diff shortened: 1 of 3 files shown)


danielkroeni:
I just incremented the `ttcVersion` and force pushed an amended commit. I am not sure how stupid this was.
And I also saw https://github.com/idris-lang/Idris2/pull/1986 which may interact with this PR.

danielkroeni:
> > This PR fixes #1965.
> 
> For a proper fix shouldn't we add an equality check in case the hashes match?

I was so happy having found the simple `asHex` problem that I already forgot the possible deeper problem.
I even suggested it here https://github.com/idris-lang/Idris2/issues/1965#issuecomment-938450485. But I would need help doing it. I currently do not see where to add the check. If someone could fix it, I would appreciate it.

danielkroeni:
> One of us will have to merge, doubt it will be a problem. Or you can replace #1986 if that is preferred.

I doesn't look like this PR is ready soon. I will merge.

PR #2007, opened 2021-10-14
[ fix #2002 ] implicits used in record update

### LINKED ISSUE #2002: Implicits used in record update are undefined
Compiling the following program..

```idris
module Bug

record Rec where
    n : Nat

data T : Rec -> Type where
    C : T ({ n := Z } r)
```

results in this error:
```
1/1: Building Bug (Bug.idr)
Error: While processing constructor C. Undefined name r. 

Bug:9:23--9:24
 5 |     n : Nat
 6 | 
 7 | 
 8 | data T : Rec -> Type where
 9 |     C : T ({ n := Z } r)
```

Explicitly introducing an implicit for `r` (e.g. `forall r. T ({ n := Z } r)`) solves the issue. I would've expected Idris to automatically infer that an implicit is necessary since `r` begins with a lowercase letter.

eayus:
Fixes #2002 . The reason why I've left this is a draft is that my fix seems to have exposed/introduced some strange behaviour. I've tried to implement a generalised fix to the original issue, where the implicit may appear in the record field updates as well. However, compiling the following program...

```idris
module Bug

record Rec where
    n : Nat

data T : Rec -> Type where
    C : T ({ n := a } r)
```
gives the very confusing error of:
```
1/1: Building Bug (Bug.idr)
Error: While processing constructor C. Linearity checking failed on
    ?{P:a:2443}_[] ((pat 0 r : Bug.Rec) => (pat 0 n375 : Prelude.Types.Nat)
    => Prelude.Types.Nat not a function type)

Bug:8:19--8:20
 4 | record Rec where
 5 |     n : Nat
 6 | 
 7 | data T : Rec -> Type where
 8 |     C : T ({ n := a } r)
                       ^
```

Before my changes, this bugged program would be rejected with an error saying `a` and `r` are not bound. Therefore I am not sure whether I have made a mistake implementing the previous fix (very possible), or whether this is a separate bug that I have uncovered.

### DIFF
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -141,7 +141,7 @@ idrisTestsData = MkTestPool "Data and record types" [] Nothing
        "data001",
        -- Records, access and dependent update
        "record001", "record002", "record003", "record004", "record005",
-       "record006", "record007", "record008", "record009"]
+       "record006", "record007", "record008", "record009", "record010"]
 
 idrisTestsBuiltin : TestPool
 idrisTestsBuiltin = MkTestPool "Builtin types and functions" [] Nothing
--- a/tests/idris2/record010/expected
+++ b/tests/idris2/record010/expected
@@ -1,0 +1,1 @@
+1/1: Building record (record.idr)
--- a/tests/idris2/record010/record.idr
+++ b/tests/idris2/record010/record.idr
@@ -1,0 +1,14 @@
+-- This test ensures that implicits bound on the RHS of a
+-- record update field are correctly bound by the compiler.
+
+
+record Rec where
+    n : Nat
+
+
+data T : Rec -> Type where
+    C : T ({ n := Z } r)
+
+
+data U : Rec -> Type where
+    D : U ({ n $= S } r)
--- a/tests/idris2/record010/run
+++ b/tests/idris2/record010/run
@@ -1,0 +1,4 @@
+rm -rf build
+
+$1 --no-color --console-width 0 --no-banner --check record.idr
+


eayus:
In writing this fix, I feel like that the representation for record update fields:
```idris
data IFieldUpdate' : Type -> Type where                                                           
    ISetField : (path : List String) -> RawImp' nm -> IFieldUpdate' nm                           
    ISetFieldApp : (path : List String) -> RawImp' nm -> IFieldUpdate' nm
```
should be modified into something like:
```idris
data IFieldUpdateType = ISetField | ISetFieldApp

record IFieldUpdate nm where
    updateType : IFieldUpdateType
    path : List String
    tm : RawImp' nm
```
Since it seems that we don't actually care about the update type a lot of the time. This would avoid the need for helper functions which operate indiscriminantly of the path or `RawImp` term.

I've done a quick grep for this type and I can't see any other place where this change would make the code worse. However, I understand I might be overlooking something, and also that changes to core data structures can be an inconvenience.

eayus:
I've added a test case for #2002 which will be fixed once this PR is done.

There are still some related bugs causing Idris to reject valid programs (namely when the implicit is used in the actual field update). Those programs were rejected before this PR so it's not a regression.

For example, this program:
```idris
record Rec where
    n : Nat

R : Rec

data T : Rec -> Type where
    C : forall a. T ({ n := a } R)
```
is rejected with the error:
```
1/1: Building Bug (Bug.idr)
Error: While processing constructor C. Need to bind implicits {_:379} : Type, {_:378} : Type.

Bug:13:22--13:34
 12 | data T : Rec -> Type where
 13 |     C : forall a. T ({ n := a } R)
```

PR #2015, opened 2021-10-16
[ prelude ] Add a function for backing up the `Alternative`

buzden:


buzden:
> Does this need to exist in `prelude`? I feel like it's reaching into _nice to have_ territory which conflict with the goal of keeping `prelude` minimal

I'd agree that *maybe* `base` is a better place for tihs, but I haven't found of nice place for this function in the `base` lib.

gallais:
Do we actually need such a thing?

```
backupWith v $
(<|> pure v) $
```
have the same length.

gallais:
I expect `backup` to involve some kind of IO to store values in a (possibly remote)
file system. So the name is more confusing than the actual implementation to me.

PR #2024, opened 2021-10-17
[ base ] Add couple of properties of `either` function with mappings

buzden:


buzden:
> I'd use `Fusion` instead of `Simplifies` as these are fusion rules.

I renamed

PR #2027, opened 2021-10-18
Implemented %noinline

danielkroeni:
This PR implements the `%noinline` pragma which flags a function that it should **not** be inlined. This is useful to ensure, that a function is not eliminated by the Inliner and is similar to [Haskell's NOINLINE pragma](https://downloads.haskell.org/~ghc/9.0.1/docs/html/users_guide/exts/pragmas.html?highlight=noinline#pragma-NOINLINE).


danielkroeni:
@Z-snails Could you please give me a hint where to check for conflicting `%inline` and `%noinline` pragmas?
And please, just tell me if such PRs generated more work than they help.

PR #2044, opened 2021-10-22
[refactor] Use multiline strings for better readability

ska80:
This PR is a follow-up to #1142 as I couldn't reopen it.

### DIFF
--- a/src/Compiler/RefC/RefC.idr
+++ b/src/Compiler/RefC/RefC.idr
@@ -452,9 +452,6 @@ const2Integer c i =
         _ => i
 
 
-
-
-
 -- we return for each of the ANF a set of statements and two possible return statements
 -- The first one for non-tail statements, the second one for tail statements
 -- this way, we can deal with tail calls and tail recursion.
@@ -545,7 +542,6 @@ mutual
         constBlockSwitch alts retValVar (i+1)
 
 
-
     constDefaultBlock : {auto a : Ref ArgCounter Nat}
                      -> {auto t : Ref TemporaryVariableTracker (List (List String))}
                      -> {auto oft : Ref OutfileText Output}
@@ -564,7 +560,6 @@ mutual
         freeTmpVars
         decreaseIndentation
         emit EmptyFC "  }"
-
 
 
     makeNonIntSwitchStatementConst :
@@ -727,9 +722,6 @@ mutual
     cStatementsFromANF (ACrash fc x) = do
         emit fc $ "// CRASH"
         pure $ MkRS "NULL" "NULL"
-
-
-
 
 
 addCommaToList : List String -> List String
@@ -989,29 +981,35 @@ header : {auto c : Ref Ctxt Defs}
       -> {auto h : Ref HeaderFiles (SortedSet String)}
       -> Core ()
 header = do
-    let initLines = [ "#include <runtime.h>"
-                    , "/* " ++ (generatedString "RefC") ++" */"]
+    let initLines = """
+      #include <runtime.h>
+      /* \{ (generatedString "RefC") } */
+
+      """
     let headerFiles = Libraries.Data.SortedSet.toList !(get HeaderFiles)
     let headerLines = map (\h => "#include <" ++ h ++ ">\n") headerFiles
     fns <- get FunctionDefinitions
-    update OutfileText (appendL (initLines ++ headerLines ++ ["\n// function definitions"] ++ fns))
+    update OutfileText (appendL ([initLines] ++ headerLines ++ ["\n// function definitions"] ++ fns))
 
 footer : {auto il : Ref IndentLevel Nat}
       -> {auto f : Ref OutfileText Output}
       -> {auto h : Ref HeaderFiles (SortedSet String)}
       -> Core ()
 footer = do
-    emit EmptyFC ""
-    emit EmptyFC " // main function"
-    emit EmptyFC "int main(int argc, char *argv[])"
-    emit EmptyFC "{"
-    if contains "idris_support.h" !(get HeaderFiles)
-       then emit EmptyFC "   idris2_setArgs(argc, argv);"
-       else pure ()
-    emit EmptyFC "   Value *mainExprVal = __mainExpression_0();"
-    emit EmptyFC "   trampoline(mainExprVal);"
-    emit EmptyFC "   return 0; // bye bye"
-    emit EmptyFC "}"
+    emit EmptyFC """
+
+      // main function
+      int main(int argc, char *argv[])
+      {
+          \{
+            (if contains "idris_support.h" !(get HeaderFiles)
+                then "idris2_setArgs(argc, argv);" else "")
+          }
+          Value *mainExprVal = __mainExpression_0();
+          trampoline(mainExprVal);
+          return 0; // bye bye
+      }
+      """
 
 export
 executeExpr : Ref Ctxt Defs -> (execDir : String) -> ClosedTerm -> Core ()
--- a/src/Compiler/RefC/RefC.idr
+++ b/src/Compiler/RefC/RefC.idr
@@ -1001,9 +1001,9 @@ footer = do
       // main function
       int main(int argc, char *argv[])
       {
-          \{
-            (if contains "idris_support.h" !(get HeaderFiles)
-                then "idris2_setArgs(argc, argv);" else "")
+          \{ ifThenElse (contains "idris_support.h" !(get HeaderFiles))
+                        "idris2_setArgs(argc, argv);"
+                        ""
           }
           Value *mainExprVal = __mainExpression_0();
           trampoline(mainExprVal);
--- a/src/Compiler/RefC/RefC.idr
+++ b/src/Compiler/RefC/RefC.idr
@@ -983,7 +983,7 @@ header : {auto c : Ref Ctxt Defs}
 header = do
     let initLines = """
       #include <runtime.h>
-      /* \{ (generatedString "RefC") } */
+      /* \{ generatedString "RefC" } */
 
       """
     let headerFiles = Libraries.Data.SortedSet.toList !(get HeaderFiles)
--- a/src/Compiler/RefC/RefC.idr
+++ b/src/Compiler/RefC/RefC.idr
@@ -454,6 +454,7 @@ const2Integer c i =
 
 
 
+
 -- we return for each of the ANF a set of statements and two possible return statements
 -- The first one for non-tail statements, the second one for tail statements
 -- this way, we can deal with tail calls and tail recursion.
--- a/src/Compiler/RefC/RefC.idr
+++ b/src/Compiler/RefC/RefC.idr
@@ -452,6 +452,8 @@ const2Integer c i =
         _ => i
 
 
+
+
 -- we return for each of the ANF a set of statements and two possible return statements
 -- The first one for non-tail statements, the second one for tail statements
 -- this way, we can deal with tail calls and tail recursion.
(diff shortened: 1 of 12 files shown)


andrevidela:
May I suggest reverting to `unlines`, merging this and opening an issue ?

ska80:
> May I suggest reverting to `unlines`, merging this and opening an issue ?

Good idea.

src/Compiler/RefC/RefC.idr
```diff
@@ -452,9 +452,6 @@ const2Integer c i =
         _ => i
 
 
-
```
mattpolzin, line 455: Although I can see the temptation, I am not sure if we should be making arbitrary whitespace changes given that we don't have community or compiler-specific guidelines on the subject and the author might well have made intentional choices around these newlines based on their personal preferences.

src/Compiler/Scheme/Chez.idr
```diff
@@ -385,49 +390,56 @@ getFgnCall version (n, fc, d) = schFgnDef fc n d version
...
 startChez : String -> String -> String
-startChez appdir target = startChezPreamble ++ unlines
-    [ "export LD_LIBRARY_PATH=\"$DIR/" ++ appdir ++ ":$LD_LIBRARY_PATH\""
-    , "export DYLD_LIBRARY_PATH=\"$DIR/" ++ appdir ++ ":$DYLD_LIBRARY_PATH\""
-    , "export IDRIS2_INC_SRC=\"$DIR/" ++ appdir ++ "\""
-    , "\"$DIR/" ++ target ++ "\" \"$@\""
-    ]
+startChez appdir target = startChezPreamble ++ """
+  export LD_LIBRARY_PATH="$DIR/\{ appdir }":$LD_LIBRARY_PATH
+  export DYLD_LIBRARY_PATH="$DIR/\{ appdir }":$DYLD_LIBRARY_PATH
```
mattpolzin, line 410:
```suggestion
  export LD_LIBRARY_PATH="$DIR/\{ appdir }:$LD_LIBRARY_PATH"
  export DYLD_LIBRARY_PATH="$DIR/\{ appdir }:$DYLD_LIBRARY_PATH"
```

src/Compiler/Scheme/Chez.idr
```diff
@@ -385,49 +390,56 @@ getFgnCall version (n, fc, d) = schFgnDef fc n d version
...
-    [ "@echo off"
-    , "set APPDIR=%~dp0"
-    , "set PATH=%APPDIR%" ++ appdir ++ ";%PATH%"
-    , "set IDRIS2_INC_SRC=%APPDIR%" ++ appdir
-    , "\"" ++ chez ++ "\" " ++ progType ++ " \"%APPDIR%" ++ target ++ "\" %*"
-    ]
+startChezCmd chez appdir target progType = """
+  @echo off
+
+  rem \{ generatedString "Chez" }
```
mattpolzin, line 420: this is new. I know so little about Chez that I can't say whether I like it or not :) but it's not a direct translation.

src/Compiler/Scheme/Chez.idr
```diff
@@ -385,49 +390,56 @@ getFgnCall version (n, fc, d) = schFgnDef fc n d version
...
-    , "\"" ++ chez ++ "\" " ++ progType ++ " \"$DIR/" ++ target ++ "\" \"$@\""
-    ]
+startChezWinSh chez appdir target progType = """
+  #!/bin/sh
+  # \{ generatedString "Chez" }
+
+  set -e # exit on any error
+
+  DIR=$(dirname "$(readlink -f -- "$0" || cygpath -a -- "$0")")
+  PATH="$DIR/\{ appdir }":$PATH
```
mattpolzin, line 437:
```suggestion
  PATH="$DIR/\{ appdir }:$PATH"
```

src/Idris/Package.idr
```diff
@@ -795,6 +795,25 @@ errorMsg = unlines
   , "    --build-dir <dir>"
   , "    --output-dir <dir>"
   ]
+{-
```
mattpolzin, line 798: Left in by accident?

src/Idris/CommandLine.idr
```diff
@@ -402,13 +402,16 @@ versionMsg = "Idris 2, version " ++ show version
...
-        "\n" ++
-        "Environment variables:\n" ++
-        envsUsage
+usage = """
+  \{ versionMsg }
+  Usage: idris2 [options] [input file]
+
+  Available options:
+  \{ optsUsage }
+
```
mattpolzin, line 411: This creates an extra newline right?

src/Core/Options.idr
```diff
@@ -40,17 +40,18 @@ outputDirWithDefault d = fromMaybe (build_dir d </> "exec") (output_dir d)
...
-          , "+ Build Directory        :: " ++ show bdir
-          , "+ Local Depend Directory :: " ++ show ldir
-          , "+ Output Directory       :: " ++ show (outputDirWithDefault d)
-          , "+ Installation Prefix    :: " ++ show dfix
-          , "+ Extra Directories      :: " ++ show edirs
-          , "+ Package Directories    :: " ++ show pdirs
-          , "+ CG Library Directories :: " ++ show ldirs
-          , "+ Data Directories       :: " ++ show ddirs]
+toString d@(MkDirs wdir sdir bdir ldir odir dfix edirs pdirs ldirs ddirs) = """
+  + Working Directory      :: \{ wdir }
```
mattpolzin, line 44: I suppose I don't personally mind one way or the other, but why the change to not quoting these strings (wdir, bdir, ldir, dfix)?

mattpolzin (commented):
Looked over most changes for consistency. Didn't quite make it through all of the Chez compiler files yet because with no Chez background it takes me some time to parse whether any unintended changes have been introduced.

src/Compiler/Scheme/Chez.idr
ska80: This change just adds a comment to a batch script with CG's name.

src/Compiler/Scheme/Chez.idr
ska80: Shouldn't we just assume that the rest of the `$PATH` should be already correctly escaped, and we only care about the path that we prepend?

src/Idris/Package.idr
ska80: See issue #2087 

I decided to keep the multiline version.  Once #2087 is fixed, it can then be used in place of the 'unlines' version. If it is not acceptable I will delete it.

src/Idris/CommandLine.idr
ska80: This is a direct translation.  With every change, I tested the output to make sure it was the same.

src/Compiler/Scheme/Chez.idr
ska80: I tried to keep a direct translation.

src/Compiler/RefC/RefC.idr
ska80: Fair enough.  I can revert those vertical whitespace changes if needed.  I agree I made it look more pleasing according to my personal preferences. We do need community guidelines.

src/Idris/CommandLine.idr
ska80: My bad, you are right.  Will fix it.

src/Core/Options.idr
ska80: Will fix.

src/Compiler/Scheme/ChezSep.idr
```diff
@@ -39,59 +39,67 @@ import Libraries.Utils.String
...
-  , unlines ["  (load-shared-object \"" ++ escapeStringChez lib ++ "\")" | lib <- libs]
-  , ")"
-  ]
+schHeader libs compilationUnits = """
+  (import (chezscheme) (support)
+      \{ unwords ["(" ++ cu ++ ")" | cu <- compilationUnits] })
+  (case (machine-type)
+    [(i3le ti3le a6le ta6le tarm64le) (load-shared-object "libc.so.6")]
+    [(i3osx ti3osx a6osx ta6osx tarm64osx) (load-shared-object "libc.dylib")]
+    [(i3nt ti3nt a6nt ta6nt) (load-shared-object "msvcrt.dll")]
```
mattpolzin, line 48: Missing `(load-shared-object \"ws2_32.dll\")`

src/Idris/Package.idr
mattpolzin: I think we should remove the comment in the meantime. Until the multiline version is actually made to be the version in-code, it creates an extra place where any modifications to the message needs to be made and for the same reason the commented out version will need to be checked for consistency after the bug is fixed so it doesn't really save time at that point IMO.

src/Compiler/RefC/RefC.idr
mattpolzin: Personally, I would say these newline changes should be reverted before merging, but if you would like to keep them and another maintainer decides this is ready to merge with these changes in place, I won't put up a fuss over such a minor detail. I just consider it a courtesy to the original author of the file.

mattpolzin (changes requested):
I've looked over the remaining files, so to the extent that my eyes can be trusted to spot things, I've only found 1 remaining meaningful difference.

Looks really close!

src/Compiler/Scheme/ChezSep.idr
ska80: I removed this line due to the removal of the same line in `Chez.idr`.

src/Compiler/RefC/RefC.idr
ska80: Ok, I have reverted these whitespace changes.

src/Idris/Package.idr
ska80: Done.  Thanks for the review!

mattpolzin (approved):
This is looking ready to me! Many of those blocks of text are definitely easier to read using multiline string literals.

src/Compiler/Scheme/Racket.idr
```diff
@@ -330,42 +335,48 @@ getFgnCall : {auto f : Ref Done (List String) } ->
...
-startRacketCmd racket appdir target = unlines
-    [ "@echo off"
-    , "set APPDIR=%~dp0"
-    , "set PATH=%APPDIR%" ++ appdir ++ ";%PATH%"
-    , racket ++ "\"%APPDIR%" ++ target ++ "\" %*"
-    ]
+startRacketCmd racket appdir target = """
+  @echo off
+
+  rem \{ generatedString "Racket" }
```
andrevidela, line 360: Is this line extra?

andrevidela:
I went through it too, didn't notice anything except for 1 comment

src/Compiler/Scheme/Racket.idr
ska80: Yup

andrevidela: should it be removed? I'm not clear why it's been added

ska80: Are we talking about the vertical whitespace or the comment with a function that inserts CG's name?

If it is the latter, all shell startup wrapper scripts have this function that inserts CG's name, so I added it here for consistency.

andrevidela: It was the latter, seems good to me!

ska80: @andrevidela Thanks for the review!

PR #2052, opened 2021-10-23
Generalize `doParse` to work in any underlying `MonadToken`

gergoerdi:
The expectation of a `MonadToken` is that it implements two effects:
  * Supply the `nextToken`
  * Allow checkpointing of token stream state

The old `doParse` is recovered using `State (List (WithBounds tok))`,
but we can also create new parsers that use e.g. an unpacked byte buffer
and an index into it. This is especially useful for the JS backend, where
just transforming the input into a cons-list representation can lead to a
stack overflow from as little as 40k elements, in my experience.

### DIFF
--- a/libs/contrib/Text/Parser/Core.idr
+++ b/libs/contrib/Text/Parser/Core.idr
@@ -5,6 +5,7 @@ import Data.List
 import Data.List1
 
 import public Control.Delayed
+import Control.Monad.State
 import public Text.Bounded
 
 %default total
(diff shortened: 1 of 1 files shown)


gergoerdi:
So what is the plan with this one? Is there some changes I need to do, is there disagreement that it's a good idea, is it unclear who should even decide on it?

andrevidela:
Given how old this is and the current [contributing guidelines](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md#things-we-probably-wont-accept) I'm going to close this

PR #2060, opened 2021-10-25
Save warnings to TTC

michaelmesser:
Exposes warnings to the LSP by saving them to TTC. Makes it so https://github.com/idris-community/idris2-lsp/pull/103 works.

### DIFF
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -8,6 +8,7 @@ import Core.Case.CaseTree
 import Core.Context
 import Core.Context.Log
 import Core.Core
+import Core.Env
 import Core.Hash
 import Core.Name.Namespace
 import Core.Normalise
@@ -18,6 +19,7 @@ import Core.UnifyState
 
 import Data.Buffer
 import Data.List
+import Data.List1
 import Data.String
 
 import System.File
@@ -33,7 +35,7 @@ import public Libraries.Utils.Binary
 ||| (Increment this when changing anything in the data format)
 export
 ttcVersion : Int
-ttcVersion = 66
+ttcVersion = 67
 
 export
 checkTTCVersion : String -> Int -> Int -> Core ()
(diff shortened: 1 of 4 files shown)


gallais:
I'd say print the warnings for a file when you're loading it. When you're reloading the
file you're currently interacting with it shouldn't reload its dependencies so that would
not be too noisy IMO.

michaelmesser:
I'm very confused by your comment. I believe that dependencies are reloaded on `:reload`. My understanding is that `:reload` calls `resetContext` (clears the dependencies) and `readAsMain` (loads the dependencies). Am I misunderstanding something?

gallais:
I was talking about using `C-c C-l` in emacs to reload a file after having made some edits.

michaelmesser:
Sorry. I'm still not understanding.
Should the current behavior change?
Should if warnings are printed depend on if a TTC file is present?
If TTCs are present should it only print for the main file (no dependencies)?
I personally don't care when warnings are printed, I just need them to be exposed through the API so LSP can access them. I'm not sure if I should expose a `List Warning` for the main file, a `List Warning` containing warnings for the main file and dependencies, or a `SortedMap` of files names to `List Warning`.
Or if I should allow LSP to pass a `(filename : String) -> List Warning -> List Error -> Core ()` function to the API. Now that I think about it more, having LSP pass the function to the API might be the correct way. If I go that route, should I put the function on `REPLOpts` or should it be passed as an argument to `buildDeps`?

Currently this PR keeps Idris behavior the same and exposes a `List Warning` containing warnings for the main file and dependencies.

bkomuves:
I suggest that if there are more than one reasonably expected behaviour (for example LSP needs are different from a command line user's needs), then just make the behaviour configurable.

michaelmesser:
@bkomuves I know what behavior the LSP needs. I'm asking about how the CLI should behave. Even if I make it configurable, I still have to pick a default. The main question is: should warnings be printed for file if it is already compiled?

michaelmesser:
I am redoing this in a different PR.

PR #2074, opened 2021-10-30
Cumulativity preparation: Add field for universe level to TType

edwinb:
This doesn't do anything yet, other than introduce new universe variables whenever we introduce a new type, but it's the first step towards checking the universe hierarchy. Next step is to add constraints when checking pi, unifying/converting types, and when adding data constructors.

I'd like to at least merge this step now, without having completed the implementation, because it's a change to a core data type and I'm hoping it'll mean fewer merge conflicts later.

### DIFF
--- a/tests/idris2/basic049/expected
+++ b/tests/idris2/basic049/expected
@@ -151,7 +151,7 @@ Fld:226:1--226:10
 
 Error: While processing right hand side of r2_shouldNotTypecheck1. Ambiguous elaboration. Possible results:
     Main.R2.MkR
-    Main.R1.MkR Type
+    Main.R1.MkR ?field
 
 Fld:72:26--72:29
  68 | r1 : R1 -- explicit fields access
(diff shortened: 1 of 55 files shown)


tests/idris2/basic049/expected
```diff
@@ -151,7 +151,7 @@ Fld:226:1--226:10
 
 Error: While processing right hand side of r2_shouldNotTypecheck1. Ambiguous elaboration. Possible results:
     Main.R2.MkR
-    Main.R1.MkR Type
+    Main.R1.MkR ?field
```
buzden, line 154: This looks suspicious  Can this PR influence badly on type inference?

edwinb: I think the old output is wrong here. There's no reason for the name to be resolved to a `Type`, and I think it's probably because we need to have `Context` in another place in `Error`.

edwinb: I suppose I should probably test this theory while I have the chance...

edwinb: ...and that is indeed what's going on. There's a few places in the `Error` structure that don't record the context at the point of error. It should be there whenever there's a `Term`. If it's not, we have to be sure the term has been normalised.

Anyway, the updated output is correct. I'll fix the error message display separately.

edwinb:
Actually just to keep things easier for myself I'm going to close this and send an updated version, with a tiny bit more progress, once #2075 is merged.

I don't know how long it'll be before I get around to implementing the rest of the rules, but it'll be nice to have the universe variables as part of TT anyway.

PR #2077, opened 2021-10-31
[base] add IsRight and IsLeft proofs to Data.Either

kbertalan:
IsRight and IsLeft types, like IsJust and IsYes.

I have used IsRight instead of IsJust to provide more meaningful error messages, when the value is a Left.

I have no particular use case for IsLeft, it is included for the sake of completeness.


gallais:
Thanks!

PR #2081, opened 2021-11-01
[ performance ] bitops arithmetic speedup

benhormann:
`modulo` is slow, `bitwise-and` is fast, `fxand` is fastest.

We could specialise small widths to use fixnum operators.
64-bit Schemes have ~61 bits to play with (32-bit has ~30). So that excludes Bits64 entirely, and `fx*` for Bits32 (Bits32 shifts should be possible by truncating beforehand instead of afterwards).

We could also use macros to optimise out `(sub1 (ash 1 bit))`, and clean up the syntax. Macros could even select the right op for the machine size.

Let's get at least this much in before the next release? It opens up options for fast hashing.

---

`b+` appears to be unused as of v0.4.0, superseded by `bs+` and `bu+`.


benhormann:
Hmm, the Gambit bitops test is disabled in Golden. Is there a reason we can't install Gambit to run that bitops test, using one the Chez builds?
Or add `make test IDRIS2_TESTS_CG=gambit only=basic001` somewhere.

gallais:
>  Is there a reason we can't install Gambit to run that bitops test,

The gambit test were *extremely* slow so we had to turn them off. The backend is not actively maintained AFAIK.

Issue #2082, opened 2021-11-01
[Proposal] A new `%deprecate` pragma


mattpolzin:
- [x] I have read [CONTRIBUTING.md](../../CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

I propose we add a `%deprecate` pragma that has the behavior of marking the following definition such that when that definition is used the compiler produces a warning to the effect of "_x_ is deprecated and will be removed in a future version." 

## Motivation

It can be very useful for API and package developers to let consumers know that it is no longer recommended to use part of the API and furthermore that part of the API will be removed in the future.

This allows a package author to phase out old code over the course of more than one release version while giving consumers a heads up that they should change their code.

## The proposal

I propose we add a no-argument `%deprecate` pragma

I suggest "deprecate" as a command much like "nomangle" and "noinline" rather than using the past-tense ("deprecated") that otherwise would also feel natural to me.

The `%deprecate` pragma would be used just prior to a definition that the developer wants to mark as deprecated. This means the following stylistic choices are all valid:
```idris
%deprecate
export
foo : String -> String
foo x = x
```

```idris
export
%deprecate
foo : String -> String
foo x = x
```

```idris
export %deprecate
foo : String -> String
foo x = x
```

```idris
%deprecate export
foo : String -> String
foo x = x
```

The `%deprecate` pragma causes the compiler to produce a warning at the call site, not the definition site. This means the above examples do not produce compiler warnings, but the following example does:
```idris
bar : String
bar = foo "hello"
```

The exact wording of the warning is less critical to the proposal, but I propose the above produce the warning:
```shell
Warning: Deprecation warning: Main.foo is deprecated and will be removed in a future version.
```

It is often useful to provide some guidance to the end-user to help them modify their code to produce the same results without using the now-deprecated function. I propose we co-opt the existing documentation mechanism for this purpose.

For example, suppose that instead the above `foo` was defined as:
```idris
||| Please use @id@ instead. `foo` was always just a specialized form of @id@ anyway.
%deprecate
export
foo : String -> String
foo x = x
```

The warning produced by the above `bar` definition would now be:
```shell
Warning: Deprecation warning: Main.foo is deprecated and will be removed in a future version.
  Please use @id@ instead. `foo` was always just a specialized form of @id@ anyway.
```

### Technical implementation

[...]

PR #2086, opened 2021-11-02
Add %deprecate pragma

mattpolzin:
Add the `%deprecate` pragma as proposed [here](https://github.com/idris-lang/Idris2/issues/2082).

Closes https://github.com/idris-lang/Idris2/issues/2082.

In short, add `%deprecate` before a definition to cause the compiler to generate a warning whenever the given function is applied. Modify or add code documentation (`||| docs`) for the definition to give any additional suggestions you want printed along with the warning.

- Adds `%deprecate` pragma and associated warning.
- Updates CHANGELOG
- Adds beginnings of a "pragmas" page in the reference documentation.


src/Idris/Doc/String.idr
```diff
@@ -454,7 +466,8 @@ getDocsForName fc n config
                   in annotate DocStringBody
                      (concatWith (\l, r => l <+> hardline <+> r) docs)
                      <$ guard (not (null docs))
-             pure (vcat (docDecl :: docBody))
+             let maybeDocDecl = [docDecl | showType]
+             pure . vcat . catMaybes $ maybeDocDecl :: (map Just $ docBody)
```
gallais, line 470: Should the `:doc` output also include a `DEPRECATED` string?

mattpolzin: Yeah, that seems smart. Will do.

gallais (commented):
Does the fact `checkDeprecation` is called from `CompileExpr` mean that we get away with
using deprecated functions in values that have a `0` quantity and are not compiled?
Should this be checked during variable lookup in `TTImp.Elab.App` instead?

mattpolzin:
> Does the fact `checkDeprecation` is called from `CompileExpr` mean that we get away with
> 
> using deprecated functions in values that have a `0` quantity and are not compiled?

Not sure, I'll try that and see. 

> Should this be checked during variable lookup in `TTImp.Elab.App` instead?

Quite possibly. Thanks for the idea. Where to check this was roughly a guess on my part so I only know that it works for the cases I tested but not that it's the best place to put it.

mattpolzin:
Much better results; in broad strokes moving to `Elab.App` was the prefect plan, but let me know if the way I surfaced it there seems reasonable to you!

src/TTImp/Elab/App.idr
```diff
@@ -91,17 +98,27 @@ getNameType rigc env fc x
...
                            (throw (LinearMisuse fc !(getFullName x) lhs rhs))
 
+    checkDeprecation : FC -> GlobalDef -> Core ()
+    checkDeprecation fc gdef =
+      do when (Deprecate `elem` gdef.flags) $
+           recordWarning $
+             Deprecated
+               "\{show gdef.fullname} is deprecated and will be removed in a future version."
+               (Just (fc, gdef.fullname))
+    checkDeprecation _ _ = pure ()
```
buzden, line 108: Isn't this line an unreachable case?

mattpolzin: It certainly is. Fixed!

gallais (approved):

Issue #2032, opened 2021-10-19
Slow typechecking on `Int` operation when `Data.Fin.fromInteger` is in scope


cypheon:
# Steps to Reproduce

```idris
import Data.Fin

addMillion : Int -> Int
addMillion x = 2000000 + x
```

`idris2 --timing --check AddMillion.idr`

# Expected Behavior

Program should typecheck within a reasonable time.

# Observed Behavior

Idris2 uses lots of memory and takes a long time. With a larger number (e.g. 20 million), it uses >8GB ram and is not finished after 5 minutes.

```
1/1: Building AddMillion (AddMillion.idr)
TIMING ++ Parsing AddMillion.idr: 0.000s
TIMING ++ Reading imports: 0.077s
TIMING +++ Check RHS AddMillion:4:1--4:27: 14.894s
TIMING +++ Building compile time case tree for Main.addMillion: 0.000s
TIMING +++ Building size change graphs Main.addMillion: 0.000s
TIMING +++ Checking Coverage Main.addMillion: 0.000s
TIMING +++ Totality check overall: 0.000s
TIMING ++ Processing decls: 14.895s
TIMING ++ Compile defs: 0.000s
TIMING + Elaborating AddMillion.idr: 14.973s
TIMING + Build deps: 15.076s
TIMING + Loading main file: 15.078s
```

# Additional Notes

A bisect shows this was introduced by d4263441b7fe13e483d0363878b05ddef913e280 and is probably related to (the type of?) the new `fromInteger` implementation.
Maybe there is a way to somehow "deprioritize" the `fromInteger` from `Fin` so that `fromInteger : Integer -> Int` is tried first?

# Workarounds

There are a few workarounds available, all of which make typechecking finish within ~200ms:

Reverse the operands:
```idris
addMillion x = x + 2000000
```

Force type of the integer literal:
```idris
addMillion x = the Int 2000000 + x
```

Hide `fromInteger`:
```idris
%hide Data.Fin.fromInteger
addMillion x = 2000000 + x
```

gallais:
> Maybe there is a way to somehow "deprioritize" the fromInteger from Fin so that fromInteger : Integer -> Int is tried first?

I don't think that's possible: the `Int` one comes from the `Num` implementation and the monomorphic
alternative will always be tried first.

andrevidela:
Maybe a warning that large numbers might be slow to compile would help?

gallais:
[...]

PR #2090, opened 2021-11-02
[ re #2032 ] faster version of fromInteger

gallais:
This should be twice as fast


Z-snails:
Faster at runtime or compile time?
It was already the identity function at runtime. (and in many cases, removed entirely)

PR #2078, opened 2021-11-01
[ fix ] loading of dylib on macOS Monterey

chrrasmussen:
Hi,

After upgrading to macOS 12 (Monterey) I was unable to run the `idris2` executable that I had previously installed. Instead I was shown the following error:
```
Exception: (while loading libidris2_support.dylib) dlopen(libidris2_support.dylib, 0x0002): tried: 'libidris2_support.dylib' (no such file), '/usr/local/lib/libidris2_support.dylib' (no such file), '/usr/lib/libidris2_support.dylib' (no such file), '/Users/USER/Idris/Idris2-main/libidris2_support.dylib' (no such file), '/usr/local/lib/libidris2_support.dylib' (no such file), '/usr/lib/libidris2_support.dylib' (no such file)
```

Thanks to @ska80's clarification in INSTALL.md (commit [880981d](https://github.com/idris-lang/Idris2/commit/880981d5b4de57b55f762ef6b13a11ce17c5421d#diff-09b140a43ebfdd8dbec31ce72cafffd15164d2860fd390692a030bcb932b54a0R30-R31)) I was able to fix the problem. It appears that macOS Monterey requires `DYLD_LIBRARY_PATH` to be set (instead of `LD_LIBRARY_PATH`).

In order to make it possible to bootstrap Idris 2 on macOS Monterey from the `main` branch I also needed to update the Chez and Racket bootstrap files: The previous bootstrap files would generate the `idris2` shell script with only `LD_LIBRARY_PATH` specified.

---

Similar error when bootstrapping from Racket:
```
idris2 --build prelude.ipkg
ffi-lib: couldn't open "libidris2_support.dylib" (dlopen(libidris2_support.dylib, 0x0006): tried: 'libidris2_support.dylib' (no such file), '/usr/local/lib/libidris2_support.dylib' (no such file), '/usr/lib/libidris2_support.dylib' (no such file), '/Users/USER/Idris/Idris2-main/libs/prelude/libidris2_support.dylib' (no such file), '/usr/local/lib/libidris2_support.dylib' (no such file), '/usr/lib/libidris2_support.dylib' (no such file))
  context...:
   /Applications/Racket v7.9/collects/ffi/unsafe.rkt:131:0: get-ffi-lib
   '#%mzc:idris2-boot: [running body]
   top-level: [running body]
make[2]: *** [all] Error 1
make[1]: *** [prelude] Error 2
make: *** [bootstrap-racket] Error 2
```

PR #2093, opened 2021-11-03
[ fix ] Detect OS as Darwin when running ARM version of Chez on a Mac

chrrasmussen:
Hi! This change allows one to bootstrap and run idris2 on Macs using an ARM version of Chez (See more info below).

Similar error as #2078, but in this case the error was caused by `System.os` returning the wrong OS.
```
Exception: (while loading libidris2_support.so) dlopen(libidris2_support.so, 0x0002): tried: '/Users/USER/Idris/Idris2-main/build/exec/idris2_app/libidris2_support.so' (no such file), '/libidris2_support.so' (no such file), 'libidris2_support.so' (no such file), '/usr/local/lib/libidris2_support.so' (no such file), '/usr/lib/libidris2_support.so' (no such file), '/Users/USER/Idris/Idris2-main/build/exec/idris2_app/libidris2_support.so' (no such file), '/libidris2_support.so' (no such file), '/Users/USER/Idris/Idris2-main/libs/prelude/libidris2_support.so' (no such file), '/usr/local/lib/libidris2_support.so' (no such file), '/usr/lib/libidris2_support.so' 
```

Note that there is only a single line of change in `bootstrap/idris2_app/idris2.ss`.

---

For anyone interested:
The [Racket organization](http://github.com/racket) has a fork of Chez Scheme that supports ARM. I installed it by running the following commands:
1. `git clone https://github.com/racket/ChezScheme`
2. `cd ChezScheme`
3. `git submodule init`
4. `git submodule update`
5. `./configure --pb`
6. `make tarm64osx.bootquick`
7. `./configure --threads` (`--threads` is important to make it work with Idris 2)
8. `make`
9. `sudo make install`

This installs `scheme` to `/usr/local/bin`.

### DIFF
--- a/support/chez/support.ss
+++ b/support/chez/support.ss
@@ -4,7 +4,7 @@
     [(i3ob ti3ob a6ob ta6ob) "unix"]  ; OpenBSD
     [(i3fb ti3fb a6fb ta6fb) "unix"]  ; FreeBSD
     [(i3nb ti3nb a6nb ta6nb) "unix"]  ; NetBSD
-    [(i3osx ti3osx a6osx ta6osx) "darwin"]
+    [(i3osx ti3osx a6osx ta6osx tarm64osx) "darwin"]
     [(i3nt ti3nt a6nt ta6nt) "windows"]
     [else "unknown"]))
 


Russoul:
Thanks a lot for this, @chrrasmussen! I had no idea we have an ARM chez.

PR #2096, opened 2021-11-03
Add the `System` `run` function

madman-bob:
Adds the `System.run` function, which takes a shell command, and runs it, returning the stdout and return code of that run.

Also includes:

- The `escapeCmd` utility function for escaping Idris strings as shell string literals.
- `system`, `popen`, and `run` now have "escaped" equivalents, which take lists of arguments, and escapes them.
  eg. `run ["printf", "Hello, %s", "$PATH"]` doesn't expand the `$PATH`.
- `pclose` now returns the return code of the closed process.
- The `fRead` function, which reads all remaining contents of an open file handle.
- A `fastConcat` implementation for `SnocList`.
- Tests for all of the above.


madman-bob:
Good grief Windows.

I was not able to find an official source for escaping rules in the Windows command prompt, so I'm not certain that what I have is comprehensive.

Further, there seems to be no certain way of escaping the newline character. So if you want to use it in a portable shell command... good luck.

mattpolzin (commented):
This is great! I think it warrants a CHANGELOG entry, not only because it's a substantial feature but also because it's a breaking change to pclose.

PR #2100, opened 2021-11-05
[ opt ] transform rule making List.length tail-recursive.

falsifian:


libs/prelude/Prelude/Types.idr
```diff
@@ -413,6 +413,21 @@ namespace List
...
+  ||| Returns the first argument plus the length of the second.
+  public export
+  lengthPlus : Nat -> List a -> Nat
+  lengthPlus n [] = n
+  lengthPlus n (x::xs) = lengthPlus (S n) xs
+
+  ||| `length` implementation that uses tail recursion. Exported so
+  ||| tailRecLengthIsLength can see it.
+  public export
+  tailRecLength : List a -> Nat
```
Z-snails, line 425: Before we get too many tail recursive equivalents, lets discuss naming.

I'm a fan of `lengthTR` because it's got the important bit at the start of the name, and not too much space is taken up by the fairly unimportant tail-recursive bit.

andorp: I like the initiative, and we need tail-rec functions. I think this is a symptom of a wider problem. We should implement an automatic CPS transformation of recursive functions as an optimization at the NamedCExp level. I think not having tail-recursive functions in the Scheme backend is not a problem, but it is a problem for other backends, and that problem may be better solved at the IR level, rather than the source level. I gave it a go but has failed, I couldn't allocate focused time on implementing such transformation, but maybe someone else is able to pick that up?

falsifian: > I'm a fan of `lengthTR` because it's got the important bit at the start of the name, and not too much space is taken up by the fairly unimportant tail-recursive bit.

I'm not a fan of abbreviations except really obvious ones, but will change to `lengthTR` if others prefer; that's a nice suggestion.

> We should implement an automatic CPS transformation of recursive functions as an optimization at the NamedCExp level.

+cc @stefan-hoeck who was looking into something Javascript-specific a couple of months ago.

I would be more than happy to see the transformation rules go away after they are no longer needed. My motivation right now is that I'm working on a web toy I'm hoping to share at some point, and keep having to make changes like this. So my ideal would be to have this PR (and possibly future ones) accepted and then rolled back once someone implements a better fix.

andorp: | My motivation right now is that I'm working on a web toy 
Yep, that is a good motivation and drive for any addition we do. Of course this solves the issue you face with and it is a good contribution that helps other too. 

| someone implements a better fix.
Shamefully I failed that, maybe one day... ;)

stefan-hoeck: > > We should implement an automatic CPS transformation of recursive functions as an optimization at the NamedCExp level.
> 
> +cc @stefan-hoeck who was looking into something Javascript-specific a couple of months ago.

To no avail so far, I'm afraid. Problem is, I have no experience with CPS transformations. Also, I'm not sure how this would affect performance (of the compiler *and* the generated code), nor am I even sure we need this. For instance, I wonder why a JS oriented purely functional language like PureScript does not perform such transformations. Even in PureScript there is the need of a specialized library for tail-call optimized monadic recursion. And yes, just because others don't do it does not mean we shouldn't do this, but as I said, without some guidance I don't know where to start.

Right now Idris2 produces JavaScript that performs really well in the browser, and we have the ability to write tail-call optimized monadic recursive functions as described in the idris-tailrec project (a demonstration of this can be seen [here](https://stefan-hoeck.github.io/idris2-rhone-js)). I use these in my personal projects and it works really well. Again, no reason not to got for a better solution, but please let us not sacrifice the performance we get right now.

@andorp: Since you already gave this a try: Do you have resources about how to perform CPS transformations (probably followed by a defunctionalization step), if possible for a functional language similar to Idris2? Or do you have code snippets to share with what you already tried? Finally, @falsifian , it might be easier to take the discussion about how to implement such a thing to the code-gen channel on discord.

buzden: > I wonder why a JS oriented purely functional language like PureScript does not perform such transformations

As far as I know, they force every piece of library code to use CPS and trampolining, so no automatic transformations exist. They even have special definitions of complex interfaces like monad, which are significantly different and a bit harder to implement than haskell-like ones. This paper describes this: http://functorial.com/stack-safety-for-free/index.pdf
Other eager languages, like Scala, also adopted this in their FP libraries.

stefan-hoeck: > As far as I know, they force every piece of library code to use CPS and trampolining, so no automatic transformations exist. They even have special definitions of complex interfaces like monad, which are significantly different and a bit harder to implement than haskell-like ones. This paper describes this: http://functorial.com/stack-safety-for-free/index.pdf Other eager languages, like Scala, also adopted this in their FP libraries.

That's correct and it is the reason why I wrote [idris2-tailrec](https://github.com/stefan-hoeck/idris2-tailrec), which takes the ideas from the article you linked to, and adapts them for a language with a totality checker. The question of course is, what we in Idris should be doing about this. Should we a) go the same way as PureScript and Scala, putting the burden on programmers to come up with tail-recursive versions of their functions and writing transform rules for the ones in our standard libs, or should we b) try and have the compiler do these optimizations for all recursive functions? Since I'm uncertain about the impact of b) on performance and readability/debugability of the generated code, I'm tending towards a) for the time being.

ohad:
How do these relate to `contrib`'s `Data.List.TailRec`? (https://github.com/idris-lang/Idris2/blob/main/libs/contrib/Data/List/TailRec.idr)?

falsifian:
@ohad I hadn't heard of that.

I don't know if that library will solve my problem. Libraries I depend on might themselves use the default, non-tail-recursive implementations. So if the prelude/base versions don't get changed, I might have to send pull requests to the other libraries I use, changing them to use `Data.List.TailRec`.

ohad:
We might want to change the prelude versions, but then remove the duplication in `contrib`, perhaps by 'upgrading' the `contrib` code into `base/prelude`.

libs/prelude/Prelude/Types.idr
falsifian: > I'm a fan of `lengthTR` because it's got the important bit at the start of the name, and not too much space is taken up by the fairly unimportant tail-recursive bit.

Changed. I kept the name of the `%transform` rule as `tailRecLength` but changed the other occurrences to `lengthTR`.

falsifian: For right now, is there still debate about whether this PR should go in?

The broader conversation about what to do with recursion is interesting. @stefan-hoeck thanks for the Discord invitation; I don't know if I'll have time (and don't normally use Discord) but maybe I will drop by some time.

falsifian:
> We might want to change the prelude versions, but then remove the duplication in `contrib`, perhaps by 'upgrading' the `contrib` code into `base/prelude`.

That sounds reasonable. Maybe the `contrib` library could have `length = lengthTR`, in case someone looks for the code in that library.

It doesn't seem urgent, though. Is anything blocking this PR?

ohad:
> I don't know if that library will solve my problem. Libraries I depend on might themselves use the default, non-tail-recursive implementations. So if the prelude/base versions don't get changed, I might have to send pull requests to the other libraries I use, changing them to use `Data.List.TailRec`.

Is there a general mechanism here?
That is, can a `%transform` retroactively (perhaps through recompilation) change the implementation of already compiled libraries?
In that case, maybe the right solution is to have a library of `%transform`s that once you import, turns functions into their tail-recursive counterparts, even retroactively.

(In the future, when we have stronger notions of equality, maybe we could even have a `%safe_transform` that requires a proof that the source and target of the transformation behave the same.)

falsifian:
> Is there a general mechanism here? That is, can a `%transform` retroactively (perhaps through recompilation) change the implementation of already compiled libraries?

Earlier, I briefly tried just putting my `%transform` in a separate library within my own project, but that didn't seem to do anything. So unless there's some other method I don't know about, I think it's currently necessary to make the change right here.

falsifian:
I rebased on `main` and force-pushed. Still works for me.

The failing check looks like a bug in the check as far as I can tell: `/action/lib/linter.sh: line 57: /action/lib/functions/tapLibrary.sh: No such file or directory`.

Anything I can do to help get this merged?

falsifian:
Monthly-ish ping.
* I have released [some code](https://git.sr.ht/~falsifian/s3d) that depends on this PR. It may be possible to work around if the PR is not accepted, but it would require me to either patch a library I depend on which uses `length`, or re-implement the function I am using from that library.
* Is it possible to re-run the checks? The failing one gives this weird error which I would guess is unrelated: `/action/lib/linter.sh: line 57: /action/lib/functions/tapLibrary.sh: No such file or directory`

mattpolzin:
> Is it possible to re-run the checks? The failing one gives this weird error which I would guess is unrelated

The linter failure should be resolved if you merge `main` into this branch again. The version of linter was updated in the CI config in a later commit to address these spurious linter failures.

PR #2101, opened 2021-11-05
Use escaped version of `system` internally

madman-bob:
Follow on from #2096. Use the new escaped version of `system` in the source of Idris itself.

This closes a couple of potential security holes, and makes the handling of these sorts of issues more uniform.

This PR will fail tests until the next version bump. I'm submitting the PR now, so that I don't forget about it.

### DIFF
--- a/.github/workflows/ci-idris2.yml
+++ b/.github/workflows/ci-idris2.yml
@@ -33,7 +33,7 @@ on:
       - '.github/workflows/ci-super-linter.yml'
 
 env:
-  IDRIS2_VERSION: 0.5.1 # For previous-version build
+  IDRIS2_VERSION: 0.6.0 # For previous-version build
 
 jobs:
 
--- a/src/Compiler/RefC/CC.idr
+++ b/src/Compiler/RefC/CC.idr
@@ -41,8 +41,8 @@ findLDFLAGS
          | Just ldflags => pure ldflags
        pure ""
 
-clibdirs : List String -> String
-clibdirs ds = concat (map (\d => "-L" ++ d ++ " ") ds)
+clibdirs : List String -> List String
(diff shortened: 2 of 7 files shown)


libs/base/Data/SnocList.idr
```diff
@@ -30,6 +30,8 @@ export
 Cast (List a) (SnocList a) where
   cast xs = Lin <>< xs
 
+%transform "fastConcat" concat {t = SnocList} {a = String} = fastConcat . cast
```
mattpolzin, line 33: Can you pull this out into a separate PR (and get it merged sooner)?

mattpolzin: Now I see that was part of the system run PR and therefore will get merged sooner.

madman-bob:
As we've now had a version bump, I've rebased this. I thought that these functions should now be available to the core, but the tests say no. Have I misunderstood something?

mattpolzin:
CI is still enforcing compatibility with v0.5.1 until we decide it's worth requiring v0.6.0 (dropping support to build the compiler with v0.5.1). 

I'm of the opinion it's worth it for almost any quality of life, code cleanliness, or efficiency gains in the compiler code base right now, but the idea is solid: eventually it would be nice to have source code that supports an arbitrary number of versions back (not just one, but as many as possible). 

Anyway, just update the IDRIS_VERSION environment variable in this file: https://github.com/idris-lang/Idris2/blob/main/.github/workflows/ci-idris2.yml

madman-bob:
I've bumped the version number. Now only the Windows CI is failing.

My understanding is that the checked-in Scheme version is for bootstrapping, while the Idris version is for development. So there is little holding back the version number. I'd want to move to the new version as soon as possible. To avoid confusion in future - maybe even make it part of the release process.

PR #2102, opened 2021-11-05
Add List singleton function

madman-bob:
Adds the `List` singleton function. Honestly, I'm suprised it's not already in `base`.

We could use `pure` instead, but that's not always the semantic meaning we try to get across. Further, the presence of `singleton` for a variety of list-like types, but not `List` itself, could be confusing.


andrevidela (approved):
looks good to me!

PR #2103, opened 2021-11-05
Add Alternating List fst and snd functions

madman-bob:
Adds the Alternating List `fst` and `snd` functions, which take the sublists of alternating lists corresponding to the odd and even elements, respectively. Named for consistency with the `Bi...` interfaces, and by analogy with `Pair`.

That is, `fst` and `snd` are the obvious non-trivial functions of type:

```idris
fst : Odd a b -> List a
snd : Odd a b -> List b
```

Also adds tests for these functions.

### DIFF
--- a/libs/contrib/Data/List/Alternating.idr
+++ b/libs/contrib/Data/List/Alternating.idr
@@ -214,6 +214,30 @@ Traversable (Odd a) where
 mutual
     namespace Odd
         public export
+        odds : Odd a b -> List a
+        odds (x :: xs) = x :: evens xs
+
+    namespace Even
+        public export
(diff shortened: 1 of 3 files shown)


mattpolzin:
I'm actually not very opinionated about this name choice but to throw out one other idea for consideration (inspired by `lefts`/`rights`): `odds`/`evens`.

madman-bob:
That makes sense. I had considered `odd`/`even`, but thought that might be confusing as we also have `Odd`/`Even`. Phrasing it instead as `odds`/`evens` seems to work well.

gallais:
This feels like some kind of `unzip`.

PR #2114, opened 2021-11-10
Injective interface and its implementations

andrevidela:
Add an `Injective` interface in `base` that you can call with `injective` or with `inj f` where `f` is the function over which the injection happens.

This is a copy of @nickdrozd 's PR with my commits added at the end, the preivous PR is here: #1761

### DIFF
--- a/libs/base/Control/Function.idr
+++ b/libs/base/Control/Function.idr
@@ -1,0 +1,5 @@
+module Control.Function
+
+public export
+interface Injective (f : a -> b) where
+  injective : f x = f y -> x = y
--- a/libs/base/Control/Function.idr
+++ b/libs/base/Control/Function.idr
@@ -6,5 +6,5 @@ interface Injective (f : a -> b) where
   injective : {x, y : a} -> f x = f y -> x = y
 
 public export
-inj : (f : a -> b) -> Injective f => {x, y : a} -> f x = f y -> x = y
-inj _ = injective
+inj : (0 f : a -> b) -> Injective f => {0 x, y : a} -> (0 _ : f x = f y) -> x = y
+inj _ eq = irrelevantEq (injective eq)
--- a/libs/base/Control/Function.idr
+++ b/libs/base/Control/Function.idr
@@ -6,5 +6,5 @@ interface Injective (f : a -> b) where
   injective : {x, y : a} -> f x = f y -> x = y
 
 public export
-inj : (0 f : a -> b) -> Injective f => {0 x, y : a} -> (0 _ : f x = f y) -> x = y
+inj : (0 f : a -> b) -> {auto 0 _ : Injective f} -> {0 x, y : a} -> (0 _ : f x = f y) -> x = y
 inj _ eq = irrelevantEq (injective eq)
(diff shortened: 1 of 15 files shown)


libs/base/Data/List1.idr
```diff
@@ -164,6 +165,16 @@ export
 consInjective : (x ::: xs) === (y ::: ys) -> (x === y, xs === ys)
 consInjective Refl = (Refl, Refl)
 
+%hint
+export
+consHeadInj : {x : a} -> Injective (x :::)
+consHeadInj = MkInjective (\Refl => Refl)
```
gallais, line 171: Is there a reason for naming these?

andrevidela: Not particularly, I can try to implement them as `implementation` if that's what you're asking

gallais (commented):
ugh that last comment somehow stayed in the drafts

libs/base/Control/Function.idr
```diff
@@ -0,0 +1,10 @@
+module Control.Function
+
+public export
+interface Injective (f : a -> b) where
+  constructor MkInjective
+  injective : {x, y : a} -> f x = f y -> x = y
+
+public export
+inj : (f : a -> b) -> Injective f => {x, y : a} -> f x = f y -> x = y
```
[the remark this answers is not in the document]
andrevidela: Oh absolutely right, it should!

nickdrozd:
Hey @andrevidela, thanks for working on this. The extra implementations look good. My only comment is that the usual implementation style should be used whenever possible, rather than defining functions that build implementations. (That's what's happening, right?)

libs/contrib/Data/Nat/Factor.idr
```diff
@@ -124,8 +123,7 @@ multOneSoleNeutral (S k) (S (S j)) prf =
         rewrite plusCommutative k j in
         rewrite sym $ plusAssociative j k (k * S j) in
         rewrite sym $ multRightSuccPlus k (S j) in
-        succInjective k (j + (S (S (j + (k * (S (S j))))))) $
-        succInjective (S k) (S (j + (S (S (j + (k * (S (S j)))))))) prf
+        inj S $ injective prf
```
nickdrozd, line 126: I think it would be clearer if this were two direct calls to `injective`, rather than one to the `inj` wrapper.

andrevidela: The `inj` wrapper is necessary because `f` cannot be inferred here. We could pass it implicity, but this is exactly what the `inj` wrapper is for

libs/base/Control/Function.idr
```diff
@@ -6,5 +6,5 @@ interface Injective (f : a -> b) where
   injective : {x, y : a} -> f x = f y -> x = y
 
 public export
-inj : (f : a -> b) -> Injective f => {x, y : a} -> f x = f y -> x = y
-inj _ = injective
+inj : (0 f : a -> b) -> Injective f => {0 x, y : a} -> (0 _ : f x = f y) -> x = y
```
gallais, line 9:
```suggestion
inj : (0 f : a -> b) -> {auto 0 _ : Injective f} -> {0 x, y : a} -> (0 _ : f x = f y) -> x = y
```

Given that we have implementations like `{x : a} -> Injective (x :::) where`, it does make a difference to have `Injective` at 0.

gallais (approved):

PR #2119, opened 2021-11-11
[ re #238 ] Fix program error condition of `git diff` call in `Golden`

buzden:
Manage the fact that `system()` can (and, actually, have to) use non-negative exit codes to mark an error, `127` in particular.

Also, manage the fact that some `libc` implementations (`glibc` in particular) do shift left the exit code by 8 bits for some reason:
- [how bad call is managed using `127` and `W_EXITCODE`](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/system.c;h=48668fb392e67d1a84bb8ffef480e4237b7b4645;hb=HEAD#l180)
- [how good call is managed, using `__waitpid`](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/system.c;h=48668fb392e67d1a84bb8ffef480e4237b7b4645;hb=HEAD#l171)
- [how `__waitpid` is implemented through `wait4`](https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/waitpid.c;h=d46f57ea133454b5034473ba6a855494636f74cd;hb=HEAD#l38)
- [how `wait4` returns its code through `W_EXITCODE`](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/wait4.c;h=c8f9d3d0c6af0d4347a3e54afcdf959232340c5b;hb=HEAD#l78)
- [how `W_EXITCODE` is defined through `__W_EXITCODE`](https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/sys/wait.h;h=5325bfa4d2e5586739d7dc84878dc58c9feca50b;hb=HEAD#l68)
- [how `__W_EXITCODE` shifts by 8 bits](https://sourceware.org/git/?p=glibc.git;a=blob;f=bits/waitstatus.h;h=baaf96ac04490de6aea6719acdd27bfd88fded65;hb=HEAD#l56)

These treatments actually are needed when `--interactive` mode is used, but no `git` is available. The current exit code treatment does not show expected and existing values at all in this case.

### DIFF
--- a/support/c/idris_system.c
+++ b/support/c/idris_system.c
@@ -2,5 +2,10 @@
 #include "idris_system.h"
 
 int idris2_system(const char* command) {
-    return system(command);
+    int status = system(command);
+    #ifdef _WIN32
+      return status;
+    #else
+      return WIFEXITED(status) ? WEXITSTATUS(status) : -1;
+    #endif
 }
--- a/libs/test/Test/Golden.idr
+++ b/libs/test/Test/Golden.idr
@@ -253,6 +253,9 @@ runTest opts testPath = forkIO $ do
     expVsOut : String -> String -> List String
     expVsOut exp out = ["Expected:", maybeColored Green exp, "Given:", maybeColored Red out]
 
+    badSystemExitCode : Int -> Bool
+    badSystemExitCode code = code < 0 || code == 127 {- 127 means shell couldn't start -}
+
     mayOverwrite : Maybe String -> String -> IO ()
     mayOverwrite mexp out = do
       case mexp of
@@ -267,7 +270,7 @@ runTest opts testPath = forkIO $ do
             escapeArg testPath ++ "/expected " ++ escapeArg testPath ++ "/output"
           putStr . unlines $
             ["Golden value differs from actual value."] ++
-            (if (code < 0) then expVsOut exp out else []) ++
+            (if badSystemExitCode code then expVsOut exp out else []) ++
             ["Accept actual value as new golden value? [y/N]"]
       b <- getAnswer
       when b $ do Right _ <- writeFile (testPath ++ "/expected") out


madman-bob:
As the exit code shifting affects all uses of `system`, I'd prefer it be fixed at that level, rather than just patching around it afterwards.

I recently did [something similar](https://github.com/idris-lang/Idris2/pull/2096/files#diff-28d1143aa61bfad6e5a2dde636d861e490b1f213d1ce8f6188b3bcfa633f4719) for `pclose`, which I expect can be transfered easily enough to `system`.

buzden:
> As the exit code shifting affects all uses of `system`, I'd prefer it be fixed at that level, rather than just patching around it afterwards.

I agree with you. I force-pushed a fix.

However, still, check in `Golden` needs to be fixed because not only code shifting plays role here, but special `127` code is returned according to docs when shell fails.

support/c/idris_system.c
```diff
@@ -2,5 +2,10 @@
 #include "idris_system.h"
 
 int idris2_system(const char* command) {
-    return system(command);
+    int status = system(command);
+    #ifdef _WIN32
+      return status;
+    #else
+      return WIFEXITED(status) ? WEXITSTATUS(status) : -1;
```
mattpolzin, line 9: I am unclear (perhaps you already know the answer) based on the documentation about whether `WIFEXITED()` will return 0 or non-zero when passed `-1`. At least one page I read indicated it might result in `255` when given `-1`.

The reason I ask, is because the documentation for `system()` indicates you get `-1` if the shell cannot be created or no exit status can be determined (and that `-1` is _without_ applying `WEXITSTATUS()`) so I wonder if we want to explicitly check for `-1` prior to doing the ternary you have here right now.

ref. https://man7.org/linux/man-pages/man3/system.3.html#RETURN_VALUE

buzden: `WIFEXITED` returns `true` (i.e. non-zero) in the case when status is produced by calls to `exit` or `_exit`. Those functions cannot be used to make the status to be `-1` because only least significant 8 bits are passed according to documentation and `int` type cannot be less than two bytes according to the language spec. Also, if you test `WIFEXITED`  on `-1`, you expectedly get `0`. That said, I think the present check is sufficient.

mattpolzin: Sounds good. Thanks!

mattpolzin (approved):

PR #2121, opened 2021-11-14
[ mkdoc ] Wrap declaration part in <code> tag

gemmaro:
In this pull request, the declaration code in the HTML document generated by `--mkdoc` option will be changed to be enclosed in `<code>`.

This should improve the semantic meaning of the generated HTML markup.

The "declaration part" here means the area shown in the screenshot below:

Screenshot 2021-11-14 at 22-47-41 Libraries Text PrettyPrint Prettyprinter SimpleDocTree

## See also

* A very similar past example was found in Rustdoc.
  * Issue: [rustdoc, when I use google translate, that translate the doc as well as the code.](https://github.com/rust-lang/rust/issues/88020)
  * Pull request: [[rustdoc] Wrap code blocks in `<code>` tag](https://github.com/rust-lang/rust/pull/88093)


gallais:
I wonder whether this should be done with a new `Code` annotation in `Idris.Doc.String` or is it
the case that `Decl` always wraps (and is the only thing wrapping) code?

PR #2123, opened 2021-11-16
Remove unused imports

mattpolzin:
I'm working on a PR that will introduce a warning for when an import is not used in a module. That work is not done, but in the meantime I thought it would be smart to open a PR that removes the majority of those unused imports.

If others think this kind of cleanup is beneficial, it is nice to have a standalone PR removing the imports because there are so many of them -- this way, reviewing the PR is actually manageable because beyond making sure CI still passes you just need to skim the removals to verify no fishy changes are being made.

A couple of questions:
1. Is this a desirable thing to do? I personally like removing unused imports from my own code because I find it makes the import list easier to work with and comprehend. But should it be forced upon the compiler codebase?
2. Does it make things faster? I would have thought so, but it turns out to be negligible with respect to speed. The current `main` branch compiles (using version 0.5.1 of the compiler) in 3 minutes 20 seconds on my machine. This branch, by contrast, compiles in 3 minutes 19 seconds. A whopping 1 second improvement! These durations were _very_ consistent, so the 1 second does appear to be _real_, just not very significant.

I strongly suggest that review of this PR takes the form of skimming the changes to make sure they are all removed imports (except for the 1 import that is made public to make it more ergonomic to use `List.NonEmpty` without requiring explicit import of `Data.List` when only `Data.List.Extra` is needed). It would be a massive endeavor with no clear benefit to try to reason about whether the imports I removed make sense to remove. That's the whole reason for a warning after all -- let the computer figure out what imports aren't needed.

### DIFF
--- a/libs/base/Control/App/FileIO.idr
+++ b/libs/base/Control/App/FileIO.idr
@@ -2,8 +2,6 @@ module Control.App.FileIO
 
 import Control.App
 
-import Data.List
-import Data.String
 import System.File
 
 %default covering
--- a/libs/base/Control/WellFounded.idr
+++ b/libs/base/Control/WellFounded.idr
@@ -2,7 +2,6 @@ module Control.WellFounded
 
 import Control.Relation
 import Data.Nat
-import Data.List
 
 %default total
 
--- a/libs/base/Data/Colist.idr
+++ b/libs/base/Data/Colist.idr
@@ -1,6 +1,5 @@
 module Data.Colist
 
-import Data.Maybe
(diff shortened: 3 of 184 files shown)


gallais (approved):

benhormann:
I second this 👍. Sort them at the same time?

I was wondering about implementing this with a linter / formatter.

Rebuilding with changes in unused imports cause otherwise unneeded recompilation.

gallais:
Let's ship before we start creating conflicts x)

PR #2134, opened 2021-11-19
[ fix ] Forbid "." as namespace identifier

claymager:
#2124 allowed `idris2 --find-ipkg <srcfile>` when `sourcedir  =  "."`, but broke `idris2 --build <ipkg>`. This allows both.


gallais:
:scream: 

I guess we should add a bunch of tests for these.

mattpolzin:
Can we support and test "./" while we're at it? That's the most common way I spell "current directory" (although I've never specified the source directory to be the current directory in an idris package file myself).

claymager:
Currently testing:
-  "."
-  "./"
- "./."
- "./src/"
- "../src" (without --find-ipkg)

PR #2136, opened 2021-11-20
Fixes problem where when compiling using "load-file" in ide mode causes compilation to run slower and slower each time.

redfish64:
When using idris2 --ide-mode, with a package file that depends on other packages, every time "LoadFile" is run, extra_dirs gets the packages being depended on tacked onto it.
After awhile, extra_dirs becomes extremely long, and compilation will slow down because of this, and the only way to fix it is to kill idris2 and start again.

This happens because "process (LoadFile fname_in _)" in IDEMode/REPL.idr calls findIpkg in Pakckage.idr. findIpkg calls addPkgDir for each of the depended on packages. addPkgDir calls addExtraDir, which adds the dir to extra_dirs

This commit fixes this, but I'm not sure if it's the best way to go about doing it. The problem is that process (LoadFile fname_in _) needs findIpkg to populate extra_dirs at least once, so I'm unsure of what would be a better solution. Maybe searching extra_dirs and only adding packages that weren't already there might be better. Or another idea would be right before loading the file, reset Ctxt to the initial state as it was when idris was started.


gallais:
Is the principled solution to instead use a `SortedSet` of extra dirs to avoid duplicates?

redfish64:
> Is the principled solution to instead use a `SortedSet` of extra dirs to avoid duplicates?

Even if that was done, it would still allow findIpkg to add entries to extra_dirs that stick around as long as the Idris process running. So say the ipkg file contained a dependency on package A. If the user runs "LoadFile", then changes the ipkg file to no longer contain package A, but instead package B, then runs "LoadFile" again, now the extra_dirs would contain both packages A and B, but only as long as the process is running. When the process is killed and restarted, when running "LoadFile", "extra_dirs" would only contain package B. This behavior seems non-obvious to the user.

IMO, the principled solution would be to make calling "LoadFile" return the same result regardless if the process is killed and restarted or not between calls.

gallais:
Sorry this took so long to come back to. Your thinking makes sense to me now.

redfish64:
No problem,  I realize how busy you probably are.

PR #2137, opened 2021-11-20
[base]: `minimum` and `maximum` for `Vect`

alissa-tung:
Some (and other) combinators I thought helpful, especially implement them would meet some rewriting issues.


andrevidela:
Do we want this in base @gallais @edwinb ?

alissa-tung:
> Do we want this in base @gallais @edwinb ?

GHC has `minimum` and `maximum` in base, seems they are somehow idiomatic? Although they are implemented differently (those take terms on another combinator specialised for `Vect` and `Fin`).

gallais:
The implementation looks way over-complicated to me. I'm also not sure
it's necessary to return the `Fin n`: you can always zip your vec with
all fins & then use the comparison operator that only looks at the second
component if you want to grab the index too.

alissa-tung:
> The implementation looks way over-complicated to me. I'm also not sure it's necessary to return the `Fin n`: you can always zip your vec with all fins & then use the comparison operator that only looks at the second component if you want to grab the index too.

Turns out that it can be done by zip and fold.

PR #2140, opened 2021-11-22
[ type inference ] use determining parameter in relation interfaces

stefan-hoeck:
Type inference for relation interfaces like `Reflexive` can be improved by using `rel` as a determining parameter for those interfaces. For instance, the following proof (taken from a [discussion on discord](https://discord.com/channels/827106007712661524/838742669953204224/912149038814601277)) requires explicit type annotations in the call to `reflexive`:

```idris
import Data.Nat
import Control.Relation

minusLTE : (n : Nat) -> (m : Nat) -> LTE (n `minus` m) n
minusLTE 0 m = LTEZero
minusLTE (S k) 0 = LTESucc (reflexive {ty = Nat, rel = LTE})
minusLTE (S k) (S j) = lteSuccRight (minusLTE k j)
```

With the changes from this PR, this will no longer be necessary.

### DIFF
--- a/libs/base/Control/Relation.idr
+++ b/libs/base/Control/Relation.idr
@@ -16,57 +16,57 @@ Rel ty = ty -> ty -> Type
 
 ||| A relation is reflexive if x ~ x for every x.
 public export
-interface Reflexive ty rel where
+interface Reflexive ty rel | rel where
   constructor MkReflexive
   reflexive : {x : ty} -> rel x x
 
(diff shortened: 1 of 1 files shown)


gallais:
Strange because `reflexive {rel = LTE}` is enough for `ty` to be
reconstructed but somehow idris cannot figure `rel` out unless it's
marked as a determining parameter?

PR #2146, opened 2021-11-24
Make RefC search for files in data dirs

claymager:
Previously, the RefC files were located in IDRIS2_PREFIX. This is decoupled to allow users to change the prefix away from where Idris2 is installed (for ad-hoc library install locations, for example).

This is done by moving `refc/` and `include/` into `support/` with the data for the other backends, which can be included in IDRIS2_DATA.

My primary motivation is that this (or something like it) will allow using the refc and derivative backends in Nix, while still letting users install local packages the normal way. This is probably also relevant for other package managers.


claymager:
So it's saying `macos-self-host::Test self-hosted` failed, but I can't see the logs, and I've found a (late-model) macbook and tests are passing locally. Any ideas what might have gone wrong?

gallais:
I downloaded the logs & the test ones (that the build seemed stuck on) were not present.
I restarted the build and this time it looks like it succeeded.

madman-bob:
Looks good to me. My only comment is that it might make sense to also rename the `include` folder to `c` for consistency with the other folder names in `support`.

Note that we would then have both the folders `c` and `refc` in `support`. The `c` folder contains support functions common across all backends that use C, while the `refc` folder contains support functions specific to the RefC backend.

gallais:
Okay after merging `main` back into `refc` we're finally seeing github recognise that the PR is
now essentially empty. I'm not sure what happened. This was really strange.

PR #2155, opened 2021-11-26
Injective follow-up

nickdrozd:
A few additions and changes related to the new `Injective` interface.

### DIFF
--- a/libs/base/Control/Function.idr
+++ b/libs/base/Control/Function.idr
@@ -23,3 +23,7 @@ public export
 [InjFromComp] {f : a -> b} -> {g : b -> c} ->
   Injective (g . f) => Injective f where
     injective prf = injective {f = (g . f)} $ cong g prf
+
+public export
+[IdInjective] Injective Prelude.id where
+  injective = id
--- a/libs/base/Control/Function.idr
+++ b/libs/base/Control/Function.idr
@@ -1,5 +1,6 @@
 module Control.Function
 
+||| An injective function maps distinct elements to distinct elements.
 public export
 interface Injective (f : a -> b) where
   constructor MkInjective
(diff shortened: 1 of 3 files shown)


libs/base/Control/Function.idr
```diff
@@ -1,10 +1,21 @@
...
+  (Injective f, Injective g) => Injective (g . f) where
+    injective prf = injective $ injective prf
+
+||| If (g . f) is injective, so is f.
 public export
-inj : (0 f : a -> b) -> {auto 0 _ : Injective f} -> {0 x, y : a} -> (0 _ : f x = f y) -> x = y
-inj _ eq = irrelevantEq (injective eq)
+{f : a -> b} -> {g : b -> c} ->
+  Injective (g . f) => Injective f where
+    injective prf = injective {f = (g . f)} $ cong g prf
```
gallais, line 21: These should probably be named instances because if you take `id` as one of the functions, they will always be candidates.

Also, `inj` should be kept.

nickdrozd: Added back `inj` and named the implementations.

nickdrozd: As far as I can tell it isn't possible to implement `Injective id` because these interfaces can only be implemented for type constructors. Is that intended or is it a bug? Or is there a way to do it?

buzden: > As far as I can tell it isn't possible to implement `Injective id`

Mmmm. This typechecks fine:

```idris
import Control.Function
                       
Injective Prelude.id where
  injective = id
```

nickdrozd: Ah, `Prelude.id` works, but `id` doesn't.

buzden: `id` is a lower-case identifier. Being unqualified, it is quantified and expectedly does not work. It is the same as `Injective a`.

libs/base/Control/Function.idr
```diff
@@ -23,3 +23,7 @@ public export
 [InjFromComp] {f : a -> b} -> {g : b -> c} ->
   Injective (g . f) => Injective f where
     injective prf = injective {f = (g . f)} $ cong g prf
+
+public export
+Injective Prelude.id where
```
gallais, line 28: Again I don't think it's a good idea to have this as an unnamed instance because it will always be a valid candidate.

nickdrozd: Added name

nickdrozd:
Rebased on latest main. The lint failure is unrelated.

PR #2163, opened 2021-12-01
Things I wish we had for AoC

gallais:


andrevidela:
Shall we keep this open until Dec 25 and keep adding onto it?

gallais:
Yep that was my plan.

gallais:
Between the Idris meeting and then the holidays, I stopped working on AoC
super early this year so I agree there's no point in waiting longer.

PR #2173, opened 2021-12-06
[ fix #2098 ] Allow unclosed comment blocks

### LINKED ISSUE #2098: Allow unclosed comment blocks
Idris currently supports comment blocks that start with `{-` and end with `-}`. Idris does not currently have a way to say "comment out the rest of the file."

We should teach the parser to accept an unclosed `{-` as simply a comment that lasts the rest of the file.

The motivation here is both
1. A common way to work on a refactor is to hide the rest of the file and then reintroduce one definition at a time. 
2. A reasonable way to build out a new file might be to throw notes and code that you can't quite get working yet into a comment at the bottom as you go. 

Some languages call this strategy a "fold" in the file. Things above the fold are compiled and things below the fold are ignored.

bkomuves:
This implementes request #2098: allow source files ending with unclosed comment blocks.

### DIFF
--- a/src/Libraries/Text/Lexer/Core.idr
+++ b/src/Libraries/Text/Lexer/Core.idr
@@ -15,6 +15,7 @@ export
 data Recognise : (consumes : Bool) -> Type where
      Empty : Recognise False
      Fail : Recognise c
+     EOF : Recognise False
      Lookahead : (positive : Bool) -> Recognise c -> Recognise False
      Pred : (Char -> Bool) -> Recognise True
      SeqEat : Recognise True -> Inf (Recognise e) -> Recognise True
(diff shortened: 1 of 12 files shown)


Z-snails:
Correct me if I'm wrong, but this doesn't appear to actually test unclosed block comments, could you add a test please?

bkomuves:
> Correct me if I'm wrong, but this doesn't appear to actually test unclosed block comments, could you add a test please?

Ok, will do. I didn't do it initially because to properly test it you would need like 100 test cases... But I can certainly add one or two.

ska80:
@bkomuves Thanks!

This great addition could be mentioned in the `CHANGELOG.md` :)

Issue #1387, opened 2021-05-07
[ pretty-print ] Unexcepted erased linear names when querying docs of constructors.


alissa-tung:
# Steps to Reproduce
```idris
data Issue : Nat -> Type where
  MkCons : (n : Nat) -> (0 p : n = 42) -> Issue n
```
# Expected Behavior
```idris
Main> :doc Issue
Main.Issue : Nat -> Type
  Totality: total
  Constructor: MkCons : (n : Nat) -> (0 p : n = 42) -> Issue n
```
# Observed Behavior
```idris
Main> :doc Issue
Main.Issue : Nat -> Type
  Totality: total
  Constructor: MkCons : (n : Nat) -> (0 _ : n = 42) -> Issue n
```

gallais:
* Red is constructor
* Green is definition
* Blue is type constructor

gallais:
A possible fix:

1. Modify `PiInfo` so that `Explicit` carries a `Maybe Name` recording whether the user had provided an explicit name
2. Patch the pretty printer in `Idris.Pretty` to use that name if it present

Step 1. will break a lot of code I presume but they should be super easy to fix. You'll just have to follow the compiler
as it chases type error after type error. I may eventually do it on a day I'm looking for a zen extended refactoring.

PR #2175, opened 2021-12-06
[ fix #1945 ] Add `default` support for record fields

### LINKED ISSUE #1945: Add `default ` support for record fields
```idris
-- this works
aNat : {default Z theNat : Nat} -> Nat
aNat = theNat

-- this does not but should
record ANat where
  constructor MkANat
  {default Z theNat : Nat}

-- I then expect to have:
-- MkANat : {default Z theNat : Nat} -> ANat
-- theNat : ANat -> Nat
```

bkomuves:
This PR implements #1945: "Add `default` support for record fields"

One difference from the expected behaviour as described in #1945 is that in the example
```
record ANat where
  constructor MkANat
  {default Z theNat : Nat}
```
when asking for the type of `:t MkANat`, we get
```
MkANat : {default 0 _ : Nat} -> ANat
```
instead of
```
MkANat : {default Z theNat : Nat} -> ANat
```
That is, the field name `theNat` is not printed. I'm not sure what do to about that; it seems independent from accepting the definition.


gallais:
The caveat you mention is related to #1387

ska80:
@bkomuves Thanks!

I think this deserves to be mentioned in the `CHANGELOG.md` too :)

PR #2181, opened 2021-12-07
[ fix #2138] Add an %unhide pragma

### LINKED ISSUE #2138: Add an %unhide pragma
- [x] I have read [CONTRIBUTING.md](../../CONTRIBUTING.md).

## Summary

I propose to add an `%unhide` counterpart of the existing `%hide` pragma (I have no strong feelings about the name of the new pragma).

## Motivation

`%hide` is not scoped, so once a name is hidden, it cannot be accessed in the remaining part of the given source file. However, it may happen that one would like to hide a name only locally. Unhiding an already hidden name is a very lightweight solution for this.

Another motivation is a simple symmetry argument: If you can hide things, it makes sense to be able to unhide them, too.

## Motivating example

I would like to be able to write something like:
```
namespace Foo

  public export
  data T 
    = A Int
    | B T

namespace Bar

  public export
  data T 
    = A Int
    | B T
    | C Bool

foo : Foo.T
bar : Bar.T
```
without making the recursive `T` in the definition of `Bar.T` qualified (because with complicated types that would become very ugly and noisy). I can hide `Foo.T` before defining `Bar.T`, but then I cannot use `Foo.T` in the remaining of the file (in particular, I cannot define `foo`). 

I could put the two datatype definitions into separate modules, but I don't want to - it seems to me that namespaces were introduced as a lightweight solution for exactly this type of problem, but currently they work better with functions than with datatypes.

## The proposal

Add an `%unhide` pragma, which reveals names previously hidden by `%hide`.

### Examples

```
namespace Foo

  public export
  data T 
    = A Int
    | B T

namespace Bar

  %hide Foo.T

  public export
  data T 
    = A Int
    | B T
    | C Bool

  %unhide Foo.T

foo : Foo.T
foo = ...

bar : Bar.T
bar = ...
```

### Technical implementation

I have a proof-of-concept implementation here: https://github.com/bkomuves/Idris2/commit/bcff144371a01bb0f304b7c7c008cd72f2c391ea

It seems to work, but I'm not familiar with the codebase, so I'm not confident enough that it does exactly what I want.

## Alternatives considered

Alternative solutions to the motivating problem:

- use qualifed names in the definition of `Bar`. Pros: This works already; cons: the code will be ugly with more complicated types.
- put the two definitions into separate modules, and use qualified imports. Pros: This too works already; cons: a bit heavyweight for the given problem. Namespaces are lightweight, and I think that's the main reason for them.
- make `%hide` scope-aware. Pros: makes `%hide` a more proper namespace manipulation tool; cons: Possibly very complicated change; other pragmas are not scope-aware either
- in case of ambiguity within a namespace, default to the local name. Pros: makes sense to me; cons: consequences of such a change are not clear to me.

## Conclusion

This is a very lightweight change which solves a problem and makes the `%hide` construct more symmetric.

bkomuves:
This PR implements proposal #2138: "Add an %unhide pragma".


src/Core/Context.idr
```diff
@@ -1415,6 +1435,16 @@ hide fc n
               | res => ambiguousName fc n (map fst res)
          put Ctxt (record { gamma $= hideName nsn } defs)
 
+-- Set a name as Public that was previously hidden
+export
+unhide : {auto c : Ref Ctxt Defs} ->
```
mattpolzin, line 1440: it would be nice, though I think not a blocker to merging this work, if there was a warning when unhiding a name that was not hidden to begin with.

src/Core/Context.idr
```diff
@@ -1435,6 +1435,11 @@ hide fc n
               | res => ambiguousName fc n (map fst res)
          put Ctxt (record { gamma $= hideName nsn } defs)
 
+-- `unhide` tries to use `recordWarning`, which is defined below
+-- so I have to forward-declare it here...
+export
+recordWarning : {auto c : Ref Ctxt Defs} -> Warning -> Core ()
```
gallais, line 1441: Can't you just move the whole definition rather than implicitly declaring a mutual block?

bkomuves: Ok, moved `unhide` to the bottom (and `hide` too, to have them next to each other)

mattpolzin (approved):

mattpolzin:
Thanks!

ska80:
@bkomuves Thanks!

It would be great if `docs/source/reference/pragmas.rst` and `CHANGELOG.md` were also updated.

PR #2183, opened 2021-12-08
[ fix #1778 ] TCO

### LINKED ISSUE #1778: RefC tail-call optimization
The RefC backend is not performing tail-call optimization properly.

# Steps to Reproduce

Consider this elaborate identity function on `Nat`s. We're treating `Nat`s as `List ()`, and reversing them.

```idris
reverse : Nat -> Nat
reverse = reverseOnto 0
    where
      reverseOnto : Nat -> Nat -> Nat
      reverseOnto j 0 = j
      reverseOnto j (S k) = reverseOnto (S j) k

main : IO ()
main = printLn $ reverse $ the Nat 100000
```

# Expected Behavior

This should trivially print `100000`.

# Observed Behavior

Instead, we get a Segfault, due to running out of stack space.

# Attempted Diagnosis

The trampolining support code looks correct, if messy.

As `reverseOnto` is tail-recursive, it should not be launching another trampoline, instead returning the complete closure. If we replace the line `tmp_4 = trampoline(closure_6);` with `tmp_4 = closure_6;` in the generated code for `reverseOnto`, then it compiles and runs without issue.

A superficial reading of `RefC.idr` suggests that we do actually distinguish between tail-recursive and non-tail-recursive function calls, so something must be getting lost along the way.

dagit:
This could definitely use better testing. In particular, I'm not super confident that I've used the correct value for `TailPositionStatus` in all places (would love some code review on that from someone else who understands this code). The intention is to track (outside in) whether the next call we encounter is in tail position, with the intuition that, in ANF, the tail position is the outer most call of the term.

However, things aren't as bleak as I'm making it sound. For example, it does appear to be more correct than the version it replaces. All the existing tests pass and I added a test based on the example from #1778. And it works on the program where I noticed the issue.

I expect the performance to be essentially the same for programs that terminate before/after this patch and infinitely better for programs that don't terminate without it :)

Also, tagging @madman-bob as he's probably a good reviewer for this.


dagit:
I was just looking over this paper:

> [A Functional Perspective on SSA Optimisation Algorithms by Manuel M. T. Chakravarty, Gabriele Keller and Patryk Zadarnowski](https://www.jantar.org/papers/chakravarty03perspective.pdf)

And they mention that in ANF, tail position should correspond to any call made in the body of a let. Their ANF notation has `if-then-else` instead of `case`, and looking at the grammar for it to detect tail calls in the branches of an `if-then-else` you would need to propagate some information from the outside like this change does with threading `TailPositionStatus`.

Specifically, this is what they say:

> Tail calls are made explicit by their placement within a let expression: function applications appearing on the right-hand side of a variable binding represent normal calls, while those appearing in the body represent tail calls (jumps).

But if you look at the grammar it's clear that if you processed ANF in a syntax directed way, then when you're looking at an `if-then-else` (or for us a `case`) you'd need to know if it's the body of a `let` or not.

Anyway, the take away for me is that code in this PR could be improved. Specifically, in the `ALet` case of `cStatementFromANF` we have this line:

```idris
bodyAssignment <- cStatementsFromANF body tailPosition
```

And I believe it should become:

```idris
bodyAssignment <- cStatementsFromANF body InTailPosition
```

It's already pretty late here, so I'll try to test this tomorrow and possibly update the code in the PR.

dagit:
Never mind the change I mentioned about lets. The tests fail with that change. I think I see why but I'll have to think about it more.

madman-bob (commented):
Ok, I've read this PR as carefully as I can.

From a behaviour perspective, I think it's correct.

Focusing on the `ALet` comment in particular, it makes sense to me that the body of a `let` clause is in a tail position precisely when the `let` itself is. So the `bodyAssignment` line should indeed pass through the tail position, rather than try to force it into one position.

From a maintenance perspective, the `ReturnStatement` and `TailPositionStatus` types both try to do the same thing, and in `cStatementsFromANF`, they step on each others toes. While it works at the moment, it could easily cause confusion. I would recommend that we only keep one or the other.

For simplicity, it would probably be easier to remove the `ReturnStatement` type, as it is only returned by `cStatementsFromANF`. This function should instead return a `String`, and case split on the passed `TailPositionStatus` to determine which string to return.

It may be possible to simplify further, having `cStatementsFromANF` instead return `()` and use `emit` - as all the other functions do - but that may be a little bit more involved.

gallais (approved):
Looks good to me!
Edit: I guess there's not much point in waiting for more feedback: it's already been up for more than a month.

PR #2185, opened 2021-12-08
Fix compilation on OpenBSD

octeep:
Other BSDs probably need this too but I've only tested on OpenBSD.

### DIFF
--- a/support/c/idris_file.c
+++ b/support/c/idris_file.c
@@ -13,6 +13,7 @@
 #include "windows/win_utils.h"
 #else
 #include <sys/select.h>
+#include <sys/wait.h>
 #endif
 
 #include "idris_util.h"


support/c/idris_file.c
```diff
@@ -15,6 +15,10 @@
 #include <sys/select.h>
```
cypheon, line 15: It seems (from reading some manpages), that `<sys/wait.h>` can (and should) be included on all OSs (except Windows). So maybe the include would be a better fit up here?

```suggestion
#include <sys/select.h>
#include <sys/wait.h>
```

support/c/idris_file.c
```diff
@@ -15,6 +15,10 @@
...
 #endif
 
+#ifdef __OpenBSD__
+#include <sys/wait.h>
+#endif
+
```
cypheon, line 21: Then this additional `#ifdef` would not be necessary
```suggestion
```

cypheon (commented):
This might not be that specific to OpenBSD.

mattpolzin:
This looks good to go once you apply the other suggestion of removing the second import of `sys/wait` (because it is now in the else statement that is executed everywhere but Windows)
```
#ifdef __OpenBSD__
 #include <sys/wait.h>
 #endif
```

mattpolzin (approved):

mattpolzin:
Thanks!

dagit:
I ran into this on FreeBSD as well. So I'm glad the resolution was to use the include on more than OpenBSD.

PR #2188, opened 2021-12-09
[fix #2033] determining when to rebuild modules

### LINKED ISSUE #2033: failure to recompile modules, seemingly randomly
let's call this file `bug.sh`:

```bash
IDRIS=idris2-0.5.1

cat > Module1.idr<< EOF
module Module1
export msg : String; msg="Hello world"
EOF

cat > module1.ipkg<< EOF
package pkg1
modules = Module1
EOF


cat > Main.idr<< EOF
module Main
import Module1

main : IO ()
main = do putStrLn ("The message is: " ++ msg)
EOF

cat > main.ipkg<< EOF
package app

main = Main
executable = test
EOF

${IDRIS} --build module1.ipkg
${IDRIS} --build main.ipkg

./build/exec/test

# overwrite msg
cat > Module1.idr<< EOF
module Module1
export msg : String; msg="New Message"
EOF

# this doesn't help, i.e. it's not due to low resolution of file modification times
sleep 1

# this helps
#rm build/ttc/Module1.tt*

${IDRIS} --build module1.ipkg
${IDRIS} --build main.ipkg

./build/exec/test
```

if i run it multiple times, it behaves inconsistently?!

```
$ ./bug.sh 
1/1: Building Module1 (Module1.idr)
2/2: Building Main (./Main.idr)
The message is: Hello world
1/1: Building Module1 (Module1.idr)
The message is: New Message

$ ./bug.sh 
1/1: Building Module1 (Module1.idr)
2/2: Building Main (./Main.idr)
The message is: Hello world
The message is: Hello world

$ ./bug.sh 
1/1: Building Module1 (Module1.idr)
2/2: Building Main (./Main.idr)
The message is: Hello world
1/1: Building Module1 (Module1.idr)
The message is: New Message

$ ./bug.sh 
1/1: Building Module1 (Module1.idr)
2/2: Building Main (./Main.idr)
The message is: Hello world
1/1: Building Module1 (Module1.idr)
The message is: New Message

$ ./bug.sh 
1/1: Building Module1 (Module1.idr)
2/2: Building Main (./Main.idr)
The message is: Hello world
1/1: Building Module1 (Module1.idr)
The message is: New Message

$ ./bug.sh 
1/1: Building Module1 (Module1.idr)
2/2: Building Main (./Main.idr)
The message is: Hello world
1/1: Building Module1 (Module1.idr)
The message is: New Message

$ ./bug.sh 
1/1: Building Module1 (Module1.idr)
2/2: Building Main (./Main.idr)
The message is: Hello world
The message is: Hello world
```

what's going on?

the actual context where this comes up is that i have moved `IdrisPaths.idr` into a separate module, so that i can generate it under `build/generated/`, and `idris2.ipkg` just imports it. but the updated `yprefix` never seems to reach the `idris2.ss` file, not even once when i was testing.

in this reproducer it sometimes happens. sorry if it's my fault, i have no clue what's going on here.

mattpolzin:
- Makes it so that `ProcessIdr` and `ModTree` agree about when a module needs to be rebuilt (there was a `>=`/`<=` disagreement that made test scripts liable to give confusing results even if in reality the bug would rarely occur -- the fix is strictly beneficial to code maintainability regardless).
- Makes `%inline` definitions part of a module's interface even if the definition is not `public`.
- Does not inline heuristically if a definition is not `public`; this may be a regression to some runtime performance in some cases, but we cannot safely inline defs that are not explicitly marked for inlining across module boundaries without (a) adding that def to the source module's interface or (b) using some information other than import interface hashes to determine when a module should be rebuilt.

Closes https://github.com/idris-lang/Idris2/issues/2033


mattpolzin:
> Can we benchmark whether only using the heuristics if the name is public boostraps to a measurably slower idris2?

Seems worth knowing. I'll do some checking. I suspect that the Chez backend in particular will do a good job of optimizing this kind of thing on its own once you get all of the way to scheme code being run by the scheme runtime, but we may see an impact to other backends for which I would have a harder time testing performance impacts.

> If so we could instead consider running the heuristics earlier and making inlinable but non-public definitions part of
the interface.

I'd have to look into this a bit more. I'm not 100% sure that the current inline heuristic code happens too late in the process to add it to the module interface, but it definitely might; My original thinking was that we don't want to inline across module boundaries without that being the intention of the module author anyway because it feels subversive to the fact that the author specifically did not publicly export the definition (which by definition means they did not want it to be part of the interface), but we are getting into some subtle distinctions now between "part of interface because changes will cause other code that uses compile-time logic to stop compiling" or "part of interface because changes will cause other modules using the definition to be recompiled."

mattpolzin:
Ok, I did some benchmarking. The following only applies to the Chez backend, so I still wonder whether the impact of the inlining work has a larger impact on other backends.

I measured the results of `make` having installed Idris from (a) the HEAD of main and (b) the branch for this PR.

There was absolutely no difference in the amount of time it took to run a full `make clean && make` between the two compiler versions; Across 4 runs with each compiler, the average `make` time was `3m33.0s` with 5 seconds of variance for HEAD of main and `3m32.9s` with 4 seconds of variance for this branch.

There was a perceptible difference in the size of the TTC/binary output, but it was very small. With HEAD of main, the TTC output for the compiler was `424K` smaller than the TTC output of this branch. The executable directory was similarly `172K` smaller with the HEAD of main than with this branch.

Given the lack of a difference in runtime performance of code produced by the Chez backend and the very small difference in binary size (even for a very large codebase like the compiler itself), I propose that we merge this change (as it fixes a known bug that has come up more than a few times in Discord and now the issue tracker) and we can reintroduce more aggressive inlining in a future PR when someone has time to investigate the right approach.

I can think of two possible approaches and I don't know which one is better: (a) inline earlier (as suggested above) and make inlined definitions part of a module's interface or (b) allow non-public inlining of definitions _specifically when compiling executables_ so that we can avoid inlining between arbitrary modules but still gain benefits when building executables.

attila-lendvai:
i can confirm that when i'm rebasing https://github.com/idris-lang/Idris2/pull/1990 on this PR, then it fixes my headaches of not incorporating the changes made to `IdrisPaths.idr` into the final executable.

attila-lendvai:
@mattpolzin what's the ETA of this PR? i'm asking because it would be very helpful to me if it got either merged or rebased to `main` (it now has merge conflicts).

mattpolzin:
I'm not sure. It's the holidays so everything is moving more slowly but I'm afraid I cannot tell you when this PR will get reviewed; I can only say I won't be merging it myself without another reviewer having a chance to take a close look and weigh in on the changes.

src/Core/Binary.idr
```diff
@@ -510,27 +510,37 @@ readFromTTC nestedns loc reexp fname modNS importAs
...
+         sourceFileHash <- fromBuf b
+         interfaceHash <- fromBuf b
+         pure (sourceFileHash, interfaceHash)
+
+-- Implements a portion of @readTTCFile@ from where @getHashes@ leaves off.
+-- The next field is the import hashes. The fields must be read in order.
+getImportHashes : String -> Ref Bin Binary ->
+                  Core (List (Namespace, Int))
+getImportHashes file b
+    = do (sourceFileHash, interfaceHash) <- getHashes file b
```
gallais, line 541: For all of these, I'm tempted to suggest `ignore $ getX file b` for the prefix we're not even inspecting

src/Idris/ProcessIdr.idr
```diff
@@ -223,26 +223,45 @@ gc = primIO $ prim__gc 4
...
+                Core Bool
+isTTCOutdated ttcFile sourceFiles
+  = do ttcTime  <- modTime ttcFile
+       srcTimes <- traverse modTime sourceFiles
+       log "module.hash" 20 $
+         unwords $
+           [ "Checking whether source code mod times are newer than \{show ttcTime};"
+           , "src times:"
+           , show srcTimes
+           ]
```
gallais, line 246:
```suggestion
         unlines $
           "Checking whether source code mod times are newer than \{show ttcTime}; src times:"
           :: zipWith (\ src, tm => "\{src} : \{show tm}") sourceFiles srcTimes
```

src/Idris/ProcessIdr.idr
```diff
@@ -223,26 +223,45 @@ gc = primIO $ prim__gc 4
...
              | _ => pure False
        (Just storedSourceHash, _) <- readHashes ttcFileName
              | _ => pure False
        pure $ sourceCodeHash == storedSourceHash
 
+||| Determine if any of the given source files have changed.
+export
+unchangedHashes : (hashFn : Maybe String) -> (ttcFileName : String) -> (sourceFileNames : List String) -> Core Bool
+unchangedHashes hashFn ttcFileName
+  = map (all id) . traverse (unchangedHash hashFn ttcFileName)
```
gallais, line 263: Does this mean we keep re-reading the hash for `ttcFileName`? Is that expensive?

gallais (commented):
A welcomed refactoring.

src/Compiler/Opts/InlineHeuristics.idr
```diff
@@ -46,5 +46,14 @@ inlineHeuristics fn = do
         | Nothing => pure ()
     let True = inlineCDef fn cdef
         | False => pure ()
-    log "compiler.inline.heuristic" 25 $ "inlining heuristic decided to inline: " ++ show fn
-    unless (NoInline `elem` gdef.flags) $ setFlag EmptyFC (Resolved fnIdx) Inline
+    -- we can't inline non-public defs (unless they are marked as Inline by developer)
```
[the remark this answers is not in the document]
mattpolzin: I'll rework this comment to be clearer -- the "unless marked Inline" does not com into play here since this code only handles inlining by heuristic; marking something inline is explicit so it'll already be marked for inlining before getting to this code. Anyway, you've made me realize this comment could be better.

src/Core/Binary.idr
```diff
@@ -510,27 +510,37 @@ readFromTTC nestedns loc reexp fname modNS importAs
...
     = do hdr <- fromBuf {a = String} b
-         when (hdr /= "TT2") $ corrupt ("TTC header in " ++ file ++ " " ++ show hdr)
+         when (hdr /= "TT2") $
+           corrupt ("TTC header in " ++ file ++ " " ++ show hdr)
          ver <- fromBuf @{Wasteful} b
          checkTTCVersion file ver ttcVersion
          fromBuf b
 
+-- Implements a portion of @readTTCFile@ from where @getTotalReq@ leaves off.
+-- The fields must be read in order.
```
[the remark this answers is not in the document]
mattpolzin: Yeah, I agree that the comment is confusing. I'll rework it.

src/Core/Binary.idr
mattpolzin: I like it. Will do.

src/Idris/ProcessIdr.idr
mattpolzin: This code does re-read the hash for each `sourceFileName`, but I think that optimizing that away here would be premature given that this function is only called with a single item in `sourceFileNames` in practice.

Now that it's been a bit and I am looking back, I don't like that I chose to create this function to match the signature of the `isTTCOutdated` function just for the sake of the call site below. I am going to push a change that entirely deletes this function and instead reworks the call site slightly.

PR #2189, opened 2021-12-09
[ fix #2032 ] Slow typechecking on `Int` operation when `Data.Fin.fromInteger` is in scope

### LINKED ISSUE #2032: Slow typechecking on `Int` operation when `Data.Fin.fromInteger` is in scope

bkomuves:
This PR fixes #2032 while being minimally invasive.

An example of the problem this is fixing is that following program typechecks very slowly (linear in the value of the constant):
```
import Data.Fin

addMillion : Int -> Int
addMillion x = 1000000 + x
```
The cause of the problem that even though `Data.Fin.fromInteger` is not called by this program, the typechecker tries to unify its type with the hole representing the type of `fromInteger` inserted before that literal constant. 

`Data.Fin.fromInteger` had the following type before this PR:
```
fromInteger : (x : Integer) -> {n : Nat} -> {auto 0 prf : So (fromInteger x < n)} -> Fin n
```
The problem is that `fromInteger x : Nat` in the proof obligation converts 1,000,000 to an unary `Nat` during typechecking, before finally rejecting it. That takes linear time (and memory) in the size of the constant.

This fix introduces a new comparison function which compares an `Integer` and a `Nat` directly. This way we can still pattern match on the `Nat` and things like `fromInteger 1 : Fin (S (S n))` still work. The new `fromInteger` then calls the old one (with a fake proof), so the rest of the conversion is unchanged.

One disadvantage of this change is that the error message when the auto proof search fails is a bit uglier (see test `basic042`).

### DIFF
--- a/libs/base/Data/Fin.idr
+++ b/libs/base/Data/Fin.idr
@@ -199,23 +199,23 @@ natFromInteger x = natToFinLt (fromInteger x)
 -- but we cannot convert the `Integer` to `Nat` either, because that slows down typechecking
 -- even when the function is unused (issue #2032)
 public export
-(<) : Integer -> Nat -> Bool
-(<) x n with (x < the Integer 0)
-  (<) _ _     | True  = True           -- if `x < 0` then `x < n` for any `n : Nat`
-  (<) x (S m) | False = (x-1) < m      -- recursive case
-  (<) x Z     | False = False          -- `x >= 0` contradicts `x < Z`
+integerLessThanNat : Integer -> Nat -> Bool
+integerLessThanNat x n with (x < the Integer 0)
+  integerLessThanNat _ _     | True  = True                            -- if `x < 0` then `x < n` for any `n : Nat`
+  integerLessThanNat x (S m) | False = integerLessThanNat (x-1) m      -- recursive case
+  integerLessThanNat x Z     | False = False                           -- `x >= 0` contradicts `x < Z`
 
 ||| Allow overloading of Integer literals for Fin.
 ||| @ x the Integer that the user typed
 ||| @ prf an automatically-constructed proof that `x` is in bounds
 public export
 fromInteger : (x : Integer) -> {n : Nat} ->
-              {auto 0 prf : So (x < n)} ->
+              {auto 0 prf : So (integerLessThanNat x n)} ->
               Fin n
 fromInteger x = natFromInteger x {prf = lemma prf} where
   -- to be minimally invasive, we just call the previous implementation.
   -- however, having a different proof obligation resolves #2032
-  0 lemma : {x : Integer} -> {n : Nat} -> So (x < n) -> So (fromInteger {ty=Nat} x < n)
+  0 lemma : {x : Integer} -> {n : Nat} -> So (integerLessThanNat x n) -> So (fromInteger {ty=Nat} x < n)
   lemma oh = believe_me oh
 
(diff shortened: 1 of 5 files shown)


libs/base/Data/Fin.idr
```diff
@@ -188,14 +188,35 @@ public export
 maybeLT : (x : Nat) -> (y : Nat) -> Maybe (x `LT` y)
 maybeLT x y = maybeLTE (S x) y
 
+public export
+natFromInteger : (x : Integer) -> {n : Nat} ->
```
mattpolzin, line 192: `finFromInteger` perhaps?

tests/base/data_fin001/fromInteger.idr
```diff
@@ -0,0 +1,22 @@
+
+-- this is to test that changes introduced for #2032 do not ruin other things
+-- (#2032 itself is a performance issue, I don't know how to properly test for that)
```
mattpolzin, line 3: There's a `benchmarks` folder with some benchmarks in it, but I don't know the history of that stuff and I can't seem to run the benchmarks locally using the `bench.sh` script (might be because I am on a mac). If that is not being maintained, I don't think there is any other way for you to create tests beyond these which ensure functionality.

libs/base/Data/Fin.idr
bkomuves: ah, good catch! Apparently my brain stops working when there are more than integer-like 2 types...

gallais:
I was going to ask for a test in `base` but I see it's already here. :partying_face:
Thanks for the PR!

PR #2190, opened 2021-12-09
[ new ] explicit import lists

gallais:
* [X] fix parsing issue
* [ ] fix renaming to a namespaced name (disallow?)
* [ ] add test for namespaced names


mattpolzin (commented):
This is exciting! My preference would be for the warning about unfound names to be an error, even.

src/Idris/Doc/Keywords.idr
```diff
@@ -534,7 +552,8 @@ keywordsDoc =
   :: "else" ::= ifthenelse
   :: "forall" ::= forallquantifier
   :: "rewrite" ::= rewriteeq
-  :: "using" ::= ""
+  :: "using" ::= importDirectives
```
buzden, line 555: `using` keyword is used also for named interface implementations that use another named implementation as a superimplementation. Maybe, if should be mentioned here also.

gallais: I'm happy to merge a PR that someone would assemble to improve the `:doc` thing.
I originally left `using` blank because I did not know what to write. There are
least two other uses in the parser separate from the one added (and documented)
in this PR.

gallais:
> My preference would be for the warning about unfound names to be an error, even.

I don't like errors because they (currently) break interactivity and you don't get e.g. semantics highlighting.
You can always use `-Werror` in the CI if you want to ensure

gallais:
Strangely enough I'm not getting the renamed definitions in the REPL *unless*
there is an error in the file being loaded. And I'm getting *all* the things (even the
ones filtered out) if there is no error.

Looks like the REPL is somehow reloading the dependencies and bypassing my directives?

PR #2192, opened 2021-12-13
[ base ] Use `Fin n` as index in `Bits`

buzden:
This was originally created during discussions at IDM but is was blocked by #2032. Since it's now fixed, we can consider this change.

Also, this PR contains a small cleanup as a separate commit.

### DIFF
--- a/docs/source/typedd/typedd.rst
+++ b/docs/source/typedd/typedd.rst
@@ -333,7 +333,7 @@ now uses a safer type for the number of shifts:
 
     randoms : Int -> Stream Int
     randoms seed = let seed' = 1664525 * seed + 1013904223 in
-                       (seed' `shiftR` fromNat 2) :: randoms seed'
+                       (seed' `shiftR` 2) :: randoms seed'
 
 
 In ``ArithCmd.idr``, update ``DivBy``, ``randoms``, and ``import Data.Bits``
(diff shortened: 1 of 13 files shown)


stefan-hoeck:
I'm not sure this is the way to go. I agree that the current situation with `Subset Nat (LT n)` is even worse, but I think the index should actually be of the same type (possibly with an erased proof of being in bounds) as the bits value itself. The reason is simple: `Fin n` like `Nat` is converted to an `Integer` at the backends, and at least in case of the JS backends this leads to a significant overhead for integer types like `Int32`. But bitwise operations are all about performance, aren't they? And going via arbitrary precision integers all the time imposes a significant performance overhead at least on some backends.

buzden:
> but I think the index should actually be of the same type

I don't agree with you. I think that logical interface shouldn't sacrifice from runtime representation issues (in ideal world, of course).

Also, unbounded integer representation problem can have more solutions than you propose. For example, we can have a compiler optimisation for representing `Fin`s of constant size with bounded ints (e.g. using those `Bitx` types). For this, having `Fin` rather than `Subset Nat` is even better as a simple sign for the compiler to optimise runtime here.

stefan-hoeck:
> I don't agree with you. I think that logical interface shouldn't sacrifice from runtime representation issues (in ideal world, of course).
> 
> Also, unbounded integer representation problem can have more solutions than you propose. For example, we can have a compiler optimisation for representing `Fin`s of constant size with bounded ints (e.g. using those `Bitx` types). For this, having `Fin` rather than `Subset Nat` is even better as a simple sign for the compiler to optimise runtime here.

Fair enough. This change is definitely better than the current situation.

PR #2196, opened 2021-12-13
[ fix #1959 ] use modern record update syntax

### LINKED ISSUE #1959: Fix "TODO deprecated" in the parser
We should make this an actual warning:

https://github.com/idris-lang/Idris2/blob/a9ccf4db4f69677c5287a1278ab167247dac161d/src/Idris/Parser.idr#L798-L800

Especially now that we have a deprecation mechanism:

https://github.com/idris-lang/Idris2/blob/a9ccf4db4f69677c5287a1278ab167247dac161d/src/Idris/Parser.idr#L567

teggotic:
fixes https://github.com/idris-lang/Idris2/issues/1959


docs/source/reference/records.rst
```diff
@@ -176,14 +176,14 @@ Finally, the examples:
 
       -- record update syntax uses dots now
       -- prints 3.0
-      printLn $ (record { topLeft.x = 3 } rect).topLeft.x
+      printLn $ ({ topLeft.x := 3 } rect).topLeft.x
 
       -- but for compatibility, we support the old syntax, too
-      printLn $ (record { topLeft->x = 3 } rect).topLeft.x
+      printLn $ ({ topLeft->x = 3 } rect).topLeft.x
```
mattpolzin, line 182:
```suggestion
      printLn $ ({ topLeft->x := 3 } rect).topLeft.x
```

mattpolzin (commented):
This is great. I've thought about it a few times but never actually followed through!

teggotic:
I see that TTImp can't compile new record syntax and this causes ttimp/record002 to fail
image

teggotic:
Seems like for some reason tests on windows and everything else produces different result in tests/ideMode/ideMode005/expected8

teggotic:
@mattpolzin maybe you could help me?

src/Idris/Parser.idr
```diff
@@ -793,15 +793,27 @@ mutual
...
+                 bounds (do
+                   decoratedKeyword fname "record"
+                   body True
+                 ))
+             <|>
+               bounds (body False))
            pure (PUpdate (boundToFC fname b) b.val)
+    where
+      oldSyntaxWarning : String
+      oldSyntaxWarning = "DEPRECATED: old record update syntax. Use \"{ f := v } p\" instead of \"{ f = v } p\""
```
mattpolzin, line 808:
```suggestion
      oldSyntaxWarning = "DEPRECATED: old record update syntax. Use \"{ f := v } p\" instead of \"record { f = v } p\""
```

src/Idris/Parser.idr
```diff
@@ -793,15 +793,27 @@ mutual
...
-                           commit
-                           fs <- sepBy1 (decoratedSymbol fname ",") (field kw fname indents)
-                           decoratedSymbol fname "}"
-                           pure $ forget fs)
+      = do
+           b <- (
+               withWarning oldSyntaxWarning (
+                 bounds (do
+                   decoratedKeyword fname "record"
+                   body True
```
mattpolzin, line 801: Maybe we should commit here between the record keyword and the body.

mattpolzin (changes requested):
This is looking pretty good.

I think we should have at least one test that uses the old record syntax both to prove that we've only deprecated (and not broken) that syntax and also to show that the warning you've added shows up!

src/Idris/Parser.idr
teggotic: Maybe we should use a better? message
image

mattpolzin: That might be a better message. The thing that might be most confusing to a user is the subtle difference between an equals sign and a colon-equals or the fact that the new syntax suggestion could be interpreted as simply not showing the "record" keyword again instead of indicating that the keyword should not be used. At the end of the day I think either of the messages you've suggested would be good.

teggotic: I'll stick to the first message then

teggotic:
> This is looking pretty good.
> 
> I think we should have at least one test that uses the old record syntax both to prove that we've only deprecated (and not broken) that syntax and also to show that the warning you've added shows up!

tests/idris2/record002/ now has uses both versions
added tests/idris2/warning003/ for deprecation warning

mattpolzin (approved):
Thanks! This looks good to go once CI is all passing.

gallais:
Thanks for the PR!

jfdm:
As a quick note, this PR created a 'breaking' change in how Idris2 operates (i.e. syntax) can we ensure that such deprecations are reported in the CHANGELOG, and that older entries in the CHANGELOG.are not erroneously updated. My PR #2251 fixes this.

Issue #2178, opened 2021-12-07
REPL command loadpackage only loads packages that are already loaded


ohad:
# Steps to Reproduce

```bash
$ idris2 --no-banner -p contrib
Main> :lp "contrib"
Done
Main> :quit
Bye for now!
$ idris2 --no-banner
Main> :lp "contrib"
```

# Expected Behavior
Package `contrib` loads fine:
```
Done
```

# Observed Behavior
Package contrib doesn't load, other, already loaded, packages load fine:
```
Package not found in the known search directories
Main> :lp "base"
Done
Main> :quit
Bye for now!
```

Russoul:
This is expected behavior (w.r.t what I had in mind implementing it). But that would be a nice addition :)

ohad:
I see. 
Does that make this into a feature request?
Also, docs for `:lp` should be clearer, I think.

Russoul:
Yeah, I guess it does. I'll try to PR a docs update soonish.

PR #2198, opened 2021-12-15
Rename `lp/loadpackage` to `package`

Russoul:
In response to #2178

 I moved the doc-string up a few lines. Now it's right next to `module` which is close semantically to the command in question.
 Also renamed the command to `package` to dispel confusion about its semantics (Assuming the word 'import' conveys the intention better than 'load').


mattpolzin:
If we do this I think it should be accompanied by a changelog entry to help folks find it if they are trying to track down why the old name isn't working.

Issue #2207, opened 2021-12-19
Node is missing some FFI functions from base


Z-snails:
If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!

# Steps to Reproduce
Try to compile various `System.*` functions with node codegen

# Expected Behavior
Works

# Observed Behavior
Fails with
```
Uncaught error: INTERNAL ERROR: No supported backend found in the definition of System.Errno.prim__getErrno. Supported backends: "node","javascript". Backends in definition: "C".
```

Here's every FFI function in base that's missing a node version if someone wants to implement these:
* [ ] System
  * [ ] prim__sleep
  * [ ] prim__usleep
  * [ ] prim__getEnvPair
  * [ ] prim__setEnv
  * [ ] prim__unsetEnv
  * [ ] prim__system
  * [ ] prim__getPID
* [ ] Data.Buffer
  * [ ] prim__setBits64
  * [ ] prim__getBits64
  * [ ] stringByteLength
* [ ] System.Clock
  * [ ] prim__clockTimeMonotonic
  * [ ] prim__clockTimeUtc
  * [ ] prim__clockTimeProcess
  * [ ] prim__clockTimeThread
  * [ ] prim__clockTimeGcCpu
  * [ ] prim__clockTimeGcReal
  * [ ] prim__osClockValid
  * [ ] prim__osClockSecond
  * [ ] prim__osClockNanosecond
* [ ] System.Concurrency
  * Not all these can be implemented
* [ ] System.Directory
  * [ ] prim__openDir
  * [ ] prim__closeDir
  * [ ] prim__removeDir
  * [ ] prim__dirEntry
* [ ] System.Errno
  * [ ] prim__getErrno
  * [ ] prim__strerror
* [ ] System.FFI
  * [ ] prim__malloc
  * [ ] prim__free
* [ ] System.Info
  * [ ] prim__getNProcessors
* [ ] System.Signal
  * [ ] prim__sighup
  * [ ] prim__sigint
  * [ ] prim__sigabrt
  * [ ] prim__sigquit
  * [ ] prim__sigill
  * [ ] prim__sigsegv
  * [ ] prim__sigtrap
  * [ ] prim__sigfpe
  * [ ] prim__sigusr1
  * [ ] prim__sigusr2
  * [ ] prim__ignoreSignal
  * [ ] prim__defaultSignal
  * [ ] prim__collectSignal
[...]

PR #2212, opened 2021-12-22
[ nodejs ] add pid

alissa-tung:


Z-snails:
Addresses #2207

Great thanks!

PR #2213, opened 2021-12-22
[ libs ] move propaganda out of contrib

gallais:
The plan is to keep these around because they can uncover regressions in the compiler.


mattpolzin:
Love it. Sets us up well to explore moving to another repo and continuing to test against it a la Collie and Frex.

libs/linear/linear.ipkg
```diff
@@ -0,0 +1,11 @@
+package linear
+version = 0.5.1
+
+depends = contrib
```
mattpolzin, line 4: It doesn't look like this dependency is needed.

gallais:
That `plusAssociative` proof was fairly horrible. Glad to see [it's not just me](http://agda.github.io/agda-stdlib/Data.Integer.Properties.html#27895) though.

PR #2221, opened 2021-12-25
[ ci ] update previous version to v0.5.1, and super-linter to v4.8.5

attila-lendvai:
The current super-linter v3 line is broken, they recommend to use the
v4 line: https://github.com/github/super-linter/issues/2253

Also pick a specific version, so that such transient linter bugs won't
camouflage as Idris CI failures.

### DIFF
--- a/.github/workflows/ci-idris2.yml
+++ b/.github/workflows/ci-idris2.yml
@@ -33,7 +33,7 @@ on:
       - '.github/workflows/ci-super-linter.yml'
 
 env:
-  IDRIS2_VERSION: 0.5.0 # For previous-version build
+  IDRIS2_VERSION: 0.5.1 # For previous-version build
 
 jobs:
 
@@ -50,9 +50,9 @@ jobs:
         id: get_commit_message
         run: |
           if   [[ '${{ github.event_name }}' == 'push' ]]; then
-            echo ::set-output name=commit_message::$(git log --format=%B -n 1 HEAD)
+            echo "::set-output name=commit_message::$(git log --format=%B -n 1 HEAD)"
           elif [[ '${{ github.event_name }}' == 'pull_request' ]]; then
-            echo ::set-output name=commit_message::$(git log --format=%B -n 1 HEAD^2)
+            echo "::set-output name=commit_message::$(git log --format=%B -n 1 HEAD^2)"
           fi
 
     outputs:
@@ -82,7 +82,7 @@ jobs:
           echo "deb http://security.ubuntu.com/ubuntu hirsute universe" | sudo tee -a /etc/apt/sources.list
           sudo apt-get update
           sudo apt-get install -y -t hirsute chezscheme
-          echo "$HOME/.idris2/bin" >> $GITHUB_PATH
+          echo "$HOME/.idris2/bin" >> "$GITHUB_PATH"
 
       # Get our hands on the released version either by using the cache
       # or by rebuilding it if necessary.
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-idris2.yml
```diff
@@ -300,8 +300,8 @@ jobs:
           echo "deb http://security.ubuntu.com/ubuntu hirsute universe" | sudo tee -a /etc/apt/sources.list
           sudo apt-get update
           sudo apt-get install -y -t hirsute chezscheme
-          echo "$HOME/.idris2/bin" >> $GITHUB_PATH
-          chmod +x $HOME/.idris2/bin/idris2 $HOME/.idris2/bin/idris2_app/*
+          echo "$HOME/.idris2/bin" >> "$GITHUB_PATH"
+          chmod +x "$HOME/.idris2/bin/idris2" "$HOME/.idris2/bin/idris2_app/"*
```
mattpolzin, line 304: I'm not at the computer to verify right now but does this glob outside the quotes work as intended? Looks a bit off but I might have just never seen the asterisk put after the quotes before.

mattpolzin (commented):
Was it intentional to make the changes here unrelated to the linter update? The quoting is inconsistent and I am not sure why the `false` that permanently disables a test was added.

.github/workflows/ci-idris2.yml
attila-lendvai: i added the `false` to disable a one last racket bootstrap run, sorry for not commenting on that! IIRC racket was previously disabled in the CI because it takes too long and it just times out.

the new quotes are because the new linter version is smart enough to understand that those are pieces of shell scripts, and it complains that the ENV vars may contain spaces and/or globs.

how/where is the quoting inconsistent? note that i left the star out of the quotes, so that it is interpreted as an anything-marker.

mattpolzin: I don't think we should disable the windows racket check prematurely. As far as I know it has been working consistently for quite some time and it seems to usually complete in 15 minutes. 

> how/where is the quoting inconsistent?

I was reviewing without my laptop to hopefully get this merged right away because it looks like the right solve for failing linting but I wasn't able to verify all of the quotation changes acted the same way (and I'm not a bash expert). The inconsistently is around what exactly gets quoted. Sometimes just the variable (which is probably the only thing that needs to be quoted for the linter to be happy), sometimes it's the variable and the whole path up to the asterisk, sometimes the last forward slash is outside the quotes.

attila-lendvai: i see, thanks!

ok, i've updated the PR: eliminated the racket thing, and made the quotes uniform.

let me know if there's anything else.

attila-lendvai:
sidenote: a bug in the windows action, manifested by a transient network issue:

```
Run git clone --depth 1 https://github.com/cisco/ChezScheme
  git clone --depth 1 https://github.com/cisco/ChezScheme
  c:\msys64\usr\bin\bash -l -c "pacman -S --noconfirm tar make"
  echo "*** -u $(pwd))" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    IDRIS2_VERSION: 0.5.1
    IDRIS2_CG: racket
    MSYSTEM: MINGW64
    MSYS2_PATH_TYPE: inherit
    CC: gcc
    RACKET_RACO: raco
    PWD: /d/a/Idris2/Idris2
Cloning into 'ChezScheme'...
error: RPC failed; curl 56 OpenSSL SSL_read: Connection was reset, errno 10054
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
```

and yet  the action's execution continues, and fails at a later step (at `cd ChezScheme`: no such directory).

attila-lendvai:
the hung elab CI run is also due to a network issue (stuck at repo checkout).

mattpolzin:
This looks good to me! Just waiting for CI to re-run. I might try to fix that quirk of the CI run where it can fail during cloning ChezScheme and still continue on to inevitably also fail in a future step, but that shouldn't be the concern of this PR.

PR #2224, opened 2021-12-26
[WIP] Request for discussion: Remove coreLift_

dagit:
Preface: I imagine there needs to be some discussion and consensus reached about these changes. I find having a concrete PR to point at helps with the discussion.

I was helping someone on the Discord server work through some build failures. What we noticed is that it seemed like the step that was failing was doing so silently despite having `set -x` in scripts and using `make`. Then a later step was failing because a file, `idris2.so`, was missing.

Looking at `src/Compiler/Scheme/Chez.idr`, I realized it's possible that the uses of `coreLift_` could be masking errors. Particularly when combined with `system` or file system operations. Then I discovered this related discussion: https://github.com/idris-lang/Idris2/issues/1244#issuecomment-856714168

I'm not sure if the space in the path was causing an exit code of `-1` or if the path was simply causing a file to be created in the wrong directory. So I can't say for sure that the changes I'm proposing would have made that failure more obvious, but it seems likely. And I think these changes will make it possible for us to report an error to the user as soon as something has failed instead of catching it later.

With all that out of the way, what I'm proposing is to avoid functions with a type like `m a -> n ()` that are implement in terms of `ignore` as much as possible. In this case, I decided to completely remove `coreLift_` so that I could easily find/audit all the uses. In the end, I decided to just leave it gone, but I could imagine adding it back and just being cautious about using it.

For functions with types like `system : String -> io Int` or `a -> io (Either Error ())` that are being used in the context of `Core`, it's probably better to wrap them in something that checks the error condition and `throws` the error. To that end, I created `system_` and `chmodRaw_`, that use `throw` and return `Core ()`. I just arbitrarily put them in `Core.System` and I haven't added their doc strings yet.

In other cases, it seemed better to check the result at the call site. I guess the main reason to do that is so that the error message can be customized to fit the use site. And maybe that implies I should be using some function like `Either a b -> (a -> Error) -> Core b` that `throw`s the `Error`. Although, idris makes it convenient to handle those cases right at the call site.

I started auditing uses of `ignore` too, but I decided I made a big enough mess for one day and that I should probably talk to others before spending a long time on it.

As an aside, I see that `coreLift_` was introduced when fixing #758 and I'm hope I'm not breaking anything by not ignoring return values. All the tests passed locally with these changes, so that is a good sign.

### DIFF
--- a/src/Idris/REPL.idr
+++ b/src/Idris/REPL.idr
@@ -14,6 +14,7 @@ import Core.LinearCheck
 import Core.Metadata
 import Core.Normalise
 import Core.Options
+import Core.System
 import Core.TT
 import Core.Termination
 import Core.Unify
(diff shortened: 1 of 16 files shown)


src/Idris/REPL.idr
```diff
@@ -238,8 +239,10 @@ updateFile update
              | Nothing => pure (DisplayEdit emptyDoc) -- no file, nothing to do
          Right content <- coreLift $ readFile f
                | Left err => throw (FileErr f err)
-         coreLift_ $ writeFile (f ++ "~") content
-         coreLift_ $ writeFile f (unlines (update (forget $ lines content)))
+         Right () <- coreLift $ writeFile (f ++ "~") content
```
alrunner4, line 242: There's a `Core.Core.writeFile` with this behavior.

alrunner4 (commented):
My 2¢: I'm in favor of discouraging the use of return-ignoring `_` functions, especially with the nice alternate case syntax available in `do` notation. So many `coreLift $ putStrLn ...` in the diff that I almost want to suggest a `corePutStr`, too, but that's stretching the scope of a patch like this.

src/Idris/REPL.idr
dagit: Thanks.

gallais:
I don't think it makes sense to remove it entirely: people will just write `_ <- ...` instead.
One possible stopgap would be for `ignore` (and friends) to take their type argument
explicitly this way you'd at least be forced to reconsider an `ignore Int funcall` if `funcall`
starts returning a `Maybe Error` instead of `Int`.

andrevidela:
Given the timeline and the lack of interest I'm going to close this. Feel free to reopen if someone wants to revive the topic

PR #2230, opened 2021-12-29
[cleanup] Unhide string module

mattpolzin:
When the effect of the `base` `String.lines` and `String.unlines` functions changed subtly a while back, a compatibility layer was added to the compiler and these `String` module functions were hidden.

This PR unhides and switches back to the `String` module versions of these functions. It also removes `replicate` from `Libraries.Data.String.Extra` because that function has since been ported to `base`.

### DIFF
--- a/libs/base/Data/String.idr
+++ b/libs/base/Data/String.idr
@@ -112,7 +112,7 @@ lines' s = linesHelp [] s
 ||| Splits a string into a list of newline separated strings.
 |||
 ||| The emptry string becomes an empty list. The last newline, if not followed by
-||| any additional characters, is eaten (there will never be an empt string last element
+||| any additional characters, is eaten (there will never be an empty string last element
 ||| in the result).
 |||
 ||| ```idris example
--- a/libs/base/Data/String.idr
+++ b/libs/base/Data/String.idr
@@ -10,7 +10,7 @@ singleton : Char -> String
 singleton c = strCons c ""
 
 ||| Create a string by using n copies of a character
-export
+public export
 replicate : Nat -> Char -> String
 replicate n c = pack (replicate n c)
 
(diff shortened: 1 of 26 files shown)


libs/base/Data/String.idr
```diff
@@ -90,6 +90,10 @@ unwords = pack . unwords' . map unpack
 
 ||| Splits a character list into a list of newline separated character lists.
 |||
+||| The emptry string becomes an empty list. The last newline, if not followed by
+||| any additional characters, is eaten (there will never be an empt string last element
```
gallais, line 94:
```suggestion
||| any additional characters, is eaten (there will never be an empty string last element
```

libs/base/Data/String.idr
```diff
@@ -107,6 +111,10 @@ lines' s = linesHelp [] s
 
 ||| Splits a string into a list of newline separated strings.
 |||
+||| The emptry string becomes an empty list. The last newline, if not followed by
+||| any additional characters, is eaten (there will never be an empt string last element
```
gallais, line 115:
```suggestion
||| any additional characters, is eaten (there will never be an empty string last element
```

gallais (commented):
Looks good!

libs/base/Data/String.idr
```diff
@@ -90,6 +90,10 @@ unwords = pack . unwords' . map unpack
 
 ||| Splits a character list into a list of newline separated character lists.
 |||
+||| The emptry string becomes an empty list. The last newline, if not followed by
```
mattpolzin, line 93:
```suggestion
||| The empty string becomes an empty list. The last newline, if not followed by
```

libs/base/Data/String.idr
```diff
@@ -107,6 +111,10 @@ lines' s = linesHelp [] s
 
 ||| Splits a string into a list of newline separated strings.
 |||
+||| The emptry string becomes an empty list. The last newline, if not followed by
```
mattpolzin, line 114:
```suggestion
||| The empty string becomes an empty list. The last newline, if not followed by
```

mattpolzin (commented):
Found two more typos. I can be all butter fingers sometimes. No wonder I like static type systems so much ;)

PR #2235, opened 2021-12-31
Refactor string functions, rewrite tests

### LINKED ISSUE #2172: Refactor string functions using join
`unwords`, `unlines`, etc. have ad-hoc implementations that should be refactored via

```idris
join : String -> List String -> String
join sep ws = concat (intersperse sep ws)
```

dignissimus:
Resolves #2172


Z-snails:
You need to add the test to `tests/Main.idr` for it to actually run.

libs/base/Data/String.idr
```diff
@@ -132,7 +133,8 @@ unlines' (l::ls) = l ++ '\n' :: unlines' ls
 ||| ```
 export
 unlines : List String -> String
-unlines = pack . unlines' . map unpack
+unlines [""] = "\n"
+unlines = join "\n"
```
buzden, line 137: All patterns in pattern matching must have the same count of arguments they are matching at, so you'd better to have `unlines ss = join "\n" ss` for the second line.

Anyway, this implementation is not equivalent to the current `unlines` implementation. For example, `unlines ["", ""]` now gives `"\n\n"`, your implementation gives `"\n"`.

libs/base/Data/String.idr
```diff
@@ -1,5 +1,6 @@
 module Data.String
 
+import Data.String.Extra
```
buzden, line 3: This looks like a dependency of `base` to `contrib` which should not be. I think it's better to move `join` function from `contrib` to `base` for that case.

libs/base/Data/String.idr
mattpolzin: Just to further the point made in the above comment, here's a reference to when the semantics of lines/unlines were most recently changed: https://github.com/idris-lang/Idris2/pull/1585

That is, the current semantics were intentional instead of accidental so we should honor them in the changes made in this PR. That might make the suggestion of using `join` less obvious of a win, at least for unlines, though I wish I'd thought to mention that on the open GitHub ticket in the first place.

libs/base/Data/String.idr
mattpolzin: Big +1 from me on moving `join` from contrib to base.

dignissimus: https://github.com/idris-lang/Idris2/blob/54332752f0cb43645f365d6780a00c3348a4377e/libs/base/Data/String.idr#L41-L52

How's this?

libs/base/Data/String.idr
dignissimus: Should unlines be written to use `join`? I ended up writing the following which doesn't include join so I'm considering reverting this specific change since it won't be any different to the original code.

https://github.com/idris-lang/Idris2/blob/54332752f0cb43645f365d6780a00c3348a4377e/libs/base/Data/String.idr#L147-L149

mattpolzin: Since the function isn't publicly exported and it is transformed to `fastUnlines` at runtime, I suppose it's neither here nor there whether it unpacks and runs on characters or not. I think I slightly prefer your new approach personally.

mattpolzin:
You've gone a different direction with your test file -- making a single file and name spacing it by folder in a way that reminds me of unit tests instead of sequentially increasing test folders. 

Your approach to the rename both makes sense to me and at the same time makes me wonder if it forces an awkward middle ground between the golden testing we have today and the unit testing we would ideally want for a module in the base library but cannot have yet because our test harness only supports golden.

So, I'm on the fence about the file renames and relocation within the test folders, but aside from that (as noted above by @Z-snails) , you'll need to add the test to the Main file in the test harness for it to be picked up (it isn't running right now).

mattpolzin:
Right now you've got some build failures in the compiler where you've added functions to Data.String that were previously added to the compiler's internal Libraries.Data.String.Extra. 

We can't remove the definitions from Libraries.Data.String.Extra until after the next release because the compiler depends on base and needs to be able to build against the _previous_ version of base (before you added those functions to base). 

This is awkward, but not uncommon when working in the compiler codebase. You should be able to add `%hide Data.String.join` (and for drop) to the top of files that currently use those functions. You could also modify those files in the compiler to fully qualify their calls to the functions as `Libraries.Data.String.Extra.join`. 

After the next release, that workaround can be removed (as is done in another currently open PR for a similar reason: https://github.com/idris-lang/Idris2/pull/2230).

mattpolzin:
I'm immediately thinking the circumstance is different here and the compiler should I think complain if you try to hide the version you're adding now before it's available in the previous version of base. 

You'll probably need to use the second approach I mentioned of fully qualifying the calls to Libraries.Data.String.Extra.

andrevidela:
Given that this has gone stale for quite a while I'm going to close this. If someone else wants to try again, feel free (@ProofOfKeags)

PR #2238, opened 2021-12-31
[ refactor ] More IDE protocol

ohad:
Include the changes in the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,24 @@
   `prim__void`.
 * Adds `%deprecate` pragma that can be used to warn when deprecated functions are used.
 
+### IDE protocol changes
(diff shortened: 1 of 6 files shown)


gallais (approved):

ohad:
Thanks!

I'm not sure what to do about `idris2protocols.ipkg.hide`. It's currently in a no-good state, but getting it into a consistent state will take some work, and in any case it interferes with interactive editing at the moment. 

Is it better to just delete it and re-introduce it when we can make it consistent?

gallais:
Ah I see, I hadn't noticed the commented out bits were in the .hide file.
I guess we can merge it as is given it's in a (so far) unused file.

PR #2240, opened 2021-12-31
Change associativity for System.Path operators

eayus:
Currently the operators for combining paths associate to the right, but it would be more convenient if they associated to the left instead. For example, currently you cannot write:
```idris
root /> "usr" /> "lib"
```
As this is parsed as:
```idris
root /> ("usr" /> "lib")
```
But `/>` takes a `Path` on the LHS and so this is invalid.

By making the operators associate to the left, you can write this example without any extra brackets.

### DIFF
--- a/libs/contrib/System/Path.idr
+++ b/libs/contrib/System/Path.idr
@@ -13,7 +13,7 @@ import Text.Quantity
 
 import System.Info
 
-infixr 5 </>, />
+infixl 5 </>, />
 infixr 7 <.>
 
 
--- a/src/Libraries/Utils/Path.idr
+++ b/src/Libraries/Utils/Path.idr
@@ -19,7 +19,7 @@ import System.File
 
 %default total
 
-infixr 5 </>, />
+infixl 5 </>, />
 infixr 7 <.>
 
 


mattpolzin:
Makes sense. I wonder if we want to make the same change to the version at src/Libraries/Extra/Path.idr

On one hand, if it's a good idea here I would expect it to be a good idea there. On the other, we'd be fixing something that isn't broken (for the compiler) so always a danger of breaking existing code.

buzden:
> On one hand, if it's a good idea here I would expect it to be a good idea there. On the other, we'd be fixing something that isn't broken (for the compiler) so always a danger of breaking existing code.

I think such incoherence between actual library and `src/Libraries` is a deferred broken code. Some time later when refactoring libraries (e.g. factoring out system library out of `contrib` with dependency of the compiler on it) or just moving code between libraries and compiler...

mattpolzin:
I definitely tend to agree. It will be a bit different I suspect once things like this are moved out of the compiler repo because at that point drift between the compiler and some external package that happens to have been born of the same code will seem inevitable if not desirable (to allow the external package to gain features not needed by the compiler). Nevertheless, here we are today with the two files clearly intended to mirror each other at their creation.

gallais:
Yes, can we please port the fix to the compiler libs too?

eayus:
Ok the same change has been made to the compiler lib

PR #2243, opened 2022-01-02
[ fix ] mutual tail recursion

stefan-hoeck:
The tail recursion optimizer used in the JS backends generates n copies of the same optimized functions for n mutually tail recursive functions as discussed on [discord](https://discord.com/channels/827106007712661524/827109190266257409/926986207366361138). This fix makes sure each function is only generated once.

Thanks @dunhamsteve, for finding this and suggesting the fix.

### DIFF
--- a/src/Compiler/ES/TailRec.idr
+++ b/src/Compiler/ES/TailRec.idr
@@ -195,16 +195,13 @@ hasTailCalls g (x ::: Nil) = maybe False (contains x) $ lookup x g
 hasTailCalls _ _           = True
 
 -- Given a strongly connected group of functions, plus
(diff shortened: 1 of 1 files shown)


gallais:
Can we get a test case for this please?

stefan-hoeck:
I'll need some help with this. Node and browsers like firefox seem to cope with this bug just fine (that's why it went undetected so far), so in a test case, we'll need to look at the generated JS. In Linux, I'd use command line utilities like `grep`, `sort`, and `count` for this. Can I assume that these will be available on all platform this test will be run on? Or is there another way to go about this?

gallais:
Let's not bother for now then.

PR #2245, opened 2022-01-02
[ prelude ] Give every `Show` type a default `Interpolation`

andylokandy:
Now every type implemented `Show` is allowed to be interpolated into strings without following a `show`.

Previous discussion on https://github.com/idris-lang/Idris2/pull/1967#issuecomment-1003534258

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,11 +43,14 @@
   concat [interpolate "hello ", interpolate world]
   ```
 
-  This allows you to write expressions within slices without having to call `show`
-  but for this you need to implement the `Interpolation` interface for each type
-  that you intend to use within an interpolation slice. The reason for not reusing
-  `Show` is that `Interpolation` and `Show` have conflicting semantics, typically
-  this is the case for `String` which adds double quotes around the string.
+  This allows you to write expressions within slices without having to call `show`.
+  The prelude provides a default `Interpolation` implementation for each type that
+  implments `Show`. The reason for not reusing `Show` is that `Interpolation` and
+  `Show` have conflicting semantics, typically this is the case for `String` which
+  adds double quotes around the string.
+
+  In case that you'd like to implement customized `Interpolation` to a type, you
+  may have to avoid ambiguous by `%hide interpDefault`.
 
 ### Compiler changes
 
--- a/libs/prelude/Prelude/Interpolation.idr
+++ b/libs/prelude/Prelude/Interpolation.idr
@@ -1,5 +1,7 @@
 module Prelude.Interpolation
 
+import Prelude.Show
+
 ||| Interpolated strings are of the form `"xxxx \{ expr } yyyy"`.
 ||| In this example the string `"xxxx "` is concatenated with `expr` and
 ||| `" yyyy"`, since `expr` is not necessarily a string, the generated
@@ -14,6 +16,17 @@ interface Interpolation a where
   interpolate : a -> String
 
 ||| The interpolation instance for Strings is the identity
-export
-Interpolation String where
+public export
+[interpId] Interpolation String where
   interpolate x = x
+
+public export
+[interpShow] Show a => Interpolation a where
+  interpolate x = show x
+
+%hint
+%inline
+public export
+interpDefault : {a : _} -> Show a => Interpolation a
+interpDefault {a = String} = interpId
+interpDefault {a = _} = interpShow
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -503,7 +503,9 @@ mutual
     where
       toRawImp : PStr -> Core RawImp
       toRawImp (StrLiteral fc str) = pure $ IPrimVal fc (Str str)
-      toRawImp (StrInterp fc tm) = desugarB side ps tm
+      toRawImp (StrInterp fc tm)
+        = let vfc = virtualiseFC fc in
+          pure $ IApp vfc (IVar EmptyFC (UN $ Basic "interpolate")) !(desugarB side ps tm)
 
       -- merge neighbouring StrLiteral
       mergeStrLit : List PStr -> List PStr
@@ -524,12 +526,8 @@ mutual
       strInterpolate []
         = IVar EmptyFC (NS preludeNS $ UN $ Basic "Nil")
       strInterpolate (x :: xs)
-        = let xFC = virtualiseFC (getFC x) in
-          apply (IVar xFC (NS preludeNS $ UN $ Basic "::"))
-          [ IApp xFC (IVar EmptyFC (UN $ Basic "interpolate"))
-                     x
-          , strInterpolate xs
-          ]
+        = let vFC = virtualiseFC (getFC x) in
+          apply (IVar vFC (NS preludeNS $ UN $ Basic "::")) [x, strInterpolate xs]
 
   trimMultiline : FC -> Nat -> List (List PStr) -> Core (List PStr)
   trimMultiline fc indent lines
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -45,7 +45,7 @@ idrisTestsBasic = MkTestPool "Fundamental language features" [] Nothing
        "basic066", "basic067", "basic068",
        "idiom001",
        "interpolation001", "interpolation002", "interpolation003",
-       "interpolation004"]
+       "interpolation004", "interpolation005"]
 
 idrisTestsCoverage : TestPool
 idrisTestsCoverage = MkTestPool "Coverage checking" [] Nothing
--- a/tests/idris2/docs001/expected
+++ b/tests/idris2/docs001/expected
@@ -164,7 +164,6 @@ Hints:
   Cast String Nat
   Eq String
   FromString String
-  Interpolation String
   Monoid String
   Ord String
   Semigroup String
--- a/tests/idris2/interpolation003/Test.idr
+++ b/tests/idris2/interpolation003/Test.idr
@@ -1,5 +1,10 @@
 import Data.String
 
+%hide interpDefault
+
+Interpolation String where
+  interpolate x = x
+
 Interpolation Bool where
   interpolate True = "TT"
   interpolate False = "FF"
@@ -11,5 +16,10 @@ Interpolation Nat where
 Interpolation Integer where
   interpolate = show
 
-test : String
-test = "\{2}\{True}: \{True}\{True}"
+test : List String
+test = [
+    "This is \{True}",
+    "This is \{False}",
+    "This is a number : \{S (S (S Z))}",
+    "\{2}\{True}: \{True}\{True}"
+  ]
--- a/tests/idris2/interpolation003/expected
+++ b/tests/idris2/interpolation003/expected
@@ -1,8 +1,5 @@
 1/1: Building Test (Test.idr)
-Main> "This is TT"
-Main> "This is FF"
-Main> "This is a number : S S S 0"
-Main> "2TT: TTTT"
-Main> Main.test : String
-test = concat [interpolate 2, interpolate True, interpolate ": ", interpolate True, interpolate True]
+Main> ["This is TT", "This is FF", "This is a number : S S S 0", "2TT: TTTT"]
+Main> Main.test : List String
+test = [concat [interpolate "This is ", interpolate True], concat [interpolate "This is ", interpolate False], concat [interpolate "This is a number : ", interpolate 3], concat [interpolate 2, interpolate True, interpolate ": ", interpolate True, interpolate True]]
 Main> Bye for now!
--- a/tests/idris2/interpolation003/expected
+++ b/tests/idris2/interpolation003/expected
@@ -1,5 +1,5 @@
 1/1: Building Test (Test.idr)
 Main> ["This is TT", "This is FF", "This is a number : S S S 0", "2TT: TTTT"]
 Main> Main.test : List String
-test = [concat [interpolate "This is ", interpolate True], concat [interpolate "This is ", interpolate False], concat [interpolate "This is a number : ", interpolate 3], concat [interpolate 2, interpolate True, interpolate ": ", interpolate True, interpolate True]]
+test = [concat ["This is ", interpolate True], concat ["This is ", interpolate False], concat ["This is a number : ", interpolate 3], concat [interpolate 2, interpolate True, ": ", interpolate True, interpolate True]]
 Main> Bye for now!
--- a/tests/idris2/interpolation003/input
+++ b/tests/idris2/interpolation003/input
@@ -1,6 +1,3 @@
-"This is \{True}"
-"This is \{False}"
-"This is a number : \{S (S (S Z))}"
 test
 :printdef test
 :q
--- a/tests/idris2/interpolation005/Test.idr
+++ b/tests/idris2/interpolation005/Test.idr
@@ -1,0 +1,4 @@
+import Data.String
+
+test : String
+test = "\{2}\{True}: \{True}\{True}"
--- a/tests/idris2/interpolation005/expected
+++ b/tests/idris2/interpolation005/expected
@@ -1,0 +1,8 @@
+1/1: Building Test (Test.idr)
+Main> "This is True"
+Main> "This is False"
+Main> "This is a number : 3"
+Main> "2True: TrueTrue"
+Main> Main.test : String
+test = concat [interpolate 2, interpolate True, interpolate ": ", interpolate True, interpolate True]
+Main> Bye for now!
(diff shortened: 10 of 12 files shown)


andylokandy:
cc @mattpolzin @gallais @andrevidela

src/Idris/Parser.idr
```diff
@@ -1397,6 +1397,8 @@ fnDirectOpt : OriginDesc -> Rule PFnOpt
 fnDirectOpt fname
     = do pragma "hint"
          pure $ IFnOpt (Hint True)
+  <|> do pragma "chaser"
```
andrevidela, line 1400: Why is this necessary compared to not using the pragma? I assumed every definition was `Hint False` by default?

andylokandy: Actually, every definition isn't `Hint _` by default.

andylokandy: Hmmm, seems that this can't be merged until `chaser` is available on the new boostrap.

andylokandy:
Thanks to @ohad for the [suggestion](https://github.com/idris-lang/Idris2/pull/2247#issuecomment-1046009280), we're now resolving the ambiguity explicitly by runtime pattern matching on the type `a`. In most cases, the type `a` should be able to be erased during the compilation.

gallais:
I'm not sure I'm comfortable merging this without good evidence it does not imply hanging onto
more types at runtime.

andylokandy:
I tried to codegen two samples into chez scheme, it seems that the type indeed get earased, but because I'm not familiar with scheme I'm not 100% sure about it.

```idris
main : IO ()
main = putStr "\{2}\{True}: \{True}\{True}"
```

> ```(define Main-main(blodwen-lazy (lambda () (PreludeC-45IO-putStr (cons (vector (vector (lambda (u--b) (lambda (u--a) (lambda (u--func) (lambda (arg-8478) (lambda (eta-0) (PreludeC-45IO-u--map_Functor_IO u--func arg-8478 eta-0)))))) (lambda (u--a) (lambda (arg-9181) (lambda (eta-0) arg-9181))) (lambda (u--b) (lambda (u--a) (lambda (arg-9187) (lambda (arg-9194) (lambda (eta-0) (let ((act-17 (arg-9187 eta-0))) (let ((act-16 (arg-9194 eta-0))) (act-17 act-16))))))))) (lambda (u--b) (lambda (u--a) (lambda (arg-9661) (lambda (arg-9664) (lambda (eta-0) (let ((act-24 (arg-9661 eta-0))) ((arg-9664 act-24) eta-0))))))) (lambda (u--a) (lambda (arg-9675) (lambda (eta-0) (let ((act-51 (arg-9675 eta-0))) (act-51 eta-0)))))) (lambda (u--a) (lambda (arg-10912) arg-10912))) (PreludeC-45Interfaces-concat (cons (lambda (arg-8067) (lambda (arg-8070) (string-append arg-8067 arg-8070))) "") (vector (lambda (u--acc) (lambda (u--elem) (lambda (u--func) (lambda (u--init) (lambda (u--input) (PreludeC-45Types-u--foldr_Foldable_List u--func u--init u--input)))))) (lambda (u--elem) (lambda (u--acc) (lambda (u--func) (lambda (u--init) (lambda (u--input) (PreludeC-45Types-u--foldl_Foldable_List u--func u--init u--input)))))) (lambda (u--elem) (lambda (arg-10161) (PreludeC-45Types-u--null_Foldable_List arg-10161))) (lambda (u--elem) (lambda (u--acc) (lambda (u--m) (lambda (i_con-0) (lambda (u--funcM) (lambda (u--init) (lambda (u--input) (PreludeC-45Types-u--foldlM_Foldable_List i_con-0 u--funcM u--init u--input)))))))) (lambda (u--elem) (lambda (arg-10190) arg-10190)) (lambda (u--a) (lambda (u--m) (lambda (i_con-0) (lambda (u--f) (lambda (arg-10204) (PreludeC-45Types-u--foldMap_Foldable_List i_con-0 u--f arg-10204))))))) (cons (PreludeC-45Interpolation-u--interpolate_Interpolation_interpShowC-36a (cons (lambda (u--x) (PreludeC-45Show-u--show_Show_Integer u--x)) (lambda (u--d) (lambda (u--x) (PreludeC-45Show-u--showPrec_Show_Integer u--d u--x)))) 2) (cons (csegen-36) (cons ": " (cons (csegen-36) (cons (csegen-36) '()))))))))))```

---

```idris
%hide interpDefault

Interpolation String where
  interpolate x = x

Interpolation Bool where
  interpolate = show

Interpolation Integer where
  interpolate = show

main : IO ()
main = putStr "\{2}\{True}: \{True}\{True}"
```

> ```(define Main-main(blodwen-lazy (lambda () (PreludeC-45IO-putStr (cons (vector (vector (lambda (u--b) (lambda (u--a) (lambda (u--func) (lambda (arg-8478) (lambda (eta-0) (PreludeC-45IO-u--map_Functor_IO u--func arg-8478 eta-0)))))) (lambda (u--a) (lambda (arg-9181) (lambda (eta-0) arg-9181))) (lambda (u--b) (lambda (u--a) (lambda (arg-9187) (lambda (arg-9194) (lambda (eta-0) (let ((act-17 (arg-9187 eta-0))) (let ((act-16 (arg-9194 eta-0))) (act-17 act-16))))))))) (lambda (u--b) (lambda (u--a) (lambda (arg-9661) (lambda (arg-9664) (lambda (eta-0) (let ((act-24 (arg-9661 eta-0))) ((arg-9664 act-24) eta-0))))))) (lambda (u--a) (lambda (arg-9675) (lambda (eta-0) (let ((act-51 (arg-9675 eta-0))) (act-51 eta-0)))))) (lambda (u--a) (lambda (arg-10912) arg-10912))) (PreludeC-45Interfaces-concat (cons (lambda (arg-8067) (lambda (arg-8070) (string-append arg-8067 arg-8070))) "") (vector (lambda (u--acc) (lambda (u--elem) (lambda (u--func) (lambda (u--init) (lambda (u--input) (PreludeC-45Types-u--foldr_Foldable_List u--func u--init u--input)))))) (lambda (u--elem) (lambda (u--acc) (lambda (u--func) (lambda (u--init) (lambda (u--input) (PreludeC-45Types-u--foldl_Foldable_List u--func u--init u--input)))))) (lambda (u--elem) (lambda (arg-10161) (PreludeC-45Types-u--null_Foldable_List arg-10161))) (lambda (u--elem) (lambda (u--acc) (lambda (u--m) (lambda (i_con-0) (lambda (u--funcM) (lambda (u--init) (lambda (u--input) (PreludeC-45Types-u--foldlM_Foldable_List i_con-0 u--funcM u--init u--input)))))))) (lambda (u--elem) (lambda (arg-10190) arg-10190)) (lambda (u--a) (lambda (u--m) (lambda (i_con-0) (lambda (u--f) (lambda (arg-10204) (PreludeC-45Types-u--foldMap_Foldable_List i_con-0 u--f arg-10204))))))) (cons (Main-u--interpolate_Interpolation_Integer 2) (cons (Main-u--interpolate_Interpolation_Bool 1) (cons ": " (cons (Main-u--interpolate_Interpolation_Bool 1) (cons (Main-u--interpolate_Interpolation_Bool 1) '()))))))))))```

andylokandy:
Simpler exmaple:

```
Main> :let test = "\{2}\{True}: \{True}\{True}"
Main> :di test
```

> `Compile time tree: [0] (Prelude.Interfaces.concat Prelude.Basics.List String Prelude.Types.Monoid implementation at Prelude.Types:663:1--665:15 Prelude.Types.Foldable implementation at Prelude.Types:443:1--456:59 (Prelude.Basics.:: String (Prelude.Interpolation.interpolate Integer (Prelude.Interpolation.interpDefault Integer Prelude.Show.Show implementation at Prelude.Show:110:1--112:50) (Prelude.Num.fromInteger Integer Prelude.Num.defaultInteger 2)) (Prelude.Basics.:: String (Prelude.Interpolation.interpolate Prelude.Basics.Bool (Prelude.Interpolation.interpDefault Prelude.Basics.Bool Prelude.Show.Show implementation at Prelude.Show:196:1--199:23) Prelude.Basics.True) (Prelude.Basics.:: String (Prelude.Interpolation.interpolate String (Prelude.Interpolation.interpDefault String Prelude.Show.Show implementation at Prelude.Show:188:1--190:57) ": ") (Prelude.Basics.:: String (Prelude.Interpolation.interpolate Prelude.Basics.Bool (Prelude.Interpolation.interpDefault Prelude.Basics.Bool Prelude.Show.Show implementation at Prelude.Show:196:1--199:23) Prelude.Basics.True) (Prelude.Basics.:: String (Prelude.Interpolation.interpolate Prelude.Basics.Bool (Prelude.Interpolation.interpDefault Prelude.Basics.Bool Prelude.Show.Show implementation at Prelude.Show:196:1--199:23) Prelude.Basics.True) (Prelude.Basics.Nil String)))))))
Run time tree: [0] (Prelude.Interfaces.concat [__] [__] Prelude.Types.Monoid implementation at Prelude.Types:663:1--665:15 Prelude.Types.Foldable implementation at Prelude.Types:443:1--456:59 (Prelude.Basics.:: [__] (Prelude.Interpolation.interpolate [__] (Prelude.Interpolation.interpDefault Integer Prelude.Show.Show implementation at Prelude.Show:110:1--112:50) (Prelude.Num.fromInteger [__] Prelude.Num.defaultInteger 2)) (Prelude.Basics.:: [__] (Prelude.Interpolation.interpolate [__] (Prelude.Interpolation.interpDefault Prelude.Basics.Bool Prelude.Show.Show implementation at Prelude.Show:196:1--199:23) Prelude.Basics.True) (Prelude.Basics.:: [__] (Prelude.Interpolation.interpolate [__] (Prelude.Interpolation.interpDefault String Prelude.Show.Show implementation at Prelude.Show:188:1--190:57) ": ") (Prelude.Basics.:: [__] (Prelude.Interpolation.interpolate [__] (Prelude.Interpolation.interpDefault Prelude.Basics.Bool Prelude.Show.Show implementation at Prelude.Show:196:1--199:23) Prelude.Basics.True) (Prelude.Basics.:: [__] (Prelude.Interpolation.interpolate [__] (Prelude.Interpolation.interpDefault Prelude.Basics.Bool Prelude.Show.Show implementation at Prelude.Show:196:1--199:23) Prelude.Basics.True) (Prelude.Basics.Nil [__])))))))`

stefan-hoeck:
Sorry for chiming in here, but I'm not sure that this is really an improvement over the status quo. I think there was an agreement that `Show` and `Interpolation` have different semantics, but with this PR we make it *harder* for programmers to write and use custom implementations instead of using the default, which is based on `Show`. If I understand correctly, custom implementations will require a `%hide interpDefault` to work correctly if they also come with a `Show` implementation, in which case all data types using `interpDefault` will no longer be available during interpolation. I wonder if it's such a bad thing to require a single line of code for every implementation of `Interpolation`:

```idris
Interpolation Foo where interpolate = show
```

stefan-hoeck:
> @stefan-hoeck In 99.99% of the cases, you won't add your own `Interpolation` implementation, you will explicitly convert it into `String`, e.g. to print an integer in hexadecimal while another one in binary: `"\{hex 100} \{bin 100}"`. The most significant reason for having `Interpolation` instead of `Show` is to specially handle `String`.

If this is true, why do we even need the `Show a => Interpolation a` default case at all? My fear is that this adds a significant layer of complexity and obscurity to the Prelude, which is - according to your numbers - hardly ever needed. Why can't we just have

```idris
Interpolation String where interpolate x = x
```
and nothing else in the Prelude?

andrevidela:
> In other corner cases, you'd better format it into String rather than defining your own Interpolation

I'm not sure I understand this correctly. The `Interpolation` interface is precisely here to handle corner cases where you want to format your data specially for interpolation. It's not hard to come by either: Typically record types have multiple fields that you want to show across multiple lines, but if you insert it within an interpolated string, the line breaks will mess up your formatting, so the `Interpolation` instance should not have any line breaks. Here is a very real example:

```
record HTTPRequest where
  method : String
  uri: String
  header : String
  body : String

Show HTTPRequest where
  show req = """
    method: \{req.method}
    uri: \{req.uri}
    header: \{req.header}
    body: \{req.body}
    """
Interpolation HTTPRequest where
  interpolate req = "\{req.Method} \{req.uri}"

...

-- Use show to render Requests by themselves
sendRequest : HTTPRequest -> IO ()
sendRequest req = server.send (show req)

...

-- Use interpolate to insert in other strings
parseRequest : RequestParser a => Request -> Log a
parseRequest req = case parse req of Nothing => LOG Error "could not parse \{req}"; Just v => pure v
```

I agree that having to type `%hide interpDefault` is a usability regression. Maybe the opposite would be more desirable? that is, having to import the default interpolation

```
import Interpolation.Default

-- now we have access to Show a => Interpolation a
```

This also affords is the option to backpedal without making a breaking change if we find a solution to ambigious interfaces, what do you think?

andylokandy:
@andrevidela The users may also want to use the multiline version `HTTPRequest ` in interpolation. I feel uncomfortable making `Interpolation` a formatting-behavior variant against `Show`, like `Debug` and `Display` in Rust.

To me, the first reasonable solution to your example should be to add a `Display` interface and somehow control whether `Show a => Interpolation a` or `Display a => Interpolation a` to use in string literal interpolation.

Thus, it leads to a design decision of Idris: How to idiomatically add more new format behavior in Idris? Currently, Idris only has `Show` and `Pretty`. Should we learn Rust and then add a bunch of formatter interfaces, e.g. [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html), [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html), [Octal](https://doc.rust-lang.org/std/fmt/trait.Octal.html), [Binary](https://doc.rust-lang.org/std/fmt/trait.Binary.html) and [more](https://doc.rust-lang.org/std/fmt/index.html#:~:text=Traits-,Binary,X%20formatting.,-Write) and add builtin notations in string literal to tell compiler which one to use?

I think functions are more flexible and idiomatic in Idris rather than builtin notations, so I think we should add formatting functions to add new format behavior. Then the `Show` is really a structure printer for developers like `Debug` in Rust.

To give a concrete example on top of yours:
```idris
record HTTPRequest where
  method : String
  uri: String
  header : String
  body : String

Show HTTPRequest where
  show req = """
    method: \{req.method}
    uri: \{req.uri}
    header: \{req.header}
    body: \{req.body}
    """

display : HTTPRequest -> String
display req = "\{req.Method} \{req.uri}"
...

-- Use show to render Requests by themselves
sendRequest : HTTPRequest -> IO ()
sendRequest req = server.send (show req)

...

-- Use interpolate to insert in other strings
parseRequest : RequestParser a => Request -> Log a
parseRequest req = case parse req of Nothing => LOG Error "could not parse \{display 
 req}"; Just v => pure v
```

andylokandy:
> that is, having to import the default interpolation ... This also affords is the option to backpedal without making a breaking change if we find a solution to ambigious interfaces, what do you think?

Once imported in a module, you have no way to use any type that manually implements `Interpolation` in string literal because of ambiguity. And you aren't even able to `%hide` that unnamed implementation. Which means we have to reach an agreement on whether do or not to do.

stefan-hoeck:
There more I read about this, the more this sounds like a code smell. Why do we have an `Interpolation` interface in the first place if client code is not supposed to implement it? I agree with you, @andylokandy, that the best solution would be to use explicit string conversion functions anyway. It's slightly more verbose but comes with much less mental overhead and therefore, fewer surprises. So: If users are not supposed to implement `Interpolation`, I'd suggest to get rid of it altogether and only allow expressions of type `String` in interpolation. From my own experience, this works very well.

andylokandy:
> If users are not supposed to implement Interpolation, I'd suggest to get rid of it altogether and only allow expressions of type String in interpolation. From my own experience, this works very well.

@stefan-hoeck  Yes! Exactly the same as [my previous comment](https://github.com/idris-lang/Idris2/issues/555#issuecomment-924753627). Anyway, I now believe a default `Show a => Interpolation a` also works well in many cases, especially for enums. So I still think `Interpolation` is necessary but can be hidden for most users.

gallais:
> If users are not supposed to implement `Interpolation`,

I don't understand where this impression came from.

Adding your own implementation is only a problem if we accept the idea of adding this `Show a => Interpolation a`.
Which to me indicates the issue is with this overly ambiguous implementation rather than with `Interpolation`.

andylokandy:
> > If users are not supposed to implement Interpolation,
>
> I don't understand where this impression came from.

The idea comes from [this comment](https://github.com/idris-lang/Idris2/pull/2245#:~:text=Therefore%2C%20it%20leads,Binary%20and%20more%3F)

stefan-hoeck:
> > If users are not supposed to implement `Interpolation`,
> 
> I don't understand where this impression came from.

This impression comes from the discussion in this PR. Namely:

> In other corner cases, you'd better format it into String rather than defining your own Interpolation

> Adding your own implementation is only a problem if we accept the idea of adding this `Show a => Interpolation a`. Which to me indicates the issue is with this overly ambiguous implementation rather than with `Interpolation`.

Agreed.

Issue #2091, opened 2021-11-02
Split `public export` into `public` and `export`


ohad:
- [X] I have read [CONTRIBUTING.md](../../CONTRIBUTING.md).
- [X] I have checked that there is no existing PR/issue about my proposal.

## Summary

Introduce a new access modifier, `public` which means 'the type-checker can use the definition of the name outside the current namespace`, so for example the following type-checks:

```idris
namespace A
  public
  F : Nat
  F = 42

  public export
  G : Nat
  G = F

Foo : G = 42
Foo = Refl
```

## Motivation

When writing code that involves a lot of type-level computation, we introduced auxiliary definitions and notations that we don't want to expose to client code. However, client code should be able to use the body of these definitions.

This can be particularly painful when we introduce notation inside parameters blocks. For example, here's a little snippet that defines monoids:
```idris
public export
record MonoidStructure where
  constructor MkMonoidStructure
  0 U : Type
  mult : U -> U -> U
  unit : U

parameters (M : MonoidStructure)
  public export
  (ford : a === U M) => Semigroup a where
    (<+>) {ford = Refl} = M .mult

  public export
  (ford : a === U M) => Monoid a where
    neutral {ford = Refl} = M .unit

  record MonoidLaws where
    constructor Validate
    lftNeutrality : (a : U M) ->       Interfaces.neutral <+> a = a
    rgtNeutrality : (a : U M) -> a <+> Interfaces.neutral       = a
    associativity : (a,b,c : U M) ->
                    a <+> (b <+> c) = (a <+> b) <+> c
```

The use of `Prelude`'s Monoid` and `Semigroup` notation makes the development clear and clean.

The problem: any client code that imports this module, also import the `Monoid` instances, and that may clash with existing notations.

Simply `export`ing the implementations won't work, since we use the notation inside types.

## The proposal

Introduce a new modifier, `public`, which means 'the definition will be publicly available to client code, and will be substituted when type-checking if needed`.

### Examples
In the monoid example:
```idris
public export
record MonoidStructure where
  constructor MkMonoidStructure
  0 U : Type
  mult : U -> U -> U
  unit : U

parameters (M : MonoidStructure)
  public
  (ford : a === U M) => Semigroup a where
    (<+>) {ford = Refl} = M .mult

  public
  (ford : a === U M) => Monoid a where
    neutral {ford = Refl} = M .unit

  record MonoidLaws where
    constructor Validate
[...]

PR #2247, opened 2022-01-03
[ parser ] Add pragma `defaulthint` for allowing specialization on hint

andylokandy:
Idris2 supports several types of hints for auto searching, and they, represented by pragmas, settle in the table:

|   | always tried | search by goal type and tried
-- | -- | --
First try | `%globalhint` | `%hint`
Last resort | `%defaulthint` | 

In https://github.com/idris-lang/Idris2/pull/2245 I'm trying to implement `Show a => Interpolation a` which is currently ambiguous with `Interpolation String` where the compiler can't decide which implementation to use for `String`. To solve this problem, we need to support specialization for `%hint` so that overlapping on implementation is allowed and the compiler can pick the most specific one when it is ambiguous.

The missing piece in the table above is the last resort for hints that `search by goal type and tried` which quite fulfills the general purpose of specialization. For example, `Interpolation String`, which is `%hint`, is a sealed and concrete implementation, while `Show a => Interpolation a`, which is the last resort for `%hint`, is only applied when `a` isn't `String`.

As @andrevidela proposed, the missing type of hint sound more reasonable with the occupied name `%defaulthint`, so the hint pragma is going to be renamed to:

|   | always tried | search by goal type and tried
-- | -- | --
First try | `%globalhint` | `%hint`
Last resort | `%defaultglobalhint` | `%defaulthint`

But such modification can't happen without breaking the bootstrap build because the prelude uses the old `%defaulthint` and the bootstrap compiler doesn't recognize `%defaultglobalhint`. Therefore, a workaround is to give `%defaulthint` both old and new sematic in the new compiler, and remains it `%defaulthint` in the prelude. After a new release, we step forward to move `%defaulthint` to the new semantic and rename `%defaulthint` to `%defaultglobalhint` in the prelude. To be clear, the hint semantic in this PR is:

|   | always tried | search by goal type and tried
-- | -- | --
First try | `%globalhint` | `%hint`
Last resort | `%defaultglobalhint`, `%defaulthint` | `%defaulthint`


gallais:
What's an "indirect hint"?

More generally, we'll need some documentation for that feature.

andylokandy:
If I don't get it wrong, the indirect hint is currently the internal mechanism for resolving the constraint of *parent interface*. For #2245, it's basically to avoid ambiguity between `Show a => Interpolation a` and `Interpolation String` on `String`. 

Maybe the `%defaulthint` is better for this purpose. But currently, it won't resolve the `Show a => Interpolation a` with `%defaulthint` because all non-`%defaulthint` are excluded while resolving holes in a `%defaulthint`. Particularly auto-search can't find an implementation for `Show a`. We can enable non-`%defaulthint` while resolving holes but I don't know if it breaks any assumptions.

andylokandy:
> What's an "indirect hint"?

From the comment written by @edwinb, global hints (`%globalhint`) are searched by name and tried one by one, and type hints (`%hint`) are searched by the hint's function type. 
https://github.com/idris-lang/Idris2/blob/a09c5082c5aea25a6d3643f523efdf8fa8efb2ae/src/Core/Context.idr#L737-L752

As we know interface implementations are records under the hood, and the compiler will annotate the anonymous implementation by `%hint` which makes it available when implementation is required. If multiple valid type hints are found, there will be an ambiguity error.

The pragma `%chaser` added by my PR exposes the indirect type hint which is only tired if all direct type hint is falling and the compiler will choose any of the indirect type hints if more than one is viable.

The use case of `%chaser` may be giving an interface a default general implementation while allowing it to be overridden by a more specific implmentation.

andrevidela:
This is probably ok, but as @gallais says it would require documentation, could you please fill up the description for `chaser` in the pragma section? https://github.com/idris-lang/Idris2/blob/main/docs/source/reference/pragmas.rst

This would be a good opportunity to also document `%hint` `%globalHint` and `%defaultHint`. There is also an argument to be made for the name `chaser`. I'm not sure what it would mean and given the description you've given it sounds like the name `defaultHint` would be more descriptive, but it's already taken. Maybe `fallbackHint` would be better?

if I understand correctly

|   | search by name | search by type |
| ------------- | ------------- | ----|
| Search space  |  `globalHint` | `hint` |
| Last resort | `defaultHint` | `chaser`|

is that right?

andylokandy:
@andrevidela Yes, that's how I understand hints. 

> I'm not sure what it would mean and given the description you've given it sounds like the name defaultHint would be more descriptive, but it's already taken. Maybe fallbackHint would be better?

I personally prefer retaking the `%defaulthint`, and renaming the old `%defaulthint` to `%autohint` because it will be tried when the type of given value mismatches with the goal. 

For example:

search by name | search by type
-- | --
Search space | `%globalhint` | `%hint`
Last resort | `%autohint` | `%defaulthint`


This may be viable because `%autohint` is only supposed to be used for primitives and used in the prelude, therefore we have a chance to make this breaking change without breaking users' code.

andrevidela:
A little bit of search reveals:

- https://github.com/search?q=%25globalhint+extension%3Aidr
- https://github.com/search?q=%25defaultHint+extension%3Aidr

I think that's a viable breaking change, we're only breaking custom backends, but there are less than 10 of them so we could even issue pull requests for those to streamline the change, what do you think @gallais ?

andylokandy:
@gallais @andrevidela I've updated according to our last discussion. Anyway, it has a pitfall that, if there are multiple viable `%defaulthint`s, the default hints will get picked randomly by the import order. So I'm going to implement a mechanism to pick the most specific one in that case in another PR if the design in this PR is accepted.

ohad:
I've ran into similar issues/considerations when using implementations to define sets of notation (eg, in frex).

I'm not sure there is a good canonical solution for overlapping instances, including avoiding them. The idea that the type will completely determine the notation globally is not compositional and is at odds with reasoning, especially in the absence of η-equality. (This is a soft argument, based on encountering many examples and designs that are at odds with it. But I could be wrong about it, of course!)

So currently the best I hope for is an ergonomic interface to prioritising `auto`-hints. That was the rational behind adding local `%hint`s.

As you write, it's useful to have a set of notation apply to a whole module, but once the notation starts being imported into other modules, you start getting weird overlapping instance errors.

In #2091, I propose we split `public export` into `public` and `export` (or alternatively into `public export` and `public nonexport`), for both definitions and module imports. 

One reason for doing this is that if one is working internally to a module, you can add hints as `public nonexport`, which means your types and definitions can use the neater notation, but that doesn't pollute the namespace in the rest of the module hierarchy.

Similarly, this extends to sub-modules that set-up the notation. Concretely, a `Foo.Notation` module defines hints (including `Interpolation` instances) as `public export` in a `Foo.Notation` module, but then the main module `Foo` only `import public nonexport Foo.Notation`.

This means that module `Foo` has full benefit of the overloaded notation, but the rest of the module hierarchy that imports `Foo` doesn't have to buy into this notation.

This way the user has finer control over which overloaded notation is in effect, rather than trying to control a heuristic about most specialised types.

andylokandy:
@ohad From the example you've given, the visibility control seems to be orthogonal to the specialization. Correct me if I've got it wrong.

> So currently the best I hope for is an ergonomic interface to prioritising auto-hints. That was the rational behind adding local %hints.

It sounds like you'd want to allow implementation overlapping (either by specialization or visibility control) while hiding all syntax noise. My experience tells that it's still better to distinguish between sealed and overridable implementations to prevent frustrating ambiguity in the distance. See also this discussion for the idea: https://rust-lang.github.io/rfcs/1210-impl-specialization.html#the-default-keyword

ohad:
No, I don't think it's orthogonal, and I'm not sure linking to a long discussion about how Rust is structured helps here.
I'd also hesitate to bring into the picture the terminology 'sealed' and 'overridable' implementations, since our current interface/implementation mechanism is extremely limited in this respect: the decision which implementations satisfy the constraints of an interface happens at the call site, which leads to very frustrating behaviour the moment types become slightly dependent.

The issue isn't having the error 'overlapping instances'. That's fine, and should happen when we have an ambiguity.
What I'm saying is that using the most specialised type to resolve this ambiguity is **not** a solution, and in fact can make matters worse: because notation suites lead through module imports, you can have 'silent' resolutions where a higher priority hint leaks through an import (necessarily, due to type-dependency). This leads to very confusing behaviour/errors, and will lead to situation that you just can't solve without heaps of boilerplate.

From my perspective, the solution isn't going to be more nuanced rules about how ambiguity gets resolved using subtle unification techniques. The solution is going to be to equip users with the language to tell the ambiguity resolution mechanism how to manage the hints so that they don't leak through abstraction barriers.

If this is too abstract, I can come up with more concrete examples, but it might take some time (deadlines....)

andylokandy:
I still think they are orthogonal because any of them can solve all problems here. I conclude the problems into these two:

1. Implementations/hints are easily get leaked into other modules where the module author does not intend to.
2. Multiple overlapping implementations/hints are intended to be imported into a module and willing the compiler to resolve the ambiguity following the intuition.

Even if we support `public noexport`, the `Show a => Interpolation a ` will still conflict with `Interpolation String` because they are intended to import into the module at the same time.

andylokandy:
> This is correct. This will not resolve situations where one has two different overloaded structures on the same type. Moreover, one can cook-up real situations where you'd want one choice, and other situations where you'd want the other one.

In most cases, I'd prefer to let it throw an ambiguity error and let the user explicitly choose one either by local hint or `%hide` or `public noexport`. My point is that if one of the hints is explicitly denoted as overridable by pragma, we choose the sealed one (seal is the default behavior) if both are viable. So long as you don't arbitrarily denote hints as overridable by `%defaulthint`, everything is fine and the de-ambiguous is strict, thus the bad situation you described will not happen.

The `the most specific one` only comes in in only one corner case: there is no viable sealed hint but multiple viable overridable hints. In this PR, Idris will randomly pick one, and I'd like it to pick the most specific one, that's it.

andylokandy:
I'm afraid that such sugar has to hardcode implementations into the compiler, thus disallowing users to define a new implementation of `Interpolation` in their library. On the contrary, if we provide some syntax (or the plain local hint) for users to specify which implementation to use for each type, we're enforcing the user to do the compiler's job ---- to figure out all occurring types in the expression, for gaining the ability of fine-grained control which the user may not desire. I believe this is a trade-off of fine-grained control and ergonomics.

Besides, even if we eventually decide to use hint specialization for interpolation, the user is still able to control the hint pool used for specialization, thus, the specialization can be disabled by using a sealed `%hint` local hint.

 And further, one more solution that just popped out from my mind: Enforce the overridable hint to specify the more general hint it prioritizes to, otherwise throws an ambiguity error, instead of the compiler computing the priority by some mystery algorithm/unifications. For example:

```idris
Interpolation String where
    interpolate x = x

%defaulthint[defaultInterpolation]
[prettyterpolation] Pretty a => Interpolation a where
    interpolate x = show (pretty x)

%defaulthint[]
[defaultInterpolation] Show a => Interpolation a where
    interpolate x = show x
```

ohad:
> I'm afraid that such sugar has to hardcode implementations into the compiler, thus disallowing users to define a new implementation of Interpolation in their library. 

This is not true. Users can define new implementations of interpolation, these can be exported to client code, and so on. But if someone wants the functionality of a library but doesn't want the set of notation it uses because they'd rather use a different interpolation, then they can do so too.

> On the contrary, if we provide some syntax (or the plain local hint) for users to specify which implementation to use for each type, we're enforcing the user to do the compiler's job ---- to figure out all occurring types in the expression, for gaining the ability of fine-grained control which the user may not desire.

This is not true. The compiler cannot figure it out by itself, because the situation is inherently ambiguous. Sometimes we will want to use `One`, other times we will want to use `Two`. If the compiler enforces one way or the other, even if it's based on some intricate type information, that will cause the users who need the other behaviour no end of frustration.

What you are suggesting seems to me to be an imperative mutation of the hierarchy of hints. This may potentially be very confusing, since users cannot inspect or see this kind of state at all. I'm proposing to keep the overloading resolution mechanism simple, but find a clean set of primitives to contextualise the current collection of notations.

ohad:
I'm confused why you think you cannot do that already.

The main issue that it will cause is that it will create an inherent ambiguity with `Interpolation String`.
I can see situations where users will want to use one, or the other. We need to give them a clean mechanism to express this choice.

andylokandy:
Let me clearify the question: what mechainism the user should use in the latest release to state that he want to use `Intepolation String` and use `Show a => Interpolation a` where a is not String, without listing all possible types that a can be.

ohad:
I'm not sure what you mean. Can you please give a full example?

andylokandy:
```idris
Intepolation String where
    intepolation x = x

-- Currently, this doesn't work because `Nat`, `Int`, `Bool` and so on don't have `Intepolation` implementations
-- But luckly, they all have `Show` implementations,
-- so, is there a way to make the following successfully complied, without having to give `Intepolation Nat`, `Intepolation Bool` and so on one by one?
str : String
str = "\{the Nat 1} \{the Int 1} \{the Double 1.2} \{True} \{"string"} \{[1, 2, 3]} \{the (Vect ? Int) [1, 2, 3]}"
```

ohad:
This spec doesn't make sense, because the situation is inherently ambiguous.

I can write a program that takes a `Show a` instance and interpolates it:
```idris
foo : Show a => a -> String
foo x = "The argument is \{a}"
```
and I might want the value interpolated using the show instance, because this is a metaprogram that produces well-formed idris code:
```idris
||| Assumes Pollack-consistency of `Show a`
emit : Show Type => (a : Type) -> Show a => a -> String
emit x = """
value : \{a}
value = \{x}
"""
```

If the type-checker resolves the inherent ambiguity one of or the other, the other kind of program cannot be written. So if we want to insist writing these kinds of programs, we'll need to find a solution that lets us write both.

ohad:
My solution for this would be to not export `Show a => Interpolate a` as a hint, but as a named hint. Then users can resolve the ambiguity locally:
```idris
[InterpolateWithShow] Show a => Interpolation a where
  interpolate x = show x

Splice : {a : Type} -> (Show a)  => a -> String
Splice {a} x =
    let %hint useThis : ?
        useThis = case a of {String => %search; _ => InterpolateWithShow {a}}
    in "The argument is \{x}"


Emit : Show Type => (a : Type) -> Show a => a -> String
Emit a x =
    let %hint useThis : ?
        useThis = InterpolateWithShow {a}
    in """
val : \{show a}
val = \{x}
"""

Show Type where
  show String = "String"
  show Integer = "Integer"
  show x = "Unknown, cause a type error please"

val1, val2 : String
val1 = Splice 1
val2 = Emit Integer 1
```

andylokandy:
I Applied @ohad's suggestion to https://github.com/idris-lang/Idris2/pull/2245 and I am closing this accordingly.

PR #2254, opened 2022-01-08
[ tests ] Add test case for #2243

dunhamsteve:
This PR adds a test case for #2243, which was a bug in the javascript code generation.  It uses Idris to compile a small program `EvenOdd.idr` to javascript and then runs an Idris program to check the generated javascript file for duplicate function names.


edwinb:
I'm now getting a stack overflow in this test on Mac - is anyone else seeing this? It's clearly okay on CI so I guess we must be set up differently. The stack overflow is when executing `Main` and appears to be in the primitive `__prim_js2idris_array` so possibly this isn't related to the thing being tested.

edwinb:
Ah, the solution appears to be to rewrite `__prim_js2idris_array` iteratively. I'm no JS expert but I'll test it properly and send a PR.

dunhamsteve:
Hmm, I'm not seeing a failure on Mac when run directly or in `make test` (node 16.13.1).

I wasn't expecting the test would be run in node - it doesn't pass in a `--cg node` to the Idris invocation with the `-x`, but it appears that is is indeed running it in node under `make test`.

Looking at `__prim_js2idris_array`, I'm surprised it doesn't blow the stack for me. If I concatenate two copies of `build/exec/app.js`, it does fail for me. Maybe there is something subtly different causing `build/exec/app.js` to be larger on your system or your node has a smaller stack for some reason. 

I think `__prim_js2idris_array` should be rewritten to be iterative, it looks problematic as-is. I can try to put together a fix and test case later today.

dunhamsteve:
Ignore my previous comment, I didn't see that you'd already fixed it.

Issue #1792, opened 2021-07-27
A language version field in package files


mattpolzin:
- [x] I have read [CONTRIBUTING.md](../../CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

I'd like to see an optional language version field in ipkg files. Perhaps `langversion`? The field should work like the depends entries (accepting either a constant version or a range that the language must fall within). 

## Motivation

Right now there is no good way that I know of for a package to indicate it requires a certain range of Idris 2 versions. A package might use language features only available after a certain version or it might only be useful prior to the language features of another Idris 2 version.

## The proposal

The proposal is just to add a single new optional field like the above-suggested `langversion` field to Package files. Then package managers can choose to use that information if desired and perhaps future work in the Idris 2 compiler can check that the version of itself meets the criteria in a package being `--install`ed. 

### Examples

The following example would indicate that the Idris 2 compiler itself at version 0.5.0 cannot be built with a language version less than 0.4.0. It's a bit problematic at this point to see into the future and know that upcoming versions of Idris 2 will be backwards compatible, so it also specifies a conservative upper bound. 
```
package idris2

...
version = 0.5.0
langversion >= 0.4.0 && < 0.6.0
...
```

andrevidela:
What do you think about using commits (and tags) in that field rather than the compiler's version numbers? In my experience I tend to work against a specific commit and build from there, specially when working with experimental features.

In terms of core functionality it should be the same since the tag and the version number should always be in sync.

This would also allow us to define some sort of "bootstrap history" and say " the compiler at commit X can be built with the compiler from commit Y which itself can be compiled with commit Z, but you can't compile X with Z"

I guess it would make ranges a bit more difficult to work with since commits have no guarantee to happen in chronological order but it doesn't feel too surprising to me that ranges would be disabled for commits and only kept for plain version numbers.

Maybe this is a bad idea because it leaks too much of git's functionality into the compiler!

kenaryn:
Hello,
I'm unsure with it could help you a bit but Julia use a built-in mechanic, that is a version number literal like the following `v"..."`, e.g. `if "v0.4" < VERSION < "v0.5-" # do something with 0.4 series end`.
In Julia, the reserved word VERSION is a CONSTANT, which in my case returns `v"1.6.2"`.

Here is the brief description in documentation: https://docs.julialang.org/en/v1/manual/strings/#man-version-number-literals
[...]

PR #2256, opened 2022-01-10
[ new ] optional language version field to ipkg files.

mattpolzin:
Closes https://github.com/idris-lang/Idris2/issues/1792.

Adds the optional `langversion` field. If used, you can specify `>`, `<`, `>=`, and `<=` (the same criteria as with dependency version constraints).

```
package mypackage
version = 0.0.1

langversion >= 0.5.1

depends = contrib >= 0.5.1
        , network
...
```

- [x] Add field.
- [x] Add error when Idris version is not supported.
- [x] Add tests.
- [x] Add documentation.


src/Idris/Package/Types.idr
```diff
@@ -85,6 +98,39 @@ inBounds mv bounds
...
+  export
+  inBounds : Version -> PkgVersionBounds -> Bool
+  inBounds (MkVersion (maj, min, patch) tag) bounds
+     = let v = MkPkgVersion (maj ::: [min, patch]) in
+       maybe True (\v' => if bounds.lowerInclusive
+                             then v >= v'
+                             else v > v' || (v == v' && tag /= Nothing)) bounds.lowerBound &&
+       maybe True (\v' => if bounds.upperInclusive
+                             then v < v' || (v == v' && tag == Nothing)
+                             else v < v' || (v == v' && tag /= Nothing)) bounds.upperBound
```
gallais, line 120:
```suggestion
                             else v < v') bounds.upperBound
```

otherwise `2.0-abc < 2.0` which seems to go against the logic that 

> a Version with a tag is always considered larger than one with the
> same major, minor, and patch numbers but no tag.

src/Idris/Package/Types.idr
```diff
@@ -173,49 +221,52 @@ export
...
-    , comment "script to run before cleaning"
-    , strField "preclean"    (snd <$> desc.preclean)
+    , comment     "script to run before cleaning"
+    , strField    "preclean"    (snd <$> desc.preclean)
 
-    , comment "script to run after cleaning"
-    , strField "postclean"   (snd <$> desc.postclean)
+    , comment     "script to run after cleaning"
+    , strField    "postclean"   (snd <$> desc.postclean)
     ]
```
gallais, line 270: Can we drop the extra space introduced by aligning labels?
This obscures diffs.

mattpolzin: Done!

PR #2262, opened 2022-01-14
[ totality ] Data.String.words should be total

stefan-hoeck:
This looks like an oversight: `Data.String.words` is provably total, as all white space is dropped after each invocation of `break`. While it would be nice to actually have a proper proof of totality, this wasn't done for other functions like `Data.List.split` either, so I just went for the sledgehammer here.

### DIFF
--- a/libs/base/Data/String.idr
+++ b/libs/base/Data/String.idr
@@ -2,6 +2,7 @@ module Data.String
 
 import Data.List
 import Data.List1
+import Data.SnocList
 
 %default total
 
@@ -49,13 +50,20 @@ fastUnlines = fastConcat . unlines'
 ||| Splits a character list into a list of whitespace separated character lists.
 |||
 ||| ```idris example
-||| words' (unpack " A B C  D E   ")
+||| words' (unpack " A B C  D E   ") [<] [<]
 ||| ```
-words' : List Char -> List (List Char)
-words' s = assert_total $ case dropWhile isSpace s of
-            [] => []
-            s' => let (w, s'') = break isSpace s'
-                  in w :: words' s''
+words' :  List Char
+       -> SnocList Char
+       -> SnocList (List Char)
+       -> List (List Char)
+words' (c :: cs) sc css =
+  if isSpace c
+     then case sc of
+       [<] => words' cs [<] css
+       _   => words' cs [<] (css :< (sc <>> Nil))
+     else words' cs (sc :< c) css
+words' [] [<] css = css <>> Nil
+words' [] sc  css = (css :< (sc <>> Nil)) <>> Nil
 
 ||| Splits a string into a list of whitespace separated strings.
 |||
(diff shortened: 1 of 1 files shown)


gallais (approved):

libs/base/Data/String.idr
```diff
@@ -51,9 +51,8 @@ fastUnlines = fastConcat . unlines'
 ||| ```idris example
 ||| words' (unpack " A B C  D E   ")
 ||| ```
-covering
 words' : List Char -> List (List Char)
-words' s = case dropWhile isSpace s of
+words' s = assert_total $ case dropWhile isSpace s of
             [] => []
             s' => let (w, s'') = break isSpace s'
                   in w :: words' s''
```
CodingCellist, line 58:
```suggestion
                  in w :: words' (assert_smaller s s'')
```

libs/base/Data/String.idr
```diff
@@ -51,9 +51,8 @@ fastUnlines = fastConcat . unlines'
 ||| ```idris example
 ||| words' (unpack " A B C  D E   ")
 ||| ```
-covering
 words' : List Char -> List (List Char)
-words' s = case dropWhile isSpace s of
+words' s = assert_total $ case dropWhile isSpace s of
```
CodingCellist, line 55:
```suggestion
words' s = case dropWhile isSpace s of
```

CodingCellist (changes requested):
Is there any benefit to using a slightly smaller sledgehammer?  : )

libs/base/Data/String.idr
stefan-hoeck: I thought about this too. However, strictly speaking, `s''` is not necessarily smaller than `s`. This is only true after removing the leading space(s) (left over from the call to `break`), which happens, when `dropWhile isSpace` is invoked. So, while it is true that `words'` is total, it is not true that `s''` is always smaller than `s`.

stefan-hoeck: So, I dropped the sledgehammer altogether and wrote a tail recursive implementation, which satisfies the totality checker without cheating.

dunhamsteve: In `let (w,s'') = break isSpace s'`, we have that either `s'' = (' ' :: cs)` or `s'' = []`. Could we just drop the first character of `s''` before recurring?

stefan-hoeck: Sure, but since this is already done when calling `dropWhile isSpace`, I didn't think it was worth the hassle. Anyway, I completely rewrote the implementation, as I think it's preferable to have a tail call optimized version anyway.

CodingCellist (approved):

libs/base/Data/String.idr
```diff
@@ -49,24 +50,29 @@ fastUnlines = fastConcat . unlines'
...
+       -> SnocList (List Char)
+       -> List (List Char)
+words' (c :: cs) sc css =
+  if isSpace c
+     then case sc of
+       [<] => words' cs [<] css
+       _   => words' cs [<] (css :< (sc <>> Nil))
+     else words' cs (sc :< c) css
+words' [] [<] css = css <>> Nil
+words' [] sc  css = (css :< (sc <>> Nil)) <>> Nil
```
gallais, line 66: Should this also case on whether `sc` is empty?

stefan-hoeck: I extracted a `wordsHelper` function to do what you described above.

PR #2264, opened 2022-01-16
[ new ] Add fixities' FC to the syntax info.

mattpolzin:
This small change retains origin information for the fixity declaration for an operator.

I think this could be generally useful for various things, but it is specifically useful for the unused warning import feature I am working on -- modules like prelude's `Prelude.Ops` (declares nothing but fixities) can be imported and indeed required for an operator to be fully defined but without knowing that the operator's fixity came from that module it is not possible to determine the module has been "used."


mattpolzin:
Real struggle fest with the tagging over here 😓

PR #2266, opened 2022-01-16
Update typedd.rst - Exercise 2 of 8.2.5

trevarj:
The skeleton code in exercise 2 of Chapter 8.2.5 no longer type-checks with idris2.

From ohad in the Discord:
"Idris2 treats unbound implicits in a more sophisticated way than idris1, and the n that appears in your code is constrained to be the same n as the n that appears in the type of myReverse.
You want a helper function that's more general, where the length of the accumulator changes with each call"

### DIFF
--- a/docs/source/typedd/typedd.rst
+++ b/docs/source/typedd/typedd.rst
@@ -149,6 +149,9 @@ In ``ReverseVec.idr``, add ``import Data.Nat`` for the ``Nat`` proofs.
 In ``Void.idr``, since functions must now be ``covering`` by default, add
 a ``partial`` annotation to ``nohead`` and its helper function ``getHead``.
 
+In Exercise 2 of 8.2.5, the definition of ``reverse'`` should be changed to
+``reverse' : Vect k a -> Vect m a -> Vect (k + m) a``
+
 Chapter 9
 ---------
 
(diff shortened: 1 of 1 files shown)


trevarj:
Original:
```
myReverse : Vect n a -> Vect n a
myReverse xs = reverse' [] xs where
                reverse' : Vect n a -> Vect m a -> Vect (n + m) a
                reverse' acc [] = ?reverseProof_nil acc
                reverse' acc (x :: xs) = ?reverseProof_xs (reverse' (x :: acc) xs)
```

Proposed change (compiles):
```
myReverse : Vect n a -> Vect n a
myReverse xs = reverse' [] xs where
                reverse' : Vect k a -> Vect m a -> Vect (k + m) a
                reverse' acc [] = ?reverseProof_nil acc
                reverse' acc (x :: xs) = ?reverseProof_xs (reverse' (x :: acc) xs)
```

mattpolzin:
Looks like a great note to add to me! Might I suggest adding a bit of Ohad's description to your change as well to explain what's going on? As little as "because the `n` in `reverse'` is otherwise bound to the same value as the `n` in the signature of `reverse`".

trevarj:
@mattpolzin Added your description in another commit! Thanks!

mattpolzin (approved):
Thanks!

PR #2267, opened 2022-01-16
Make prim_js2idris_array iterative

edwinb:
I was getting a stack overflow on the tailrec002 test, caused by this primitive being recursive. I've made it iterative. I'm no JS expert but I've tested its behaviour against the old version.

### DIFF
--- a/support/js/support.js
+++ b/support/js/support.js
@@ -1,11 +1,12 @@
 class IdrisError extends Error { }
 
 function __prim_js2idris_array(x){
-  if(x.length === 0){
-    return {h:0}
-  } else {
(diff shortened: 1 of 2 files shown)


stefan-hoeck:
Thanks a lot for this. I just happen to have stumbled on a stack overflow from this yesterday, but didn't get around to fixing it yet :-). The JavaScript looks fine to me.

edwinb:
If it looks fine to you then I'll just merge it. Thanks :).

dunhamsteve:
Does there need to be a `let` in front of the `i` in the for loop?

PR #2268, opened 2022-01-17
[ breaking, totality ] fGetLine should be covering

stefan-hoeck:
For the same reason as `fRead` is `covering`, `fGetLine` and `fSeekLine` should be `covering` instead of `total`. For instance, the following program will unavoidably and quickly fill a computer's memory:

```idris
import System.File

main : IO ()
main = do
  Right f <- openFile "/dev/zero" Read | Left e => pure ()
  Right s <- fGetLine f | Left e => pure ()
  closeFile f
  putStrLn s
```

### DIFF
--- a/libs/base/System/File/ReadWrite.idr
+++ b/libs/base/System/File/ReadWrite.idr
@@ -41,6 +41,7 @@ prim__removeFile : String -> PrimIO Int
 |||
 ||| @ h the file handle to seek through
 export
+covering
 fSeekLine : HasIO io => (h : File) -> io (Either FileError ())
 fSeekLine (FHandle f)
     = do res <- primIO (prim__seekLine f)
@@ -53,6 +54,7 @@ fSeekLine (FHandle f)
 |||
 ||| @ h the file handle to get the line from
 export
+covering
 fGetLine : HasIO io => (h : File) -> io (Either FileError String)
 fGetLine (FHandle f)
     = do res <- primIO (prim__readLine f)


mattpolzin:
How does marking these as covering not trigger `readLinesOnto` to be non-total? Seems like a bug with coverage checking that it doesn't result in a compiler error, though perhaps marking something as covering does not actually mean "this is not total" but rather "this need not be total"? So if the compiler still thinks it is total, it propagates as total?

buzden:
I'm sure `fGetLine` is `total`. This is a function that produces a value of type `io (Either FileError String)` being given the values of type `Has io` and `File` and it does it always, thus totally.

@stefan-hoeck The problem you are pointing at happens inside the particular real-world `IO` interpreter. Thus, it should be solved some other way.

Plus, indeed, as @mattpolzin stated, this `covering` marks do not help because totality checker rechecks any non-total function in any total use site and will ensure that `fGetLine` is total again, despite the `covering` mark. I don't think this is a bug in the totality checker, it just repeats some checks over and over. And, by the way, this was (and probably is) used in standard libraries when functions without `total` marks were used in total functions -- I don't know whether was it intentional or not, but the current totality checker mechanism allowed it safely.

stefan-hoeck:
> How does marking these as covering not trigger `readLinesOnto` to be non-total? Seems like a bug with coverage checking that it doesn't result in a compiler error, though perhaps marking something as covering does not actually mean "this is not total" but rather "this need not be total"? So if the compiler still thinks it is total, it propagates as total?

Maybe we need to changed the totality of the corresponding prim_ function as well?

stefan-hoeck:
> I'm sure `fGetLine` is `total`. This is a function that produces a value of type `io (Either FileError String)` being given the values of type `Has io` and `File` and it does it always, thus totally.

I get your meaning here, but if we interpret totality for `IO` actions this way, shouldn't functions like `readFile` also be considered total, as they are guaranteed to return a value of the given type, which will only upon being executed potentially run forever?

Consider also the following case:

```idris
forever : a -> LazyList a
forever v = v :: forever v
```

According to your reasoning, `forever` should be total, since it will always return a value of the given type. However, Idris treats this as a partial function (rightfully so, in my opinion), as consuming such a value will loop forever.

Personally, I think totality annotations should tell us something about a program's behavior at runtime. Otherwise, function `main` can almost always be considered to be total, although we all know that we are not interested in the value of type `IO ()` in the end but in the program that runs when `main` is being executed.

stefan-hoeck:
Just one more note on this: Under the hood, `IO a` is a wrapper around `(1 _ : %World)  -> IORes a`. Surely, in case of `fGetLine`, this wrapped function must be considered to be partial, as in case of some infinite data streams like `/dev/zero` it will never yield a value of type `IORes a`. Should it then be considered to be a total operation to wrap a partial function in a record wrapper, thus hiding the underlying partiality from the totality checker? I think not.

buzden:
> Under the hood, `IO a` is a wrapper around `(1 _ : %World) -> IORes a`. Surely, in case of `fGetLine`, this wrapped function must be considered to be partial, as in case of some infinite data streams like `/dev/zero` it will never yield a value of type `IORes a`.

I agree with you, I was wrong. *Returning* partial function (doesn't matter wrapped or not) is the same as *being* a patrial function. Probably, we should say even "having a partial function in a covariant position" instead of "returning".

Then it all means that according `prim__` functions are not total. And marking them so seems to solve problem of rechecking `fGetLine` to be total because all ends in a foreign partial primitive.

gallais:
> since it will always return a value of the given type. 

In both this case and the `fGetLine` cases it's not actually true that the function returns a value of the
given type: Idris, unlike Haskell, distinguishes data and codata. Data is necessarily finite to be able to
guarantee that structural induction eventually terminates.

I agree that `readFile` & other functions assuming a file is finite should only be covering. If you want
a guaranteed total variant, you'll have to return chunks in succession using something à la

```
data Hose : Type where
  Nil : Hose
  (::) : String -> Inf (IO Hose) -> String
```

(not sure whether the `Inf` should be inside or outside the `IO` tbh).

PR #2271, opened 2022-01-18
Add a few missing NodeJS FFI functions to `System`

mattpolzin:
Specifically, `setEnv`, `unsetEnv`, and `system`.


libs/base/System.idr
```diff
@@ -117,6 +125,9 @@ getEnvironment = getAllPairs 0 []
 
 ||| Add the specified variable with the given value string to the environment,
 ||| optionally overwriting any existing environment variable with the same name.
+||| Returns True whether the value is set, overwritten, or not overwritten because
+||| overwrite was 0. Returns False if a system error occurred. You can `getErrno`
```
mattpolzin, line 129:
```suggestion
||| overwrite was False. Returns False if a system error occurred. You can `getErrno`
```

Z-snails:
Partial fix for #2207.

PR #2273, opened 2022-01-19
Add Fin modular arithmetic Num instance

madman-bob:
Adds `Num` and `Neg` instances for `Fin (S n)`, implementing modular arithmetic, with test.

This seems to me to be the natural `Num` and `Neg` implementations for `Fin`, so they are unnamed.

I've not optimized these implementations, but it should be simple enough to do once we represent `Fin`s as `Integer`s under-the-hood. These operations are just the equivalent operations on `Integer`, composed with (Euclidean) `mod n`.


octeep:
Should there be a division function as well, given a proof that the divisor and the modulus are coprime?

PR #2274, opened 2022-01-19
make readBufferData return the number of bytes that have been read

octeep:
This commits make `readBufferData` return the number of bytes that have been read. Without it, you cannot read the content of a file by chunks.

An example of how returning the number of read bytes can be used:
```haskell
||| Construct a `Stream` reading from a File
export
fromFile : HasIO m => File -> Stream (Of Bits8) m (Either FileError ())
fromFile file = (newBuffer 1024) >>= loop where
  loop : Maybe Buffer -> Stream (Of Bits8) m (Either FileError ())
  loop (Just buffer) = do
    False <- liftIO $ fEOF file
    | True => pure (Right ())
    Right cap <- liftIO (rawSize buffer >>= Test.readBufferData file buffer 0)
    | Left err => pure (Left err)
    data' <- traverse (getBits8 buffer) [0..(cap-1)]
    fromList_ data' *> loop (Just buffer)
```

### DIFF
--- a/libs/base/System/File/Buffer.idr
+++ b/libs/base/System/File/Buffer.idr
@@ -30,11 +30,11 @@ export
 readBufferData : HasIO io => (fh : File) -> (buf : Buffer) ->
                  (offset : Int) ->
                  (maxbytes : Int) ->
-                 io (Either FileError ())
+                 io (Either FileError Int)
 readBufferData (FHandle h) buf offset max
     = do read <- primIO (prim__readBufferData h buf offset max)
          if read >= 0
-            then pure (Right ())
+            then pure (Right read)
             else pure (Left FileReadError)
 
 ||| Write the data from the buffer to the given `File`.
--- a/tests/refc/buffer/TestBuffer.idr
+++ b/tests/refc/buffer/TestBuffer.idr
@@ -38,7 +38,7 @@ main = do
         | Nothing => pure ()
     Right f <- openFile "testRead.buf" Read
         | Left err => put $ pure err
-    Right () <- readBufferData f readBuf 0 8
+    Right ok <- readBufferData f readBuf 0 8
         | Left err => put $ pure err
     put $ bufferData readBuf
 


ska80:
> This commits make `readBufferData` return the number of bytes that have been read

Can we do the same for `writeBufferData`?

octeep:
> > This commits make `readBufferData` return the number of bytes that have been read
> 
> Can we do the same for `writeBufferData`?



I've glanced at the code and C documentation just now. The OpenBSD manual and Linux states that fwrite would return the number of objects that have been written, which in our case would be the `max` argument passed to `writeBufferData`.

In the case where the number of objects returned is less than the number of objects specified, it would mean than an error has occurred while writing to the file.

So if we are going to return the number of bytes that have been written, I think the type signature would probably be something like:
```haskell
writeBufferData : HasIO io => (fh : File) -> (buf : Buffer) ->
                  (offset : Int) ->
                  (maxbytes : Int) ->
                  io (Either (FileError, Int) ())
```
Since the returned value should only be different from `maxbytes` unless an error has occurred. 

This also means that naming the argument "maxbytes" in `writeBufferData` is a bit of a misnomer, since we always expect "maxbytes" number of bytes to be written. It should probably be renamed to something else.

I also suspects that if you specify `maxbytes` to something greater than `buffer size - offset`, it will probably cause some form of segfault.

That said, all these are just what I've gathered while browsing documentation on my phone. I'm about to sleep right now so I can't perform any tests on how the functions actually behaves, I'd need to do some tests tomorrow.

octeep:
So I've dug up the C89 and C99 specification, which MacOS, Linux, OpenBSD conforms to, on how `fwrite` should behaves:
```
4.9.8.2 The fwrite function
Synopsis
         #include <stdio.h>
         size_t fwrite(const void *ptr, size_t size, size_t nmemb,
                  FILE *stream);
Returns
The fwrite function returns the number of members successfully written, which will be less than nmemb only if a write error is encountered.
```
This means that the returned value would only be different from `maxbytes` when a write error has occurred. Then the new type signature for `writeBufferData` would probably be something like:
```
writeBufferData : HasIO io => (fh : File) -> (buf : Buffer) ->
                  (offset : Int) ->
                  (maxbytes : Int) ->
                  io (Either (FileError, Int) ())
```

mattpolzin:
@octeep do you mind opening another PR to add this to the CHANGELOG? It's possible this breaks some code and that will make it much easier for folks to track down why.

PR #2276, opened 2022-01-20
make writeBufferData return the number of bytes that have been written

octeep:
Continued from #2274


src/Libraries/System/Directory/Tree.idr
```diff
@@ -200,9 +200,17 @@ copyDir src target = runEitherT $ do
     MkEitherT $ createDir $ show target
     copyDirContents !(liftIO $ explore src) target
   where
+    -- Glue for the new and old definition of copyFile
+    partial
+    extract_error : (a : Type) -> a -> FileError
+    extract_error (Pair FileError _) (x, _) = x
+    extract_error FileError x = x
```
gallais, line 207: This looks pretty dubious to me. I'd rather we use the usual approach of

1. put the new version in a `Libraries/` module
2. `import Libraries.Whatever as Lib`
3. use `Lib.copyFile'`
4. remove the `Lib` nonsense after the next release

gallais (changes requested):
Same remark for both type-case based solutions.

src/Libraries/System/Directory/Tree.idr
octeep: removed it :D

mattpolzin:
Let's be sure to mention this in the CHANGELOG as a breaking change.

octeep:
> Let's be sure to mention this in the CHANGELOG as a breaking change. Just something easy to forget that I've been forgetting about a lot lately!

I've added mention to the changes to both read and write buffer data in the changelog.

octeep:
> > Let's be sure to mention this in the CHANGELOG as a breaking change. Just something easy to forget that I've been forgetting about a lot lately!
> 
> I've added mention to the changes to both read and write buffer data in the changelog.

I've splitted the read buffer data change in CHANGELOG into #2283, which also updates unit test to check for `readBufferData` return value. Forgot to do that in my last PR. Sorry for that.

PR #2280, opened 2022-01-21
[ fix ] add let in JS array conversion

stefan-hoeck:
This is a follow-up of #2267 . As noted correctly by @dunhamsteve, there should be a `let` in the `for` statement. While this is not strictly necessary, it can lead to nasty scoping bugs as described in [this stackoverflow reply](https://stackoverflow.com/questions/5717126/var-or-no-var-in-javascripts-for-in-loop).

Sorry, for not spotting this myself when I gave Edwin my OK on this.

### DIFF
--- a/support/js/support.js
+++ b/support/js/support.js
@@ -3,7 +3,7 @@ class IdrisError extends Error { }
 function __prim_js2idris_array(x){
   let acc = { h:0 };
 
-  for (i = x.length-1; i>=0; i--) {
+  for (let i = x.length-1; i>=0; i--) {
       acc = { a1:x[i], a2:acc };
   }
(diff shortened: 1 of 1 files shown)


gallais:
This is wild.

dunhamsteve:
Even more wild - tweaking the example from the stack overflow answer - this counts as a declaration for the `i` in the for loop (scoping it to the body of the function). 

```js
function f (){
    for (i=0; i<5; i++);
    var i;
}
var i = 2;
f ();
alert (i);
```

gallais:
Please stahp :fearful:

andrevidela:
@gallais 
image

Issue #2279, opened 2022-01-21
Racket unsafe-fx functions missing in bootstrap/idris2_app/idris2.rkt


jfoutz:
If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!
docs have updated location - https://docs.racket-lang.org/reference/unsafe.html

# Steps to Reproduce
compile 
```
module Main

main : IO ()
main = putStrLn "Hello world"
``` 
with racket 8.3

I would suggest, add `(require racket/unsafe/ops)` to the racket prolog for 8.3 right about here - https://github.com/idris-lang/Idris2/blob/877025d5247dadd45b14aec8a9c751e93db566c5/src/Compiler/Scheme/Racket.idr#L42
but I don't see an obvious pattern for checking the require against a given racket version.

```
host % racket build/exec/hello_app/hello.rkt 
build/exec/hello_app/hello.rkt:52:20: unsafe-fx+: unbound identifier
  in: unsafe-fx+
  location...:
   build/exec/hello_app/hello.rkt:52:20
host idris2 % vi build/exec/hello_app/hello.rkt    #<- add (require racket/unsafe/ops)
host idris2 % racket build/exec/hello_app/hello.rkt
Hello world
```
# Expected Behavior
Hello world

# Observed Behavior
idris2 hello.idr -o hello         
build/exec/hello_app/hello.rkt:52:20: unsafe-fx+: unbound identifier
  in: unsafe-fx+
  location...:
   build/exec/hello_app/hello.rkt:52:20

jfoutz:
I might be a fool. it's there in the source. I think this might only apply to the bootstrap compiler.

jfoutz:
I think this will fix it, but it's possibly a very silly thing to do. Mainly because it'll break bootstrapping with 8.2
https://github.com/idris-lang/Idris2/pull/2281

anyway, that change works for me. I'm not in a position to easily compare with 8.2 behavior 

I realize there are no additional tests. I'd be happy to add them with some vague guidance. It's not super obvious how to, like, validate without actually running against 8.3

jfoutz:
https://github.com/idris-lang/Idris2/pull/2286
probably a better solution, use the generated bootstrap from 8.3

benhormann:
Apologies, the cause is my change. Try:
```
[...]

PR #2281, opened 2022-01-21
fix 2279

jfoutz:
builds and self hosts locally, with racket 8.3, but is dangerous for 2 reasons.
I'm not really set up to compare behavior against 8.2 but I _believe_ this will break bootstrapping with 8.2. 

There are failing tests. I forked from main rather than the release, so I don't know if that's expected or a side effect of this change. (although quite a few look totally unrelated to floating point numbers, those that do reference floats are scary).

but I don't think this
```
index 4482a376..e69de29b 100644
--- a/idris2/basic068/expected
+++ b/idris2/basic068/output
@@ -1,2 +0,0 @@
[B [A 5]]
{B {A 6}}
Golden value differs from actual value.
```
could be related.

Also, it sure seems like `bootstrap/idris2_app/idris2.rkt` is auto generated, so perhaps this is a pointless pr.


jfoutz:
Since the tests passed, this seems pretty safe. This just corrects the path for the float functions in racket 8.3. I do not know, but I suspect the ubuntu path was disabled, because it couldn't bootstrap.And (I'm still guessing here) ubuntu has the latest racket. Without this change (or something like it), you can't bootstrap with the current release of racket.

attila-lendvai:
whatever needs to be fixed, it needs to be done in the compiler.

the files you have changed are compiler generated files comitted into the git repo as _bootstrap shortcuts_, i.e. to speed up bootstrapping the Idris compiler.

please open an issue that describes the problem, preferably with some copy-pasted output in its description, and/or describing what needs to be changed in the output, and this PR should probably be closed (once the opened issue has been cross-referenced in the comments).

attila-lendvai:
scratch that, i see that you have already created an issue for this: https://github.com/idris-lang/Idris2/issues/2279

jfoutz:
Yeah, you're right. https://github.com/idris-lang/Idris2/pull/2286

PR #2283, opened 2022-01-21
update refc/buffer to check how many bytes have been read

octeep:
Just realized I missed this in #2274. This PR ensures that `readBufferData` returns the correct result.


attila-lendvai:
looks good. a changelog entry and an extra testcase.

PR #2286, opened 2022-01-23
fix: bring bootstrap to racket 8.3

jfoutz:
The compiler was updated here - https://github.com/idris-lang/Idris2/blob/35a84e505d66a11b84845e31f869456cade36848/src/Compiler/Scheme/Racket.idr#L48

but the compiler output for bootstrap wasn't updated.

better approach than
https://github.com/idris-lang/Idris2/pull/2281


jfoutz:
Actually, I'm not 100% sure I did this correctly.

Issue #2288, opened 2022-01-25
[ performance ] (>>=) and (<*>) for List have quadratic runtime complexity.


stefan-hoeck:
# Steps to Reproduce

Build and run the following program:

```idris
import Data.SnocList

replicateTR : Nat -> a -> List a
replicateTR n v = go n Nil
  where go : Nat -> List a -> List a
        go 0     xs = xs
        go (S k) xs = go k (v :: xs)

main : IO ()
main = printLn . length $ [1..5000] >>= (\n => replicateTR n n)
```

# Expected Behavior

Runs in a few seconds on reasonably modern hardware.

# Observed Behavior

Still not done while I write this issue... But I'll submit a PR with a fix soonish.

gallais:
Fixed by #2289

PR #2289, opened 2022-01-25
[ fix ] issue #2288

stefan-hoeck:
New implementation of `(>>=)` and `(<*>)` for `List` with linear runtime complexity.

### DIFF
--- a/libs/prelude/Prelude/Types.idr
+++ b/libs/prelude/Prelude/Types.idr
@@ -455,10 +455,17 @@ Foldable List where
 
   foldMap f = foldl (\acc, elem => acc <+> f elem) neutral
 
+-- tail recursive O(n) implementation of `(>>=)` for `List`
+listBind : List a -> (a -> List b) -> List b
+listBind as f = go Nil as
+  where go : List b -> List a -> List b
+        go xs []        = reverse xs
+        go xs (y :: ys) = go (reverseOnto xs (f y)) ys
+
 public export
 Applicative List where
   pure x = [x]
-  fs <*> vs = concatMap (\f => map f vs) fs
+  fs <*> vs = listBind fs (\f => map f vs)
 
 public export
 Alternative List where
@@ -467,7 +474,7 @@ Alternative List where
 
 public export
 Monad List where
-  m >>= f = concatMap f m
+  (>>=) = listBind
 
 public export
 Traversable List where
--- a/libs/prelude/Prelude/Types.idr
+++ b/libs/prelude/Prelude/Types.idr
@@ -456,11 +456,18 @@ Foldable List where
   foldMap f = foldl (\acc, elem => acc <+> f elem) neutral
 
 -- tail recursive O(n) implementation of `(>>=)` for `List`
+public export
 listBind : List a -> (a -> List b) -> List b
 listBind as f = go Nil as
   where go : List b -> List a -> List b
         go xs []        = reverse xs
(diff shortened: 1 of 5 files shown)


libs/prelude/Prelude/Types.idr
```diff
@@ -455,10 +455,17 @@ Foldable List where
 
   foldMap f = foldl (\acc, elem => acc <+> f elem) neutral
 
+-- tail recursive O(n) implementation of `(>>=)` for `List`
+listBind : List a -> (a -> List b) -> List b
+listBind as f = go Nil as
+  where go : List b -> List a -> List b
+        go xs []        = reverse xs
+        go xs (y :: ys) = go (reverseOnto xs (f y)) ys
+
```
gallais, line 464: Should we have  a `%transform` rule turning `concatMap` into `listBind`?

stefan-hoeck: I like the idea, but I'm not very fluent with `%transform` rules. Can we specialize such a rule for a constrained function (in case of `concatMap` this would require specializing both `m` and `t` to `List`, yes?). Also, if we used such a transform rule for `foldMap` instead, would this then automatically also apply to `concatMap` and `concat`? Or would all three require their own transform rule?

gallais: Yes we can cf.

> libs/prelude/Prelude/Types.idr:%transform "fastConcat" concat {t = List} {a = String} = fastConcat

`concatMap` does seem to be defined in terms of `foldMap` so adding that more fundamental `%transform` rule should indeed be enough.

stefan-hoeck: OK. Thanks for the clarifications. I'm going to give this a try and add some additional test cases to make sure the transform rules apply for all three functions. But this will have to wait till tonight.

stefan-hoeck: So, I need some help with this. I added the transform rules to `Prelude.Types` (a single rule for `foldMap` only seems to be not enough). But unless I also add them to the source file in the `prelude/bind001` test (which I did in the latest commit), they don't seem to apply.

gallais: I wouldn't expect the `concatMap`  and `concat` ones to be needed as their own bodies are defined in terms of
`foldMap`  and so would be optimised just fine.

Is the fact the rule doesn't fire for `foldMap` due to the fact that the `Foldable` is public export and things get inlined?
Is it possible to attach a `%noinline` to that definition of `foldMap`?

stefan-hoeck: I can't annotate the implementation of `foldMap` of `List` with `%noinline`: The parser won't accept it. Also, removing any of the transform rules from the test source file (test/prelude/bind001/bind.idr) will result in the corresponding function not being transformed, so at least there, all of them are needed. The rule for `foldMap` alone does not seem to affect `concatMap` or `concat`.

I also tried annotating the implementation of `Foldable List` with `%noinline` but to no avail.

gallais: Ok, so my suggestion is to
1. abandon the `%transform` proposal & merge this PR
2. open a bug report about the strange behaviour we are seeing

No need to tie this particular contribution to the bug (?) being resolved.

stefan-hoeck: I ditched the transform rules and will open an issue later on.

PR #2290, opened 2022-01-25
[ fix ] Fix Makefile targets for `linear` and `papers` libraries

ska80:


gallais:
It's on purpose that the `papers/` library is not installed with the other ones as it's not meant to be
reusable content but rather propaganda / complex constructions taken from papers.

ska80:
> It's on purpose that the `papers/` library is not installed with the other ones as it's not meant to be reusable content but rather propaganda / complex constructions taken from papers.

I removed `papers` library from `install` targets.  This PR still has other fixes.

Makefile
```diff
@@ -49,7 +49,9 @@ TEST_PREFIX ?= ${IDRIS2_CURDIR}/build/env
 IDRIS2_BOOT_PREFIX := ${IDRIS2_CURDIR}/bootstrap-build
 
 # These are the library path in the build dir to be used during build
-export IDRIS2_BOOT_PATH := "${IDRIS2_CURDIR}/libs/prelude/build/ttc${SEP}${IDRIS2_CURDIR}/libs/base/build/ttc${SEP}${IDRIS2_CURDIR}/libs/contrib/build/ttc${SEP}${IDRIS2_CURDIR}/libs/network/build/ttc${SEP}${IDRIS2_CURDIR}/libs/test/build/ttc${SEP}${IDRIS2_CURDIR}/libs/linear/build/ttc"
+IDRIS2_LIBRARIES = prelude base contrib network test linear
+$(foreach library,$(IDRIS2_LIBRARIES),$(eval IDRIS2_BOOT_PATH := $(IDRIS2_BOOT_PATH)$(IDRIS2_CURDIR)/libs/$(library)/build/ttc$(SEP)))
```
ska80, line 53: Hmm... This works on my macOS and Linux (Ubuntu 21.10) machines but fails when building on CI.

attila-lendvai:
FTR, before merging this please consider https://github.com/idris-lang/Idris2/pull/1990 because they will clash with each other.

also, note to self: this possibly has stuff that should/could be incorporated there.

CodingCellist:
With the closure of #1990, what does the status of this become? Should it still be considered for merging (possibly with some changes), or should it be closed?

Makefile
```diff
@@ -49,7 +49,11 @@ TEST_PREFIX ?= ${IDRIS2_CURDIR}/build/env
 IDRIS2_BOOT_PREFIX := ${IDRIS2_CURDIR}/bootstrap-build
 
 # These are the library path in the build dir to be used during build
-export IDRIS2_BOOT_PATH := "${IDRIS2_CURDIR}/libs/prelude/build/ttc${SEP}${IDRIS2_CURDIR}/libs/base/build/ttc${SEP}${IDRIS2_CURDIR}/libs/linear/build/ttc${SEP}${IDRIS2_CURDIR}/libs/network/build/ttc${SEP}${IDRIS2_CURDIR}/libs/contrib/build/ttc${SEP}${IDRIS2_CURDIR}/libs/test/build/ttc"
+IDRIS2_LIBRARIES = prelude base linear network contrib test
```
gallais, line 52: I think this approach is now actively going against #2677

ska80: Both lines have the same set of libraries.  Or am I missing something?

CodingCellist: As long as it's not installing `papers` it's fine, no? I recently forgot to install `linear` as part of a `make install-libs` and it caused a bunch of problems...
CI seems happy as well?  : )

ska80: Could someone please, point to a line where `papers` gets installed?  I went through the changes one more time, and I just can't see it.

The only change regarding `papers` is building (but not installing) documentation: https://github.com/ska80/Idris2-SH/blob/9ebfa2e0509b984d673b3d2c4eb736a4ac2dfff1/Makefile#L109

If this PR is pointless, let's close it :)

CodingCellist: @ska80, afaict `papers` never gets installed as part of this PR, which is why I'm mostly happy to merge it  ^^

The only concern I have is whether `papers` should be removed from line 109, since the docs are not installed as part of the `install-libdocs` target; they're built but never used. (Alternatively: should they be added to the `install-libdocs` target? My gut instinct is "No", with the reasoning being that if you want to use the `papers` library, you have to do everything yourself, but that might not be the best approach...)

ska80: @CodingCellist Ok, I have removed the line that builds docs for `papers`.  Thanks for reviewing!

CodingCellist:
Right, with that I think this is good to go. There's a target for `linear`, which is something we want to let people experiment with, and if people want to be (even more) on the bleeding edge, they can install `papers` manually.

Thanks for all your efforts ska80!  : )

PR #2291, opened 2022-01-26
[ fix ] expose Data.Seq.Internal

michaelmesser:
I get `Module Data.Seq.Internal not found` with `import Data.Seq.Unsized` if I do not have this change.


gallais:
I guess we have no way of hiding internal modules?! :cry:

PR #2293, opened 2022-01-27
[ linear ] add utilities for linear datastructures

andrevidela:
I've renamed `Pow` to `Copies` for a more descriptive name, you can now write 
``n `Copies\` v`` and read it as "n copies of v"

I've also added `LNat` with addition and multiplication, `power` was too hard to implement

`Copies` also allows to implement `map`, `foldl` and `replicate` for `LVect`.

The API of functions using `Copies` is trying to emulate the API of non-linera functions so that 

```
map f [a,b,c]
```
does what you expect and the copies of `f` are built using proof-search.


libs/linear/Data/Linear/LList.idr
```diff
@@ -0,0 +1,17 @@
...
+  Nil : LList a
+  (::) : a -@ LList a -@ LList a
+
+%name LList xs, ys, zs, ws
+
+export
+length : (c : Consumable a) => LList a -@ LNat
+length [] = Zero
+length (x :: xs) = let () = consume x in
+                    Succ (Linear.LList.length @{c} xs)
```
gallais, line 17: Do you really need to explicitly pass the implementation of `Consumable` in the recursive call?
```suggestion
length (x :: xs) = consume x `seq` Succ (LList.length xs)
```

libs/linear/Data/Linear/LNat.idr
```diff
@@ -0,0 +1,39 @@
...
+||| because we canot call `S` with an argument that is expected to be used exactly once
+public export
+0 toNat : LNat -@ Nat
+toNat Zero = Z
+toNat (Succ n) = S (toNat n)
+
+export
+Consumable LNat where
+  consume Zero = ()
+  consume (Succ n) = let () = consume n in ()
```
gallais, line 22:
```suggestion
  consume (Succ n) = consume n
```

libs/linear/linear.ipkg
```diff
@@ -7,4 +7,7 @@ modules = Data.Linear.LEither,
           Data.Linear.LMaybe,
           Data.Linear.LVect,
           Data.Linear.Notation,
-          Data.Linear.Pow
+          Data.Linear.Copies,
+          Data.Linear.LNat,
+          Data.Linear.LList
```
gallais, line 12: We usually try to keep these sorted alphabetically.

libs/linear/Data/Linear/LVect.idr
```diff
@@ -48,3 +53,35 @@ export
...
+map f {fns=[]} [] = []
+map f {fns=(f :: fs)} (x :: xs) = f x :: map f {fns=fs} xs
+
+export
+length : Consumable a => LVect n a -@ LNat
+length [] = Zero
+length (x :: xs) = let () = consume x in Succ (length xs)
+
+export
+foldl : {0 f : acc -@ elem -@ acc} -> n `Copies` f -@ acc -@ (LVect n elem) -@ acc
```
gallais, line 69: The convention is a bit inconsistent between `map f {fns}` and `foldl {f} fns`.
Can we rely on auto search for these or should we uniformly take the `fns` explicitly?

andrevidela: Oh right, auto-search works fine for both, we should use `foldl f {fns}`, will fix

gallais:
I'd say the obvious example is:

```idris

prefix 5 !*
||| Prefix notation for the linear unrestricted modality
public export
record (!*) (a : Type) where
  constructor MkBang
  unrestricted : a
```

which we should probably add to `Data.Linear.Notation` anyway.

andrevidela:
Ah good point let me add it

andrevidela:
All done! I've used `dup` for duplicate, which is slightly awkward because it overlaps with `Prelude.dup`, maybe there is a better name for it that doesn't clash?

I was tempted by adding a `Bifunctor` interface but the only type that could implement it is `LPair` so I've just implemented it as a function on `LPair`

gallais:
I suppose we could use `Duplicable / duplicate` like we have `Consumable / consume`.
Also: should `duplicate` have type ``(1 x : a) -> 2 `Copies` x`` rather than `a -@ LPair a a`?

andrevidela:
Oh right that's definitely a more precise type, `Duplicable/duplicate` sounds good

andrevidela:
all done, I had to move the interfaces away from notation because `Copies` was importing notation already and `duplicate` needs to import `Copies`.

The applicative-ish instances are actually instrumental to this implementation because matching on `Copies` is really buggy with the current linearity checking in patterns

```
lol : 2 `Copies` "hello" -> String
lol v = let [v1, v2] = v in v1 ++ v2
```
fails because `v1` unifies with `v2` but if we write `[v1, v1]` linearity checking fails because `v1` is used twice.

This is actually an old known bug, it just never really showed up. you can trigger it with any two linear variables that unify

gallais:
I don't know much about comonoids but is it necessarily the case that the result of
a comult of a comonoid is the same value twice? I think it makes sense to have that
extra restriction for a duplication operation but for a comonoid in its full generality?

After a bit of googling, I found a good example in some of [Heunen's lecture notes](http://www.inf.ed.ac.uk/teaching/courses/cqi/6-monoids.pdf):

> In Rel, any group G forms a comonoid with comultiplication `g ∼ (h, h⁻¹g)` for all g, h ∈ G, and counit `1 ∼ •`. 

I guess here in Idris we have the extra restriction that we're in `Set` & total linear
functions so it may be true that `comult` happens to be duplication. :thinking:

Still if `comult : a -@ LPair a a` then it makes more sense to have `Comonoid` as a
separate interface rather than just an alias for `(Consumable a, Duplicable a)`.

andrevidela:
Oh I didn't know that indeed,
I think we would still have to write it a `LPair a a` because we can implement a group interface for finite fields for example the inverse of 1 mod 3 is 2 with respect to addition so the comult of `1` is `1 # 2`

will fix later today, and will rebase too 😄

andrevidela:
All done, in addition to changing `Comonoid`, I've added `%assert total` where it was missing, and I've added an alias import `Data.Linear`, it looks like there wasn't anything using this namespace, so it seemed ideal to export `Notation` `Interface` and `Copies` from it.

libs/papers/Data/Linear/Inverse.idr
```diff
@@ -8,7 +8,9 @@ import Data.INTEGER
 import Data.Linear.LEither
 import Data.Linear.LMaybe
 import Data.Linear.LVect
-import Data.Linear.Notation
+import Data.Linear
+
+import Data.Linear
```
gallais, line 13: Typo I assume:

```suggestion
import Data.Linear
```

PR #2294, opened 2022-01-28
[ refactor ] Abstract Prelude.elem to work with all Foldables

madman-bob:
Abstract `Prelude.elem` to work with all `Foldable`s, not just `List`s.

Also adds `Prelude.elemBy`, for all `Foldable`s, and removes the `elem`s and `elemBy`s for `List`, `SnocList`, and `Vect`.

This introduced a handful of ambiguities in `src`, which have also been fixed.


mattpolzin:
Do you mind adding a CHANGELOG entry (that mentions the potential need for disambiguating code after the next release)? I might start to have a reputation as the CHANGELOG police 😬 but maybe that's not so terrible 🙂.

PR #2302, opened 2022-02-02
[ base ] Split a list according to a decidable property

CodingCellist:
The datatype for this was provided by @gallais as help for a project I'm working on. But I thought it seemed useful enough that it should possibly be part of `Data.List.Quantifiers`; for all to use. Let me know you disagree or if there is a better place to put it  : )


libs/base/Data/List/Quantifiers.idr
```diff
@@ -187,3 +187,63 @@ pushOutInInverse [] = Refl
...
+record Split {a : Type} (p : a -> Type) (xys : List a) where
+  constructor MkSplit
+  {xs, ys : List a}
+  {auto interleaving : Interleaving xs ys xys}
+  ||| A proof that all elements in `xs` satisfies the property used for the
+  ||| split.
+  prfs    : All p xs
+  ||| A proof that all elements in `ys` do not satisfy the property used for the
+  ||| split.
+  contras : All (Not . p) ys
```
gallais, line 218: oof yeah you'll probably want to use long names for these to avoid getting warnings about shadowing
for all of the uses of xs and ys in type signatures

CodingCellist: I took the names from the original Discord version for `xs` and `ys`: `ayes` and `naws` respectively.
Not great, not terrible? ^^

PR #2303, opened 2022-02-02
Allow Type-level String operations

madman-bob:
Allow `String` operations to reduce at the Type-level, with test.


madman-bob:
Your guess is correct. A proper fix for this would involve some serious thought around the `import` mechanism, but I don't know precisely what that would end up looking like.

PR #2304, opened 2022-02-02
[ libs ] Strengthen some totality checks

nickdrozd:


nickdrozd:
Things I wanted to change but couldn't:

- Some `prim__` functions are covering and some aren't, and I don't know why.
- In `TailRec`, the `with` blocks somehow interfere with `assert_smaller`.

stefan-hoeck:
>     * Some `prim__` functions are covering and some aren't, and I don't know why.

Totality of primitive functions is defined [here](https://github.com/idris-lang/Idris2/blob/ad3a1c153c62522b65cb0106de3660e648aa466c/src/Core/Primitives.idr#L678).

PR #2305, opened 2022-02-03
[ re #1990 ] Use `which` to make exec path absolute

gallais:


melted:
This will definitely not work on Windows outside of a Unix-like environment where 'which' doesn't exist (it may or may not work inside, I'm thinking of pathname conversion issues there, but that will be answered by the tests). There is a program filling the same function there called 'where.exe', the difference is that it can return several lines if there is more than one instance of that program in the path.

benhormann:
@gallais You could use `command -v` instead, which is more portable. What it really needs is an `isAbsolutePath` check.

Having `/usr/bin/env scheme` as a default is bad; `#!/usr/bin/env scheme --program` is the same as `/usr/bin/env "scheme --program"`. Idris should abort if it can't find the compiler.

benhormann:
@gallais With #1992, this becomes moot — an absolute path is not required. It could test the given command with `--version` if you want it to fail early.

PR #2310, opened 2022-02-05
[ cleanup ] Remove unused `get`'s, replace `get+put` with `update`

buzden:
When inspecting the `Core` code I found that there are a few places where `Ref` is read using `get` for no visible reason and can be eliminated. While finding such places, I found a lot of repeating consequent `get`+`put` calls which could be replaced with existing `update` one.

### DIFF
--- a/src/Compiler/ES/Codegen.idr
+++ b/src/Compiler/ES/Codegen.idr
@@ -672,7 +672,7 @@ mutual
     as <- traverse (map (insertBreak r) . alt) alts
     d  <- traverseOpt stmt def
     nm <- get NoMangleMap
-    pure $  switch (minimal nm sc <+> ".h") as d
+    pure $ switch (minimal nm sc <+> ".h") as d
     where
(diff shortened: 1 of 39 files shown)


mattpolzin (commented):
Having not read over the PR yet, I'll just speculate out loud that sometimes it may be meaningful to read a value prior to some other effectful code and then put something back rather than using update -- while uncommon, I imagine, was that on your mind while doing this refactor?

buzden:
> was that on your mind while doing this refactor?

Sure! To my knowledge I changed only those places that
1) does not use the read value between `get` and `put`
2) does not contain calls to functions that change the ref between `get` and `put`
3) does not use read value after the `put`.

buzden:
Some mystery is going on with windows testing job after merging...

gallais:
The windows CI is flaky atm. We typically get a fail and then a success after a restart. Strange.

PR #2311, opened 2022-02-06
[ new ] A couple of nearly trivial proofs about list length.

mattpolzin:
Not totally sure on the names I used for these functions. Open to different spellings.

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -954,3 +954,10 @@ export
 lengthMap : (xs : List a) -> length (map f xs) = length xs
 lengthMap [] = Refl
 lengthMap (x :: xs) = cong S (lengthMap xs)
+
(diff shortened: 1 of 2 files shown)


libs/base/Data/List.idr
```diff
@@ -954,3 +954,10 @@ export
 lengthMap : (xs : List a) -> length (map f xs) = length xs
 lengthMap [] = Refl
 lengthMap (x :: xs) = cong S (lengthMap xs)
+
+||| Proof that replicate produces a list of the requested length.
+export
+replicateLength : (n : Nat) -> length (replicate n x) = n
```
gallais, line 960: I think we should stick to the convention introduced above: `lengthMap`  for the `length (map ...)`  lemma
so `lengthReplicate`  for the `length (replicate ...)` lemma.

PR #2313, opened 2022-02-07
[ fix ] Flush standard when prompting for package information

nmeum:
On Unix-like operating systems stdio.h is usually line-buffered. As
putStr uses fputs(3) from stdio.h internally, output will be written to
standard out after a newline character is written to the buffer. Since
the various prompts written by `idris2 --init` don't contain a newline
character they are buffered indefinitely. As such, the prompt strings
themselves are never presented to the user.

To fix this issue, this commit introduces a prompt utility procedure
which prints a prompt string and afterwards flushes standard output
before receiving input using getLine. This ensures that the prompt
string is always written before input is retrieved via getLine.

This is similar to c725b11c891b9f9f1e25ead66d9fc75a91b69836.

### DIFF
--- a/src/Idris/Package/Init.idr
+++ b/src/Idris/Package/Init.idr
@@ -9,6 +9,7 @@ import Data.String
 
 import Idris.Package.Types
 import System.Directory
+import Control.App.FileIO
 
 import Libraries.Utils.Path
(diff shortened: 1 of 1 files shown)


gallais:
Thanks!

PR #2314, opened 2022-02-07
[ doc ] Add documentation for lambda-lifted representation

magnostherobot:
The code behind the "Lifted" intermediate representation used in Idris 2's compilation has been left mostly uncommented, until now. All features marked as "public" have been given doc-strings, with sections on the majority of arguments passed to functions.

A downside to having this much documentation is that the :doc REPL command becomes quite verbose when used on e.g. Lifted: over a screen of text is printed!

There didn't seem to be a consistent line wrap limit in this file, so all code and documentation have been made to fit in a 80-character-wide limit. Some function arguments were also given names, as a prerequisite for writing about them in doc-strings.

There are a few TODOs added to this file: there are some parts of the lambda-lifted code that I don't fully understand, and I'd appreciate some advice on where I could look to fill in the gaps in my knowledge.  Generally, it's probably a good idea to have someone who understands this code well to read through this new documentation, as I'm not entirely confident in my own comprehension.

This PR is marked as WIP since it contains some TODOs. If there is any interest in this PR being completed I'd be more than happy to take the last couple steps and make this a PR proper :)


Z-snails (commented):
Overall very good and clear, I've provided some more info for the bits you marked TODO

jfdm:
> A downside to having this much documentation is that the :doc REPL command becomes quite verbose when used on e.g. Lifted: over a screen of text is printed!

This is an annoying aspect of documentation. Idris1 had a settable option to display brief (first sentence/paragraph) or long doc strings (everything!). Internally both the brief and long were stored separately. 

Maybe it is time that this option came back, a good first issue...

gallais:
> then this PR should be good to go I think?

I can still see some unresolved comments by @Z-snails which is why I haven't merged
anything. But I'm happy to accept the PR even if the documentation is not 100%. It's already
a net improvement.

Z-snails:
@gallais I think most or all of my comments have been resolved, I'm happy for this to be merged, as it's a massive improvement over what was already there!

PR #2317, opened 2022-02-10
add deleteAt for List

joelberkeley:


joelberkeley:
@buzden ok, thanks. That's confusing. I'd always thought that was the case until recently when I kept seeing Idris error when i pattern matched on an erased value. I guess there must have been something else going on. Do you mean to erase just the `Inbounds` or also the index?

joelberkeley:
oh i see - they duplicate the data so only one needs to be runtime to pattern match

PR #2319, opened 2022-02-11
fix tanh value at infinity

joelberkeley:
tanh should be plus or minus 1 at plus or minus infinity, but because it's calculated by dividing `sinh` and `cosh`, we get `nan`. This fixes that. I could perhaps make it cleaner if there's a `sign` function somewhere

I've not had a thorough think if there are other special cases that need accounting for here


joelberkeley:
desired behaviour is not obvious - replaced with https://github.com/idris-lang/Idris2/issues/2320

PR #2324, opened 2022-02-16
[ prelude ] Ease `Compose` compositions

buzden:
This enables us to do `map f @{Compose @{Compose}}` instead of `map f @{Compose @{(Compose, %search)}}` and `(>=>) a 
b @{Compose @{Compose}}` instead of `(>=>) a b @{Compose @{(Compose, %search, %search)}}` when the level is three.

### DIFF
--- a/libs/prelude/Prelude/Interfaces.idr
+++ b/libs/prelude/Prelude/Interfaces.idr
@@ -118,7 +118,7 @@ ignore = map (const ())
 namespace Functor
   ||| Composition of functors is a functor.
   public export
-  [Compose] Functor f => Functor g => Functor (f . g) where
+  [Compose] (l : Functor f) => (r : Functor g) => Functor (f . g) where
     map = map . map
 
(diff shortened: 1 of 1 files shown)


gallais:
Is it not possible to name the arguments so that you can `map @{Compose @{g = Compose}}` if it's not
associated the same way?

buzden:
> Is it not possible to name the arguments so that you can `map @{Compose @{g = Compose}}` if it's not associated the same way?

If I understand what you asked, we can do the following.

We can add names sub-interfaces to the definitions of `Compose`, like this:

```idris
namespace Functor
  [Compose] (l : Functor f) => (r : Functor g) => Functor (f . g) where
    ...
```

If so, we can compose the other way around:

```idris
C : Functor f => Functor g => Functor h => Functor (f . (g . h))      
C = Compose {r=Compose}                                                                                                                                                 
```

Issue #2237, opened 2021-12-31
chez/chez014 test does not terminate on docker


youngmysteriouslight:
# Steps to reproduce
```
$ cat Dockerfile
FROM ubuntu:20.04

ARG REPO=https://github.com/idris-lang/Idris2/
ARG COMMIT=HEAD

RUN apt-get update && apt-get install -y gcc git make chezscheme libgmp3-dev build-essential clang && rm -rf /var/lib/apt/lists/*

RUN echo "cloning $REPO" && \
    git clone $REPO /opt/idris2 && \
    cd /opt/idris2 && git checkout $COMMIT

RUN true

WORKDIR /opt/idris2

ENV SCHEME=scheme

RUN export PATH="${PATH}:/root/.idris2/bin" && make bootstrap && make install
RUN make test
```
```
$ docker build -f Dockerfile .
```

# Expected behavior
`make test` must terminate.

# Actual behavior
`make test` is stuck on `chez/chez014`.

# Investigation
Adding `print` or `putStrLn` anywhere in `chez/chez014/Echo.idr` fixes the problem:
```
main : IO ()
main = do
  Right serverPort <- runServer True
    | Left err => putStrLn $ "[server] " ++ err
  runClient True serverPort
  Right () <- runServer False
    | Left err => putStrLn $ "[server] " ++ err
  runClient False 0
  ignore $ removeFile testSocket
  print "test finishes"
```

# Another way to reproduce the problem
```
$ git clone https://github.com/idris-lang/idris2 && cd idris2 && docker run -it --rm -v "$PWD":/opt/idris2 -w /opt/idris2 ubuntu:20.04 bash
[container]$ apt-get update && apt-get install --yes gcc make chezscheme libgmp3-dev git && rm -rf /var/lib/apt/lists/*
[container]$ export SCHEME=scheme
[container]$ make bootstrap
[container]$ make install
[container]$ make test
```

# Acknowledgement
Thanks to @alexhumphreys for reproducing the problem and discussions.

alexhumphreys:
Also of note, removing the localhost test by deleting the lines:
```
Right serverPort <- runServer True
    | Left err => putStrLn $ "[server] " ++ err
  runClient True serverPort
```
also causes the test to terminate and not hang.

PR #2326, opened 2022-02-16
Add Dockerfile

jhvst:
What: A Dockerfile to build a container which creates an endpoint to Idris. With this approach, it is possible to e.g. type-check an Idris file by running: `podman run -v "$PWD:$PWD":z -w "$PWD" idris -c Vect.idr`. What happens here is the current folder is mounted to the container and the workspace is changed to the mounted folder inside the container. Then, the container runs Idris, and you can pass arguments as if you would have local version of Idris. If no parameters is given, help is printed.

Why: I wanted to have an ARM build on the Apple M1 and needed an easy way to run the latest changes. Coincidentally, I occasionally use Fedora Silverblue which requires everything to be a container.

Most of this code is from #2237 but uses a more recent version of Linux. I also noticed that containers were discussed in #1507.

I used racket because chez-scheme did not exist on ubuntu:22.04. I also looked into something more lightweight than ubuntu, in particular alpine linux, but neither chez-scheme nor racket is currently in the package manager.

Do you think providing a Dockerfile within the repository to run Idris is a good idea?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM ubuntu:22.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt update && apt install -y gcc git make racket libgmp3-dev build-essential clang
+RUN git clone https://github.com/idris-lang/Idris2 /opt/idris2
+
+WORKDIR /opt/idris2
+
+ENV IDRIS2_CG=racket
+RUN make bootstrap-racket
+RUN make install
+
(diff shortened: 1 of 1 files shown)


buzden:
Have you seen #789 and [this](https://github.com/alexhumphreys/idris2-dockerfile)?

jhvst:
No I did not, I used search function only on issues, not PRs.

For posteriority, I will have my Idris 2 Dockerfile here: https://github.com/jhvst/dockerfiles

PR #2330, opened 2022-02-18
add a proof that any `Nat` is `LTE` itself

joelberkeley:


gallais:
It's just a bit further up: https://github.com/idris-lang/Idris2/pull/2330/files#diff-e7afeb0c3b56f9e7d37b44a16c812633b29bd811c31e90b37e79f34b481a0989R102

joelberkeley:
oh ok thanks, today I learned about Reflexive

PR #2337, opened 2022-03-02
Fix for lexer colums

jutaro:
The lexer under contrib Text.Lexer.Core assigns wrong colum numbers to tokens after line breaks. This patch fixes this bug.

### DIFF
--- a/libs/contrib/Text/Lexer/Core.idr
+++ b/libs/contrib/Text/Lexer/Core.idr
@@ -147,7 +147,7 @@ tokenise pred line col acc tmap str
 
     getCols : List Char -> Int -> Int
     getCols x c
-         = case span (/= '\n') (reverse x) of
+         = case span (/= '\n') x of
                 (incol, []) => c + cast (length incol)
                 (incol, _) => cast (length incol)
 


andylokandy:
FYI, `src/Libraries/Text/Lexer/Core.idr` may has the same issue.

dunhamsteve:
I just hit the same issue today. If I lex `"let x = 1\n    y = 2\n in x + y"`, the first `y` comes back with a column of 0.

It looks like it also occurs in `src/Libraries/Text/Lexer/Core.idr`. This issue doesn't affect the compiler because it uses a similar function from `src/Libraries/Text/Tokenizer/Core.idr` that already has the fix.

andrevidela:
@dunhamsteve so we just need to make the same change to `src/Libraries/Text/Lexer/Core.idr`?

dunhamsteve:
Yeah, I have a branch with both changes and a test case, but I wasn't sure if it was kosher to snipe this PR.

https://github.com/idris-lang/Idris2/compare/main...dunhamsteve:Idris2:lexer-issue

dunhamsteve:
The ipkg parser uses the buggy lexer, so you can see this bug in action if you try to load this ipkg:
```
package foo
    xxx = 123
```
and get:
```
idris2 --build foo.ipkg
Uncaught error: Parse errors ([("foo.ipkg":2:1--2:4, "Expected end of input.")])
```

andrevidela:
@dunhamsteve let's take your branch and merge it. it's also got the shared commit authors so attribution is accounted for

PR #2338, opened 2022-03-02
[ ci ] install mingw64 gcc for windows

joelberkeley:
~~currently commented out to check CI is actually broken~~ not any more

### DIFF
--- a/.github/workflows/ci-idris2.yml
+++ b/.github/workflows/ci-idris2.yml
@@ -202,7 +202,7 @@ jobs:
         run: |
           git clone --depth 1 https://github.com/cisco/ChezScheme
           c:\msys64\usr\bin\bash -l -c "pacman -S --noconfirm tar make"
-          c:\msys64\usr\bin\bash -l -c "# pacman -S mingw-w64-x86_64-gmp"
+          c:\msys64\usr\bin\bash -l -c "pacman -S mingw-w64-x86_64-gmp"
           echo "PWD=$(c:\msys64\usr\bin\cygpath -u $(pwd))" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
       - name: Configure and Build Chez Scheme
         run: |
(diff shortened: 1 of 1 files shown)


joelberkeley:
**I DON'T THINK CI RAN, SO MAYBE DON'T MERGE YET?**

.github/workflows/ci-idris2.yml
```diff
@@ -201,7 +201,7 @@ jobs:
       - name: Get Chez Scheme
         run: |
           git clone --depth 1 https://github.com/cisco/ChezScheme
-          c:\msys64\usr\bin\bash -l -c "pacman -S --noconfirm tar make"
+          c:\msys64\usr\bin\bash -l -c "pacman -S --noconfirm tar make mingw-w64-x86_64-gmp"
```
benhormann, line 204:
```suggestion
          c:\msys64\usr\bin\bash -l -c "pacman -S --noconfirm tar make mingw-w64-x86_64-gcc"
```

GCC depends on GMP. So if `mingw-w64-x86_64-gmp` is missing, we need to install `mingw-w64-x86_64-gcc`.
Add it for windows-self-host-racket too, line 419.

BTW, the first warning gives a clue to this: `idris_net.c:211:3: warning: implicit declaration of function 'inet_ntop'; did you mean 'inet_ntoa'?` (normally there are no warnings).

benhormann (changes requested):
Seems the MINGW64 compiler is no longer installed by default?

.github/workflows/ci-idris2.yml
gallais: Let's try that!

benhormann: @gallais The other windows job will need it too. (pity I can't add a suggestion there too)

benhormann:
```
git reset 1011cc61 --soft
git commit -m 'install mingw64 gcc for windows CI jobs'
git push -f
```
would clean up these commits.

Though gallais will probably squash the PR merge, so maybe just update the title instead?

joelberkeley:
@benhormann i believe the person merging can choose the commit message

gallais:
Thank you both for looking into this!

PR #2339, opened 2022-03-03
[ fix ] Euclidian div and mod for JS and Scheme backends

### LINKED ISSUE #1571: The modulo function incorrectly behaves as a remainder function.
# Steps to Reproduce
Modulo:
```idris
map (\d => d `mod` 3) [-3..3] 
```
Remainder:
```idris
map (\d => d `rem` 3) [-3..3] 
```
# Expected Behavior
Modulo:
```idris
[0, 1, 2, 0, 1, 2, 0]
```
Remainder:
```idris
[0, -2, -1, 0, 1, 2, 0]
```
# Observed Behavior
Modulo:
```idris
[0, -2, -1, 0, 1, 2, 0]
```
Remainder:
```
Error: Undefined name rem. 
```

# Commentary
The function for `mod` seems to be incorrectly implemented as the remainder function.
This is in disagreement with a number of languages - Chez Scheme and Haskell all behave as in the expected behavior above.
This issue is also present in the original Idris 1.

I'm moderately worried that just straight up swapping the definition may result in some problems in the core, but presumably if appropriate care is taken to rename all the usages of `mod` to `rem` as described above (and as used by Haskell), everything should work fine.

stefan-hoeck:
This is a long overdue followup on #1480, where @madman-bob added Euclidian `div` and `mod` to the RefC backend. With this PR, the JS and Scheme backends follow after, and the corresponding test cases in the `integers` test have finally been activated. This should also fix #1571.

One thing I'm not sure about: I didn't look at or change the code of the Scheme evaluator. Any thoughts on that?


stefan-hoeck:
I guess the Windows CI failing is related to #2338, yes?

stefan-hoeck:
I rebased this to make sure the windows CI works as well...

PR #2340, opened 2022-03-04
[ doc ] `:printdef` for interface implementations

gallais:


mattpolzin:
This is super useful.

The expanded behavior (of printing the implementation itself), feels akin to the existing `:printdef` (which I am sure does not support printing interface implementations). This just makes me wonder: is this functionality more fitting as an extension to `:printdef`? I'm not sure it is, but curious if you considered that as an option here.

gallais:
Good point. It should be a fairly simple transformation of the parser & moving the call to
`getDocsForImplementation` to the `PrintDef` case in `process`.

PR #2348, opened 2022-03-10
[ refactor, breaking ] Factor out a type for primitive types out of `Constant`

buzden:
There are several reasons to do this:

- there are a lot of functions that actually expect or return only primitive types, not any `Constant`;
- there were couple of places in the compiler (in different modules) where distinction between value constants and type constants was done through comparing some integer tag with 0, which is fragile;
- better exhaustiveness check; sometimes types were wrongly managed as `_` and some other types were put into `_` for unknown reason;
- sometimes there is a need to have some type of primitive types as a type index; I ran into such a need in some elaboration scripts and in my being done improvements in the scheme evaluator.

This is a breaking change for backends and for elaboration scripts that are working with primitive types, however fixes are rather trivial.


src/Core/Ord.idr
```diff
@@ -44,22 +65,8 @@ Ord Constant where
...
-        tag Bits8Type = 20
-        tag Bits16Type = 21
-        tag Bits32Type = 22
-        tag Bits64Type = 23
-        tag StringType = 24
-        tag CharType = 25
-        tag DoubleType = 26
-        tag WorldType = 27
-
+        tag (PrT _) = 13
```
mattpolzin, line 68: It may be inconsequential, but you’ve switched the order of worldval compared to any prim.

buzden: Yes, I thought that this order is arbitrary anyway, so it does not really matter. Do you think it's worth to make worldval to be the least again?

mattpolzin: I'm fairly uncertain. I doubt it matters, but thought it was worth the note. It's possible that some code somewhere does need world to get sorted to the end of something or other; it is a pretty special type of thing.

gallais:
Yes I'm happy to merge this as is.

stefan-hoeck:
I merged the PR at elab-util.

PR #2349, opened 2022-03-10
[ base ] quantifiers for Snoclists

gallais:
This would enable e.g. https://github.com/gallais/idris2-table/commit/82f04f829e9bde3b6afca1d9e97fe9df9ab2a500 in case @madman-bob is interested in having
fewer things to maintain.


madman-bob:
I'm always a fan of having fewer things to maintain.

Though I am concerned about duplication between the `List` `All`, and the `SnocList` `All`. How difficult is it to get them under a common interface?

jfdm:
Out of curiosity, Is there any particular reason why the decidable functions are called `decide` and not `any` & `all`. In the other quantifier modules there is a clear naming pattern. I wonder if it would make sense to continue this pattern, or update the other instances to the pattern used here.

stefan-hoeck:
This would also allow us to implement `all` tail recursively for both types of universal quantification. I just experimented with such a tail recursive version the other day, because I wanted to proof stuff about lengthy lists of values on the JS backends. They'd have to be defined in the same module, though.

gallais:
> Out of curiosity, Is there any particular reason why the decidable functions are called decide and not any & all.

Probably to avoid overloading the functions in the `Foldable` interface.

gallais:
I'll merge this as is for now but if someone comes up with a better name, I'm happy to rename.

PR #2350, opened 2022-03-10
[ prelude ] add proof of uninhabited for symetric equality

andylokandy:


libs/prelude/Prelude/Uninhabited.idr
```diff
@@ -34,8 +34,8 @@ absurd : Uninhabited t => (h : t) -> a
 absurd h = void (uninhabited h)
 
 public export
-Uninhabited (True = False) where
-  uninhabited Refl impossible
+Uninhabited (x = y) => Uninhabited (y = x) where
```
gallais, line 37: This seems like a great way to make the proof search loop.

andylokandy:
```idris
module Test

Uninhabited (x = y) => Uninhabited (y = x) where
  uninhabited = \h => void $ uninhabited $ sym h

data MyNat = ZZ | SS MyNat

-- Comment out for attempting to trap the compiler into an infinity loop
-- Uninhabited (ZZ = SS ZZ) where
--   uninhabited Refl impossible 

nonSense : Not (SS ZZ = ZZ)
nonSense = absurd
```

outpus:

```
Error: While processing right hand side
of nonSense. Can't find an implementation
for Uninhabited (SS ZZ = ZZ).

Test:12:12--12:18
 08 | -- Uninhabited (ZZ = SS ZZ) where
 09 | --   uninhabited Refl impossible 
 10 | 
 11 | nonSense : Not (SS ZZ = ZZ)
 12 | nonSense = absurd
                 ^^^^^^

Error(s) building file Test.idr
```

andylokandy: Seems that the complier is robust enough to not get trapped.

andylokandy: well... the test shows that the proof search indeed looped.

andylokandy:
@gallais After further investigation, I've found the proof searcher does have cycle detection. The real cause of the test failure is due to the ambiguity of multiple viable proving strategies to make.

For example, given the implementations:

```idris
1. Uninhabited (Z = S n)
2. Uninhabited (x = y) => Uninhabited (S x = S y)
3. Uninhabited (x = y) => Uninhabited (y = x)
```

and to prove `Uninhabited (S (S Z) === S Z)`, it can take `1 2 3` or `1 3 2`, ambiguous!.

So to solve the problem, I've to re-introduce the [`%defaulthint` pragma](https://github.com/idris-lang/Idris2/pull/2351) so that I can annotate `Uninhabited (x = y) => Uninhabited (y = x)` by `%defaulthint` to avoid ambiguity.

PR #2351, opened 2022-03-11
[ parser ] Add pragma %defaulthint for allowing ambiguity in proof search

andylokandy:
Previous discussion: https://github.com/idris-lang/Idris2/pull/2247, https://github.com/idris-lang/Idris2/pull/2350#issuecomment-1065319462


andylokandy:
Is there any concern I could try to address? Or is it a bad idea? 😌

gallais:
I'm weary of adding extra ambiguity so I'd let Edwin decide on this one.

PR #2357, opened 2022-03-17
[ prelude ] add `subtract`

joelberkeley:


libs/prelude/Prelude/Num.idr
```diff
@@ -30,6 +30,16 @@ interface Num ty => Neg ty where
...
   (-) : ty -> ty -> ty
 
+||| A convenience alias for `(-)`, this function enables partial application of subtraction on the
+||| right-hand operand as
+||| ```idris example
+||| (`subtract` 1)
+||| ```
+||| which with `(-)` would instead be parsed as `-1`.
+export
+subtract : Num ty -> ty -> ty -> ty
```
gallais, line 40:
```suggestion
subtract : Num ty => ty -> ty -> ty
```

joelberkeley: this is the missing semi-colon of idris

PR #2359, opened 2022-03-17
[ base ] Some properties of `List` and `SnocList`

buzden:
...or else it is hard to write even simple properties involving blend of `List1`, `SnocList` and `List`.

### DIFF
--- a/libs/base/Data/List.idr
+++ b/libs/base/Data/List.idr
@@ -1027,3 +1027,8 @@ export
 lengthReplicate : (n : Nat) -> length (replicate n x) = n
 lengthReplicate 0 = Refl
 lengthReplicate (S k) = cong S (lengthReplicate k)
+
+export
+foldlAppend : (f : acc -> a -> acc) -> (init : acc) -> (xs : List a) -> (ys : List a) -> foldl f init (xs ++ ys) = foldl f (foldl f init xs) ys
+foldlAppend f init []      ys = Refl
+foldlAppend f init (x::xs) ys = rewrite foldlAppend f (f init x) xs ys in Refl
--- a/libs/base/Data/SnocList.idr
+++ b/libs/base/Data/SnocList.idr
@@ -6,11 +6,11 @@ import Data.Fin
 
 %default total
 
-export
+public export
 Cast (SnocList a) (List a) where
   cast sx = sx <>> []
 
-export
+public export
 Cast (List a) (SnocList a) where
   cast xs = Lin <>< xs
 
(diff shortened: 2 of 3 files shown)


libs/base/Data/List.idr
```diff
@@ -1027,3 +1027,8 @@ export
 lengthReplicate : (n : Nat) -> length (replicate n x) = n
 lengthReplicate 0 = Refl
 lengthReplicate (S k) = cong S (lengthReplicate k)
+
+export
+foldlOfAppend : (xs : List a) -> (ys : List a) -> (f : acc -> a -> acc) -> (init : acc) -> foldl f init (xs ++ ys) = foldl f (foldl f init xs) ys
```
gallais, line 1032: Please use the same naming scheme as in the rest of the file:
* lemma of the shape `fun1 (fun2 xs) = ...` is named `fun1Fun2`
* no `Of` in the middle of the name
* arguments taken in the same left-to-right order they're in in the statement, here: `f, init, xs, ys

libs/base/Data/SnocList.idr
```diff
@@ -138,3 +138,50 @@ findIndex _ Lin = Nothing
...
+
+export
+appendLinLeftNeutral : (sx : SnocList a) -> [<] ++ sx = sx
+appendLinLeftNeutral [<]       = Refl
+appendLinLeftNeutral (sx :< _) = rewrite appendLinLeftNeutral sx in Refl
+
+--- Append of snoc-list and usual list ---
+
+export
+heterListAppendAsListAppend : (xs : SnocList a) -> (ys : List a) -> xs <>> ys = cast xs ++ ys
```
gallais, line 161: This `(<>>)` combinator is called chips

libs/base/Data/SnocList.idr
```diff
@@ -138,3 +138,50 @@ findIndex _ Lin = Nothing
...
+heterListAppendAsListAppend [<]       ys = Refl
+heterListAppendAsListAppend (sx :< x) ys = do
+  rewrite heterListAppendAsListAppend sx (x :: ys)
+  rewrite heterListAppendAsListAppend sx [x]
+  rewrite sym $ appendAssociative (cast sx) [x] ys
+  Refl
+
+export
+heterSnocAppendAsSnocAppend : (xs : SnocList a) -> (ys : List a) -> xs <>< ys = xs ++ cast ys
```
gallais, line 170: And this one `(<><)` is fish

buzden:
Hopefully I resolved the comments, rebased and force-pushed. Also, I've added a couple more properties, I hope that's okay.

libs/base/Data/List.idr
buzden: I wonder why `++` is already called `append`. Isn't it a concatenation, when appending is what `snoc` function does?

PR #2360, opened 2022-03-17
[ new ] failing blocks

gallais:
Blocks that let user document invalid code. Can be used in literate
mode to document things that do not typecheck.

Left to do:

* use the same error messages in pretty mode & when checking the
  substring is present.
* support syntax highlighting (I think we just need to purge the
  name information for the identifiers that won't make it to TTC)


andrevidela:
It can also be used for test at compile-time !

PR #2361, opened 2022-03-20
[ base ] Add some more properties, functions and interface implementations

buzden:
Types `List`, `List1`, `SnocList` and `Maybe` are affected. Also, definitions from `Data.List1` modules were moved to a special `Core` submodule to avoid circular dependencies while keeping backward compatibility.


gallais:
I'd rather move these crucial SnocList definitions to prelude tbh

buzden:
I've rebased upon recently merged PRs to ensure nothing breaks

gallais (approved):

PR #2363, opened 2022-03-20
[ doc ] Add docs for quotes

Z-snails:


src/Idris/Parser.idr
```diff
@@ -2065,7 +2065,12 @@ docArgCmd parseCmd command doc = (names, DocArg, doc, parse)
         <|> Keyword <$> anyKeyword
         <|> Symbol <$> (anyReservedSymbol <* eoi
                        <|> parens (Virtual Interactive) anyReservedSymbol <* eoi)
-        <|> Bracket <$> (IdiomBrackets <$ symbol "[|" <* symbol "|]")
+        <|> Bracket <$> (
+              IdiomBrackets <$ symbol "[|" <* symbol "|]"
```
mattpolzin, line 2069: IMO it would be nice for `:doc [|` (just the opening bracket) to work and similarly for the other bracket types, if you'd like to make the closing brackets optional while you're in here.

Z-snails: I think the lexer/tokeniser disallows unmatched brackets, so that would be much more complicated, and in my opinion not worth it, as the error message it helpful enough to suggest closing the bracket

PR #2364, opened 2022-03-20
Implement standard List operations for SnocLists

ohad:
The design philosophy is that we should always think of lists as going left-to-right, but switch between cons-lists `List` and snoc-lists `SnocList`, depending on how we want to access them.

In particular, the various operations that work on each side of the list should always mean that side: `drop` always drops the left-most bunch of elements, and so on.

Matters to discuss during review:

1. ~~I didn't know what to call the analogous right-operations. I opted for the reverse (`pord` for `drop`, `ekat` for `take`, and so on), just like snoc is the analogue of cons. Better names are welcome, just lets not bikeshed over it.~~
  Renamed following code review
3. the `splitOntoLeft` and `splitOntoRight` operations are definable using `take`, `drop`, their analogues, and concatenation. But maybe someone wants to get both at once with one traversal. Not sure it saves much.
4. ~~The dependency on `contrib` is the `PreorderReasoning`, so these go into `contrib`. If people start using them and the interface stabilises, we can move them to `base`.~~
  Dependency removed thanks to  #2368 
6. I made some operations `public`. I expect we're going to get more of those, as more people use standard-library functions at the type-level. I wish we had a better way to do it than rewrite the standard libraries!


gallais:
I suppose we already have `cons` vs. `snoc` to justify that naming convention but it's pretty horrible. :/
Maybe `takeR` for a take from the right? This is the convention used in Haskell's [`Data.Seq`](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Sequence.html#g:10) for instance.

libs/base/Data/SnocList.idr
```diff
@@ -217,6 +217,10 @@ toListCast (x::xs) = do
   rewrite toListCast xs
   Refl
 
+public export
```
gallais, line 220: There should be a docstring warning users this is O(n)

libs/contrib/Data/SnocList/Operations.idr
```diff
@@ -0,0 +1,133 @@
...
+-}
+
+||| Shift `n` elements from the beginning of `xs` to the end of `sx`,
+||| returning the same lists if `n` >= length `xs`.
+|||
+||| @ n  the number of elements to take
+||| @ sx the snoc-list to append onto
+||| @ xs the list to take the elements from
+public export
+shiftLeft : (n : Nat) -> (sx : SnocList a) -> (xs : List a) -> (SnocList a, List a)
```
gallais, line 55: combinations of `(take, drop)` are usually called `splitAt` IIRC

ohad: I see. How about `splitOntoLeft`/`splitOntoRight` then?

ohad:
> I think Syntax.PreorderReasoning should go in base (ideally extended with the frex gadgets
for symmetric & propositional equality steps). It would allow us to clean up a lot of proofs.

Should I include those proofs in the upcoming PR, or just the new syntax moved the base + bugfixes for any code it breaks

gallais:
> Should I include those proofs in the upcoming PR

Not unless you want to. Moving `Syntax.PreorderReasoning` to `base` is valuable in and of itself
whether or not we also perform the refactorings it makes possible.

ohad:
Since I don't want to work out what it might do to bootstrapping, I'll leave refactoring any stdlib proofs until the next release.

ohad:
If no other comments and passes tests, ready to merge

PR #2368, opened 2022-03-22
Move `Syntax.PreorderReasoning` into `base`

ohad:
Add more (generic and ordinary) preorder reasoning combinators

Developed in [frex](https://github.com/frex-project/idris-frex/blob/4a3cbb94bcd1ffae527c453df94ac4ab0cf61e35/src/Syntax/PreorderReasoning/Setoid.idr#L29)


libs/base/Syntax/PreorderReasoning.idr
```diff
@@ -35,3 +35,13 @@ example x =
    ~~ x+1  ...( plusZeroRightNeutral $ x + 1 )
    ~~ 1+x  ...( plusCommutative x 1          )
 -}
+
+-- Smart constructors
+public export
+(..<) : (y : a) -> {x : a} ->
+    y = x -> Step x y
```
gallais, line 42: Is there a reason this is not using heterogeneous equality like the following combinator?

ohad: Yes, `Builtins.sym` is homogeneous.
I've changed that now, but we need to use a local copy until the next release to enable bootstrapping

ohad:
OK, ready to merge

PR #2371, opened 2022-03-23
[ refactor ] Index Pretty over the type of annotations

gallais:
The fact that the `Pretty` interface is polymorphic in the annotations type means it's essentially useless
as an interface. So here's a refactoring that allows users to specify the type of annotations associated
to the type being rendered.

Immediate benefit: compiled trees in `:di` have a bit more colour.


gallais:
This change seems to make the build *a lot* slower, in particular two relatively small modules take ages to build:
1. `Core.CompileExpr.Pretty`
2. `Core.Context.Pretty`

`make clean && time make` using e8d3d788c1d27a97079ed7638cbf940ac5656628 takes:

* ~4m30s to compile e8d3d788c1d27a97079ed7638cbf940ac5656628 itself
* ~8m15s to compile this branch

I guess we'll have to see what @edwinb thinks.

PR #2375, opened 2022-03-25
print full, human-readable function names in error messages

joelberkeley:
Is this right? And are there other places I could do this while I'm at it?

### DIFF
--- a/src/Idris/Error.idr
+++ b/src/Idris/Error.idr
@@ -221,7 +221,7 @@ perror (PatternVariableUnifies fc env n tm)
            | Nothing => (fc1, EmptyFC)
       in if sr1 < sr2 then (fc1, fc2) else if sr1 == sr2 && sc1 < sc2 then (fc1, fc2) else (fc2, fc1)
 perror (CyclicMeta fc env n tm)
-    = pure $ errorDesc (reflow "Cycle detected in solution of metavariable" <++> meta (pretty !(prettyName n)) <++> equals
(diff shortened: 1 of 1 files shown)


src/Idris/Error.idr
```diff
@@ -257,19 +257,19 @@ perror (BadDataConType fc n fam)
...
-    = pure $ errorDesc (pretty !(prettyName n) <++> reflow "is not covering.")
+    = pure $ errorDesc (pretty !(toFullNames n) <++> reflow "is not covering.")
         <+> line <+> !(ploc fc) <+> line
         <+> reflow "Calls non covering function" <+>
         case ns of
              [fn] => space <+> pretty fn
              _ => pretty 's' <+> colon <++> concatWith (surround (comma <+> space)) (pretty <$> ns)
 perror (NotTotal fc n r)
-    = pure $ errorDesc (code (pretty !(prettyName n)) <++> reflow "is not total," <++> pretty r)
+    = pure $ errorDesc (code (pretty !(toFullNames n)) <++> reflow "is not total," <++> pretty r)
```
gallais, line 272: The reason probably needs a `toFullNames` too as that was one of the $ names reported on the discord.

joelberkeley:
> In general, it's not a good idea to blindly search & replace function calls without even investigating
what they used to do.

Indeed. Since I'm entirely unfamiliar with the compiler code, my plan was to raise the PR and see if it was what you meant, and if not, get an idea of how much effort it would be to understand the problem - i.e. whether I can be useful here. It looks like this might be more involved than I have the capacity for right now. Sorry if I've made more work.

PR #2382, opened 2022-03-29
Replace ``\d`` to ``\a`` as Vim shortcut to create definition

ngeiswei:
Note that I did not check that one (I'm not a Vim user), but it makes sense according to the table at the top of the page.

### DIFF
--- a/docs/source/proofs/patterns.rst
+++ b/docs/source/proofs/patterns.rst
@@ -109,7 +109,7 @@ concrete value for its *second argument* (in fact, the left hand side of
 the equality has been reduced from ``plus Z m``.) Again, we can prove
 this by induction, this time on ``m``.
 
-First, create a template definition with ``\d``:
+First, create a template definition with ``\a``:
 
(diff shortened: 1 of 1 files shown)


CodingCellist:
Huh, I've always been using `<LocalLeader>d` (thinking `d` for `definition`/`define`), but it turns out both `<LocalLeader>a` and `<LocalLeader>d` are mapped to "add definition": [https://github.com/edwinb/idris2-vim/blob/master/ftplugin/idris2.vim#L301-L302=](https://github.com/edwinb/idris2-vim/blob/master/ftplugin/idris2.vim#L301-L302=)

The TDD book uses `<LocalLeader>a`, but it also specifies `<LocalLeader>d` as the documentation command, which is now `<LocalLeader>h` in Vim (`h` for `help`). I'm tempted to instead change all the `\a` definitions to `\d`, to create a distinction with Idris1, but maybe that's silly ^^;

In any case, the docs should definitely be consistent  : )

PR #2383, opened 2022-03-29
[ refactor, re #2375 ] HasNames Error implementation

gallais:


joelberkeley:
alas no - only a very large test case that I've now lost

PR #2395, opened 2022-04-01
Add Unit testing framework

magnostherobot:
While strict typing can remove the need for unit testing in a lot of places, there are still situations in which a batch of unit tests would strengthen confidence in the accuracy of code tremendously. This PR introduces a new testing framework, suitable for use alongside the existing Golden test framework.

When developing this module, ease of use has been prioritised to ensure that unit tests can be introduced into any Idris 2 codebase. Further, the module's code should be simple enough that any Idris 2 programmer can understand it, and even add to it, if need be.

...

Happy April Fool's, hope I'm not too late ;)

### DIFF
--- a/libs/test/Test/Unit.idr
+++ b/libs/test/Test/Unit.idr
@@ -1,0 +1,7 @@
+module Test.Unit
+
+Test : Type -> Type
+Test a = a -> Bool
+
+isUnit : Test Unit
+isUnit unit = unit == ()


stefan-hoeck:
This is a text book example of how to write concise high performance code in Idris 2. Amazing how you managed to come up with a feature complete, tail-call optimized test runner that still satisfies the totality checker. Well done!

the-emerald (approved):
lgtm 🚀🚀🚀

PR #2407, opened 2022-04-08
[ contrib ] Implement `Show ParsingError`

CodingCellist:
When writing parsers using the `contrib` library, it is often useful to be able to `show` the parse error (provided you can `show` your tokens).

### DIFF
--- a/libs/contrib/Text/Parser/Core.idr
+++ b/libs/contrib/Text/Parser/Core.idr
@@ -271,6 +271,14 @@ position = Position
 public export
 data ParsingError tok = Error String (Maybe Bounds)
 
+public export
+Show tok => Show (ParsingError tok) where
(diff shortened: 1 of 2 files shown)


libs/contrib/Text/Parser/Core.idr
```diff
@@ -271,6 +271,14 @@ position = Position
 public export
 data ParsingError tok = Error String (Maybe Bounds)
 
+public export
```
gallais, line 274: Does it need to be `public`? I don't think we'll run this at the type level.

CodingCellist: Probably not, in that case. I tend to use `public` as a "this implementation is not secret" thing, but if it's detrimental, then it can be removed  : )

CodingCellist: Changed  : )

PR #2413, opened 2022-04-14
[ base ] Add a `trace` variant easily embeddable to point-free expression

buzden:
I suggest to add such a function for simpler code when value should be traced and is not present with a name.


gallais:
I would suggest

```idris
traceBy : (a -> Show) -> a -> a
...

traceShow : Show a => a -> a
traceShow = traceBy show
```

to stick to the naming conventions used by `sortBy` for instance.

buzden:
Well, I've changed the definitions with my implementation of your idea, what do you think?

PR #2420, opened 2022-04-19
[ doc ] Proving Propositional Equality: example fixed

ayaye:
``plusReducesR`` is introduced in "Theorem Proving / Propositions and Judgments" section [1] with the following signature:

``plusReducesR : (n:Nat) -> plus n Z = n``

But it seems that in "Proving Propositional Equality" [2]  ``n = plus n 0`` (mentioned in text after example) is proved instead.

This PR fixes ``plusReducesR`` implementation and proposition mentioned in text: ``plus n 0 = 0``

1. https://idris2.readthedocs.io/en/latest/proofs/definitional.html#propositions-and-judgments

2. https://idris2.readthedocs.io/en/latest/proofs/propositional.html#proving-propositional-equality


gallais:
Thank you!

PR #2421, opened 2022-04-19
4 changes, 3 to RefC, 1 to Racket backend

vfrinken:
RefC: onCollectAny now has 3 arguments, the pointer, the freeing fct, and the world

created function names with brackets are handeled properly

Racket: added (require racket/unsafe/ops) to deal with unsafe-fx-, unsafe-fx+


dunhamsteve:
The require statement is already there on line 48 of Racket.idr. The Racket bootstrap issue is because the compiler in `idris2.rkt` is an older version that doesn’t add that require to generated code.

benhormann:
The bootstrap issue is because my fixup PR hasn't been merged. 😞

vfrinken:
the usecase for the  functions whose names include `[]` are functions generates using the Control.App monad. They actually include `[]`

vfrinken:
this is an error message of the c compiler 
```
build/exec/my_app_test.c:4279:69: error: use of undeclared identifier 'l__eq__dollarl'
Value *Control_App__lt_star_gt_Applicative__parenOpen_parenOpenApp_[l__eq__dollarl]_parenClose__dollares_parenClose_0
```

gallais:
Oof. That's a horrible name.

vfrinken:
There is also a problem with IORef in the RefC backend. I will create a different pull request that only focuses on that (with tests) and not conflate it with the Racket problem.

PR #2422, opened 2022-04-20
[ ci ] use katla to build html doc of the libs

gallais:


libs/test/test.ipkg
```diff
@@ -3,4 +3,6 @@ version = 0.5.1
 
 opts = "--ignore-missing-ipkg"
 
+depends = contrib
```
buzden, line 6: Code of a lib does not change, code of lib's dependencies does not change, but dependencies change. How so? Was it intentional?

gallais: The dependency was missing.

PR #2423, opened 2022-04-20
[ fix ] highlight unambiguous names in `with'

gallais:


src/TTImp/TTImp.idr
```diff
@@ -123,7 +123,7 @@ mutual
        Implicit : FC -> (bindIfUnsolved : Bool) -> RawImp' nm
 
        -- with-disambiguation
-       IWithUnambigNames : FC -> List Name -> RawImp' nm -> RawImp' nm
+       IWithUnambigNames : FC -> List (FC, Name) -> RawImp' nm -> RawImp' nm
```
buzden, line 126: Shouldn't `Language.Reflection.TTImp` of `base` be changed accordingly?

gallais: I also forget to update `Language.Reflection.TTImp` when I modified `with`. :x

Issue #2424, opened 2022-04-20
RefC backend IORef in Control.App monad


vfrinken:
The IORef information is stored in the Value_World struct in the RefC backend. However, using the Control.App monad, this world struct is not passed along

# Steps to Reproduce
```
data MyStateName : Type where

fct_b : Has [PrimIO, State MyStateName Nat] es => App es ()
fct_b  = do
    g <- get MyStateName
    primIO $ putStrLn $ "value is " ++ (show g)
    pure ()


fct_a : Has [PrimIO] es => App es ()
fct_a = new 42 fct_b


main : IO ()
main = run fct_a
```

# Expected Behavior
Output "value is 42"

# Observed Behavior
Using the RefC backend, this results in a Segmentation Fault

I am going to create a PR implementing a global IORef storage, not one that is associated with the world struct.

PR #2425, opened 2022-04-21
RefC backend improvements

vfrinken:
1. OnCollect had the wrong number of arguments. The code creator expects
   3 arguments, but onCollect in prim.h expected 4 arguments. The first of which
   was an erased arguments. That is now fixed.

2. OnCollect did not call `newReference` when creating a new reference to the pointer
   and the freeing function

3. OnCollect and OnCollectAny still had a spurious printf statement

Those issues have been fixed, the test case can be found in
 ` tests/refc/garbageCollect`

4. [issue #2424 ] 
   The IORef mechanism expects that the %World token will be 
   passed around consistently. This is not the case. States in Control.App make 
   use of IORefs, but the function 
   `Value *Control_App_prim_app_bind(...)` created from Control.App.prim_app_bind
    had the world token erased to NULL.
```C
Value *Control_App_prim_app_bind
(
    Value * var_0
,   Value * var_1
,   Value * var_2
)
{
    Value * var_3 = apply_closure(var_0, var_2);             // Control.App:69:1--69:75
    Value * var_4 = apply_closure(var_1, var_3);             // Control.App:69:1--69:75
    Value *returnValue = tailcall_apply_closure(var_4, NULL);   <------------- HERE SHOULD BE var_2
    removeReference(var_4);
    removeReference(var_3);
    return returnValue;
}
```
   Now, IORefs are managed using a global variable,
```C
IORef_Storage * global_IORef_Storage;
```
   referenced in cBackend.h, defined in the created .c file, and set to NULL
   in main();

5. While multithreading and forking is still not supported, compiling a program
   that makes use of Control.App demands a C implementation of prim_fork.
   Files support/refc/threads.c and support/refc/threads.h provide a
   dummy implementation for it, so that Control.App programs compile and run.

A test for these 2 issues is given in `tests/refc/issue2424`


src/Compiler/RefC/RefC.idr
```diff
@@ -42,6 +42,8 @@ showcCleanStringChar '(' = ("_parenOpen" ++)
 showcCleanStringChar ')' = ("_parenClose" ++)
 showcCleanStringChar '{' = ("_braceOpen" ++)
 showcCleanStringChar '}' = ("_braceClose" ++)
+showcCleanStringChar '[' = ("_braketOpen" ++)
+showcCleanStringChar ']' = ("_braketClose" ++)
```
edwinb, line 46: This will bother me if I don't say anything even though they don't affect functionality - I guess these should be "bracket" rather than "braket"!

edwinb:
I've looked over this briefly, and it looks like a reasonable fix for the existing issues. On `%World` I spoke to @vfrinken about this yesterday about the intended use where it is erased to `NULL` and I believe this patch will fix it (and the tests support that). So if nobody shouts soon, I'll merge this.

vfrinken:
fixed the spelling mistake

PR #2433, opened 2022-04-25
Allow functions to be marked for foreign export

edwinb:
This relies on the backend knowing what to do with such things, but the general idea is to mark them with '%export "backend:exportedname"' then 'getCompileDataWith', given a back end, will search for every function that needs to be exported, as well as every function starting from the expression to be compiled. This will allow Idris functions to be called from other languages, where a backend supports it.

This is hard to set up a test case for, for the moment, since no backends actually do anything with it. So consider it a bit of a placeholder for now.

(For @vfrinken, at least initially!)


buzden:
I think, you didn't update `Eq FnOpt` when it should be done. I made a PR to your branch if you don't want to think about it much.

edwinb:
@buzden Oh yes, I missed it in Reflection didn't I. Thanks.

edwinb:
What is more fiddly is how to get the elab reflection utils test to work again, since it needs to be in sync with the reflected structures here. What's the standard approach here?

gallais:
>  What's the standard approach here?

Open a PR on https://github.com/stefan-hoeck/idris2-elab-util and merge it (even though it'll temporarily
break the build there) and then re-run the failed CI here to make sure it's fine to merge.

stefan-hoeck:
> What is more fiddly is how to get the elab reflection utils test to work again, since it needs to be in sync with the reflected structures here. What's the standard approach here?

You make a PR to elab-util with the necessary changes (or I do that, if you tell me the necessary steps). After I merged the changes, you rerun this PR's CI. @gallais , please correct my if I forgot something.

stefan-hoeck:
@edwinb I just cloned your github branch for this PR. If this is ready for merging I'll submit a PR to elab-util and merge it there.

edwinb:
Thanks all! Looks like @stefan-hoeck has sorted it, cheers. I think this is ready for merging (we can't really get to the interesting bits of some external backends without it).

stefan-hoeck:
PR at elab-util merged. You can rerun CI now.

edwinb:
Great, everything looks good now, thanks everyone.

PR #2438, opened 2022-04-27
[ debug ] Log the value at the "Can't reify" error

buzden:
I'm really not sure this is the right way to do and I'm not sure this should be merged as is, but I didn't find any way better.

"Can't reify as ..." erorrs are really unheplful since they do not show any clue on where the error is happening. The most frequent case I'm running into this error is when I'm trying to pretty-print some expression from an elaboration script and the expression that is meant to be finally a string does not reduce to a string, mainly because of an unreported bug that appears when a function is used from a module that is imported without `public`.

To have at least some clue on what possibly can go wrong, I suggest to log an expression that causes this error. We can't use `Show` on `NF` because it is usually a non-informative string like `"$resolved653"` and we can't include the `Show` of quoted term to the original error output because of it's potential size. Maybe, it's okay for a log message on the 10th level.

Anyway, the way I did it is through an addition of an auto implicit `Env Term var` variable needed for logging to the `Reify` interface which is probably no beautiful. However, no calling code should have been changed since such variable was always present and was always unique.


gallais:
I'd use [`mkEnv`](https://github.com/idris-lang/Idris2/blob/main/src/Core/Env.idr#L257) instead

buzden:
> I'd use [`mkEnv`](https://github.com/idris-lang/Idris2/blob/main/src/Core/Env.idr#L257) instead

Would an empty list be okay for passing to this function? I don't understand, since the given `NF`, I think, might be a not closed expression.

src/Core/Reflect.idr
```diff
@@ -68,9 +69,10 @@ reflectionttimp : String -> Name
 reflectionttimp n = NS reflectionTTImpNS (UN $ Basic n)
 
 export
-cantReify : NF vars -> String -> Core a
-cantReify val ty
-    = throw (GenericMsg (getLoc val) ("Can't reify as " ++ ty))
+cantReify : Ref Ctxt Defs => {vars : _} -> Env Term vars => NF vars -> String -> Core a
+cantReify val ty = do
+  logNF "reflection.reify" 10 "Can't reify as \{ty}" %search val
+  throw (GenericMsg (getLoc val) ("Can't reify as " ++ ty))
```
gallais, line 75:
```suggestion
cantReify : Ref Ctxt Defs => {vars : _} -> NF vars -> String -> Core a
cantReify val ty = do
  logNF "reflection.reify" 10 "Can't reify as \{ty}" (mkEnv emptyFC _) val
  throw (GenericMsg (getLoc val) ("Can't reify as " ++ ty))
```

buzden: Oh, I got lost in types, `vars` parameter determines everything. Thank you!

PR #2441, opened 2022-04-28
Remove .d files from c backends on make clean

berewt:
At the moment `make clean` on the main projects doesn't remove `.d` files from the `support` subdirectories.
It's problematic, as a classic `make clean install` would fail if we change dependencies in `c` backends.


berewt:
Am I able to make a typo on a 2 line edit? Yes. Fixing it...

PR #2442, opened 2022-04-28
Better parse error for missing paren after with

dunhamsteve:
A user on discourse was missing parens after `with` and couldn't figure out the issue from the compiler error message.

This PR changes the error message to:

```
1/1: Building ParseWith (ParseWith.idr)
Error: Expected '('.

ParseWith:2:12--2:13
 1 | foo : Nat -> Nat
 2 | foo a with a
                ^
```


gallais:
Thanks!

PR #2453, opened 2022-05-03
[ fix ] Add missing cases for BITS types in `extractInt`

### LINKED ISSUE #2452: RefC bad implementation for switch on Bits types
# Steps to Reproduce
```
x : Bits8
x = 0

main : IO ()
main = do
  putStrLn $
    case x of
      0 => "good"
      _ => "bad"
```

# Expected Behavior
`good`

# Observed Behavior
`bad`

# Notes
The `BITS` cases are missing in this function: https://github.com/idris-lang/Idris2/blob/4a761ea5235d8f4838fcf49407b5f44af2e29de1/support/refc/runtime.c#L81

#2446 seems closely related

I think what's most probable is that #2447 just forgot to add the new cases in the `extractInt` function. @j-nava, could you maybe confirm that you can also reproduce this issue?

I plan on making a PR shortly, unless someone beats me to it.

kuruczgy:
Fixes #2452

Note: Please review commit-by-commit, because I had to `clang-format` the whole file, and you won't easily see the actual changes in the unified diff.


buzden:
I think a test that was failing is valuable to be added

kuruczgy:
done :+1:

gallais:
Thank you!

Issue #2462, opened 2022-05-07
Runtime error calling floating-point math functions on chez-scheme


j-nava:
Calling floating-point functions such as `floor`, `sin`, `cos`, etc. may result in a runtime error depending on the internal representation of `Double` values on chez-scheme. 
This issue is related to #2362 and #2261. 

# Steps to Reproduce
```haskell
main : IO ()
main = printLn $ floor (cast "42")
```

# Expected Behavior
```
42.0
```

# Observed Behavior
```
Exception in flfloor: 42 is not a flonum
```

# Discussion
Casting strings to Double may result in exact or inexact numbers. We currently use the specific `flonums` versions of the math functions: `flfloor`, `flsin`, `flcos`, etc. and a runtime error is issued by scheme if those functions are not called with `flonums`. 

One possible fix for this could be ensuring that we always cast strings to `flonums`. But that raises other questions: 
- If a `flonum` changes back to `fixnum` internally in chez scheme as a result of other operations, subsequent calls to math functions will crash.
- Keeping numbers as `fixnums` whenever possible will likely result in better performance.

I propose that we replace the `fl` version of the math functions for the regular ones that work with all numbers, and add a `cast-double-string` function to the support library to handle presentation issues such as #2261 and #2362.

j-nava:
Apparently what I proposed would bring back #1200.

Issue #2261, opened 2022-01-13
Possible error in String representation of infinity


joelberkeley:
I asked on discord about what the `.0` in the string rendering of `+inf.0` is, and it was suggested it may be a bug.

# Steps to Reproduce
```
Main> the Double (1.0 / 0.0)
+inf.0
Main> the Double (-1.0 / 0.0)                    
-inf.0
```

# Expected Behavior
```
Main> the Double (1.0 / 0.0)
+inf
Main> the Double (-1.0 / 0.0)                    
-inf
```

bkomuves:
NaN behaves the same too:

```
Main> 0.0/0.0
+nan.0
```
(I would argue that here the `+` sign is a bug, too)

buzden:
> but the function responsible `prim__cast_DoubleString` doesn't appear to be in the code base. I guess there's some magic happening

This function is a primitive operation over a primitive type, thus managed by each codegen separately. I believe, [this](https://github.com/idris-lang/Idris2/blob/3ef46f639ee503583f26d9b559d317a8946ea4da/support/chez/ct-support.ss#L285) is for chez backend, [this](https://github.com/idris-lang/Idris2/blob/3ef46f639ee503583f26d9b559d317a8946ea4da/support/racket/ct-support.rkt#L285) for the racket one, [this](https://github.com/idris-lang/Idris2/blob/3ef46f639ee503583f26d9b559d317a8946ea4da/support/refc/casts.c#L378) for refc backend, and etc.

joelberkeley:
thanks @buzden . That looks like more effort than I'm up for, not knowing chez or racket

dunhamsteve:
The `+inf.0` round trips back to the right value and `+inf` does not.   
```
Main> the Double $ cast "+inf"
0
Main> the Double $ cast "+inf.0"
+inf.0
```

This appears to be backend specific.  The node gives `Infinity`, which round trips with the node backend, and the refc backend gives "inf" which round trips. Also, the value 2.0 will render as 2.0 in chez, 2 in node, and 2.000000 in refc.  

The question here might be does Idris want to let the backends do their native thing or should it make them behave uniformly. I could see arguments both ways: simplicity and compatibility with the backend ecosystem vs consistency and compatibility with other Idris backends.

My test program:

```
main : IO ()
main = do
    let foo = show (1.0/0)
    printLn foo
    printLn $ the Double (cast foo)
    printLn 2.0
```

PR #2465, opened 2022-05-08
[ fix ] Improve Double consistency when casting from String on Scheme backends

### LINKED ISSUE #2362: Cast from `String` to `Double` inconsistent on refc vs chez-scheme

# Steps to Reproduce
```idris
zero : Double
zero = cast "0"

zero_zero : Double
zero_zero = cast "0.0"

main : IO ()
main = do
    printLn $ zero == zero_zero
    printLn $ show zero == show zero_zero
```
Compile with refc or chez/racket backend

# Expected Behavior
On refc or nodejs:
```
True
True
```
As I'd expect
# Observed Behavior
On chez and racket:
```
True
False
```

# Possible fixes
1 - use a different scheme function to cast strings to floats
2 - add `0.0` to the output to make convince scheme it is a float

j-nava:
Closes #2362 and #2462. 
This PR fixes a side effect of #1203 by adding a `exact->inexact` call when casting from String to Double. 

The "cast" tests are based on the turned off test from #1203 but mostly checks for crashes, not correctness, because of this: https://github.com/idris-lang/Idris2/pull/1203#issuecomment-801831713.


gallais (commented):
Nice!

Rather than adding a chez and a racket test, can you not add a single test to the
`idrisTestsAllBackends` target?

j-nava:
> Nice!
> 
> 
> 
> Rather than adding a chez and a racket test, can you not add a single test to the
> 
> `idrisTestsAllBackends` target? This way we would ensure they both share the same
> 
> golden file!

Unfortunately the test output will differ from Node. There's a little bit of a discussion here #2261. 

Do you think it'd be worth adding something similar to "test all backends" but only for chez+racket?

gallais (approved):

PR #2469, opened 2022-05-11
make `depends` collect all transitive dependencies

### LINKED ISSUE #2464: depends should automatically add transitive dependencies
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).

## Summary

Idris2 should add dependencies of any dependencies specified with `depends` in an `.ipkg` file.
Overall is should add all transitive dependencies of the package you manually add to the `.ipkg` file.

This can also be implemented for any packages loaded with `-p my-package`, however for the rest of this proposal I will only mention `depends`

## Motivation

Here is a sample from a package I am writing
```
depends = hashable, hedgehog, sop, elab-util, contrib, pretty-show
```
I only want `hashable` and `hedgehog` but I had to add `sop`, `elab-util`, `contrib` and `pretty-show`
I could only add these packages because I already knew what the packages were, however other users might not.

The current system is definitely not scalable to many packages, each with many transitive dependencies

## The proposal

For each package in the `depends` field of an ipkg, the `depends` of that package should be added.
This should happen recursively, so that any transitive dependency is added.

### Examples
```
depends = hashable, hedgehog, sop, elab-util, contrib, pretty-show
```
could become
```
depends = hashable, hedgehog
```
### Technical implementation

The `.ipkg` file of each package (or part of it) would have to be installed alongside the `ttc` files.
Then the transitive dependencies could be collected, and added in the same way packages currently are.

## Alternatives considered

None

## Conclusion

This will make immediately make [pack](https://github.com/stefan-hoeck/idris2-pack) more usable (see https://github.com/stefan-hoeck/idris2-pack/issues/32)

Z-snails:
This happens by installing the (modified) ipkg file along with ttc files

~~I can't think immediately of how to make a test for this, as in tests a custom version of idris is used with packages located in a non-standard place~~
added tests

fixes #2464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -139,6 +139,7 @@
 ### Other changes
 
 * Adds docstrings for the lambda-lifted IR.
+* Package files are now installed along-side build artifacts for installed packages. This means all transitive dependencies of packages you specify with the `depends` field are added automatically.
 
 ## v0.5.0/0.5.1
 
--- a/src/Core/Options/Log.idr
+++ b/src/Core/Options/Log.idr
@@ -129,6 +129,7 @@ knownTopics = [
     ("metadata.names", Nothing),
     ("module", Nothing),
     ("module.hash", Nothing),
+    ("package.depends", Just "Log which packages are being added"),
     ("quantity", Nothing),
     ("quantity.hole", Nothing),
     ("quantity.hole.update", Nothing),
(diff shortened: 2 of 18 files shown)


buzden:
This is great! Maybe, it's worth adding a line of documentation about this or at least adding this to a changelog.

By the way, am I write that this change makes it easy to have additional dirs with the semantics of `depends` (e.g., to have the same `depends` dir for several tests) in `ipkg` or to have a command-line option for additional `depends` dirs?

Z-snails:
I have no idea how to fix that windows test, someone please help

buzden:
> This doesn't currently affect the `depends` directory, only the `depends` field in ipkg files. It may help with that in the future, I can't immediately see how though (let me know though if I've completely misunderstood you)

Yeah, I wasn't talking about your PR per se, I was asking about possible future change based on your change. As far as I understand, options data structure was not supporting several `depends` dirs and your PR adds this support. This can be used in the future by e.g. an additional command line option for adding more `depends` dirs for particular runs.

PR #2471, opened 2022-05-12
add `scanr` and `scanr1` for `Vect`

joelberkeley:


libs/base/Data/Vect.idr
```diff
@@ -425,6 +425,32 @@ foldl1 f (x::xs) = foldl f x xs
...
+||| It assumes the last element of the vector to be the starting value and then
+||| starts the fold with the element preceding it.
+|||
+||| ```idris example
+||| scanr1 (-) (fromList [1,2,3])
+||| ```
+public export
+scanr1 : (elem -> elem -> elem) -> Vect len elem -> Vect len elem
+scanr1 _ [] = []
+scanr1 f xs@(_ :: _) = scanr f (last xs) (init xs)
```
joelberkeley, line 452: snocVect would be ideal here, but doesn't exist

PR #2472, opened 2022-05-12
Fix the allbackends/evaluator005 test

dunhamsteve:
The existing allbackends/evaluator005 relies on `show` which varies in format per backend. Even the value `0` differs in appearance, so `show` is not going to be effective for testing float values.

Additionally, the values themselves vary slightly between backends on the same system.  When I tried pulling the bits64 of the values, I found that the last few bits differed for some of the numeric values. Also some internal bits of NaN or Inf varied between backends on the same system.

Reading #1200, it appears the intent of the test was to check for appropriate values for NaN, +Inf, -Inf.  So I took the approach of detecting those exceptional values and scaling the normal values by 100000, printing as an integer.


dunhamsteve:
I'm gonna pull this PR. It makes the test work, but I'm not sure the test is accomplishing anything useful, and it required some awkwardness to make it work.

Issue #1282, opened 2021-04-07
Parameters not substituted under `if` when displaying types


fredrikNordvallForsberg:
Instantiated parameters under an `if` seem to be displayed as their uninstantiated variable name. (This seems to be a display problem only -- definitions making use if the parameters typecheck -- but it means basically programming blind in these situations.)

# Steps to Reproduce

Load the following file in the repl and do `:t f`

```idris
0 Alias : Type -> Type
Alias a = (b : Bool) -> if b then a else a

foo : Alias ()
foo = ?f
```

# Expected Behavior

To get output 
```
Main.f : (b : Bool) -> (if b then () else ())
```

# Observed Behavior

Actual output is
```
Main.f : (b : Bool) -> (if b then a else a)
```
Note the variable name `a` rather than its instantiation `()`.

gallais:
Same issue without the `if ... then ... else ...` sugar:

```idris
0 Alias2 : Type -> Type
Alias2 a = (b : Bool) -> case b of
  True => a
  False => a

foo2 : Alias2 ()
foo2 = ?f2
```

The issue does not show up if the `case` is not actually matching anything
(i.e. using `case b of x => a` instead).

gallais:
Pretty sure `unelabCase` is the culprit here:

```idris
LOG unelab.case:20: Unelaborating case ("Alias", 356)
with arguments: [Builtin.Unit, b]
LOG unelab.case:20: Unelaborated to: (%case (b : ?) [Prelude.True = a, Prelude.False = a])
```

PR #2481, opened 2022-05-16
[ re #1282 ] Improve unelaboration of case expressions

kuruczgy:
This improves the situation with #1282, but I wouldn't consider it a complete fix.

The general idea is to make `unelabCase` take TT instead of TTImp terms, because replacing variables in TT is much easier. The code still turned out a bit hacky IMO. (And also it's my first time looking deeper into the idris internals, so I may or may not know what I am doing.) But according to the tests it works fairly well (I only had to fix up 2 test cases).

### DIFF
--- a/src/TTImp/Unelab.idr
+++ b/src/TTImp/Unelab.idr
@@ -46,7 +46,7 @@ mutual
   ||| Unelaborate a call to a case expression as an inline case.
   ||| This should allow us to eventurally resugar case blocks and if-then-else calls.
   |||
-  ||| This is really hard however because all we have access to is the elaborated
+  ||| This is really hard however because all we have access to is the
   ||| clauses of the lifted case expression. So e.g.
   |||      f x = case g x of p -> e
   ||| became
@@ -63,21 +63,24 @@ mutual
   ||| name x.
   |||
   ||| We will try to do our best...
-  unelabCase : {auto c : Ref Ctxt Defs} ->
+  unelabCase : {vars : _} ->
+               {auto c : Ref Ctxt Defs} ->
                List (Name, Nat) ->
-               Name -> List IArg -> IRawImp ->
-               Core IRawImp
-  unelabCase nest n args orig
+               Env Term vars ->
+               Name ->
+               List (Term vars) ->
+               Core (Maybe IRawImp)
+  unelabCase nest env n args
       = do defs <- get Ctxt
            Just glob <- lookupCtxtExact n (gamma defs)
-                | Nothing => pure orig
+                | Nothing => pure Nothing
            let PMDef _ pargs treect _ pats = definition glob
-                | _ => pure orig
+                | _ => pure Nothing
            let Just argpos = findArgPos treect
-                | _ => pure orig
+                | _ => pure Nothing
            if length args == length pargs
-              then mkCase pats argpos 0 args
-              else pure orig
+              then mkCase pats argpos args
+              else pure Nothing
     where
       -- Need to find the position of the scrutinee to rebuild original
       -- case correctly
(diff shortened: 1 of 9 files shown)


tests/idris2/debug001/expected
```diff
@@ -66,7 +66,7 @@ Main> Prelude.Types.SnocList.filter
 Arguments [{arg:0}, {arg:1}, {arg:2}]
 Compile time tree: case {arg:2} of
   Lin {e:0} => [<]
-  (:<) {e:1} {e:2} {e:3} => let rest = filter {arg:1} {e:2} in if {arg:1} {e:3} then rest :< x else rest
+  (:<) {e:1} {e:2} {e:3} => let rest = filter {arg:1} {e:2} in if {arg:1} {e:3} then rest :< {e:3} else rest
```
kuruczgy, line 69: This seems like a clear improvement to me.

tests/idris2/interactive017/expected
```diff
@@ -1,4 +1,4 @@
 1/1: Building RLE (RLE.idr)
 RLE> uncompress Empty = Val []
-uncompress (Run n x rest) = let Val ys = uncompress rest in Val (rep n x ++ ys)
+uncompress (Run n x rest) = let Val ys = uncompress rest in Val (rep n x ++ _)
```
kuruczgy, line 3: Not quite sure about this one. From what I can tell the term coming from the expression search contains an underscore as an argument to the case function, so this is the correct unelaboration, and the fact that the ys was correct there is just a coincidence. (But I have no idea how the expression search works, so I might be wrong.)

src/TTImp/Unelab.idr
```diff
@@ -63,98 +63,111 @@ mutual
...
       findArgPos _ = Nothing
 
-      idxOrDefault : Nat -> a -> List a -> a
-      idxOrDefault Z def (x :: _) = x
-      idxOrDefault (S k) def (_ :: xs) = idxOrDefault k def xs
-      idxOrDefault _ def [] = def
+      idxOrMaybe : Nat -> List a -> Maybe a
+      idxOrMaybe Z (x :: _) = Just x
+      idxOrMaybe (S k) (_ :: xs) = idxOrMaybe k xs
+      idxOrMaybe _ [] = Nothing
```
kuruczgy, line 94: Why don't we have something like these in the prelude btw?

gallais: Because we have to wait 1 release cycle to use it in the compiler and so that usually discourages us
from putting (a generalised version of) the function in the lib instead of just dumping it in the file.

src/TTImp/Unelab.idr
```diff
@@ -63,98 +63,111 @@ mutual
...
 
-      getNth : Nat -> Term vars -> Term vars
-      getNth n tm
-          = case getFnArgs tm of
-                 (f, as) => idxOrDefault n f as
+      -- TODO: some utility like this should probably be implemented in Core
+      substVar : List (Var vs) -> Term vs -> Term vs -> Term vs
```
gallais, line 97: Can't we get
```
substVar : List (Var vs, Term vs) -> Term vs -> Term vs
```
directly rather than using `foldl` on the single substVar with the repeated traversals this implies?

src/TTImp/Unelab.idr
```diff
@@ -63,98 +63,111 @@ mutual
...
+          substs' = map (bimap (map (embedVar {more = vars})) (weakenNs p)) substs
+        in
+          foldl (\tm, (x, new) => substVar x new tm) tm' substs'
+
+      argVars : {vs : _} -> Term vs -> List (Var vs)
+      argVars (As _ _ as pat) = argVars as ++ argVars pat
+      argVars (Local _ _ _ p) = [MkVar p]
+      argVars _ = []
+
+      concatEnv : {vs : _} -> Env Term vs -> Env Term vars -> Env Term (vs ++ vars)
```
gallais, line 132: Please move this to `Core.Env` and call it `(++)`.

gallais (commented):
Finally got around to having a proper look. Some small comments.

src/TTImp/Unelab.idr
kuruczgy: done

gallais:
I have this suggested change but you unchecked the "allow maintainers to edit the PR" so I can't push it to your branch:
```diff
diff --git a/src/TTImp/Unelab.idr b/src/TTImp/Unelab.idr
index 0270ad419..629392c87 100644
--- a/src/TTImp/Unelab.idr
+++ b/src/TTImp/Unelab.idr
@@ -94,15 +94,15 @@ mutual
       idxOrMaybe _ [] = Nothing
 
       -- TODO: some utility like this should probably be implemented in Core
-      substVars : List (Var vs, Term vs) -> Term vs -> Term vs
+      substVars : List (List (Var vs), Term vs) -> Term vs -> Term vs
       substVars xs tm@(Local fc _ idx prf)
-          = case find (\((MkVar {i} _), _) => i == idx) xs of
+          = case find (any ((idx ==) . varIdx) . fst) xs of
                  Just (_, new) => new
                  Nothing => tm
       substVars xs (Meta fc n i args)
           = Meta fc n i (map (substVars xs) args)
       substVars xs (Bind fc y b scope)
-          = Bind fc y (map (substVars xs) b) (substVars (map (bimap weaken weaken) xs) scope)
+          = Bind fc y (map (substVars xs) b) (substVars (map (bimap (map weaken) weaken) xs) scope)
       substVars xs (App fc fn arg)
           = App fc (substVars xs fn) (substVars xs arg)
       substVars xs (As fc s as pat)
@@ -121,8 +121,7 @@ mutual
       substArgs : SizeOf vs -> List (List (Var vs), Term vars) -> Term vs -> Term (vs ++ vars)
       substArgs p substs tm =
         let
-          substs' = map (bimap (embedVar {more = vars}) (weakenNs p)) $
-            foldr (++) [] $ map (\(vars, tm) => map (\v => (v, tm)) vars) substs
+          substs' = map (bimap (map $ embedVar {more = vars}) (weakenNs p)) substs
           tm' = embed tm
         in
           substVars substs' tm'
```

kuruczgy:
> I have this suggested change but you unchecked the "allow maintainers to edit the PR" so I can't push it to your branch:

Thanks, I applied it, and also checked the allow edits option.

PR #2496, opened 2022-05-22
Transitive dependencies v2

Z-snails:
See #2469 for initial idea
Hopefully actually installing the version will make pack work properly (pinging @stefan-hoeck)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,6 +141,7 @@
 ### Other changes
 
 * Adds docstrings for the lambda-lifted IR.
+* Package files are now installed along-side build artifacts for installed packages. This means all transitive dependencies of packages you specify with the `depends` field are added automatically.
 
 ## v0.5.0/0.5.1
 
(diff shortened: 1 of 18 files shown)


stefan-hoeck:
I'll modify *pack*, so that it can install Idris from a different repo, then I'll test your branch against the current db. Will report back here.

stefan-hoeck:
Look like all works now, with the exception of katla, which no longer builds with this. Not sure, how this is related.

Z-snails:
katla builds in CI, so something must be going wrong. Is your modified pack on github? If so I can (try and) do some debugging this coming week.

What is the error it's giving?

stefan-hoeck:
OK, the reason is that your PR branch is 25 commits behind Idris2 main, and I used commit [77c2c38](https://github.com/Z-snails/Idris2/commit/77c2c38d6a52409b7484c68e3aac1e4778dab190) of that branch in my tests. Some of these changes must have affected katla directly. If you like, you can rebase your branch and I can rerun the tests.

The version of pack I'm using for this is in a PR but it still requires one or two additions before I merge it.

gallais:
> Some of these changes must have affected katla directly

During IDM I did push changes to main to make Katla's new markdown & latex literate backends work.

stefan-hoeck:
Today's test was successful and all packages could be built with  the latest commit.

PR #2497, opened 2022-05-23
List proofs

kasiaMarek:
Added following proofs to `contrib`:
 - correspondence between bind and concatenation
 - if x is an element of (xs ++ ys) it is an element of xs or ys

### DIFF
--- a/libs/contrib/Data/List/Equalities.idr
+++ b/libs/contrib/Data/List/Equalities.idr
@@ -1,8 +1,10 @@
 module Data.List.Equalities
 
 import Control.Function
+import Syntax.PreorderReasoning
 
 import Data.List
+import Data.List.Reverse
 
 %default total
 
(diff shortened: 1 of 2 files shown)


libs/contrib/Data/List/Equalities.idr
```diff
@@ -115,3 +117,33 @@ export
 appendNonEmptyRightNotEq : (zs, xs : List a) -> NonEmpty xs => Not (zs = zs ++ xs)
 appendNonEmptyRightNotEq []      (_::_)  Refl impossible
 appendNonEmptyRightNotEq (_::zs) (x::xs) prf = appendNonEmptyRightNotEq zs (x::xs) $ snd $ biinjective prf
+
+export
+bindOnEmptyList : (f : a -> List b) -> ([] >>= f) === []
+bindOnEmptyList f = Refl
```
ohad, line 123: Is this necessary, or merely for completeness' sake?
The equation holds judgementally, so it's probably easier to just use `Refl` instead of calling `bindOnEmptyList`.

libs/prelude/Prelude/Types.idr
```diff
@@ -490,13 +490,14 @@ Foldable List where
 
   foldMap f = foldl (\acc, elem => acc <+> f elem) neutral
 
+public export listBindOnto : (a -> List b) -> List b -> List a -> List b
+listBindOnto f xs []        = reverse xs
+listBindOnto f xs (y :: ys) = listBindOnto f (reverseOnto xs (f y)) ys
```
ohad, line 495: Follow style, seems to put access modifiers in separate lines.

libs/contrib/Data/List/Equalities.idr
```diff
@@ -115,3 +117,33 @@ export
...
+    ~~ ys ++ listBindOnto f (reverseOnto (reverse zs) (f x)) xs ... cong (\n => ys ++ listBindOnto f n xs) (sym $ reverseOntoSpec _ _)
+
+||| Proof of correspondence between list bind and concatenation.
+export
+bindConcatPrf : (xs: List a) -> (x: a) -> (f: a -> List b) -> ((x::xs >>= f) = (f x) ++ (xs >>= f))
+bindConcatPrf xs x f =
+  Calc $
+    |~ listBindOnto f ([] ++ reverse (f x)) xs
+    ~~ listBindOnto f (reverse [] ++ reverse (f x)) xs ... cong (\n => listBindOnto f (n ++ reverse (f x)) xs) (sym reverseNil)
+    ~~ f x ++ listBindOnto f [] xs ... listBindOntoPrf xs f [] (f x)
```
ohad, line 149: Break lines for narrower editors

libs/prelude/Prelude/Types.idr
```diff
@@ -490,13 +490,14 @@ Foldable List where
 
   foldMap f = foldl (\acc, elem => acc <+> f elem) neutral
 
+public export listBindOnto : (a -> List b) -> List b -> List a -> List b
+listBindOnto f xs []        = reverse xs
```
ohad, line 494: It would be natural to use a `SnocList a` for the accumulator and then chip the result:
```suggestion
listBindOnto f xs []        = xs <>> []
```
but I think we might need to move more stuff into `Prelude` from `base`, so maybe it's fine to keep as lists.

kasiaMarek: I think you mean to have `listBindOnto` return `SnocList` and use `<><`. Otherwise we iterate over the accumulator each time.

kasiaMarek: I think we just to need to change the order, so `listBindOnto` is defined after fish and chip (which are both in this file as well).

libs/contrib/Data/List/Elem/Extra.idr
```diff
@@ -44,3 +44,15 @@ notElemAppRight : (ys, xs : List a)
...
+public export
+hereOrThereConcat : (xs: List a)
+                  -> (ys: List a)
+                  -> (e `Elem` (xs ++ ys))
+                  -> Either (e `Elem` xs) (e `Elem` ys)
+hereOrThereConcat [] ys pos = Right pos
+hereOrThereConcat (e :: xs) ys Here = Left Here
+hereOrThereConcat (x :: xs) ys (There pos) = case hereOrThereConcat xs ys pos of
+                                              (Left pos) => Left $ There pos
+                                              (Right pos) => Right pos
```
gallais, line 58: Isn't this a duplicate of
```
elemAppLorR : (xs, ys : List a)
           -> (prf : Elem k (xs ++ ys))
           -> Either (Elem k xs) (Elem k ys)
```
a few lines above?

kasiaMarek: It is... Thanks for spotting this.

PR #2499, opened 2022-05-24
[ fix #2493 ] Put `%runElab` smartly in a call of `%macro`

### LINKED ISSUE #2493: [ macro ] Applying arguments to the result of macro in the same application series does not work
Have a macro that returns a function, e.g.:

```idris
import Language.Reflection

%language ElabReflection

%macro
coolMacro : Nat -> Elab (Nat -> Nat)
coolMacro n = lambda Nat $ \m => pure $ n + m + 1
```

and use it:

```idris
x : Nat
x = let z = coolMacro 4 in z 5
```

All typechecks and works.

# Steps to Reproduce

Now, inline the let:

```idris
x : Nat
x = coolMacro 4 5
```

# Expected Behavior

All works as previous.

# Observed Behavior

```
Error: While processing right hand side of x. When unifying:
    Elab (Nat -> Nat)
and:
    ?argTy -> Elab Nat
Mismatch between: Elab (Nat -> Nat) and ?argTy -> Elab Nat.

MacroRetFunc:13:5--13:18
 11 | 
 12 | x : Nat
 13 | x = coolMacro 4 5
          ^^^^^^^^^^^^^
```

I.e., `coolMacro 4 5` desugars to `%runElab <coolMacro> 4 5`, but should go to `(%runElab <coolMacro> 4) 5` (which works).

By the way, defining `x` as `(coolMacro 4) 5` does not help too, because this is literally the same definition as the former one.

buzden:
This should be a fully backward-compatible change. In all previously working cases elaboration should produce expressions with the same meaning. But when it fails with `%runElab macroF arg1 arg2 arg3`, it successively tries `(%runElab macroF arg1 arg2) arg3`, `(%runElab macroF arg1) arg2 arg3`, etc. The longest argument lists inside `%runElab` are taken because of backward-compatibility **and** possible dependent types in a type of macro function (see tests).

Fixes #2493

### DIFF
--- a/src/TTImp/Elab/Ambiguity.idr
+++ b/src/TTImp/Elab/Ambiguity.idr
@@ -124,24 +124,21 @@ expandAmbigName mode nest env orig args (IVar fc x) exp
             else IMustUnify fc NotConstructor tm
     wrapDot _ _ _ _ _ _ tm = tm
 
+    notLHS : ElabMode -> Bool
+    notLHS (InLHS _) = False
+    notLHS _ = True
 
     mkTerm : Bool -> EState vars -> Name -> GlobalDef -> RawImp
     mkTerm prim est n def
         = let tm = wrapDot prim est mode n (map (snd . snd) args)
                        (definition def) (buildAlt (IVar fc n) args) in
-              if Context.Macro `elem` flags def
-                 then case mode of
-                           InLHS _ => tm
-                           _ => IRunElab fc (ICoerced fc tm)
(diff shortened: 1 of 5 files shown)


gallais:
> possible dependent types in a type of macro function (see tests).

Fair enough. If we find out it's too slow in general, we can always come back and add a special
case using `underPis (-1)` to see whether the `IVar` corresponds to a function with an `Elab`-headed
return type and bypass the whole iterative search in that easier case.

PR #2505, opened 2022-05-26
[ prelude ] Add Range implementation for Char

xerz-one:
Hey there! This is my first time fiddling around with Idris, so I apologize if the PR is not up to expectations or requirements. Please request any necessary changes, and feel free to e.g. ask for an RFC or a rewrite if so needed.

This implementation of `Range Char` merely takes care of converting to `Int` in order to use the existing `Range Nat` implementation, then mapping back the results to `List Char`.


gallais:
Looks good. Thanks for the contribution!

PR #2511, opened 2022-05-28
[ IDE ] auto completions

gallais:
TODO:
* [x] support for pragmas
* [x] support for REPL commands
* [x] support for logging topics


gallais:
The current representation of REPL commands makes defining completion for their arguments kind of annoying.
I think I'll leave it to future work.

PR #2516, opened 2022-06-01
[ error ] Make failing `IAlternative` w/ `FirstSuccess` print all errors

buzden:
At the moment if all alternatives of `IAlternative` with `FirstSuccess` rule are failing, only the last error is printed, which can be misleading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 * Adds documentation for unquotes `~( )`.
 
+### Compiler changes
+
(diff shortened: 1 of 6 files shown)


CodingCellist:
Changelog update please  : )
And is this something we'd want to have a test for?

buzden:
> Changelog update please : )

Done.

> And is this something we'd want to have a test for?

That's a good idea, I've added one.

CodingCellist (approved):

PR #2517, opened 2022-06-02
[ re #2499 ] Fix possible bad error messages in result-polymorphic scripts

buzden:
While #2499 fixed one bug, it (together with behaviour described in #2516) introduced an unpleasant behaviour in a quite non-trivial case.

TLDR: *the logic of alternatives run from #2499 was tuned to work well when elaboration script fails; i.e. as previous, if at least one alternative succeeds, the first one is used; but if all alternatives fail _but elaboration script has started_, the first error of elaboration script run is printed; if elaboration script was not started at all, errors from all the alternatives are printed (usually, unification errors).*

Consider an macro elaboration script which is polymorphic on the result and which analyses its current goal type (this was the only way to get a type with 0-args as input before #1847) having also its own `auto` params, i.e.
```idris
%macro
m : Externals => Elab a
m = do
  Just g <- goal
    | Nothing => fail "no goal"
  -- analyse the goal, fail if it can't manage it, produce the result if can
  -- ...
```

Consider now that we are calling this macro with a particular `auto` parameter:
```idris
x : Nat
x = m @{ext}
```

If the `Nat` goal is not appropriate for the `m` macro (and it should fail), then alternative mechanism tries to apply `@{ext}` to the result of the elaboration script, in which case it cannot have any particular goal type yet. This but due to problem in #2516, it prints only the last error, i.e. no goal (a side note that before a fix in this PR, a no-goal situation was done as if goal is a `?scriptTy` hole, not nevertheless).

We could just print all the errors, as it is done in #2516, but there is a problem: the value of errors from inside the elaboration script is more than of unification errors saying "I couldn't run the elaboration script". So, if the elaboration script has started, we shouldn't print errors that were trying to start it, but didn't succeed (anyway, those alternatives are blind guesses with the only pupose: to start the elaboration script).

So, I suggest to distinguish all the errors from inside elaboration scripts and print only them if elaboration script started.

Then, probably, the most controversal part, I suggest to print only the first such because 1) if script is not polymorphic on its goal type, it probably would start only in one alternative anyway 2) if script analyses its goal type, there is anyway probably won't be any good goals for more than one alternatives.

---

Thia PR depends on #2516.

Also, it contains a little fix: `goal` function in the `Elaboration` interface returns `Maybe` because there may be no decided goal yet. However, the running code (I guess, for technical reasons and finally by mistake) replaces it with a hole with a `scriptTy` name. This goal is essential for unification with `Elab ?scriptTy` but shouldn't be passed inside the script (since, we have `Nothing` case for this already).

Oh, and wrapping all errors from inside the elaboration script changes error messages from elaboration scripts, but I think it is an improvement: previously errors outside the elaboration script with of its run attempt and from inside could look similarly. Now, even ordinary typechecking errors from inside the elaboration scripts are prepended with the "error during reflection" string.

Also, this PR requires a patch to `elab-utils`, since it adds an alternative to a pretty-printable data type, it will follow.

This PR contains all logically separate changes in separate commits, I hope this will ease the review.

### DIFF
--- a/libs/base/Language/Reflection/TTImp.idr
+++ b/libs/base/Language/Reflection/TTImp.idr
@@ -89,6 +89,7 @@ mutual
        FirstSuccess : AltType
        Unique : AltType
        UniqueDefault : TTImp -> AltType
+       FirstReflection : AltType
 
   public export
   data NoMangleDirective : Type where
@@ -362,6 +363,7 @@ parameters {auto eqTTImp : Eq TTImp}
     FirstSuccess    == FirstSuccess     = True
     Unique          == Unique           = True
     UniqueDefault t == UniqueDefault t' = t == t'
+    FirstReflection == FirstReflection  = True
     _ == _ = False
 
   public export
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -29,7 +29,7 @@ import public Libraries.Utils.Binary
 ||| version number if you're changing the version more than once in the same day.
 export
 ttcVersion : Int
-ttcVersion = 20220425 * 100 + 0
+ttcVersion = 20220531 * 100 + 0
 
 export
 checkTTCVersion : String -> Int -> Int -> Core ()
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -778,6 +778,7 @@ HasNames Error where
     = BadDotPattern fc <$> full gam rho <*> pure x <*> full gam s <*> full gam t
   full gam (BadImplicit fc x) = pure (BadImplicit fc x)
   full gam (BadRunElab fc rho s desc) = BadRunElab fc <$> full gam rho <*> full gam s <*> pure desc
+  full gam (RunElabFail e) = RunElabFail <$> full gam e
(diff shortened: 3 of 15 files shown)


buzden:
Something's wrong again with CI, as it fails with "service unavailable"...

gallais:
This feels like piling hacks on top of hacks.

Can we first go back to the elaboration of `%runElab` and use `underPis` to pick the correct number
of arguments in all the cases where the type is known statically?

buzden:
> This feels like piling hacks on top of hacks.
> 
> Can we first go back to the elaboration of `%runElab` and use `underPis` to pick the correct number of arguments in all the cases where the type is known statically?

Well, my answer will be not a direct answer to your question, because I don't really get what `underPis` is doing and how to use it. But I have an idea, which might be a better solution all-in-all.

`checkRunElab`, actually, does two things: it checks that given expression reduces to `Elab ?something` and "executes" it. We may actually change it to return the result of the first check and in case of a macro (we would need a boolean flag for this in `IRunElab`), we can cut applications in `checkTerm` (the main place of call for `checkRunElab`).

Does it look less hacky?

PR #2518, opened 2022-06-03
[ base ] Add mapProperty to Data.Vect.Quantifiers.All

ngeiswei:
such function is present in Data.List.Quantifiers.All but not in
Data.Vect.Quantifiers.All

### DIFF
--- a/libs/base/Data/Vect/Quantifiers.idr
+++ b/libs/base/Data/Vect/Quantifiers.idr
@@ -63,6 +63,12 @@ namespace All
     Nil  : All p Nil
     (::) : {0 xs : Vect n a} -> p x -> All p xs -> All p (x :: xs)
 
+  ||| Modify the property given a pointwise function
+  export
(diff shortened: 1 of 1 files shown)


ngeiswei:
Please note the following
1. `Data.List.Quantifiers.Any.mapProperty` also exists but is not ported to `Data.Vect.Quantifiers.Any` in that PR.  I only needed `All.mapProperty` and since it is my first real PR I though I'd go small since I'm still lacking confidence.
2. I tried to insert `All.mapProperty` in place as to maximize consistency between the order in which functions are written in `Data.List.Quantifiers` and `Data.Vect.Quantifiers`, the thing however is that there is no real consistent order that they both follow.
3. I did not add a unit test because other `mapProperty` implementation do not have any.  But that is obviously a poor excuse, if you think I should add a unit test, just let me know.

ngeiswei:
Oh, someone already ported it!!!  I'm closing that PR then.

PR #2525, opened 2022-06-07
Make help topics more flexible

locriacyber:
There are only two topics currently, but why not dream big?

This PR also remfinal newline when printing usage.

### DIFF
--- a/src/Idris/CommandLine/HelpTopic.idr
+++ b/src/Idris/CommandLine/HelpTopic.idr
@@ -1,0 +1,46 @@
+module Idris.CommandLine.HelpTopic
+
+import Core.Options.Log     -- loggingHelpMessage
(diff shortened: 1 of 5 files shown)


src/Idris/CommandLine/HelpTopic.idr
```diff
@@ -0,0 +1,46 @@
...
+||| Help topics
+public export
+record HelpTopic where
+  constructor MkHelpTopic
+  ||| Used to match topic in `--help [topic name]`
+  name        : String
+  ||| Printed with `--help topics`
+  description : String
+  ||| Content to print
+  content     : Inf String
```
gallais, line 19: Why the `Inf`?

locriacyber: Oh it should be `Lazy`

Issue #2513, opened 2022-05-29
Incorrect rejection of erased irrefutable patterns


ohad:
(spotted by locria)

# Steps to Reproduce

```idris
f : (0 x : (a,b)) -> Nat
f x =
  let 0 (u,v) = x in
  0
```

# Expected Behavior
Type-check fine, with `0 u : a` and `0 v : b` in scope inside the let body.

# Observed Behavior
```
     While processing right hand side of f. x is not accessible in this
     context.
     
      09 | f : (0 x : (a,b)) -> Nat
      10 | f x =
      11 |   let 0 (u,v) = x in                           ^
```

If we change `0` to `1`, this type-checks fine:
```idris
f : (1 x : (a,b)) -> Nat
f x =
  let 1 (u,v) = x in
  0
```
So you'd expect consistent behaviour between the two quantities.

locriacyber:
Here is a test case: https://github.com/locriacyber/idris2-sorted-ds/blob/05386b6f242432d21ea3ad1a6addf696b9403823/src/Data/BinaryTree.idr#L84

This should compile when the bug is fixed.

gallais:
The crux is here IMO:

https://github.com/idris-lang/Idris2/blob/d786b5d1533bf509ab867b2eb4bef03fb85e61e9/src/Idris/Desugar.idr#L219-L222

`PCase` does not carry a `RigCount` which means that you cannot state which quantity
to use to check the scrutinee hence the discarding of `rig`. Idris tries to guess instead and
inevitably will make mistakes.

Two possible avenues to a fix:
* [ ] Try to elaborate `let 0 pat = e in b` to `let 0 x = e in case x of pat => b` where `x` is a fresh machine name
       with the hope that Idris will adequately guess `0` now that the scrutinee is a variable known to be `0`.
* [ ] Add a `RigCount` field to `PCase` (and so probably to `ICase` too), propagate it here and use it
       during elaboration instead of guessing.

ohad:
@gallais @dunhamsteve  Does the`RigCount` on `ICase` need to be a `Maybe RigCount`?

We don't always know the ambient rigcount during desugaring, and maybe that means we need to keep guessing in those cases.

(But I haven't yet delved into the details of quantity checking, so maybe the elaborator can deal with that.)

dunhamsteve:
[...]

Issue #66, opened 2020-05-20
Can't unify an interface with its instance


edwinb:
<a href="https://github.com/berewt"><img src="https://avatars2.githubusercontent.com/u/1143168?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [berewt](https://github.com/berewt)**
_Monday Apr 27, 2020 at 12:46 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/330_

----

The compiler seems to struggle to unify interfaces with dependent types with their implementation.

# Steps to Reproduce

try to compile this module:

```
module Test

data Dummy : (lbl : Type) -> (st : List lbl) -> Type -> Type where
  MkDummy : a -> Dummy b st a


public export
interface IxPure (f : (lbl : Type) -> (st : List lbl) -> Type -> Type)
          where
  ixPure : a -> f lbl st a

IxPure Dummy where
  ixPure = MkDummy

pure12 : Dummy String xs Integer
pure12 = ixPure $ 12
```

Fails with:

# Expected Behavior

It should compile.

# Observed Behavior

It fails with:

```
Test.idr:16:10--17:1:While processing right hand side of pure12 at Test.idr:16:1--17:1:
Can't find an implementation for IxPure ?f
```

# Workarounds

Specifying explicitly the value of `f` in `pure12` solve the issue:

```
pure12 : Dummy String xs Integer
pure12 = ixPure {f = Dummy} $ 12
```

And the problem disappears as well if we remove the `st` parameters.

edwinb:
Poking around inside here, the problem is that it can't solve the unification problem `?f ?lbl ?st = Dummy String xs`, and telling it any one of `f`, `lbl` or `st` resolves it. It's supposed to notice that `f` is an argument to an implementation so should be treated as invertible, but somehow it seems that isn't happening...

edwinb:
Ah, it's using the conversion check (which doesn't solve holes) for arguments to invertible holes, not unification, which would resolve the holes `lbl` and `st`. I believe I have a fix.

Issue #39, opened 2020-05-20
AddClause on an interface implementation adds only one clause


edwinb:
<a href="https://github.com/petithug"><img src="https://avatars3.githubusercontent.com/u/2779694?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [petithug](https://github.com/petithug)**
_Sunday Mar 29, 2020 at 18:10 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/248_

----

AddClause on an interface implementati9on should add all the clauses.

See also #247.
Loading the following in the REPL and executing ":ac 4 Show"

```
data Blah : Type where
  MkBlah : Blah

Show Blah where
```
  
# Expected Behavior

```
data Blah : Type where
  MkBlah : Blah

Show Blah where
  showPrec d x = ?showPrec_rhs
  show x = ?show_rhs
```

# Observed Behavior

```
data Blah : Type where
  MkBlah : Blah

Show Blah where
showPrec d x = ?showPrec_rhs
```

PR #2535, opened 2022-06-11
Issue 2513 - handle erased, irrefutable lets

dunhamsteve:
This PR addresses #2513 by adding a rig to ICase. It includes a test case named `idris2/linear016`. 

It needs to be double checked by someone familiar with elaboration of case statements.

Note that it changes the shape of `ICase`, which could break some elaboration scripts (including the frex build). Should this be mentioned in the Changelog? What section / wording would be appropriate?


ohad:
Question for the review: Does the`RigCount` on `ICase` need to be a `Maybe RigCount`?

We don't always know the ambient rigcount during desugaring, and maybe that means we need to keep guessing in those cases.

(But I haven't yet delved into the details of quantity checking, so maybe the elaborator can deal with that.)

ohad:
I added tentative PR's for the two failing CI:
* [Frex #66](https://github.com/frex-project/idris-frex/pull/66)
* [Elab utils #39](https://github.com/stefan-hoeck/idris2-elab-util/pull/39

Both would need to be tweaked once this PR is merged, so we'll probably want to disable the frex CI for the duration of the merge into master, and disable the Elab utils CI for 1 day (until the nightly build updates).

dunhamsteve:
Yeah, double check the Maybe thing. I thought about a Maybe, but saw the code for `ILet` in `Binders.idr` that it just multiplied the optionally supplied rig with ambient. I think is equivalent to having a Nothing, since RigW is a multiplicative identity.  This is kinda how things work everwhere, right?  If the user doesn't say anything it parses as RigW, but we pick up ambient. It doesn't look like we have `Maybe RigCount` anywhere else in the code base aside from two helper functions. 

The other thing to review is that the code in `Case.idr:checkCase` and `Binders.idr:checkLet` is both complicated and almost identical, but `ILet` adds a `{preciseInf := True}` to the `elabinfo` in one spot and has an extra check instead of an error under the `branchOne`.

There is an additional test case from @ohad on discord that I'd like to look at later today. We solved the original error in that code with this PR, but it has another error in a slightly different spot now.

dunhamsteve:
It's interesting that this fails:
```idris
data Foo : Bool -> Type where
    MkInt    : Int -> Foo True
    MkString : String -> Foo False

blah : (0 f : Foo True) -> Int
blah f = let 0 (MkInt _) = f in 0
```
with 
```
1/1: Building issue2513c (issue2513c.idr)
Error: While processing right hand side of blah. @-patterns only allowed in pattern clauses

issue2513c:7:14--7:29
 3 |     MkString : String -> Foo False
 4 | 
 5 | 
 6 | blah : (0 f : Foo True) -> Int
 7 | blah f = let 0 (MkInt _) = f in 0
 ```
But this works:
```idris
data Bar : Bool -> Type where
    BarA : Bar True
    BarB : Bar False

blah : (0 f : Bar True) -> Int
blah f = let 0 BarA = f in 0
```
as does the original
```idris
f : (0 x : (a,b)) -> Nat
f x =
  let 0 (u,v) = x in
  0
```
And this works too:
```idris
data Foo : Bool -> Type where
    MkInt    : Int -> Foo True
    -- MkString : String -> Foo False

blah : (0 f : Foo True) -> Int
blah f = let 0 MkInt _ = f in 0
```

ohad:
I think the BarA example shadows the constructor  with a freshly bound variable.

dunhamsteve:
Yeah, I wasn't used to binding capitalized names. With the parens, it fails:
```idris
data Bar : Bool -> Type where
    BarA : Bar True
    BarB : Bar False

blah : (0 f : Bar True) -> Int
blah f = let 0 (BarA) = f in 0
```

```
issue2513d:7:14--7:26:@-patterns only allowed in pattern clauses
Error: While processing right hand side of blah. @-patterns only allowed in pattern clauses

issue2513d:7:14--7:26
 3 |     BarA : Bar True
 4 |     BarB : Bar False
 5 | 
 6 | blah : (0 _ : Bar True) -> Int
 7 | blah f = let 0 (BarA) = f in 0
                  ^^^^^^^^^^^^
```

dunhamsteve:
So, with the example above, I'm getting these logs
```
LOG elab:10: Checking application of Main.case block in blah ($resolved2512) to [f@(BarA)]
	Rig Rig1
	Function type (0 f : (Main.Bar Prelude.Basics.True)) -> Int)
	Expected app type Nothing
LOG elab.app.dot:50: Found Just 2 constructors for type: (Main.Bar Prelude.Basics.True)
LOG unify.meta:5: Adding new meta (Main.{f:831}, (issue2513d:8:14--8:26, Rig0))
LOG unify.meta:10: New meta type Main.{f:831}: (Main.Bar Prelude.Basics.True)
LOG elab:10: ...as: ?Main.{f:831}_[]
LOG elab:10: Now trying Main.{f:831} f@(.(f@(BarA))): (Main.Bar Prelude.Basics.True)
LOG elab.ambiguous:10: No ambiguity f@(.(f@(BarA)))
LOG elab.ambiguous:10: No ambiguity .(f@(BarA))
LOG elab.ambiguous:10: No ambiguity f@(BarA)
```
A dot is being constructed because It's finding two constructors - but only one of those constructors is valid because the type is `Bar True`:
https://github.com/idris-lang/Idris2/blob/128099fd47fc76d2796a9d19082dde4ab66580b8/src/TTImp/Elab/App.idr#L367-L380

So if it knew only one constructor was valid, this would be fine. But it'd still have the weird @ error if there happened to be two constructors. (Instead of a not covering error.) 

Ok, I think there is a typo on App.idr:406, it's trying to rewrite a IAs, but stuffing the full term back inside instead of just that unification of the arg:

```
diff --git a/src/TTImp/Elab/App.idr b/src/TTImp/Elab/App.idr
index cda4a587..71db9f83 100644
--- a/src/TTImp/Elab/App.idr
+++ b/src/TTImp/Elab/App.idr
@@ -403,7 +403,7 @@ mutual
                  Implicit _ _ => tm
                  IAs _ _ _ _ (IBindVar _ _) => tm
                  IAs _ _ _ _ (Implicit _ _) => tm
-                 IAs fc nameFC p t arg => IAs fc nameFC p t (IMustUnify fc ErasedArg tm)
+                 IAs fc nameFC p t arg => IAs fc nameFC p t (IMustUnify fc ErasedArg arg)
                  _ => IMustUnify (getFC tm) ErasedArg tm
   dotErased _ _ _ _ _ tm = pure tm
```
That resolves the @ error, but I still get:
```
Error: While processing right hand side of blah. Can't match on BarA (Erased argument).
```
I'll need to convince it that it doesn't need to dot that term.

src/Idris/Desugar.idr
```diff
@@ -209,21 +209,22 @@ mutual
   desugarB side ps (PLam fc rig p pat argTy scope)
       = pure $ ILam EmptyFC rig !(traverse (desugar AnyExpr ps) p)
                    (Just (MN "lamc" 0)) !(desugarB AnyExpr ps argTy) $
-                 ICase fc (IVar EmptyFC (MN "lamc" 0)) (Implicit fc False)
+                 ICase fc top (IVar EmptyFC (MN "lamc" 0)) (Implicit fc False)
```
gallais, line 212: This should probably `rig` rather than top. Same comment for above.

src/Idris/Desugar.idr
```diff
@@ -209,21 +209,22 @@ mutual
...
                      [snd !(desugarClause ps True (MkPatClause fc pat scope []))]
   desugarB side ps (PLet fc rig (PRef prefFC n) nTy nVal scope [])
```
gallais, line 214: Here you should probably add a `if (isPatternVariable nm)` like in the `PLam` case on line 191.

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -435,7 +435,7 @@ Eq TTImp where
     c == c' && (assert_total $ i == i') && n == n' && a == a' && r == r'
   ILet _ _ c n ty val s == ILet _ _ c' n' ty' val' s' =
     c == c' && n == n' && ty == ty' && val == val' && s == s'
-  ICase _ t ty cs == ICase _ t' ty' cs'
+  ICase _ _ t ty cs == ICase _ _ t' ty' cs'
     = t == t' && ty == ty' && (assert_total $ cs == cs')
```
mjustus, line 439: I would have expected quantity `c` to part of the equality.

dunhamsteve: Yeah, thanks for catching that.

src/Idris/Desugar.idr
dunhamsteve: Thanks, should be fixed in the latest commit, with a test case for each.

src/Idris/Desugar.idr
dunhamsteve: This change ended up breaking the test `idris2/real002`, which was doing:

```
     32 | storeProg : Has [Console, StoreI] e => App e ()
     33 | storeProg
     34 |     = let (>>=) = bindL in
                    ^^^^^
```

That looks like something people would want to do, so I'm also checking for `isOpName`.

dunhamsteve: Oh, but Data.Fin.Extra is also broken by this change.  This function is easily repaired, but the `let` change may break other code. Do we still want to do it?
```
Error:While processing right hand side of invFinSpec. Undefined name H.
Data.Fin.Extra:57:25--57:26
53 |
54 | export
55 | invFinSpec : {n : _} -> (i : Fin n) -> 1 + finToNat i + finToNat (invFin i) = n
56 | invFinSpec {n = S k} FZ = cong S finToNatLastIsBound
57 | invFinSpec (FS k) = let H = invFinSpec k in
```

gallais: I think we do want to do it, so that we're consistent about considering that upper case letters are constructor patterns (cf. LHS).

src/Idris/Desugar.idr
```diff
@@ -211,16 +211,22 @@ mutual
                    (Just (MN "lamc" 0)) !(desugarB AnyExpr ps argTy) $
                  ICase fc rig (IVar EmptyFC (MN "lamc" 0)) (Implicit fc False)
                      [snd !(desugarClause ps True (MkPatClause fc pat scope []))]
+  desugarB side ps (PLet fc rig pat@(PRef prefFC n@(UN nm)) nTy nVal scope [])
+      = if isPatternVariable nm || isOpName n
```
gallais, line 215: If this is the right notion then it probably is for `PLam` too!

src/Idris/Desugar.idr
dunhamsteve: The original change for the `PLam` case replaced `side` with `AnyExpr` in all of the recursive calls. I wasn't sure why. Should that be done in the `PLet` case? Does it have an effect aside from allowing `?` - I couldn't find any other code that matched or compared LHS / AnyExpr.

dunhamsteve:
> I added tentative PR's for the two failing CI:
> 
> * [Frex #66](https://github.com/frex-project/idris-frex/pull/66)
> * [Elab utils #39](https://github.com/stefan-hoeck/idris2-elab-util/pull/39)
> 
> Both would need to be tweaked once this PR is merged, so we'll probably want to disable the frex CI for the duration of the merge into master, and disable the Elab utils CI for 1 day (until the nightly build updates).

I've checked in a change to temporarily disable these two tests. I also merged the latest from main.

ohad:
Can we please summarise (perhaps in the original description) which cases work and which don't, and make sure there's a remaining issue for each unintended non-compiling case?

dunhamsteve:
@ohad I think I might have that last case.  

I took my example:
```idris
data Bar : Bool -> Type where
    BarA : Bar True
    BarB : Bar False

test5 : (0 _ : Bar True) -> Unit
test5 x = let 0 BarA = x
          in ()
```

And ran both it and a manually desugared version with a lot of logging (and an additional temporary log statement), and compared the two. It turned out that the manually desugared version had the arg marked as safeErase, but Case.idr wasn't passing the safeErase info when creating the function for the case.

If I pass along the safeErase info, both my test case and the one you posted to discord work.

dunhamsteve:
Thanks for finding those issues.  I think I have the first one fixed locally and am looking into the others, but I'm not sure the last two are correctly stated:
```
 1 o : Nat
 1 l : Nat
 0 e : Nat
   r : R
foo_1 : Nat
   o : Nat
   l : Nat
 0 e : Nat
   r : R
foo_w : Nat
```
I don't fully grok the QTT stuff, but it appears that in Idris2, when you take apart a constructor with linear quantity, the stuff inside doesn't necessarily get the linear tag.  I believe the `case` for `r1`  desugars to:
```idris
r1_case : (1 _: R) -> Nat
r1_case (MkR e l o) = ?foo_1
```
and at `?foo_1` the current Idris compiler reports:
```
Main> :t foo_1
   o : Nat
 1 l : Nat
 0 e : Nat
foo_1 : Nat
```
And `r2` would desugar to
```
r2_case : R -> Nat
r2_case (MkR e l o) = ?foo_w
```
for which the Idris compiler reports:
```
Main> :t foo_w
   o : Nat
   l : Nat
 0 e : Nat
foo_w : Nat
```
Based on that I think we'd expect:
```
   o : Nat
 1 l : Nat
 0 e : Nat
   r : R
foo_1 : Nat
```
and
```
   o : Nat
   l : Nat
 0 e : Nat
   r : R
foo_w : Nat
```

I guess the last one makes sense, since you have ω copies of the argument, you can make ω copies of `l`, one from each of them. And in the previous one, you take apart one R, but it has ω copies of o inside it.  

Let me know if I'm missing something here.

buzden:
>  but I'm not sure the last two are correctly stated

First, I'm not sure `let`-binding with particular quantity has direct analogy in QTT itself, but unfortunately, I'm not competent enough to judge.

What you are saying, as far as I understand, is effectively "ignore any quantity except for 0 in `let`". I can't agree with this idea.

I understand (and my understanding can be wrong, of course) `let q lhs = rhs` as "pattern match `rhs` as `lhs` and limit all produced values to the quantity no more than `q`".

Say, you have `let 1 x = y` where `y` is omega. In this case despite `y` can have infinite copies, we are explicitly saying that we want `x` to have quantity of `1`.

But take, say, a record with the only omega field, like `Singleton` type with `Val` constructor from `base`. To my view, nothing should change, i.e `let 1 Val x = Val y` should give the same result as `let 1 x = y`. But if I understand you correctly, you say that in the `Singleton` case we should get omega `x` when in the pure case we should have an `x` with quantity `1` (please, correct me if my understanding of your position is wrong).

But as far as I understand, having `0` in those `let`s makes them work the same even in your understanding. I'd consider this to be inconsistent.

dunhamsteve:
> I understand (and my understanding can be wrong, of course) let q lhs = rhs as "pattern match rhs as lhs and limit all produced values to the quantity no more than q".

That is maybe the root of our issue. The current implementation in this PR is that `let q lhs = rhs` means "pattern match rhs as lhs at quantity q".   Given the current interpretation, I wonder if it should accept `let 1 MkR e l o`, given that it's being interpreted as `let 1 (MkR e l o)`.

With the current interpretation, the object you're scrutinizing (rhs) would be observed at quantity q (like passing to a function as a quantity q argument). Assuming the Singleton case that you mentioned, it's like passing `rhs` to the function:

```
blah : (1 _ : Singleton a) -> SomeType
blah (Val x) = ...
```
(a `let` that pattern matches turns into a `case`, which turns into a function, so this essentially is the desugaring).

I wasn't suggesting that we ignore 1 in `let` (although the current PR lets Rig1 follow the existing code path), but rather that we treat it as a match of the whole scrutinee at linear quantity.  (I.e. `let 1 (Val x) = Val y in ...`) In your test case, a 1 on the `R` times a 1 on the `l` gives us a 1. But a 1 on the `R` times a ω on the `o` gives us ω. 

(You probably know the stuff below, but just summarizing how I understand linear to work.)

Looking at functions of a linear `Singleton` if we match with a linear quantity:
```idris
import Data.Singleton
foo : (1 _ : Singleton a) -> Nat
foo (Val x) = ?hole
```
we get:
```
Main> :t hole
   x : a
 0 a : a
hole : Nat
```
The `Val x` has quantity 1, but if we take it apart we have a quantity ω `x` inside.  We're taking it apart once, to get the x, of which there are ω quantity. We end up with `1 • ω = ω` of them. 

Likewise, if we have a singleton with a linear field:
```idris
data Bar : Type -> Type where
  MkBar : (1 _ : a) -> Bar a

bar : Bar a -> Nat
bar (MkBar x) = ?hole2
```
then we're taking apart ω copies of a `Bar` which has a linear `x` inside, and we get `ω•1 = ω` of them. 
```
Main> :t hole2
 0 a : Type
   x : a
hole2 : Nat
```

dunhamsteve:
I've run through all of the combinations of variable quantity and case quantity. To the cases described above, I added ones where the variable was linear, but scrutinized at various quantities. 

There was an interesting case where you've got a linear variable and explicitly say you want to scrutinize it at Rig0. For that case, I left the linear variable unused, so the case body can have it.  So in:

```idris
r10 : (1 _ : R) -> R
r10 r = let 0 (MkR e l o) = r in ?foo_10
```

The match is irrefutable, and the user says "I want to match this at level 0, without using the variable".  The match happens (at quantity 0) and the variable is linearly available in ?foo_10:
```
 1 r : R
 0 o : Nat
 0 l : Nat
 0 e : Nat
foo_10 : R
```
Under the hood, we've got one argument to pass in the scrutinee (along with args for the rest of the environment) and another for the match:
```
LOG quantity:5: New type of Main.case block in r10: (1 r : Main.R) -> (0 {scr:1089} : Main.R) -> Main.R))
```

The extra argument also happens when we scrutinize an ω variable at 0 - we want the original variable to be available at full quantity inside the case block, but we want the match to happen at 0.  So I'm setting `splitOn` to `Nothing` in those cases, and the scrutinee gets its own pi binding. So
```idris
rw0 : R -> Nat
rw0 r = let 0 (MkR e l o) = r in ?foo_w0
```
ends up as
```
LOG quantity:5: New type of Main.case block in rw0: (r : Main.R) -> (0 {scr:1052} : Main.R) -> Prelude.Types.Nat))
```

Also it currently requires the user to be explicit (`let 0 ...`) when they want to match a Rig0 variable within a RigW context:
```idris
failing "r is not accessible in this context"
  r0w : (0 _ : R) -> Nat
  r0w r = let (MkR e l o) = r in ?foo_0w
```

dunhamsteve:
This PR appears to be stalled, I'm not sure what it was waiting on (maybe further review).  

If there is interest, I can rebase this and update to work with the latest code.  But it might be prudent to wait until new core lands and redo any bits that are still relevant.  I'm expecting the case elaboration code will change in that process.

gallais:
Yeah I'm afraid this will conflict hard with the new core. And so given the merge strategy
edwin is planning to use (basically: overwrite current idris2 with his fork), we would just end
up losing this work. So I'd rather we wait for the new core.

dunhamsteve:
I'm going to close this PR.  We still have #2535 for the underlying issue, and I'll hold onto the tests. When new-core lands, we can see what it hasn't solved already and give it another go. I believe new core will be carrying more information that may make this this a bit easier and cleaner.

PR #2536, opened 2022-06-11
Update tests using base64 utility to also work on macos

dunhamsteve:
Some of the tests fail on macos because the "base64" utility expects a `-i` switch in front of the input file.  The Linux version works both with and without the `-i`.  This PR updates the tests to add the `-i`.


Z-snails:
There's also a use of `base64` in `tests/refc/buffer`. Should that also be updated?

dunhamsteve:
Thanks for catching that. I think I missed it because refc wasn't working for me at the time.

mattpolzin:
Working on my mac 👍

Issue #2537, opened 2022-06-12
[ bug ] Effectful computations can get "optimized" away


stefan-hoeck:
# Steps to Reproduce

Run the following:

```idris2
loop : Nat -> PrimIO ()
loop Z     w = MkIORes () w
loop (S k) w =
  let MkIORes () w2 := toPrim (printLn (S k)) w
   in loop k w2

main : IO ()
main = fromPrim $ loop 10
```

# Expected Behavior

Prints a couple of numbers.

# Observed Behavior

Prints nothing.

I think this is related to the comments [here](https://github.com/idris-lang/Idris2/blob/128099fd47fc76d2796a9d19082dde4ab66580b8/src/Compiler/CompileExpr.idr#L443-L464), and I'm trying to get a fix working at the moment.

stefan-hoeck:
I'm closing this, as it has been fixed in #2538.

PR #2538, opened 2022-06-12
[ fix ] issue 2537

stefan-hoeck:
So, I tried fixing #2537, and it actually seems to work (see the new test), but I get three other test failures I'm not sure how to deal with. Pinging @Z-snails , because we've discussed this on discord.

### DIFF
--- a/src/Compiler/CompileExpr.idr
+++ b/src/Compiler/CompileExpr.idr
@@ -446,22 +446,23 @@ mutual
 -- I'm (edwinb) keeping it visible here because I plan to put it back in
 -- more or less this form once case inlining works better and the whole thing
 -- works in a nice principled way.
---                      if noworld -- just substitute the scrutinee into
---                                 -- the RHS
---                         then
+                     if noworld -- just substitute the scrutinee into
+                                -- the RHS
+                        then
                              let env : SubstCEnv args vars
                                      = mkSubst 0 scr pos args in
                                  pure $ Just (substs env !(toCExpTree n sc))
---                         else -- let bind the scrutinee, and substitute the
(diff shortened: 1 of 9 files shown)


Z-snails:
codegen/builtin001 appears to be equivalent. I can have a look at the vmcode interpreter if you'd like, although I don't remember it at all :rofl:. The other one appears to be a reordering of garbage collection stuff, so maybe that's ok

stefan-hoeck:
> codegen/builtin001 appears to be equivalent. I can have a look at the vmcode interpreter if you'd like, although I don't remember it at all rofl. The other one appears to be a reordering of garbage collection stuff, so maybe that's ok

Thanks for the quick reply. So I guess I can just take the new output for `codegen/builtin001`, yes? I'd be glad if you could have a look at the vmcode stuff.

Now, whom should I ping about the C garbage collection test?

locriacyber:
I think this is ok to merge?

Z-snails:
I don't think garbage collection order is important/should be relied on, so if you're ok with it, I'd be fine merging this if the expected output of the failing test is changed

stefan-hoeck:
> I don't think garbage collection order is important/should be relied on, so if you're ok with it, I'd be fine merging this if the expected output of the failing test is changed

I don't know enough about GC to comment on this, but it would be great if we could get to merge it. Should I adjust the remaining test output?

gallais:
What @Z-snails says makes sense to me.
Let's ping @madman-bob just in case.

stefan-hoeck:
I adjusted the last test output. This should be ready now. Thanks all for your help!

PR #2540, opened 2022-06-14
[ breaking ] Make arguments of `runRWST` be ordered like in other transformers

buzden:
Put the `RWST` argument to be the last one. This makes such functions to be easier used in point-free compositions and to be easily interchangeable with existing `runStateT`-like functions.

I ran into this when I wanted to replace `evalStateT s` by `evalRWST () s` to switch from `m a` to `m (a, w)`, but couldn't do it naturally because of different order conventions in `RWST` and the rest of transformers.

Pinging @stefan-hoeck as the person who added `RWST` into the `base` of Idris 2. Was the order intentional? I understand that this order is the original order of the Haskell implementation, but historically Idris had another (and subjectively, much better) order for `StateT` and `ReaderT` which allows partial point-free connections. I think, `RWST` should follow this convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,8 @@
   Use the type `IsYes` (with constructor `ItIsYes`) from the same module instead.
 * Adds `Data.List1.Elem`, ported from `Data.List.Elem`.
 * Adds `Data.List1.Quantifiers`, ported from `Data.List.Quantifiers`.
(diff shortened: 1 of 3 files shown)


stefan-hoeck (approved):

ysangkok:
Not sure if this is the right place to ask, but it's a bit weird that

    runStateT : stateType -> StateT stateType m a -> m (stateType, a)
    runRWST : Monoid w => r -> s -> RWST r w s m a -> m (a, s, w)

Why should the `a` be after `s` in State but before for RWST? Doesn't seem consistent.

PR #2546, opened 2022-06-16
[ fix ] holes in record parameters

gallais:
I couldn't get
```
record Bug where
  proj : ?a
```
to work because:
1. if we leave the hole `?a` it leads to duplicated hole errors
2. if we make it an implicit `?` or `_` in the getter it leads to unification issues
3. if we make it an implicit `_` in the record constructor type then we need a projection for it to form `proj : (p : Proj) -> (Proj._ p)` and we can't manufacture it because `_` is not a valid name

I think I'd like it if we could have duplicated holes in a file. Idris would just have to make sure that a solution
arising from a constraint on one instance is compatible with the other occurrences of the hole. But I'm not
going to work on that.


src/Core/Context/TTC.idr
```diff
@@ -0,0 +1,32 @@
+module Core.Context.TTC
+
+import Core.Binary
+import Core.TTC
+import Core.Context
+
+%default covering
```
buzden, line 7: Why just `covering`?

gallais: Because that's what was in the original module.

PR #2549, opened 2022-06-17
[ base ] Add a function for changing the topmost `FC` of a `TTImp`

buzden:


libs/base/Language/Reflection/TTImp.idr
```diff
@@ -236,6 +236,38 @@ getFC (IHole fc _)               = fc
 getFC (Implicit fc _)            = fc
 getFC (IWithUnambigNames fc _ _) = fc
 
+public export
```
gallais, line 239: Does it need to be `public`?

buzden: Well, no, I did so just because `getFC` was `public`

buzden: Anyway, I think it's okay for such functions to be `public` because their bodies actually are part of their specification, there is nothing to hide from the user

PR #2553, opened 2022-06-21
[ golden ] Print a minimal gap between result and time on long testnames

buzden:
or else they are printed together with no space and this does not look beautiful


gallais:
Can we not munch some of the prefix instead and print e.g.
```
(...)derivation/core/norec t-pi->..
```

buzden:
Yes, we can. It's not terrifyingly beautiful, but anyway better than status-quo

munched

buzden:
To be clear, I added this munching functionality to this PR

gallais:
:+1: Let's see if someone has a suggestion for a better name for these strange padding functions.

buzden:
> Edit: I guess we could right-align the success/FAILURE too to make things neater

I thought about it too, but decided that it may be inconvenient in the case when test names are short :shrug:

mattpolzin:
RE naming the new functions: no strong preference from me, but if you use the word “truncate” somewhere in the documentation (if not the function name) then I think it will be more discoverable. That’s the word I’d search for.

ProofOfKeags (approved):

buzden:
To be clear, I renamed functions to contain `truncate` as Mathew suggested

gallais:
"munch" more or less means the same as "truncate" but is more colloquial so I think we could keep
"truncate". But I wouldn't expect a truncate function to introduce a (pre/suf)fix. So maybe a refactoring
with

1. `truncateLeft : (width : Nat) -> String -> String`
2. `path  = "(...)" ++ truncateLeft (width `minus` (5 + msgl + length time)) path`

?

buzden:
> 1. truncateLeft : (width : Nat) -> String -> String

Hmmm, we already have in the same module functions `take` and `takeLast` with meanings of `truncateRight` and `truncateLeft` correspondingly...

buzden:
> So maybe a refactoring with
> 
> 1. `truncateLeft : (width : Nat) -> String -> String`
> 2. `path  = "(...)" ++ truncateLeft (width `minus` (5 + msgl + length time)) path`

By the way, this would lead to printing `(...)` even on short strings. All the meaning of my `truncateMunchLeft` (or `fitMunchLeft`) was to add prefix only when it's needed

gallais:
It's mostly a naming issue. :(

buzden:
> How about `(left/right)Ellipsis`? Because that's effectively what we are doing here!

Why not. I've renamed and rebased using this

PR #2555, opened 2022-06-23
Fix issue with case inside a Pi type.

dunhamsteve:
On discord, @michaelmesser reported that this code:

```idris
Test : Type
Test = (case () of () => String) -> ()
```
 gives this error message:
```
While processing right hand side of Test. Main.case block in Test is not accessible in this context.
```

The case is being turned into an erased top level function. which is then used in a non-erased context. To address the issue I changed `checkPi` in `Binders.idr` to keep the contextual quantity when processing the `argTy` of an `IPi` while in the `InExpr` elab mode (instead of forcing erased).

This passes tests and I believe it is correct, but I'd like someone more knowledgable to verify.


gallais:
Reverted as it breaks the test suite.

PR #2556, opened 2022-06-24
Support System.Signal for node backend

ProofOfKeags:
In support of #2207 

This attempts to add signal support to the node backend. This is my first attempt at any contributions to the Idris2 codebase, as such I'm going about this rather naively. You'll see that there are currently 3 stubbed out TODOs.

1. `defaultSignal`: This is tricky because it doesn't appear that node *has* default Signal handlers so we would have to decide on what the normal ones would be. and as best I can tell there isn't a way to "reset" the signal handler as best I can tell
2. `collectSignal`: Since this is stateful I wanted to check in here and see if creating a global variable in the `support_system_signal.js` file that maintains a data structure of currently pending signals is OK here or if there's a better approach
3. `handleNextCollectedSignal`: similar to number 2 I wanted to get the approach OK'ed first

Lastly, node accepts signals as string arguments, while the primops call them ints. I *think* I can just cheat here and have the primops be strings under the hood since they will also be strings when supplied as arguments to the signal functions but I also don't know if that is safe given that the primops may try to properly use the signals as integers in which case we'd have to translate inside the individual handling functions.

Would really appreciate some feedback so I can finish this and move the needle on this.

Additional reference: https://nodejs.org/api/process.html#signal-events

### DIFF
--- a/support/js/support_system_signal.js
+++ b/support/js/support_system_signal.js
@@ -42,23 +42,43 @@ function signal_int_to_string(signal) {
 }
 
 function support_system_signal_ignoreSignal(signal) {
-    support_system_signal_process.on(signal_int_to_string(signal), sig => {})
-    return 0
+    let signal_string = signal_int_to_string(signal)
+    try {
+        support_system_signal_process.on(signal_string, sig => {})
+        return 0
+    } catch (e) {
+        return -1
+    }
 }
 
 function support_system_signal_sendSignal(pid, signal) {
-    support_system_signal_process.kill(pid, signal_int_to_string(signal))
-    return 0
+    let signal_string = signal_int_to_string(signal)
+    try {
+        support_system_signal_process.kill(pid)
+        return 0
+    } catch (e) {
+        return -1
+    }
 }
 
 function support_system_signal_raiseSignal(signal) {
-    support_system_signal_process.kill(support_system_signal_process.pid, signal_int_to_string(signal))
-    return 0
+    let signal_string = signal_int_to_string(signal)
+    try {
+        support_system_signal_process.kill(support_system_signal_process.pid)
+        return 0
+    } catch (e) {
+        return -1
+    }
 }
 
 function support_system_signal_defaultSignal(signal) {
-    support_system_signal_process.removeAllListeners(signal_int_to_string(signal))
-    return 0
+    let signal_string = signal_int_to_string(signal)
+    try {
+        support_system_signal_process.removeAllListeners()
+        return 0
+    } catch (e) {
+        return -1
+    }
 }
 
 function support_system_signal_collectSignal(signal) {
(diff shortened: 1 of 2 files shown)


Z-snails:
`prim__sighup`/etc should be `Int`, you could convert the string into an int of your choosing, or even better follow what values posix uses (if that's standard, I don't actually know if posix defines them)

ProofOfKeags:
Changed them all over to int's and made it such that the default case was as close to a NOP as I could find in the nodejs API. I still need to figure out if carrying mutable state in this JS file is an acceptable way to implement the 3 remaining TODO methods.

ProofOfKeags:
I think this is ready for proper review. How should I go about testing the implementation?

mattpolzin:
I haven’t had time to properly look this over, but wanted to comment on “default cases as close to a NOP as possible” because for some signals this is pretty far from the default behavior (notably, signals that would normally cause the app to crash). I wonder if it isn’t better to leave default handling un implemented if we have to choose between surprising behavior or no support. If NodeJS doesn’t support defaulting these, maybe it makes sense that you just avoid using these functions when writing Idris with the intention of building to NodeJS (and a compile time error about the thing being unimplemented in Node is actually a very friendly option)?

ProofOfKeags:
> I haven’t had time to properly look this over, but wanted to comment on “default cases as close to a NOP as possible” because for some signals this is pretty far from the default behavior (notably, signals that would normally cause the app to crash).

Yeah perhaps the answer is to make it a NOP only when we install signal handlers for undefined signal types and if we try to send an undefined signal to another process, or raise one we just return an error (-1)?

It's unclear here because the true domain of the function is a fairly small enum, whereas the type-accepted domain is `Int`

support/js/support_system_signal.js
```diff
@@ -0,0 +1,106 @@
...
+        return 0
+    } catch (e) {
+        return -1
+    }
+}
+
+function support_system_signal_raiseSignal(signal) {
+    let signal_string = signal_int_to_string(signal)
+    try {
+        support_system_signal_process.kill(support_system_signal_process.pid)
```
Z-snails, line 67: same comment here

support/js/support_system_signal.js
```diff
@@ -0,0 +1,106 @@
...
+        return 0
+    } catch (e) {
+        return -1
+    }
+}
+
+function support_system_signal_defaultSignal(signal) {
+    let signal_string = signal_int_to_string(signal)
+    try {
+        support_system_signal_process.removeAllListeners()
```
Z-snails, line 77: This also accepts a signal (the docs suggest as a string) as an optional parameter

Z-snails (changes requested):

support/js/support_system_signal.js
ProofOfKeags: Yep this was just a mistake on my part. Good catch.

ProofOfKeags:
bump

Issue #2545, opened 2022-06-16
Add Rational Number to standard library


locriacyber:
## The proposal
1: Add `Rational` type
2: Add related helper functions to `Data.Rational`
3: Make decimal literals use `fromRational` (like `fromInteger`)
4: Add an interface for non-total division

Related: https://github.com/idris-lang/Idris2/issues/2097

Should Idris add rational literal other base (like `0x76.bf`), it can be represented with fractions fairly easily.

We can't use `Fractional` for division if we do not want `7/0` to be valid.

## Motivation
The desire to ~~create frog fractions 3~~ represent rational numbers accurately.

### Examples

```
d0 : Double
d0 = 0.3
-- d0 = fromRational (RatioOf 3 10)

x0 : Rational
x0 = 1.2

x0_accurate : x0 = RatioOf 12 10
x0_accurate = Refl
```

### Technical implementation

```
module Data.Rational

import Data.Nat

%default total

public export
data NotZero : Type where
   S : Nat -> NotZero

export
Cast NotZero Integer where
   cast (S n) = (natToInteger n) + 1

(*) : NotZero -> NotZero -> NotZero
(*) (S m) (S n) = S (m + n + m * n)

public export
record Rational where
   constructor RatioOf
   numerator : Integer
   denominator : NotZero

normalize : Rational -> Rational
normalize (RatioOf m n) = ?todo_nogcd_proof
   -- let m : Integer = cast m
   -- let n : Integer = cast n
   -- let gcd_ = gcd m n 
   -- let m = m / gcd_
   -- let n = n / gcd_
   -- RatioOf (cast m) (cast n)

export
Num Rational where
   (+) (RatioOf m n) (RatioOf m' n') =
      let 
         numer = (m * (cast n')) + (m' * (cast n))
         denom = n * n'
      in
         normalize $ RatioOf numer denom

   (*) (RatioOf m n) (RatioOf m' n') =
      let
         numer = m * m'
         denom = n * n'
      in
         normalize $ RatioOf numer denom
      
   fromInteger n = RatioOf n (S 0)

export
Neg Rational where
   negate (RatioOf m n) = RatioOf (negate m) n
   (-) m n = (+) m (negate n)

export
Abs Rational where
   abs (RatioOf m n) = RatioOf (abs m) n
```

locriacyber:
#2097 is about floating point literals like `1.76136250473`.
This issue is about a type.

locriacyber:
[...]

PR #2558, opened 2022-06-25
Add Data.Rational

locriacyber:
Current implementation is naive. I think that's why Haskell GHC has Rational as primitive.

Fix https://github.com/idris-lang/Idris2/issues/2545

### DIFF
--- a/libs/contrib/Data/Rational.idr
+++ b/libs/contrib/Data/Rational.idr
@@ -1,0 +1,91 @@
+module Data.Rational
+
+import Data.Nat.Factor
+
+%default total
+
+public export
+data Nat1 : Type where
+   S : Nat -> Nat1
+
+export
+Cast Nat1 Integer where
+   cast (S n) = (natToInteger n) + 1
+
+mult : Nat1 -> Nat1 -> Nat1
+mult (S m) (S n) = S (m + n + m * n)
+
+toNat_notbothzero : Nat1 -> Nat
+toNat_notbothzero (S n) = S n
+
+fromNat_coerce : Nat -> Nat1
+fromNat_coerce Z = S Z
+fromNat_coerce (S n) = S n
+
+public export
+record Rational where
+   constructor RatioOf
+   numerator : Integer
+   denominator : Nat1
+
+
+-- TODO: inline all this as prim__
+toUnsigned : Integer -> (Nat, Bool)
+toUnsigned n = if n < 0 then (cast n, False) else (cast (negate n), True)
+
+fromUnsigned : (Nat, Bool) -> Integer
+fromUnsigned (n, False) = cast n
+fromUnsigned (n, True) = negate (cast n)
+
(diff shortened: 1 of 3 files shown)


libs/prelude/Prelude/Types.idr
```diff
@@ -21,13 +21,12 @@ data Nat =
   | ||| Successor.
   S Nat
 
-%name Nat k, j, i
```
Z-snails, line 24: Did you forget to put this back?

libs/contrib/Data/Rational.idr
```diff
@@ -0,0 +1,91 @@
...
+   let n' = toNat_notbothzero n
+   let (m', is_negative) = toUnsigned m
+   let (n'', m'') = gcd_div (n', m')
+   RatioOf (fromUnsigned (m'', is_negative)) (fromNat_coerce n'')
+-- inline this (end)
+
+
+export
+Num Rational where
+   (+) (RatioOf m n) (RatioOf m' n') =
```
Z-snails, line 66: IMO this would be clearer with names like d1, d2 for denominator and n1, n2 for numerator.

libs/contrib/Data/Rational.idr
```diff
@@ -0,0 +1,91 @@
...
+fromNat_coerce (S n) = S n
+
+public export
+record Rational where
+   constructor RatioOf
+   numerator : Integer
+   denominator : Nat1
+
+
+-- TODO: inline all this as prim__
```
Z-snails, line 32: What do you mean by this? Are you going to resolve this todo before it is merged

gallais:
I think this would be better as an external library.

libs/prelude/Prelude/Types.idr
locriacyber: This is unnecessary. The code can still type check.

libs/contrib/Data/Rational.idr
locriacyber: It needs backend support to be fast. Right now it "should" be working.

locriacyber:
> I think this would be better as an external library.

It's in `contrib` for now. I think this data type is useful enough to be a primitive in the future.

libs/prelude/Prelude/Types.idr
stefan-hoeck: The `%name` pragma is not about type checking. Please don't remove this.

gallais:
> It's in contrib for now. 

Well, no, it's in a PR to contrib for now and I'm telling you I don't think it'll get merged.
Because

1. we're explicitly trying to [make `contrib` smaller](https://github.com/idris-lang/Idris2/wiki/1-%5BLanguage%5D-Contrib-Organisation)
2. this is explicitly a naïve implementation

locriacyber:
Moved to https://github.com/locriacyber/idris2-rational

Issue #2557, opened 2022-06-25
Incremental compilation produces bad executables when missing incremental compile data


Z-snails:
If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!

# Steps to Reproduce
https://gist.github.com/Z-snails/33bda6b1764d8d601a7680c8ccaa3e2a
run `sh build.sh`
# Expected Behavior
prints "hello world"
# Observed Behavior
`Exception: variable Mod2-main is not bound`

Either idris should give an error while compiling or return to whole program compilation

locriacyber:
The problem is that Idris refuse compile the module again in incremental mode if it finds the same module already compiled in "whole program mode".

In "whole program mode", Idris still produce 1 TTC per `.idr` file. It should be possible to continue from there.

dunhamsteve:
I could see cases where there is both a .ttc and a .so, but the .so is older than the .ttc (e.g. compile with incr on, change file, compile with incr off, then compile with incr on again). It might be necessary to check that the .so is both present and up to date when deciding whether to rebuild the module or use the existing ttc.

dunhamsteve:
Interesting. So the .ttc file says if there is a .so or not. And it collects that information to know what to link, but doesn't actually include the code in the non `.so` case.  So if there is a `Mod1.so` sitting on disk and the latest `Mod1.ttc` says it's not incremental, it won't link the stale `.so`, but also neglects to include the code from the `.ttc`.

dunhamsteve:
Ok, it was more subtle than I thought, but it looks like this will fix the issue.  The root cause is that Idris will drop the `allIncData` for a backend if the data is missing for that backend in one of the `ttc` files, but when processing a subsequent file it recreates a record for that backend.

Above, I was thinking that you could just rebuild the .ttc/.so, but hadn't considered the case of installed libraries.  There is a lot of code already in case to handle missing incremental data by switching to whole program. 

One possible future improvement would be to take a hybrid approach and do best effort incremental - use the available incremental parts and whole program the rest.

```idris
diff --git a/src/Core/Context.idr b/src/Core/Context.idr
index f40285eb..659a384c 100644
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -2474,6 +2474,10 @@ addImportedInc modNS inc
                   do recordWarning (GenericWarn ("No incremental compile data for " ++ show modNS))
[...]

PR #2560, opened 2022-06-26
Address issue #2557 Incremental compilation produces bad executables

dunhamsteve:
This PR addresses issue #2557. When there are .ttc files without corresponding .so files, incremental compilation will produce executables with undefined functions.

The root cause is that the current code attempts to drop the incremental data when it sees a non-incremental ttc (or one without support for the current codegen), but a subsequent ttc will create more incremental data, and the backend doesn't know there was an issue.

As a fix, I'm dropping the codegen from the list of `incrementalCGs` in addition to the existing code that drops the incremental data for the codegen.

The first module that comes through with an empty `modNS` and no incremental data, which we need to ignore.


Z-snails:
I think with how incremental compilation works currently, creating .so files after a missing .so file wouldn't help, as idris would recompile them anyway if you forced it to recreate the missing .so file.

Something possible better (and I think possible) is use the .ttc file to recreate the .so file if it is missing.

dunhamsteve:
That came to mind and seemed like a better solution, but there are a couple of issues:

- ttc contents files are different depending on the value of `incrementalCGs`:
    - The hash code includes function arity in incremental mode.
    - There is an embedded list of codegens (used when installing the library)
    - Some code is compiled differently (see the excepts below).
- They might be in an installed library, where we shouldn't be writing.

One possible solution: skip the `.ttc` file and rebuild both it and the `.so` _if_ it is local to the project. And if an installed `.ttc` file is not acceptable we could either bail on the entire compile or drop the backend from `incrementalCGs`.

https://github.com/idris-lang/Idris2/blob/68a144bf185d3448055c06c8e84df3cba7daaa86/src/Compiler/Inline.idr#L529-L540
https://github.com/idris-lang/Idris2/blob/68a144bf185d3448055c06c8e84df3cba7daaa86/src/Compiler/Inline.idr#L568-L572
https://github.com/idris-lang/Idris2/blob/68a144bf185d3448055c06c8e84df3cba7daaa86/src/Compiler/CompileExpr.idr#L788-L803

Z-snails:
Sorry for the delay, I think for now it might not be worth complicating it too much, so for now I am ok with just dropping the incremental CG backend, as you've implemented so far. Let me know if you've had any other thoughts in the last few weeks

dunhamsteve:
I took a look a couple of days ago and made the change to rebuild TTCs if the incremental data is missing.  

The check is added where Idris is checking the hashes &c to see if the TTC is stale.  Surprisingly, this occurs in two locations, ProcessIdr.idr and ModTree.idr.  The change to rebuild TTCs is orthogonal to the change above, so we can put it in a separate PR if desired.  I kept it in a separate branch for now, see the changes in 0120d3d. 

This case can arise when you've got idris2-lsp running without IDRIS2_INC_CGS in the environment.  It will overwrite TTCs with non-incremental versions. With the change in that branch, Idris will just rebuild those files.

dunhamsteve:
I went ahead and added the change to rebuild ttc's that were missing incremental data. I found it really helpful in the case where `idris2-lsp` was building non-incremental ttc files while I was trying to do an incremental build.

src/Core/Context.idr
```diff
@@ -2474,6 +2474,9 @@ addImportedInc modNS inc
                   do recordWarning (GenericWarn ("No incremental compile data for " ++ show modNS))
                      defs <- get Ctxt
                      put Ctxt ({ allIncData $= drop cg } defs)
+                     -- Tell session that the codegen is no longer incremental
+                     when (show modNS /= "") $
```
Z-snails, line 2478: Does this check still want to be here now that idris rebuilds from ttc?

dunhamsteve: It's still building from source, so it just considers the .ttc outdated if the .so is missing. I expect this to be triggered if a installed package is missing incremental data.  (It's possible it may consider those as having missing ttc, checking now. )

dunhamsteve: Yeah, for the case where a library is non-incremental, it will print the message and drop the incremental build.  

I made a quick test project that uses `elab-util`. 

If `elab-util` is installed non-incrementally, I get this message during build and it outputs a working, non-incremental executable:
```
Warning: No incremental compile data for Language.Reflection.Pretty
```

If I then re-install an incremental build of `elab-util` and try again.  Idris now knows to rebuild the project's `.ttc` and gives me a working incremental build.

If I try with a non-incremental `elab-util` and the main branch compiler, I get: 
```
1/1: Building Main (src/Main.idr)
Warning: No incremental compile data for Language.Reflection.Pretty
Warning: Compiling hole Main.isEqual
dunham@peri elab % build/exec/elab
Exception: variable Main-main is not bound
```

Z-snails: Ah Ok, I misunderstood, that makes sense.

PR #2562, opened 2022-06-27
[ re #2533 ] Fix printing of lazy types and values in `Show TTImp`

buzden:
The only controversial thing I see in this change is the way I decided to print `LUknown` delayed type. But this type of expression rarely appears and I think it's important to show that this type is actually a some kind of lazy type when printing. I think, we don't have any adequate Idris syntax for this, so it's printed in the comment.

### DIFF
--- a/libs/base/Language/Reflection/TTImp.idr
+++ b/libs/base/Language/Reflection/TTImp.idr
@@ -611,9 +611,11 @@ mutual
     showPrec d (IAs fc nameFC side nm s)
       = "\{show nm}@\{showPrec App s}"
     showPrec d (IMustUnify fc dr s) = ".(\{show s})"
-    showPrec d (IDelayed fc lr s) = showPrec d s
-    showPrec d (IDelay fc s) = showCon d "Delay" $ assert_total (showArg s)
-    showPrec d (IForce fc s) = showPrec d s
+    showPrec d (IDelayed fc LInf s) = showCon d "Inf" $ assert_total $ showArg s
+    showPrec d (IDelayed fc LLazy s) = showCon d "Lazy" $ assert_total $ showArg s
+    showPrec d (IDelayed fc LUnknown s) = "({- unknown lazy -} \{showPrec Open s})"
+    showPrec d (IDelay fc s) = showCon d "Delay" $ assert_total $ showArg s
+    showPrec d (IForce fc s) = showCon d "Force" $ assert_total $ showArg s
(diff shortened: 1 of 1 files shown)


gallais:
This clashes with my intention to (hopefully) print valid source code. I suppose it is convenient for debugging
if you want to check whether a `Delay` was correctly inserted here or there. Should we refactor the entire
printer to take a record of configuration toggles (e.g. printing of lazy, printing of `0` implicits)? The `Show`
instance can then be defined by feeding in a sensible default configuration while users could call the more
general function directly if they need more verbose output.

buzden:
> This clashes with my intention to (hopefully) print valid source code.

Could you please say which expression leads to a non-valid source code? :eyes: I tried to change it precisely to so do

buzden:
> Well, ok it's maybe not invalid but isn't it quite unidiomatic to write the `Delay / Force` yourself?

Well, I write `delay` and `force` as function calls as soon as I need them ;-) I do this in sufficiently lazy code, especially in places where compiler does not put when automatically and I don't want explicit lambda repacking to allow it to do so.

I think, it's okay to print all relevant information of an expression. One may want to print such an expression before and after elaboration; when you ignore printing `Delay` or `Force`, you may lose the change of the expression.

I can imagine the following situation: say, some elaboration script analyses a given expression and prints the piece it couldn't work out using `Show` (as soon as it's now present). But when `delay` or `force` are not printed at all, one may receive a strange error.

I ran onto such situation with `Lazy`: I have `Lazy` not supported by one my script and I decided to add an expression at which the script was failing, but it printed an expression which it should support because `Lazy` was not printed.

buzden:
Anyway, by the way, the whole PR was originally about `IDelayed`, not about `IDelay`.

I didn't change the logic of `IDelay` case, I just made the logic of `IForce` case to be the same as `IDelay` which you've added originally just because it was near to the piece of code I intended to change.

If you strongly think that `Delay` and `Force` should not be printed, I'm okay to remove it, removing also your original printing of `Delay`.

buzden:
While we're discussing, I've found a problem in how `case .. of` shows. I've added a commit that makes the result of `Show` for this compilable:
1. The arrow was fixed
3. Put explicit type ascription into a comment
3. Don't print type when it was not present in the expression

I've added it to the same PR because it's all about fixing `Show TTImp`, but if it is not appropriate here, please let me know.

gallais:
Okay fair enough. :+1:

Issue #1878, opened 2021-08-30
[ elab ] Quotation of a dependent pair type under value-returning `%runElab` results in a hole unexpectedly


buzden:
# Steps to Reproduce

Unified examples gist (e.g., for the future regression test) [can be found here](https://gist.github.com/buzden/454d70dfc61fbdb335eb213491b91174).

Consider a primitive elaboration script that just quotates a type and logs it:

```idris
extract : Type -> Elab Unit
extract ty = logTerm "debug.elab" 0 "type" !(quote ty)
```

Consider a macro that does the same (it can have the same body as the non-macro function, it does not matter).

```idris
%macro
extract' : Type -> Elab Unit
extract' = extract
```

Consider also couple of additional declarations:

```idris
data Y : Nat -> Type where
  Y0 : Y 0
  Y1 : Y 1

X : Type
X = (n : Nat ** Y n)
```

Then use both macro and non-macro versions:

```idris
%runElab extract X

u : Unit
u = extract' X
```

Expectedly get the same logged terms:

```
LOG debug.elab:0: type: ((Builtin.DPair.DPair Prelude.Types.Nat) (%lam RigW Explicit (Just n) Prelude.Types.Nat (Y n)))
LOG debug.elab:0: type: ((Builtin.DPair.DPair Prelude.Types.Nat) (%lam RigW Explicit (Just n) Prelude.Types.Nat (Y n)))
```

Now, inline the `X` type in the elaboration calls:

```idris
%runElab extract (n : Nat ** Y n)      
      
u : Unit      
u = extract' (n : Nat ** Y n)
```

No change in the logged terms, as expected. Good.

---

Now, change datatype definitions a little bit:

```idris
data Y : Nat -> Type -> Type where
  Y0 : Y 0 Int
  Y1 : Y 1 Nat

X : Type
X = (n : Nat ** x : Type ** Y n x)
```

Run as in the very first example:
```idris
%runElab extract X

u : Unit
u = extract' X
```

Get expectedly same logged terms:
```
LOG debug.elab:0: type: ((Builtin.DPair.DPair Prelude.Types.Nat) (%lam RigW Explicit (Just n) Prelude.Types.Nat ((Builtin.DPair.DPair %type) (%lam RigW Explicit (Just x) %type ((Y n) x)))))
[...]

PR #2568, opened 2022-07-01
[ new ] deriving Functor

gallais:
TODO

* [x] Existential types (Like `Free`)
* [x] Conditional instances (Conditional `Functor`)
* [x] Conditional instances (Conditional `Bifunctor`)

```idris
namespace Free

  data Free : (Type -> Type) -> Type -> Type where
    Pure : a -> Free f a
    Bind : f a -> (a -> Free f b) -> Free f b

  free : Functor (Free f)
  free = %runElab derive
```
  
```idris
namespace MaybeT

  record MaybeT (m : Type -> Type) (a : Type) where
    constructor MkMaybeT
    runMaybeT : m (Maybe a)

  maybeT : Functor m => Functor (MaybeT m)
  maybeT = %runElab derive
```

### DIFF
--- a/libs/base/Language/Reflection.idr
+++ b/libs/base/Language/Reflection.idr
@@ -212,7 +212,7 @@ catch : Elaboration m => Elab a -> m (Maybe a)
 catch elab = try (Just <$> elab) (pure Nothing)
 
 ||| Run proof search to attempt to find a value of the input type.
-||| Useful to check whether a given satisfies an interface constraint.
+||| Useful to check whether a give interface constraint is satisfied.
 export
 search : Elaboration m => (ty : Type) -> m (Maybe ty)
 search ty = catch $ check {expected = ty} `(%search)
--- a/libs/base/Language/Reflection.idr
+++ b/libs/base/Language/Reflection.idr
@@ -205,3 +205,14 @@ Elaboration m => MonadTrans t => Monad (t m) => Elaboration (t m) where
   getLocalType        = lift . getLocalType
   getCons             = lift . getCons
   declare             = lift . declare
+
+||| Catch failures and use the `Maybe` monad instead
+export
+catch : Elaboration m => Elab a -> m (Maybe a)
+catch elab = try (Just <$> elab) (pure Nothing)
+
+||| Run proof search to attempt to find a value of the input type.
+||| Useful to check whether a given satisfies an interface constraint.
+export
+search : Elaboration m => (ty : Type) -> m (Maybe ty)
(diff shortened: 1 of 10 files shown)


libs/base/Language/Reflection.idr
```diff
@@ -205,3 +205,14 @@ Elaboration m => MonadTrans t => Monad (t m) => Elaboration (t m) where
...
+||| Catch failures and use the `Maybe` monad instead
+export
+catch : Elaboration m => Elab a -> m (Maybe a)
+catch elab = try (Just <$> elab) (pure Nothing)
+
+||| Run proof search to attempt to find a value of the input type.
+||| Useful to check whether a given satisfies an interface constraint.
+export
+search : Elaboration m => (ty : Type) -> m (Maybe ty)
+search ty = catch $ check {expected = ty} `(%search)
```
gallais, line 218: Looks like this is not failing properly: I'm not getting `Nothing` in case the search is unsuccessful
(e.g. `search (Not Nat)`). That leads to `Deriving.Functor` generating improper code.

buzden: Is `ty` in this case just `Not Nat` or `Prelude.Basics.Not Prelude.Types.Nat`? I mean, I'm not sure `%search` expression does disambiguation and name expansion.

buzden: Oh, I read your message wrong. What do you mean by "not failing properly"? Does if fail through `Elab` instead of returning `Nothing`?

gallais: Yes exactly. It fails with "Can't find an implementation for Not Nat" (or something similar) instead of `Nothing`.

gallais: Found the culprit. And another one too. I have encountered quite a few errors involving "?postpone"
while writing the `derive` script especially with `%runElab f` when `f` has a type of the form
`Elaboration m => m ...` instead of `f : Elab ...`.

So hopefully there will be fewer of these in the future too (the `Search` test case is precisely one such case).

buzden: Wow! Would this fix #1878 by chance?

gallais: Nope you still get the hole:
> LOG debug.elab:0: type: ((Builtin.DPair.DPair Prelude.Types.Nat) ?p)

buzden: I ran into this issue on the current `main` in my complicated script :-( I didn't manage to produce a simple example yet

PR #2572, opened 2022-07-04
[ fix ] handle implicit arguments in Deriving.Functor

gallais:


buzden:
I just was playing with this:

```idris
data Y : Ord a => Type -> Type where
  MkY : Ord a => List (a, b) -> Y {a} b

YF : (ord : Ord a) => Functor $ Y @{ord}
YF = %runElab derive
```

gives

```
Error: While processing right hand side of YF. Error during reflection: While processing type of $resolved6297,mapY. Can't solve constraint between: ?_ [locals in scope: a, ord] and Ord a.

X:22:1--22:21
 18 | data Y : Ord a => Type -> Type where
 19 |   MkY : Ord a => List (a, b) -> Y {a} b
 20 | 
 21 | YF : (ord : Ord a) => Functor $ Y @{ord}
 22 | YF = %runElab derive
      ^^^^^^^^^^^^^^^^^^^^
```

gallais:
Fixed. I've even found a somewhat plausible example.

PR #2586, opened 2022-07-12
Ide random socket

dunhamsteve:
The primary issue this addresses is that there can be only one ide server running at the same time, because they all use the same socket number.  (idris-community/idris2-mode#26). 

Since the plugins read the socket number from stdout, this could be worked around by opening socket "0" which means "choose an open socket", but idris2 currently prints the configured value rather than the actual socket number that was chosen by the OS.

This PR changes `--ide-mode-socket` to print the number of the socket that was opened, rather than the configured value.  This change enables `--ide-mode-socket :0` to work correctly.

It also changes the default socket to 0.  This change allows existing clients to benefit from having more than one server running at a time, without change.  But it assumes they respect the socket number printed on stdout. I don't see any evidence via google of anyone hardcoding the number 38398, so I think this is safe.

If clients do want a specific socket, they can still specify it on the command line.

Do any plugins besides the emacs ones use ide-mode-socket?

### DIFF
--- a/src/Idris/CommandLine.idr
+++ b/src/Idris/CommandLine.idr
@@ -164,10 +164,10 @@ data CLOpt
 ||| Extract the host and port to bind the IDE socket to
 export
 ideSocketModeAddress : List CLOpt -> (String, Int)
-ideSocketModeAddress []  = ("localhost", 38398)
+ideSocketModeAddress []  = ("localhost", 0)
 ideSocketModeAddress (IdeModeSocket hp :: _) =
   let (h, p) = String.break (== ':') hp
-      port = fromMaybe 38398 (portPart p >>= parsePositive)
+      port = fromMaybe 0 (portPart p >>= parsePositive)
       host = if h == "" then "localhost" else h
   in (host, port)
   where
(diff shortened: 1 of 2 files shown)


gallais:
>  This change allows existing clients to benefit from having more than one server running at a time, without change

🥰

> But it assumes they respect the socket number printed on stdout. I don't see any evidence via google of anyone hardcoding the number 38398, so I think this is safe.

I think this is a perfectly fine to assume of modes. Can you please add this explicitly
to [the documentation of the protocol](https://github.com/idris-lang/Idris2/blob/main/docs/source/implementation/ide-protocol.rst)?
This document aims to make easy to implement new clients without having to reverse
engineer the protocol.

> Do any plugins besides the emacs ones use ide-mode-socket?

Edwin's vim mode I suspect? https://github.com/edwinb/idris2-vim

mattpolzin:
> Edwin's vim mode I suspect?

If I understand correctly, that Vim plugin actually uses Idris2 to execute 1-off commands via the `--client` flag; I’m not sure, but I believe that means it uses REPL supported syntax as opposed to the IDE protocol.

dunhamsteve:
It looks like Sphinx needs a space between a colon and the trailing `::`. Otherwise it renders `::` in the output. I adjusted the rest of the file to fix this.

PR #2591, opened 2022-07-15
[ new ] extend Deriving.Functor to (non-strictly) positive functors

gallais:


tests/idris2/reflection017/DeriveFunctor.idr
```diff
@@ -227,12 +227,18 @@ namespace Cont
   cont2 : Functor (Cont2 r e)
   cont2 = %runElab derive
 
-  ||| Continuation with short-circuiting error cont
+  ||| Uncurried version of ontinuation with short-circuiting error cont
```
dunhamsteve, line 230: continuation

PR #2594, opened 2022-07-16
[ doc ] add comments to generated javascript

gallais:
I am currently writing a paper showing some of the generated code and I'd rather
not get confused about which constructor maps to which tag. This should make
things easier.

### DIFF
--- a/src/Compiler/ES/Ast.idr
+++ b/src/Compiler/ES/Ast.idr
@@ -7,6 +7,16 @@ import Data.Vect
 
 %default total
 
+public export
+data Tag : Type where
+  ||| A data constructor. Use the tag to dispatch / construct.
+  ||| Here the Name is only for documentation purposes and should not
+  ||| be used.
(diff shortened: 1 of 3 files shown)


dunhamsteve:
I like the idea, bucklescript does something similar, but do we want these comments in minimal mode? If not maybe add a Doc node for Comment and drop it in minimal mode. (Like it does for newlines.)

gallais:
Oh, is there a minimal mode option? I assumed people would use a minifier to postprocess the file.
I'll take a look later.

dunhamsteve:
Yeah, I could see most people running it through an external  minifier anyway, but there is a compact mode.

The magic happens here:
https://github.com/idris-lang/Idris2/blob/a0d9e2bbba8b520e16dd3904fbb679a542e15324/src/Compiler/ES/Doc.idr#L84-L92

gallais:
Alright, I've added a new `Comment` constructor to `Doc` and I used this opportunity to
also print the type declaration of functions that correspond to `UN`s. :)

PR #2595, opened 2022-07-16
[ new ] Enable REPL reload command to reload ipkgs

skykanin:
Enables REPL `:reload` command to reload entire ipkgs.


mattpolzin:
One thing to note: not all packages have main files and not all modules are guaranteed to be referenced by a package’s main file. 

This seems like a beneficial change, so I think you have two choices. 
1. experiment with reloading all public modules of a package instead of just the main file. 
2. leave this change roughly as-is but change your REPL command description to be more explicit (“reloads the current file or package’s main file”).

skykanin:
> One thing to note: not all packages have main files and not all modules are guaranteed to be referenced by a package’s main file.
> 
> This seems like a beneficial change, so I think you have two choices.
> 
>     1. experiment with reloading all public modules of a package instead of just the main file.
> 
>     2. leave this change roughly as-is but change your REPL command description to be more explicit (“reloads the current file or package’s main file”).

yes, I think I'll experiment with reloading all public modules. That would be useful

PR #2596, opened 2022-07-18
add function for mapping over constrained elements of a list

joelberkeley:
not sure what to call this function

### DIFF
--- a/libs/base/Data/List/Quantifiers.idr
+++ b/libs/base/Data/List/Quantifiers.idr
@@ -89,6 +89,12 @@ namespace All
   imapProperty i f @{[]} [] = []
   imapProperty i f @{ix :: ixs} (x :: xs) = f @{ix} x :: imapProperty i f @{ixs} xs
 
+  ||| Map a constrained function over a list given a list of constraints.
+  public export
+  map : ((x : a) -> {0 ok : p x} -> b) -> (xs : List a) -> {auto 0 allOk : All p xs} -> List b
+  map f [] {allOk = []} = []
+  map f (x :: xs) {allOk = ok :: _} = f {ok} x :: map f xs
+
   ||| Forget property source for a homogeneous collection of properties
   public export
   forget : All (const type) types -> List type


libs/base/Data/List/Quantifiers.idr
```diff
@@ -89,6 +89,12 @@ namespace All
   imapProperty i f @{[]} [] = []
   imapProperty i f @{ix :: ixs} (x :: xs) = f @{ix} x :: imapProperty i f @{ixs} xs
 
+  ||| Map a constrained function over a list given a list of constraints.
+  public export
+  map : ((x : a) -> {0 ok : p x} -> b) -> (xs : List a) -> {auto 0 allOk : All p xs} -> List b
```
joelberkeley, line 94: curiously it doesn't seem to matter if I make `ok` `auto`. I would have thought it would. This seems to accept functions with and without `auto` which is what I want

joelberkeley: even more surprising, is that I can pass a function with a different name to `ok`, and it still works. Does Idris silently cast the implicit to a different name?

gallais: The difference will be visible at call sites.

joelberkeley: [here's](https://github.com/joelberkeley/spidr/blob/016c4cd9b3df1b5c991113035b97bff62fc99cae/src/Tensor.idr#L398) an example of me using it. I would have expected this to be fairly typical usage. Do you not think so?

joelberkeley: oh i see - you made the list implicit. hmmm ... well this looks like quite a different function to what I had in mind, certainly in usage. I don't know what use case that would be natural for

gallais: In your use case you're lucky because `InBounds` is injective but if you were using a `p`
that's defined by induction on its argument for instance, it could not be reconstructed by
unification.

joelberkeley: OK. This sounds like more of a complex PR than I was anticipating. I don't have the time atm to focus on an alternative approach. Happy to close or leave

PR #2600, opened 2022-07-25
[ fix ] custom backend doc

gallais:
The type Codegen was changed in #2594.


mattpolzin:
If only this code lived in a literate file instead of an rst file then we could build it in CI

Issue #2589, opened 2022-07-13
`flip` doesn't work with implicit dependent arguments


joelberkeley:
# Steps to Reproduce

Compile
```
index' : (xs : List Nat) -> (n : Nat) -> {auto 0 ok : InBounds n xs} -> Nat
index' = flip index
```

# Expected Behavior

Successfully compiles. I think it should be possible to compile this. The implicit argument is dependent on the previous two, but reordering them should be OK.

# Observed Behavior

Fails to compile with
```
Error: While processing right hand side of index'. Can't solve constraint between: ?c [no locals in scope] and {auto 0 _ : InBounds arg arg} -> Nat.

 346 | 
 347 | public export
 348 | index' : (xs : List Nat) -> (n : Nat) -> {auto 0 ok : InBounds n xs} -> Nat
 349 | index' = flip index
                ^^^^^^^^^^
```

gallais:
Are you suggesting we change the type of `flip` from

```
flip : (a -> b -> c) -> b -> a -> c
```

to

```
flip : {0 c : a -> b -> Type} -> ((x : a) -> (y : b) -> c x y) -> (y : b) -> (x : a) -> c x y
```
?

I expect that this will make type inference less likely to succeed and lead to confusing errors.

joelberkeley:
no. I can see that being unwise, for your reasons, and because it's harder to read. Would that be necessary? I had hoped/expected that it would be possible to modify the compiler/type checker to realise this is possible

gallais:
> I had hoped/expected that it would be possible to modify the compiler/type checker to realise this is possible

The type of `flip` is explicitly about non-dependent functions so there is no
hope that a change in the compiler could give you what you want.

joelberkeley:
hmmm, ok in that case I'll make a local `dflip` with that signature

joelberkeley:
btw i think `dflip` might be a useful addition to the stdlib

PR #2601, opened 2022-07-27
add `dflip`, a dependent version of `flip`

joelberkeley:
This came out of https://github.com/idris-lang/Idris2/issues/2589. I thought, since @buzden gave it a 👍🏻 I'd see if there was wider support for it

I'm very aware this is in one of the most core modules in the stdlib, so we'll want to be sure it's worth adding.

### DIFF
--- a/libs/base/Control/Function.idr
+++ b/libs/base/Control/Function.idr
@@ -2,6 +2,12 @@ module Control.Function
 
 %default total
 
+||| Takes in the first two arguments in reverse order. This is a dependent version of `flip`, where
+||| the return type can depend on the function inputs.
+public export
+dflip : {0 c : a -> b -> Type} -> ((x : a) -> (y : b) -> c x y) -> (y : b) -> (x : a) -> c x y
+dflip f x y = f y x
+
 ||| An injective function maps distinct elements to distinct elements.
 public export
 interface Injective (f : a -> b) | f where


andrevidela:
The goal of `prelude` is to be as small as possible. `dflip` hasn't been something that's been asked for before so there is little reason to conclude it needs to exist in `prelude`. It would be nice to have but in that case, I would suggest to have it in `base` or maybe even `contrib`

joelberkeley:
Happy to move it. Just let me know where

andrevidela:
To go remain consistent with #2456 We should place this in the same spot. I suggested `Basics.Dependent`, I suggest you chime in the thread to tell us what you think of this namespace for it

gallais:
`Control.Function`, maybe?
(I don't why it's in `Control` rather than `Data` like in Haskell).

andrevidela:
That would be fine but then what would be the namespace for #2456?

gallais:
Ah shoot the derive_traversable test is leaking information about the internal state of the compiler. :(
Strange that the printing of `EqMap` uses named application rather than auto's `@{}`. :thinking:

andrevidela:
Are we putting this in `Basics.Dependent`? @gallais @joelberkeley

gallais:
I don't have a strong opinion one way or the other.

joelberkeley:
I also don't have a strong opinion, though I would find it odd there being a `Basics.Dependent` without anything in `Basics`

joelberkeley:
> Ah shoot the derive_traversable test is leaking information about the internal state of the compiler. :( Strange that the printing of `EqMap` uses named application rather than auto's `@{}`. thinking

@gallais am i right in thinking that's not to do with this change? Should I wait for that to be fixed elsewhere?

gallais:
The canonical solution is to use a bit of `sed` to drop numbers in the output

joelberkeley:
> The canonical solution is to use a bit of `sed` to drop numbers in the output

sorry I'm not following at all

gallais:
For instance: https://github.com/idris-lang/Idris2/blob/main/tests/idris2/basic044/run

joelberkeley:
yeah I don't want this PR that much. Can close if no-one feels strongly

PR #2604, opened 2022-07-30
Reimplement %nomangle in terms of %export

Z-snails:
Also deprecate %nomangle


Z-snails:
@stefan-hoeck just a heads up, this appears to effect elab-utils

PR #2605, opened 2022-07-30
enforce `size` is non-negative in `malloc`

joelberkeley:
Spoke to @andrevidela about using `Nat`. This seemed preferable.

### DIFF
--- a/libs/base/System/FFI.idr
+++ b/libs/base/System/FFI.idr
@@ -58,7 +58,7 @@ prim__free : AnyPtr -> PrimIO ()
 |||
 ||| @ size the number of bytes to allocate
 export
-malloc : HasIO io => (size : Int) -> io AnyPtr
+malloc : HasIO io => (size : Int) -> {auto 0 ok : So (size >= 0)} -> io AnyPtr
 malloc size = primIO $ prim__malloc size
 
 ||| Release memory with libc `free`.


andrevidela:
I'm not sure this is actually preferable. I think our conversation was in the context of writing your own FFI, not changing the standard way we expose FFI to Idris users. In my mind it feels more natural to have the FFI exposed be as close as the actual C code and have users wrap their own API around it as they see fit.

In general I'm not sure I'm the best to weigh in since I haven't used the FFI in a long time. Maybe someone else who did use it recently can share their thoughts?

joelberkeley:
I understand what you mean, and think the goal of having it close to the C code makes sense, but I think this is closer to the actual C code
```c
void *malloc(size_t size)
```
where

> According to the 1999 ISO C standard (C99), `size_t` is an **unsigned** integer type of at least 16 bit (see sections 7.17 and 7.18.3).

(emphasis mine)

andrevidela:
I mean closer in semantics, the problem being that you can't write this anymore

```c
char* stringCopyAllocate(char * input) {
  int len = strlen(input);
  char* newStr = malloc(len * sizeof(char));
  // do the copy
  return newStr;
}
```

because you need a proof that `len > 0` which isn't what C requires you to do. Of course you can cheat your way through it with `malloc {ok = believe_me ()}` but I'm not sure we would have gained anything in that particular situation.

joelberkeley:
That's an interesting example, because both `strlen` and `sizeof` return `size_t`. I see your point, and I can see that using a proof isn't the right way to go, but I do feel quite strongly a bare `Int` is wrong. As an FFI use, I'd like Idris to catch at least the same errors that C does. Perhaps `Bits64` would be better, perhaps not. Perhaps Idris needs a practical strictly positive integer for these situations.

Issue #2552, opened 2022-06-21
Memory leak in idris2_readLine


dunhamsteve:
Idris appears to be leaking memory in FFI on scheme. This was first noticed because the LSP process was growing as I was working on Idris. Repeatedly reloading a touched `Desugar.idr` caused the process to grow in size. Further investigation revealed that both the `leaks` app on macos and `valgrind` on linux report that memory is leaked by the `delim` function in `iogetdelim.c`. I believe the malloc'd string needs to be free'd

# Steps to Reproduce

In the Idris2 source directory:

On Linux run:
```
valgrind --leak-check=full "$HOME/.idris2/bin/idris2_app/idris2.so" --find-ipkg src/Idris/Desugar.idr < /dev/null
```

On MacOS:
```
MallocStackLogging=1 idris2 --find-ipkg src/Idris/Desugar.idr
```

and while it is running,  run `leaks` with the PID.

# Expected Behavior

Expect no leaks to be reported:

# Observed Behavior

MacOS reports:
```
leaks Report Version: 4.0, multi-line stacks
Process 86396: 79532 nodes malloced for 79325 KB
Process 86396: 79229 leaks for 81130496 total leaked bytes.

STACK OF 34906 INSTANCES OF 'ROOT LEAK: <malloc>':
7   dyld                                  0x231e3fc10 start + 2368
6   scheme                                0x1046ec088 main + 2512
5   scheme                                0x10478002c run_script + 296
4   scheme                                0x10477fc64 boot_call + 44
3   ???                                   0x107929e60 0x7fffffffffffffff + 9223372041276792417
2   libidris2_support.dylib               0x104b5a4f4 idris2_readLine + 32
1   libidris2_support.dylib               0x104b5a0b4 getdelim + 72
0   libsystem_malloc.dylib                0x1a58ef42c _malloc_zone_malloc + 272
```

Linux reports:
```
==10369== LEAK SUMMARY:
==10369==    definitely lost: 8,729,128 bytes in 72,582 blocks
==10369==    indirectly lost: 0 bytes in 0 blocks
==10369==      possibly lost: 17,880 bytes in 149 blocks
==10369==    still reachable: 1,967 bytes in 8 blocks
==10369==         suppressed: 0 bytes in 0 blocks
==10369== Reachable blocks (those to which a pointer was found) are not shown.
==10369== To see them, rerun with: --leak-check=full --show-leak-kinds=all
```

CodingCellist:
[...]

PR #2606, opened 2022-07-31
[ fix ] Fixes memory leaks in currentDir, fGetLine, and fGetChars

dunhamsteve:
This PR fixes the memory leak described in #2552 in addition to two other leaks.

The `fGetLine` function leaks 1kb for every line of text read by the compiler.  This happens because scheme is not freeing pointers returned from the C FFI.  (And, in the case of the getenv FFI, we don't want it to.)  This can be a problem with long running sessions like the REPL or idris2-lsp that reload lots of files.

The leaks and the fix were verified with the MacOS `leaks` utility and Linux `valgrind` utility (see #2552 for details).

### DIFF
--- a/libs/base/System/Directory.idr
+++ b/libs/base/System/Directory.idr
@@ -3,6 +3,7 @@ module System.Directory
 
 import System.Errno
 import public System.File
+import System.FFI
 
 %default total
 
@@ -83,7 +84,9 @@ currentDir
     = do res <- primIO prim__currentDir
          if prim__nullPtr res /= 0
             then pure Nothing
-            else pure (Just (prim__getString res))
+            else do let s = prim__getString res
+                    free $ prim__forgetPtr res
+                    pure (Just s)
 
 ||| Try to open the directory at the specified path.
 export
--- a/libs/base/System/FFI.idr
+++ b/libs/base/System/FFI.idr
@@ -52,6 +52,7 @@ setField s n val = primIO (prim__setField s n fieldok val)
 prim__malloc : (size : Int) -> PrimIO AnyPtr
 
 %foreign "C:idris2_free, libidris2_support, idris_memory.h"
+         "javascript:lambda:()=>undefined"
 prim__free : AnyPtr -> PrimIO ()
 
 ||| Allocate memory with libc `malloc`.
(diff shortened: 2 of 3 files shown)


libs/base/System/FFI.idr
```diff
@@ -52,6 +52,7 @@ setField s n val = primIO (prim__setField s n fieldok val)
 prim__malloc : (size : Int) -> PrimIO AnyPtr
 
 %foreign "C:idris2_free, libidris2_support, idris_memory.h"
+         "javascript:lambda:()=>undefined"
```
andrevidela, line 55: won't this turn use of `prim_free` from a compile-time error to a runtime error? Is this what we want?

dunhamsteve: My understanding is that there isn’t a C FFI in node, so all of these pointers are memory managed. (E.g. readLine is getting a proper GC managed JavaScript string instead of a void*.)

I don’t believe there are any naked pointers with the JavaScript backend, but if there were, the failure mode here would be to leak memory (this is a noop free).

dunhamsteve:
I’ve also double checked the refc backend under valgrind to make sure this doesn’t introduce a use-after-free.

libs/base/System/Directory.idr
```diff
@@ -83,7 +84,9 @@ currentDir
     = do res <- primIO prim__currentDir
          if prim__nullPtr res /= 0
             then pure Nothing
```
gallais, line 86: Should there be freeing in the unhappy path too?

dunhamsteve: In the unhappy path the pointer is null, so it doesn't need to be freed. (The man page says that free does nothing if passed null, so we could safely call it if we want.)

dunhamsteve:
The situation as I understand it:

As long as we have an FFI exposing bare C calls, we're going to have some `void*` (scheme name for a returned type) that we're responsible for freeing, and others that we should never free (eg getenv).

When the return type is declared as `string` in scheme, we do get a fresh copy as a GCed scheme string, but the original is not freed.

This PR fixes a few leaks, but eventually we should come up with a general solution that we can apply consistently across primitives and backends. One possibility is to have a separate type (like GCPtr) that lets the backend know that it should GC and free the pointer (wrapping it if necessary). Currently GCPtrs can't be returned from FFI.

For the cases addressed here, it is our C code that is doing the malloc. The underlying API takes a buffer. So another option is to lift that buffer allocation into Idris and pass it down. But that would change the existing primitive API.

libs/base/System/FFI.idr
andrevidela: Ah yes that makes sense!

PR #2609, opened 2022-08-02
[RefC] Add 16 and 32 bit access to  base/Data.Buffer .

seagull-kamome:
Some FFI definitions are missing in base/Data.Buffer.
This PR adds 16 and 32 bit access routines.


libs/base/Data/Buffer.idr
```diff
@@ -134,7 +140,20 @@ getBits64 : HasIO io => Buffer -> (offset : Int) -> io Bits64
 getBits64 buf offset
     = primIO (prim__getBits64 buf offset)
 
+%foreign "scheme:blodwen-buffer-setint16"
+         "RefC:setBufferWord16LE"
```
gallais, line 144: Should this bit `Int16` rather than `Word16`?

gallais (commented):
What's the convention behind the `*LE` names? Why is it not uniform?

CodingCellist (changes requested):
As gallais suggested, I would also prefer if the `Word` functions were renamed to `UInt` since this describes the C primitive they operate on (unless there is a specific reason for using `Word` instead.

buzden:
> > What's the convention behind the `*LE` names? Why is it not uniform?
> 
> I suspect it's "Little Endian". It won't matter in 99% of the cases, since most modern machines are LE

Not only machines have set endianness, but also protocols. Say, if one would use `Data.Buffer` to form a network packet where usually Big Endian is used disregarding endianness of the machine where the code is executed, one would definitely want to understand the order bytes are put into the buffer.

PR #2610, opened 2022-08-02
[RefC] add missing symbols of showcCleanStringChar

seagull-kamome:
Compiler.RefC.Refc.showcCleanStringChar  forgot some symbols.  
This PR fixes that  and avoids the compiler error.


gallais:
It's hard to see what the actual changes are because some lines seem to just be
moving around for no (?) good reason.

seagull-kamome:
It is sorted in ASCII code order for readability.

PR #2618, opened 2022-08-10
[ cleanup ] Remove extraneous show and simplify printLn

ngeiswei:
The primary goal of that commit is to remove extraneous `show` calls while combined with `printLn` in some unit tests, which even though do not hurt, are a bit confusing for the novice.

The simplifications of `print` and `printLn` are questionably useless but I thought I'd include them too, for the fame perhaps.


ngeiswei:
I had to replace "5" by 5 in the expected result of test basic048.  I believe this is the correct thing to do, unless the creator of that test precisely wanted to test showing a string, which would be surprising, I suppose.

ngeiswei:
Well, I had to fix other tests.  I'm now doubting that this PR makes sense.  You'll be the judge.

ngeiswei:
It finally passes all tests.  Maybe this PR is not worth the trouble, for me it's educational, but I feel I'm a bit wasting everyone's time.  Do what you may with it.

PR #2619, opened 2022-08-10
[ fix ] more filtering of invalid datatypes

gallais:


buzden:
I thought of possibly wrongly rejected example like

```idris
data XList : (u : Nat -> Nat) -> (a : Type) -> Type where
  Nil : XList u a
  (::) : a -> XList (\a => a) a -> XList u a
```

but it seems no lambdas in types are supported, thus no shadowing possible.

PR #2621, opened 2022-08-12
[ fix ] Add missing reserved words to JS backend

emccorson:
Adds missing reserved keywords to JS backend so they are escaped properly in the output JavaScript.

Adds:

- all keywords that are always reserved (e.g. `this`...)
- all keywords that are reserved only in strict mode (e.g. `let`...)
- all keywords that are reserved only in modules or async functions (`await`)
- all future reserved keywords (`enum`)
- all future reserved keywords that are reserved only in strict mode (e.g. `public`...)

Reference list: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#keywords

### DIFF
--- a/src/Compiler/ES/Codegen.idr
+++ b/src/Compiler/ES/Codegen.idr
@@ -81,10 +81,12 @@ jsIdent s = concatMap okchar (unpack s)
 
 jsReservedNames : List String
 jsReservedNames =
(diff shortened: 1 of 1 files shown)


emccorson:
Thanks for the merge!

PR #2622, opened 2022-08-14
Auto-Implicit case split

mattpolzin:
Prior to this PR, case splitting on an auto-implicit in a function clause resulted in invalid Idris syntax.

_Before case split:_
```idris
test : Bool => ()
test @{x} = ?h
```
_After case split:_
```idris
test : Bool => ()
test @{x = False} = ?h_0
test @{x = True} = ?h_1
```

This PR fixes auto implicit case splitting so you get the desired result.

_Before case split:_
```idris
test : Bool => ()
test @{x} = ?h
```
_After case split:_
```idris
test : Bool => ()
test @{False} = ?h_0
test @{True} = ?h_1
```


CodingCellist (approved):
Nice! I didn't realise auto-implicits had different pattern-match syntax (probably because I never thought of splitting on them :sweat_smile:)

MarcelineVQ:
> Nice! I didn't realise auto-implicits had different pattern-match syntax (probably because I never thought of splitting on them sweat_smile)

I've always named them first and split on them like any other named implicit because it never occurred to me that `@{ }` could be lhs syntax, iow I'd have:
```idris
foofer : (baz : Nat) => Char
foofer {baz = S k} = ...
```
I do wonder if it's neccesary to have though, trading making a name in exchange for different syntax and the argument being positional. Guess it comes back to discussions like https://github.com/idris-lang/Idris2/pull/907 in a way, and some others I can't locate at the moment

buzden:
> I do wonder if it's neccesary to have though, trading making a name in exchange for different syntax and the argument being positional.

I think it's simply because not all `auto`-parameters are named and it is convenient.

Issue #2607, opened 2022-07-31
Warn when names in a single claim / declaration shadow each other


ohad:
# Steps to Reproduce
```idris
record Bar (a : Type) where
  constructor MkBar
  a : Nat
```

# Expected Behavior
Depends on language design. Either:

* type-check fine, and be equivalent to the alpha-converted code that type-checks fine:
```idris
record Bar (b : Type) where
  constructor MkBar
  a : Nat
```

* Report a warning that the field named `a` shadows an earlier argument called `a`.

# Observed Behavior
```
- + Errors (1)
 `-- (Interactive) line 1 col 0:
     While processing constructor MkBar. When unifying:
         Nat
     and:
         Type
     Mismatch between: Nat and Type.
```

PR #2623, opened 2022-08-15
Warning for shadowed local bindings

Z-snails:
Related to #2607 (doesn't solve it due to warnings not being emitted when errors are)


gallais:
I've fixed the typo/build issue.
@Z-snails Are we happy with the current setup? I am.

Z-snails:
yep I'm happy with this, I'll merge it shortly

PR #2630, opened 2022-08-24
[ base ] implement `decEq` for `SnocList`

joelberkeley:


libs/base/Decidable/Equality.idr
```diff
@@ -111,6 +111,17 @@ public export
...
 
+--------------------------------------------------------------------------------
+-- SnocList
+--------------------------------------------------------------------------------
+
+public export
+DecEq a => DecEq (SnocList a) where
+  decEq Lin Lin = Yes Refl
+  decEq (xs :< x) Lin = No absurd
+  decEq [] (xs :< x) = No absurd
```
buzden, line 122: Did you want `[<]` instead of `[]` here? Types should not match with this

joelberkeley: yes, fixed. Is `[<]` the same as `Lin`?

libs/base/Data/SnocList.idr
```diff
@@ -21,6 +21,14 @@ public export
 asList : SnocList type -> List type
 asList = (reverse . cast)
 
+export
+Uninhabited (Lin = x :< xs) where
+  uninhabited Refl impossible
+
+export
+Uninhabited (x :< xs = Lin) where
+  uninhabited Refl impossible
```
joelberkeley, line 30: i've made a rough guess where these and `Biinjective` should go

libs/base/Decidable/Equality.idr
buzden: > yes, fixed. Is `[<]` the same as `Lin`?

It is

libs/base/Data/SnocList.idr
buzden: It probably in a good place, but you should include `Data.SnocList` in `Decidable.Equality` then

gallais:
This seems sensible. Once the build failure is fixed, I'll be happy to merge it.

joelberkeley:
whoops - sorry I assumed this was a global build issue that needed fixing elsewhere. I see you've picked this up now

PR #2631, opened 2022-08-24
[ fix ] Fix `Show` of `TTImp` for functions with `with` clauses

buzden:
Thanks to @AlgebraicWolf for finding this.

*All* applications in LHS of `WithClause` and `PatClause` under `WithClause` are coded with `IApp`, not with `IWithApp`. Thus, show for, say

```idris
`[f a b with (expr) { f a b | x = one; f a b | y with (expr2) { f a a | y | z = e; f a a | y | zz = ee } }]
```

gives

```
"[f a b with (expr) { f a b x = one; f a b y with (expr2) { f a a y z = e; f a a y zz = ee } }]"
```

This PR fixes `Show` to print

```
"[f a b with (expr) { f a b | x = one; f a b | y with (expr2) { f a a | y | z = e; f a a | y | zz = ee } }]"
```

I decided to reuse existing utility `showIApps` for this, anyway, it was used only in one place before and fits the new purpose well.


gallais:
I suppose this should ideally be fixed in the unelaboration pass.
Because this won't catch with-apps on the RHS for instance!

buzden:
> Because this won't catch with-apps on the RHS for instance!

I don't understand the case. With-apps on the RHS are coded with `IWithApp` and are showed correctly.

Or, do you suppose to encode matches of `with`'ed expressions on the LHS with `IWithApp` too?

gallais:
> I don't understand the case. With-apps on the RHS are coded with IWithApp and are showed correctly.

Oh really? So it's strange we're failing to do it on the LHS then!

> Or, do you suppose to encode matches of with'ed expressions on the LHS with IWithApp too?

Yes, that's what I had in mind. I assumed it wasn't working on the LHS because it wasn't working anywhere else.
But if we have the machinery to produce the correct encoding then I'd rather we use it instead of changing the
printing!

buzden:
I finally sorted this thing out, and the fix was changed to be as we've discussed some time ago. Now, `Show` implementation remains untouched, and with-applications on LHS are changed to be parsed (and thus, stored at `TTImp`) as with-applications (i.e. with `PWithApp` and `IWithApp` accordingly), like they are on RHS.

Note, that generally speaking, despite this is a fix, it is a slightly breaking change: all elaboration scripts that relied on the previous strange encoding of LHS will now break. I know this from first hands, because I'm author of one of such scripts in DepTyCheck :-)

PR #2637, opened 2022-08-31
[ base ] Implement `MonadError ()` for `MaybeT`

buzden:
`Either e` and `EitherT e` have similar implementations for `MonadError e` and that's great. `Maybe` implements `MonadError ()` and this is great too. However, we don't have similar implementation for `MaybeT`.

I think, this should not clash with existing dropping-into implementation. Moreover, I think, one of the important reasons why one may use `MaybeT` in his stack is implementation for `MonadError ()`, that's why I think this implementation should be unnamed.


gallais:
Thanks!

PR #2647, opened 2022-09-06
[ test ] Fix timing issue in interactive045 test

dunhamsteve:
This is similar to #2644 - the test interactive045 randomly fails because the newly copied .idr file sometimes ends up with the same timestamp as its .ttc file.

### DIFF
--- a/tests/idris2/interactive045/run
+++ b/tests/idris2/interactive045/run
@@ -2,6 +2,7 @@ rm -rf build
 rm -f Issue1742Gen.idr
 
 cp Issue1742.idr Issue1742Gen.idr
+sleep 1
 
 $1 --no-color --console-width 0 --no-banner Issue1742Gen.idr < input
 


gallais:
Thanks. Do you know if there is any way we can use a more precise timestamp and be done with this issue once and for all?

gallais:
I see we're using `st_mtime` so we're stuck with that resolution (the js backend does get a value in ms).

dunhamsteve:
The false positives aren’t that bad for a normal compile, but it is annoying that we have to tiptoe around this for tests. 

It looks like both macOS and Linux have a nsec value available, but the names are different. We’d need a conditional and would need to change that FFI call to be in either ms or ns.  On Linux the nsec part is `buf.st_mtim.tv_nsec` and on macOS it is `buf.st_mtimespec.tv_nsec`.  I don’t have a windows system handy to check there.  Google says that the BSDs have a `st_mtimensec`. I see that in the macOS header, too, but it’s behind a conditional.

PR #2650, opened 2022-09-07
[ fix ] showTime pads nanoseconds properly

MarcelineVQ:
`showTime 4 12 $ makeDuration 312 123456` would previously give

`0312.123456000000` instead of `0312.000123456000`

Meaning the magnitude of nanoseconds were rarely accurate

### DIFF
--- a/libs/base/System/Clock.idr
+++ b/libs/base/System/Clock.idr
@@ -72,12 +72,11 @@ Show (Clock type) where
 export
 showTime : (s, ns : Nat) -> (clk : Clock type) -> String
 showTime s ns (MkClock seconds nanoseconds) =
-  let seconds = show seconds
-      quotient : Integer = cast $ 10 `power` minus 9 ns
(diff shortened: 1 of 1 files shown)


libs/base/System/Clock.idr
```diff
@@ -77,7 +77,7 @@ showTime s ns (MkClock seconds nanoseconds) =
       nanoseconds = show (cast nanoseconds `div` quotient)
   in concat [ padLeft s '0' seconds
             , "."
-            , padRight ns '0' nanoseconds
+            , padRight ns '0' (padLeft (min ns 9) '0' nanoseconds)
```
gallais, line 80: Isn't the right approach something like
```
padRight ns '0' $ take ns $ padLeft 9 '0' nanoseconds
```
?

MarcelineVQ: I see, I wasn't thorough enough in my tests, drat. Certainly your change is the correct one given more test cases. Will update with further corrections.

Issue #2616, opened 2022-08-04
Javascript backend emitted code that access global variable


locriacyber:
If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!

# Steps to Reproduce

mod.idr:
```
main : IO ()
main = printLn (mod 3 4)
```

```
> idris2 mod.idr --cg javascript
     ____    __     _         ___
    /  _/___/ /____(_)____   |__ \
    / // __  / ___/ / ___/   __/ /     Version 0.5.1-676ca0dc0
  _/ // /_/ / /  / (__  )   / __/      https://www.idris-lang.org
 /___/\__,_/_/  /_/____/   /____/      Type :? for help

Welcome to Idris 2.  Enjoy yourself!
Main> :compile out2 main
File build/exec/out2 written
```
# Expected Behavior
there should be `let` in front
```
  let r = a % b
```
# Observed Behavior
In build/exec/out2:
```

// Euclidian Modulo
const _mod = (a,b) => {
  r = a % b
  return r < 0 ? (b > 0 ? r + b : r - b) : r
}

const _modBigInt = (a,b) => {
  r = a % b
  return r < 0n ? (b > 0n ? r + b : r - b) : r
}
```

stefan-hoeck:
This is expected behavior. Idris, unlike JavaScript,  uses euclidian modulo. See the discussion here: #1048 and #1480.

locriacyber:
The problem is not about modulo. The problem is using `r` (global variable) without `let`.

locriacyber:
@stefan-hoeck don't use `%` at all.

The terrible language doesn't have `%` for BigInt.

https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-applystringornumericbinaryoperator

stefan-hoeck:
According to the tests we run, `%` works fine. So, according to your Edit: Can we use the operator or not? I'm afraid your messages are often rather cryptic...

locriacyber:
We can, probably. It works in Node.JS and Firefox for me.

It's the ECMA standard for JS that's "buggy".

PR #2653, opened 2022-09-12
[ fix ] issue 2616

stefan-hoeck:
This is a trivial fix of #2616, a potentially serious bug in the JS backends.


mattpolzin (approved):
Good stuff.

PR #2657, opened 2022-09-14
[ base ] Add a variant of `System.run` allowing processing on the fly

buzden:
We have function `system` that runs a command and its output would be just printed as an output of the caller program, and we have `run` function that allows to process the whole output of the command. But sometimes commands either have too big output or simply are long running, while there is a demand of processing their output. I suggest to have a function that allows to process program output line by line along with the execution of the command, without obligatory storage of the whole output in the memory. Also, this allows to already react appropriately on some output during the execution of long-running command.

### DIFF
--- a/libs/base/System.idr
+++ b/libs/base/System.idr
@@ -182,6 +182,36 @@ namespace Escaped
   run : HasIO io => (cmd : List String) -> io (String, Int)
   run = run . escapeCmd
 
+||| Run a shell command, allowing processing its stdout line by line.
+|||
+||| Notice that is the line of the command ends with a newline character,
+||| it will be present in the string passed to the processing function.
+|||
+||| This function returns an exit code which value should be consistent with the `run` function.
+export
+covering
+runProcessingOutput : HasIO io => (String -> io ()) -> (cmd : String) -> io Int
+runProcessingOutput pr cmd = do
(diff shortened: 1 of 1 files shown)


libs/base/System.idr
```diff
@@ -182,6 +182,36 @@ namespace Escaped
...
+|||
+||| This function returns an exit code which value should be consistent with the `run` function.
+export
+covering
+runProcessingOutput : HasIO io => (String -> io ()) -> (cmd : String) -> io Int
+runProcessingOutput pr cmd = do
+  Right f <- popen cmd Read
+    | Left err => pure 1
+  True <- process f
+    | False => pure 1 -- we do not close `f` in case of reading error, line `run` does
```
gallais, line 198: I don't understand this comment

buzden: In the original `run` function `fclose f` is called only then `fRead f` returns `Right _`. I thought this was done intentionally (I don't know for sure the rules of when `fclose` should be called or not) and I repeated the same behaviour for this function, i.e. if any file error raises *during* the processing of the output of the spawned command, `fclose` is not called for `f`.

gallais: Ah, `like` not `line

buzden: oops

PR #2658, opened 2022-09-14
[ fix #2655 ] Add support for DataOpts in records

### LINKED ISSUE #2655: Allow data opts in record definitions
At the moment if we want to pass data options to a record type, we need to do the manual desugaring ourselves.
```
record Wrap (phantom : Type) (a : Type) where
  constructor MkWrap
  [search a]
  unWrap : a
```
should be valid syntax.

Spotted by @CodingCellist

gallais:


buzden:
Looks like you forgot to push test's body

PR #2659, opened 2022-09-14
[ papers ] A Completely Unique Account of Enumeration

gallais:
TODO:

* [ ] Correctness proofs (completeness + uniqueness) [I'm not planning to do this myself]
* [ ] Examples
* [ ] Use this for performance testing


gallais:
The example is pretty horrible because you need to write the inductive family of
intrinsically typed terms in the encoding so I'm giving up on it & merging as is.

PR #2662, opened 2022-09-15
[ performance ] constant folding for all integral expressions

stefan-hoeck:
This PR enables constant folding for all integral expressions (with the exception of those of type `Int`, because they are not the same across all backends) and casts (not only lossless ones). In addition, all `Cast` implementations involving primitives as well as `believe_me` are annotated with `%inline` in the Prelude.

Background:

Consider the following minimal example:

```idris
main : IO ()
main = printLn (the Bits8 0xff + 100)
```

The parts in parentheses are converted to the following scheme expression:

```scheme
(bu+ (blodwen-toUnsignedInt 255 8) (blodwen-toUnsignedInt 100 8) 8)
```

Even worse, on the JS backend, this goes via `BigInt` literals with an even larger negative impact on performance:

```javascript
_add8u(Number(_truncUBigInt8(255n)), Number(_truncUBigInt8(100n)))
```

Therefore, every integer literal with a precision of <= 32 bits comes with the overhead of going from `BigInt` to `Number` on the JS backend.

I mark this as WIP, because I'd like to add a suitable test case for this.


stefan-hoeck:
This is now ready for review.

PR #2668, opened 2022-09-17
[ new ] Handle forward declarations of implementations

dunhamsteve:
### Problem

Idris2 has syntax for forward declaring interface implementations by omitting the `where` at the end of the line. But it is not currently usable because the subsequent definition of the implementation will throw an `AlreadyDefined` error.  This PR fixes that by not emitting a second declaration, but checking that it converts to the first one.  This is the approach taken by forward declarations for `data`.

The patch should be double checked by someone more knowledgeable than me, but I believe it is correct, and it gets us closer to the goal of not requiring mutual.

Also note that, due to a parser bug that will be fixed by #2667, forward declarations of implementations must be followed by a declaration that begins with a keyword.

### Root Cause

In the parser `implDecl` a missing `where` is indicated by a `Nothing` for the body. The elaborator in `Idris/Elab/Implementation.idr` checks for `Nothing` and skips emitting the body, with the expectation that it will be filled in later. Unfortunately, when it is filled in later the first pass is run again redefining the implementation. Forward declarations for `data` work in a similar way, but it just checks that the types are the same rather than re-declaring the constructor.

### Solution

I took inspiration from `processData` in `TTImp/ProcessData.idr` and added a check whether the definition was already present and conversion check between the existing definition and the current one to ensure that they are the same. When the types don't convert, `ProcessData` throws `AlreadyDefined`, but I chose to throw a `CantConvert` For clarity.

An alternative approach would be to add a new `ElabOpt` and push this check down into `ProcessType.idr`.

### Results

This patch will allow users to write:
```idris
interface IsEven a where
    isEven : a -> Bool

interface IsOdd b where
    isOdd : b -> Bool

implementation IsOdd Nat

implementation IsEven Nat where
    isEven 0 = True
    isEven (S k) = not $ isOdd k

implementation IsOdd Nat where
    isOdd 0 = True
    isOdd (S k) = not $ isEven k
```
instead of
```idris
interface IsEven a where
    isEven : a -> Bool

interface IsOdd b where
    isOdd : b -> Bool

mutual
  implementation IsEven Nat where
      isEven 0 = True
      isEven (S k) = not $ isOdd k

  implementation IsOdd Nat where
      isOdd 0 = True
      isOdd (S k) = not $ isEven k
```

### DIFF
--- a/src/Idris/Elab/Implementation.idr
+++ b/src/Idris/Elab/Implementation.idr
@@ -198,10 +198,9 @@ elabImplementation {vars} ifc vis opts_in pass env nest is cons iname ps named i
                                       (Just (gType vfc u))
                    let fullty = abstractFullEnvType vfc env ty
                    ok <- convert defs [] fullty (type gdef)
-                   if ok then pure ()
-                         else do logTermNF "elab.implementation" 1 "Previous" [] (type gdef)
-                                 logTermNF "elab.implementation" 1 "Now" [] fullty
-                                 throw (CantConvert (getFC impTy) (gamma defs) [] fullty (type gdef))
+                   unless ok $ do logTermNF "elab.implementation" 1 "Previous" [] (type gdef)
+                                  logTermNF "elab.implementation" 1 "Now" [] fullty
+                                  throw (CantConvert (getFC impTy) (gamma defs) [] fullty (type gdef))
 
          -- If the body is empty, we're done for now (just declaring that
          -- the implementation exists and define it later)
--- a/src/Idris/Elab/Implementation.idr
+++ b/src/Idris/Elab/Implementation.idr
@@ -13,6 +13,7 @@ import Idris.Syntax
 
 import TTImp.BindImplicits
 import TTImp.Elab.Check
+import TTImp.Elab
 import TTImp.ProcessDecls
 import TTImp.TTImp
 import TTImp.TTImp.Functor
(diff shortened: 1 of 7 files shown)


src/Idris/Elab/Implementation.idr
```diff
@@ -181,7 +182,26 @@ elabImplementation {vars} ifc vis opts_in pass env nest is cons iname ps named i
...
+                     | _ => throw (AlreadyDefined vfc impName)
+                   (ty,_) <- elabTerm tidx InType [] nest env
+                                      (IBindHere vfc (PI erased) impTy)
+                                      (Just (gType vfc u))
+                   let fullty = abstractFullEnvType vfc env ty
+                   ok <- convert defs [] fullty (type gdef)
+                   if ok then pure ()
+                         else do logTermNF "elab.implementation" 1 "Previous" [] (type gdef)
+                                 logTermNF "elab.implementation" 1 "Now" [] fullty
+                                 throw (CantConvert (getFC impTy) (gamma defs) [] fullty (type gdef))
```
gallais, line 204:
```suggestion
                   unless ok $ do logTermNF "elab.implementation" 1 "Previous" [] (type gdef)
                                  logTermNF "elab.implementation" 1 "Now" [] fullty
                                  throw (CantConvert (getFC impTy) (gamma defs) [] fullty (type gdef))
```

gallais (approved):

dunhamsteve:
That's a weird build failure. I think there is something wrong with github's macs:
```
Error: fatal: unable to access 'https://github.com/idris-lang/Idris2/': The requested URL returned error: 429
```

PR #2675, opened 2022-09-22
[ papers ] Port model-checking from "Applications of Applicative Proof Search"

CodingCellist:
Based on work by @gallais, this adds the model-checking sections of Liam O'Connor's paper ["Applications of Applicative Proof Search"](https://doi.org/10.1145/2976022.2976030), specifically:

- CTL -- Computation Tree Logic
- GCL -- Guarded Command Language

along with their examples.

This completes the implementation of the paper in Idris  : )


gallais:
I ran into the same issue a few days ago: you're using nested withs but they are unfortunately not supported
by the released version we use for bootstrapping. I wonder whether we should keep limiting ourselves to what
the bootstrapper can handle or whether we should exclude `papers/` from the bootstrapping phase and only
build it once we are running the dev version of Idris 2.

CodingCellist:
I think excluding `papers/` from the bootstrapping phase would make sense. The implementations in there are potentially on the bleeding edge, so having it build with the latest commit rather than the stable release seems sensible to me.

Alternatively, we could wait until release 0.6, which I assume would include adding multi-with to the bootstrap versions?

gallais:
I'll try to open a separate PR modifying the `Makefile` to exclude `papers/`.

CodingCellist:
Assuming #2702 fixes things, CI should pass once that's merged  : )

gallais:
:fireworks:

PR #2677, opened 2022-09-23
[ re #2675 ] Do not build libs/{contribs,papers} during bootstrap

gallais:


gallais:
Looks like this is exposing a (pre-existing?) CI failure in the racket bootstrap.
I've cherry-picked the commit in #2295 to see whether it solves the issue.

buzden:
Does this mean that `contrib` library won't be installed after `make bootstrap && make install` and to make it be installed we need to `make bootstrap && make && make install`? Or this change just makes `make bootstrap && make install` possibly fail as soon as code in `contrib` would use something not supported by the bootstrap version?

CodingCellist:
> Does this mean that `contrib` library won't be installed after `make bootstrap && make install` and to make it be installed we need to `make bootstrap && make && make install`? Or this change just makes `make bootstrap && make install` possibly fail as soon as code in `contrib` would use something not supported by the bootstrap version?

afaict, no: `contrib` _will_ still be installed by `make bootstrap && make install`:
```
[thomas@skidbladnir idris2-dev]$ export SCHEME=chez
[thomas@skidbladnir idris2-dev]$ export PREFIX="$HOME/idris2-minimal"
[thomas@skidbladnir idris2-dev]$ make distclean
  [...]
[thomas@skidbladnir idris2-dev]$ make bootstrap && make install
  [...]
[thomas@skidbladnir idris2-dev]$ $HOME/idris2-minimal/bin/idris2 --no-banner -p contrib
Main> :module Data.List.Lazy
Imported module Data.List.Lazy
Main>
Bye for now!
[thomas@skidbladnir idris2-dev]$ $HOME/idris2-minimal/bin/idris2 --no-banner
Main> :module Data.List.Lazy
Error loading module Data.List.Lazy: Module Data.List.Lazy not found

Main>
Bye for now!
```
`contrib` is found and behaves as expected when using the freshly bootstrapped install  : )

PR #2680, opened 2022-09-24
[ fix ] allow grouping pi with different types in type declaration

andylokandy:
## Observed behavior

```idris
expli : (a : String, b : Nat) -> String
      ^ Expected a type declaration.
```

## Expected behavior

type checkes


gallais:
Build error aside, I don't understand the patch: is it just moving an alternative around?
Why should that make any difference in the absence of `commit`?

andylokandy:
I didn't figure it out either... It just worked for both `(a : Nat) -> (b : Int) -> Type` and `(a : Nat, b : Int) -> Type` after the patch. And I've tried but still can't fix the ide test. 😢

andrevidela:
Are we sure we want this? wouldn't this add more confusing as to what `(,)` means?

I'm also asking this because we could also dedicate this syntax to named positional application rather than type declaration. Thought I haven't formuated a formal proposal but I worry such syntax sugar paints us in an even tighter design corner that we are in

andylokandy:
> I'm also asking this because we could also dedicate this syntax to named positional application rather than type declaration.

I think the named positional application would not overlap on the place where the type declaration could be: grouped pi must be following a `:` (beginning of a type declaration) or a `->` (a binder), while named positional application must be following an ident (a DataCon or a TyCon).

My motivation is to encourage naming all parameters in the function type because it has been painful to read function types like this one:

https://github.com/idris-lang/Idris2/blob/1402194f1405087c7e4f667579aefce2f3de1f41/src/Idris/Elab/Implementation.idr#L449-L454

But giving every parameter a name is also depressing as many parans and arrows have to be typed: `(a: Type) -> (b: Type) -> (c: Type)` comparing to `(a: Type, b: Type, c: Type)`.

gallais:
We discussed it with @edwinb last Thursday and agreed we should probably not have that extra syntax.

PR #2684, opened 2022-09-26
[ new ] put TTC files in a version-tagged directory

gallais:


buzden:
Maybe, then, `--clean` action should have a variant of cleaning files of all TTC versions?

buzden:
By the way, would old `depends` directories made as symlinks to, say, `../../build/ttc`, stop working with this PR?

gallais:
The test suite has examples with `depends`. Are these not satisfactory?

PR #2685, opened 2022-09-27
[ fix ] drop first command-line arg on node

stefan-hoeck:
As discussed on discord: The node backend comes with an additional command line argument listing the path to the `node` executable. This is inconsistent with other backends, which only list the path to the running program as their first command line argument. For consistency, this PR removes said first argument from the list of command line args on the node backend.


andrevidela:
Looks good to me, but can we quickly add it to the changelog for the next version?

stefan-hoeck:
Done.

andrevidela (approved):

PR #2686, opened 2022-09-27
[ fix ] Lexer reports incorrect column numbers

dunhamsteve:
Fixes an issue with the column numbers in lexer tokens. This builds on a fix by @jutaro in #2337 by applying the same fix to `src/Libraries/Text/Lexer/Core.idr` and adding a test case. 

The root cause is that the character array was being reversed, but it was already in reverse order.  This affect the ipkg parser. This does not affect the main idris parser, which uses a function in Tokenizer.idr which already has this fix.  

The issue can be observed by loading the ipkg:
```
package foo
    xxx = 123
```
and noting the the error message mentions column 1 instead of 3
```
idris2 --build foo.ipkg
Uncaught error: Parse errors ([("foo.ipkg":2:1--2:4, "Expected end of input.")])
```


andrevidela (approved):

andrevidela:
Looks good to me!

PR #2688, opened 2022-09-27
[ papers ] Implement a postulate hole by requiring a special interface

buzden:
Recently several modules were added to the `paper` lib, and one of them contains a postulate which leads to a warning each time one builds a compiler. The main reason for this PR is to remove this warning.

Since this is a postulate, one could just make it explicit, say with `believe_me`, and warning would be gone while postulate would stay a postulate. But I decided to suggest an implementation for this hole, since the standard library contains an interface which can be used as an external postulate. As far as I can see, this anyway does not break any idea inside this module, does it?

Unfortunately, for this trick a bunch of lines are indented, thus two-line change looks big in diff.

### DIFF
--- a/libs/papers/Data/W.idr
+++ b/libs/papers/Data/W.idr
@@ -405,64 +405,64 @@ namespace PartitionedSets
 
   namespace Examples
 
-    ORD : Type
-    ORD = PartitionedSets.W (Shape ("zero" || "succ" || "lim")) $ cases
-        ( "zero" .= mkPSet AVoid ()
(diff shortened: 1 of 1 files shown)


gallais:
I'd rather use a `parameters` block with just `etaUnit` as it's weaker than funext.
Postulating funext may send the wrong idea about the applicability of this paper.

buzden:
> I'd rather use a `parameters` block with just `etaUnit` as it's weaker than funext. Postulating funext may send the wrong idea about the applicability of this paper.

Right, that's better, done

PR #2689, opened 2022-09-28
[ fix ] natToFinLt is O(n)

stefan-hoeck:
From a discussion on [discord](https://discord.com/channels/827106007712661524/834611018775789568/1024561930020605972): `natToFinLt` and hence `Data.Fin.fromInteger` is not treated as the identity function, because both `Nat` arguments have quantity `any`. This PR adds a proof that from `So (x < n)` follows `LT x n`. With this, `natToFinLt` can invok `natToFinLT`, which is optimized away as the identity function by the compiler.

This is a draft at the moment, because I'd like to add a test case for this.

### DIFF
--- a/libs/base/Data/Fin.idr
+++ b/libs/base/Data/Fin.idr
@@ -183,12 +183,17 @@ natToFinLT : (x : Nat) -> {0 n : Nat} ->
 natToFinLT Z {prf = LTESucc _} = FZ
 natToFinLT (S k) {prf = LTESucc _} = FS $ natToFinLT k
 
+0 natToFinLemma : (m,n : Nat) -> So (m < n) => LT m n
+natToFinLemma 0 (S k)     = LTESucc LTEZero
+natToFinLemma (S k) (S j) = LTESucc (natToFinLemma k j)
+natToFinLemma (S k) 0 prf impossible
+natToFinLemma 0 0     prf impossible
+
 public export
-natToFinLt : (x : Nat) -> {n : Nat} ->
+natToFinLt : (x : Nat) -> {0 n : Nat} ->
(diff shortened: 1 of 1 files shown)


stefan-hoeck:
Test case added. This is now ready for review.

libs/base/Data/Fin.idr
```diff
@@ -184,11 +184,17 @@ natToFinLT Z {prf = LTESucc _} = FZ
 natToFinLT (S k) {prf = LTESucc _} = FS $ natToFinLT k
 
 public export
-natToFinLt : (x : Nat) -> {n : Nat} ->
+0 natToFinLemma : (m,n : Nat) -> So (m < n) => LT m n
+natToFinLemma 0 (S k)     = LTESucc LTEZero
```
mattpolzin, line 188: Since it needed to get public exported anyway and it seems at least plausibly of more general utility, how about a _slightly_ less cryptic name like `ltIsLT` (only less cryptic by an inch, I admit).

mattpolzin: On second thought, maybe not to my suggestion since this lemma is basically the existing `ltReflectsLT` from `Data.Nat`.

stefan-hoeck: OK, I moved the lemma to `Data.Nat`, giving it another name and dropping the `So`, because this makes the proof more general (in my opinion). I could of course reintroduce to `So` if that's preferred.

stefan-hoeck: I reverted this to using `So` because otherwise this would have been an unnecessary breaking change on `Data.Fin.fromInteger`. It already broke one or more tests in the test suite...

PR #2692, opened 2022-09-29
[ refactor ] use BitsXY for enum types during codegen

stefan-hoeck:
As discussed on [discord](https://discord.com/channels/827106007712661524/827109190266257409/1024579268732264488).

The idea is that `Bits32` is well defined unlike `Int`, and it allows us to get zero cost conversions from an enum type to its `Bits32` representation even for enum types with more than 100 constructors (see the discord thread for an example).

The change is trivial, but it comes with several additional considerations:

- Should we do this for all data constructor tags (using `BitsXY` instead of `Int` in `MkConAlt`)? If yes, I can try to adjust this PR accordingly. This would of course be a considerably larger piece of refactoring.
- Should we reduce the bit size to some other hard upper limit? For instance, supporting only up to 0xffff constructors for a single data type, in which case we could use `Bits16` for this?

An alternative to this would be to support constant-folding for `Int` at least up to a certain maximum value (this was also discussed on discord), in which case we should decide on a minimal bit size for `Int` that all backends should support (something we probably should do anyway, because right now, `Int` is used a lot in the compiler, and a first test showed that is perfectly fine to define an enum type with 100'000 constructors).

If the people in charge think this should first go through some discussion process/feature request in the issue tracker, I must admit that I probably don't have the time to spend on a detailed proposal right now, but I could at least copy these ravings to the issue tracker, in which case this PR can be closed and we keep things as they are for the time being.


gallais:
Should this occurrence also be `B32 (cast tag)`?

https://github.com/stefan-hoeck/Idris2/blob/ad1d9b8d15d6c83042e51bf7a9c57aa91d8eeb2a/src/Compiler/CompileExpr.idr#L313

stefan-hoeck:
> Should this occurrence also be `B32 (cast tag)`?
> 
> https://github.com/stefan-hoeck/Idris2/blob/ad1d9b8d15d6c83042e51bf7a9c57aa91d8eeb2a/src/Compiler/CompileExpr.idr#L313

Yes, I think so. I'll submit a fix later today.

stefan-hoeck:
As suggested on discord by @Z-snails , I made the `BitsXY` type to be used dependent on the number of enum constructors.

PR #2694, opened 2022-09-29
[ cleanup ] `Compiler/*` modules

Z-snails:
cleanup some messy/unclear code
Also use primitives directly where possible, instead of idris's wrappers


CodingCellist:
>  Also use primitives directly where possible, instead of idris's wrappers

Is this significant enough that it should go in the CHANGELOG?  : )

Z-snails:
@CodingCellist this isn't a (idris) user facing change, so I don't think so. After all the wrappers just called the underlying primitives.

PR #2698, opened 2022-10-01
Matching behavior for channelPut for Chez and Racket

samdphillips:
- Make Racket backend channels behave same as Chez.
- Add channel tests from Chez to Racket
- Remove note about differing behavior


tests/Main.idr
```diff
@@ -312,6 +312,7 @@ racketTests = MkTestPool "Racket backend" [] (Just Racket)
     , "semaphores001", "semaphores002"
     , "futures001"
     , "mutex001", "mutex002", "mutex003", "mutex004", "mutex005"
+    , "channels001", "channels002", "channels003", "channels004", "channels005", "channels006"
```
CodingCellist, line 315: Rather than duplicating the tests across the Chez and Racket test pools, I think we should group them under the `AllSchemes` test pool (assuming this doesn't break anything). Around `tests/Main.idr:187`.

tests/chez/channels006/Main.idr
```diff
@@ -0,0 +1,39 @@
...
+  do recv
+     recv
+     recv
+     recv
+     where
+       recv : IO ()
+       recv =
+         do usleep 100000
+            v <- channelGet ch
+            putStrLn $ "< " ++ show v
```
CodingCellist, line 29: Minor stylistic choice, but could you please change the `where` keyword to align with the `do` to make it clear that it's not part of the `do`-block, but instead a local definition used _in_ the `do`-block?  : )

tests/racket/channels006/Main.idr
```diff
@@ -0,0 +1,39 @@
...
+  do recv
+     recv
+     recv
+     recv
+     where
+       recv : IO ()
+       recv =
+         do usleep 100000
+            v <- channelGet ch
+            putStrLn $ "< " ++ show v
```
CodingCellist, line 29: (same comments as `scheme/channels006/Main.idr` (would be resolved by the `AllSchemes` change))

CodingCellist (changes requested):
Thank you for submitting this; can't believe we missed `async-channel` as an option to fix things ^^;;

A couple of "cosmetic" changes requested, but other than that lgtm!  : )

CodingCellist (approved):

PR #2700, opened 2022-10-02
[ base ] Move the worker outside of the Vect's `reverse` definition

0xd34df00d:
I've found myself needing to prove something about the behavior of `Vect`'s `reverse`, and I wasn't able to do that since `go` is a `where`-local binding in `reverse`. Moving it out (along the lines of `Data.List` implementation) simplifies proving things about `reverse`.

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -257,6 +257,17 @@ replaceAtDiffIndexPreserves (_::_) (FS z) (FS w) co y = replaceAtDiffIndexPreser
 -- Transformations
 --------------------------------------------------------------------------------
 
+||| Reverse the second vector, prepending the result to the first.
+|||
+||| ```idris example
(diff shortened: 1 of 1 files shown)


gallais:
If we're going to expose the internals, then I'd rather we use the [`foldl`-based implementation](http://agda.github.io/agda-stdlib/Data.Vec.Base.html#8726)
than this one with a lot of rewriting.

0xd34df00d:
I can do that as a part of this PR.

While we're at it, do you think it's worth adding some lemmas about `reverse` to `base` or `contrib` at all (as a separate PR, of course)? I was thinking about proving `reverse` is involutory, and `reverse (xs ++ ys) = reverse ys ++ reverse xs` — at least that's what I ended up proving in my code.

gallais:
Given the alternative implementation has perf issues & so the PR got stuck,
let's un-abandon this one.

PR #2703, opened 2022-10-07
[ admin ] Update CHANGELOG and CONTRIBUTORS

CodingCellist:
There have been a lot of changes since the last release and people, myself included, tend to forget to document these in the CHANGELOG. Given that some of these changes are potentially breaking (e.g. DataOpts in records, moving various Linear things around), I figured I'd do a bit of housekeeping to make sure missed changes were added to the CHANGELOG (I may have underestimated how many there were; it's been _a while_ since we did a release).

If I've missed anything or explained it poorly, or missed a contributor, please do pile on to this PR! Hopefully it should be smoother sailing afterwards, as long as we remember to record changes along the way ^^;;

(also, we should really do a release at some point...)


stefan-hoeck:
Actually, I think we should have a release as soon as possible. There have been so many improvements, some of which we can't yet use in the compiler or the standard libs due to bootstrapping issues. I also fear (though I have not tested this), that some of the optimizations we implemented are not applied to the compiler during bootstrapping, so that the resulting `idris2` executable might even perform poorly compared to one built with the latest compiler release.

gallais:
Linter, please stop victimising us :sob:

CodingCellist:
I can't believe that was the issue all along... NOTHING was wrong with my configs, they just weren't being read because the docs for the nl linter itself and for super-linter (which calls the nl linter) differ :upside_down_face:  :sob:

CodingCellist:
Right, I'll probably close this at the end of the week, and in the future people should try to remember to update the CHANGELOG when needed, as opposed to submitting PRs against this PR  : )

PR #2704, opened 2022-10-07
[ performance ] More stack safety in the Prelude

stefan-hoeck:
This PR adds transform rules for making `map`, `mapMaybe`, and `filter` for `List` and `SnocList` plus some other utilities tail recursive.

A note about `mapImpl`: The following transform rule is ignored during compilation:

```idris
%transform "tailRecMap" map {f = List} = List.mapTR
```
I checked this with several different variations including the follow, where all arguments are given explicitly, and none of them worked:

```idris
%transform "tailRecMap" map {b} {a} {f = List} f xs = List.mapTR f xs
```

I therefore extracted `map` for `List` and `SnocList` to utility functions `mapImpl`, and applied the transform rules to those, which works nicely as shown in the adjusted test case for the node backend.


stefan-hoeck:
This is strange. Any idea why the windows test fails?

tests/ideMode/ideMode005/expected
```diff
@@ -0,0 +1,8 @@
+51c51
+< 0000f6(:return (:ok "   xs : List Nat
+---
+> 0000fa(:return (:ok "   xs : List Nat
+53c53
+< prf2 : map S xs = ?lgjgk" ((8 4 ((:decor :type))) (13 3 ((:decor :type))) (55 3 ((:decor :function))) (59 1 ((:decor :data))) (61 2 ((:decor :bound))) (64 1 ((:decor :keyword))))) 3)
+---
+> prf2 : mapImpl S xs = ?lgjgk" ((8 4 ((:decor :type))) (13 3 ((:decor :type))) (55 7 ((:decor :function))) (63 1 ((:decor :data))) (65 2 ((:decor :bound))) (68 1 ((:decor :keyword))))) 3)
```
gallais, line 8: You should not commit this. You should run the script in the directory to update the test case.

stefan-hoeck:
Actually, I forgot the correctness proof for `SnocList`s tail recursive append, and indeed, the implementation was wrong. :-)

libs/base/Data/SnocList.idr
```diff
@@ -62,9 +62,26 @@ Show a => Show (SnocList a) where
       show' acc (xs :< x) = show' (show x :: acc) xs
 
 public export
+mapImpl : (a -> b) -> SnocList a -> SnocList b
+mapImpl f Lin = Lin
+mapImpl f (sx :< x) = (mapImpl f sx) :< (f x)
+
+-- Utility for implementing `mapTR`
+mapTR' : List b -> (a -> b) -> SnocList a -> SnocList b
```
gallais, line 70: Should this be called `mapAppend` for uniformity?

stefan-hoeck: > Should this be called `mapAppend` for uniformity?

`Data.SnocList.mapAppend` already exists. But I'm not very good at choosing useful names so I'm happy to change this to something else.

gallais: No one has piped up with a better name and this is a net gain so let's merge it now, we can always rename later on.

Issue #2166, opened 2021-12-03
Mapping a List over a rank 2 function has O(n^2) time complexity


stefan-hoeck:
I found the following (reduced to a minimal example) when mapping over the functor wrapped in a free monad:

# Steps to Reproduce

Run the following program:
```idris
mapK : {0 f,g : _} -> (forall x . f x -> g x) -> List (f a) -> List (g a)
mapK fun [] = []
mapK fun (x :: xs) = fun x :: mapK {f} {g} fun xs

main : IO ()
main = printLn . length $ mapK {f = Maybe} toList vals
  where vals : List (Maybe Bits32)
        vals = map Just [1 .. 100000]
```

# Expected Behavior
Runs in less than a second (same as the standard `map`).

# Observed Behavior
Takes more than 10 seconds on my machine.

The reason can best be explained, when looking at the code generated by the JS backends (the Scheme backends have the same issue):

```javascript
function Test_mapK($0, $1) {
 switch($1.h) {
  case 0: return {h: 0};
  case undefined: return {a1: $0(undefined)($1.a1), a2: Test_mapK($b => $0(undefined), $1.a2)};
 }
}
```
As can be seen, instead of passing the function argument `$0` as is in the recursive call, a new function with an erased argument is being created ($b => $0(undefined)). This leads to k applications of `undefined` in the k-th iteration, hence the observed quadratic time complexity.

stefan-hoeck:
Quick update: If I use an explicit erased argument, the program runs instantaneously.

```idris
mapK : {0 f,g : _} -> ((0 x : Type) -> f x -> g x) -> List (f a) -> List (g a)
mapK fun [] = []
mapK fun (x :: xs) = fun _ x :: mapK {f} {g} fun xs

main : IO ()
main = printLn . length $ mapK {f = Maybe} (\_ => toList) vals
  where vals : List (Maybe Bits32)
        vals = map Just [1 .. 100000]
```

locriacyber:
Why is `mapK` so complicated? Will this work?

```
mapK : (f -> g) -> List f -> List g
mapK fun [] = []
mapK fun (x :: xs) = fun x :: mapK fun xs

main : IO ()
main = printLn . length $ abb
  where vals : List (Maybe Bits32)
        vals = (map Just [1 .. 100000]) ++ [Nothing, Nothing]
        abb = mapK toList vals
```

gallais:
[...]

PR #2707, opened 2022-10-08
[ base ] Change how folds are defined for `Data.Vect`

0xd34df00d:
* Adds dependent folds, where the type of the accumulator depends on the current length of the `Vect`.
* Changes the `Foldable` instance for `Vect` to be expressed via the former.
* Moves the tail-recursive version of `foldr` to a separate function, `foldrTR`.
* As a sanity check, proves that `foldrTR` is equal to `foldr`.
* Changes the definition of `sumR` in `Data.Vect.Properties.Foldr` to accept just `Vect`s, since other foldables might define `foldr` differently, and this function lives in a module with `Vect` in name anyway.

### DIFF
--- a/libs/base/Data/Vect.idr
+++ b/libs/base/Data/Vect.idr
@@ -377,19 +377,41 @@ mapMaybe f (x::xs) =
        Nothing => (  len **      ys)
 
 --------------------------------------------------------------------------------
+-- Dependent folds
+--------------------------------------------------------------------------------
+
+||| Dependent right fold.
+|||
+||| Use this instead of `Foldable`'s `foldr`
+||| if your accumulator type depends on the length of the `Vect`.
+foldrD : (0 accTy : Nat -> Type) ->
+         (f : forall k. a -> accTy k -> accTy (S k)) ->
+         (acc : accTy Z) ->
+         (xs : Vect n a) ->
+         accTy n
+foldrD _     _ acc []        = acc
+foldrD accTy f acc (x :: xs) = f x (foldrD accTy f acc xs)
+
+||| Dependent left fold.
+|||
+||| Use this instead of `Foldable`'s `foldl`
+||| if your accumulator type depends on the length of the `Vect`.
+foldlD : (0 accTy : Nat -> Type) ->
+         (f : forall k. accTy k -> a -> accTy (S k)) ->
+         (acc : accTy Z) ->
+         (xs : Vect n a) ->
+         accTy n
+foldlD _     _ acc []        = acc
+foldlD accTy f acc (x :: xs) = foldlD (accTy . S) f (acc `f` x) xs
+
+--------------------------------------------------------------------------------
 -- Folds
 --------------------------------------------------------------------------------
 
 public export
-foldrImpl : (t -> acc -> acc) -> acc -> (acc -> acc) -> Vect n t -> acc
-foldrImpl f e go [] = go e
-foldrImpl f e go (x::xs) = foldrImpl f e (go . (f x)) xs
-
-public export
 implementation Foldable (Vect n) where
-  foldr f e xs = foldrImpl f e id xs
-  foldl f z [] = z
-  foldl f z (x :: xs) = foldl f (f z x) xs
+  foldr f acc xs = foldrD (const _) f acc xs
+  foldl f acc xs = foldlD (const _) f acc xs
 
   null [] = True
   null _ = False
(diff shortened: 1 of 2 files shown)


stefan-hoeck:
We have to be *very* careful about this, as it looks like it shows quadratic runtime complexity due to #2166. For instance, if I run the following program on my machine with an argument of "100000", it takes 5 seconds (!) to finish:

```idris
module Foldl

import Data.Vect
import System

foldlD : (0 accTy : Nat -> Type) ->
         (f : forall k. accTy k -> a -> accTy (S k)) ->
         (acc : accTy Z) ->
         (xs : Vect n a) ->
         accTy n
foldlD _     _ acc []        = acc
foldlD accTy f acc (x :: xs) = foldlD (accTy . S) f (acc `f` x) xs


fold : (acc -> e -> acc) -> acc -> Vect n e -> acc
fold f acc xs = foldlD (const _) f acc xs

main : IO ()
main = do
  [_,s] <- getArgs | _ => die "Invalid number of args"

  printLn $ fold (+) Z (replicate (cast s) 1)
```

If I replace `fold` on the last line with the current `foldl`, it is instantaneous even for `n = 1000000` (less than 100 ms). So, at the very least, we must not implement `foldl` in terms of `foldlD` at the moment.

stefan-hoeck:
Two more data points: The following, which uses a trick described in #2166 by using a function with an explicit erased argument, does not work here:

```idris
foldlDImpl : (0 accTy : Nat -> Type) ->
           (f : (0 k : Nat) -> accTy k -> a -> accTy (S k)) ->
           (acc : accTy Z) ->
           (xs : Vect n a) ->
           accTy n
foldlDImpl _     _ acc []        = acc
foldlDImpl accTy f acc (x :: xs) = foldlDImpl (accTy . S) (\k => f (S k)) (f _ acc x) xs
```

The only way I got this to run in linear time is by means of `believe_me`, which is clearly unsatisfactory:

```idris
foldlDImpl : (0 accTy : Nat -> Type) ->
             (f : (0 k : Nat) -> accTy k -> a -> accTy (S k)) ->
             (acc : accTy Z) ->
             (xs : Vect n a) ->
             accTy n
foldlDImpl _     _ acc []        = acc
foldlDImpl accTy f acc (x :: xs) = foldlDImpl (accTy . S) (believe_me f) (f _ acc x) xs

foldlD : (0 accTy : Nat -> Type) ->
         (f : forall k. accTy k -> a -> accTy (S k)) ->
         (acc : accTy Z) ->
         (xs : Vect n a) ->
         accTy n
foldlD at f = foldlDImpl at $ \_ => f
```

stefan-hoeck:
OK, I got a O(n) version without `believe_me` by using a helper function for the recursion:

```idris
foldlD : (0 accTy : Nat -> Type) ->
         (f : forall k. accTy k -> a -> accTy (S k)) ->
         (acc : accTy Z) ->
         (xs : Vect n a) ->
         accTy n
foldlD at f acc xs = go acc xs
  where go : at k -> Vect m a -> at (k + m)
        go           x []        =
          rewrite plusZeroRightNeutral k in x

        go {m = S l} x (y :: xs) =
          rewrite sym (plusSuccRightSucc k l) in go (f x y) xs
```

0xd34df00d:
Interesting, thanks for looking into this!

Since I'll eventually need to prove things about functions expressed with `foldlD`/`foldrD`, having a local `where`-bound helper would be unfortunate (how does one name it, after all?). Right now I have two options:

1. Move the `go` out to a top-level `foldlDhelper` or something like that, hoping that the optimizer will be happy about that. I haven't tested yet whether the optimizer is actually happy.
2. Leave `fold{l,r}` expressed directly instead of using `fold{l,r}D`, and just not care about the performance of the dependent folds until the bug is fixed (and still replacing the `foldr` implementation, having the tail-recursive version as a separate function).

Personally, I'd vote for (2), as it seems like a fair share of proofs about `fold{l,r}` implemented directly will still hold when the implementation is replaced with a call to `fold{l,r}D`, and the structure of `foldlD` proposed in the PR seems to be more obvious and more proof-friendly than whatever's needed to work around the optimizer bug.

What do you think?

gallais:
The whole point of this refactoring is that the `foldl`-based presentation does
not require the use of `rewrite`. I wonder whether I could dig up my eta-contraction
pass and see whether that's enough to get this more elegant version the perf it deserves.

PR #2669, opened 2022-09-19
refactor README and installation instructions

joelberkeley:
I've been looking at doing a good overhaul of the docs, to

* make it easier for newcomers
* reduce duplication

 and wanted to get early feedback. Do be completely honest. Is this helpful? I have an idea to work on the docs much more widely, so the approach here (short, to the point) is as important as the content.

Remaining tasks
- [x] mention  _The JavaScript codegen uses the new BigInt, hence Node.js 10.4 or higher is required._ somewhere
- [x] how to include the videos in the README?
- [x] ensure line breaks at 80 ... is this necessary? CI only requires breaks at 400.
- [ ] should we mention any specific installation instructions in the README? pack? from source?
- [ ] should we summarise the main things new to Idris2 in the README?

Issue #2233, opened 2021-12-30
Clarify installation instructions on macOS (M1)


ornamentist:
When following the standard installation instructions there are a few gotchas for installation on macOS with an M1 processor:

Using homebrew `brew install idris2` is currently not working:

```
brew install idris2
Running `brew update --preinstall`...
...
Error: idris2: no bottle available!
You can try to install from source with:
  brew install --build-from-source idris2
Please note building from source is unsupported. You will encounter build
failures with some formulae. If you experience any issues please create pull
requests instead of asking for help on Homebrew's GitHub, Twitter or any other
official channels.
```

I understand this is an issue for the Homebrew maintainer of that brew and not one for the official project, although it does have a bearing on what comes next.

Moving to installation from source with Chez scheme, the formula is also currently not working:

```
brew install chezscheme
Error: chezscheme: no bottle available!
You can try to install from source with:
  brew install --build-from-source chezscheme
Please note building from source is unsupported. You will encounter build
failures with some formulae. If you experience any issues please create pull
requests instead of asking for help on Homebrew's GitHub, Twitter or any other
official channels.
```

Again this is not the responsibility of the Idris2 developers, but it leads to installing from source with Racket. 

```
brew search racket
==> Formulae
minimal-racket           rack                     tracker                  packetq

==> Casks
brackets                                           racket ✔
Warning: Use minimal-racket instead of deprecated racket
```

It's not clear whether minimal racket is enough to build Idris2 so I went with the full racket install. I cloned the Idris2 repository and did a `gmake bootstrap-racket` which now complains about a missing `gmp.h`. 

GMP is installed via Homebrew but the make commands are not finding `gmp.h` so I modified the `config.mk` file to add `-I/opt/homebrew/include` to the `CFLAGS` variable. This allowed the build to complete.

A `gmake install` installed everything in the local directory and not the `config.mk` prefix, so I changed the `PREFIX` make variable to `/Users/stu/.idris2` which did work. This may be because I'm running fish shell and not bash.

Finally, when I run the `idris2` script I get a `idris2 -o hello hello.idr
env: scheme: No such file or directory` error. I'm guessing there's a make variable somewhere that points to a scheme executable which Racket doesn't provide?

The racket executable is at least an arm64 binary so idris2 should be able to create arm64 executables?

ornamentist:
Okay,

[...]

PR #2708, opened 2022-10-09
Update INSTALL.md

buggymcbugfix:
To successfully build Idris2 and run the tests, I had to set some environment variables, see also [this](https://discord.com/channels/827106007712661524/836186747632091148/1027971022865051679) discussion on Discord.

For the benefit of others, I think it makes sense to add this to the Installation instructions.


andrevidela:
Hi there, can you replace your link to discord with a summary of the situation? People on Github might not have access to discord, and it makes it easier to review if all the information necessary is on the PR description wihtout having to follow any links.

Thanks!

CodingCellist:
Related: #2233
and possibly also: #2669

buggymcbugfix:
I merely documented what worked for me. Feel free to close/amend this PR.

joelberkeley:
> ... might already be addressed by: #2669 ?

probably not. I don't think I edit any installation instructions in that, just move them about

gallais:
> I merely documented what worked for me.

Has someone else installed on M1/M2 since October and needed such things?
I can't test this as I don't have access to any such machines.

mattpolzin:
I'll try to confirm some of these steps soon (just got access to an Apple Silicon laptop). Some of these changes were confirmed to be needed by Steve in #2233. Others of these suggestions should AFAIK not be required regardless of architecture (e.g. the LD_LIBRARY_PATH/DYLD_LIBRARY_PATH is set by the wrapper around the Idris scheme executable and should not therefore need to be set by the user prior to install).

INSTALL.md
```diff
@@ -200,12 +210,7 @@ the Racket fork of chez scheme.
 ```sh
 git clone git@github.com:racket/ChezScheme.git
 cd ChezScheme
-git submodule init
-git submodule update
-arch=tarm64osx
-./configure --pb
-make ${arch}.bootquick
-./configure --threads
+./configure
```
mattpolzin, line 213: I'm working through these on an `aarch64-darwin` machine right now and the changes to this section raise a couple of questions for me:
1. Did the changes as listed not work?
2. Has it become a default to configure for thread support? If not, we want to leave the `--threads` in there because it is important that Chez has support for threads.

mattpolzin: I was able to spot the answer to question (2) in the configuration script for ChezScheme. That's a yes. Wouldn't hurt to be explicit, still, but shouldn't break anything not to be.

INSTALL.md
```diff
@@ -46,6 +46,16 @@ If you have an existing Idris 2, go to Step 3. Otherwise, read on...
 Make sure that:
 
 - `$PREFIX/bin` is in your `PATH`
+- `$PREFIX/lib` is in your `LD_LIBRARY_PATH` or `DYLD_LIBRARY_PATH` if on
+  `macOS` (so that the system knows where to look for library support code)
```
mattpolzin, line 50: This was not necessary for me when I bootstrapped or built from previous install on `aarch64-darwin`. It's possible this situation was made better by recent changes to the build scripts, but it's also possible that you found  this change necessary because of a different in my machine's setup and yours or a different in how we were checking whether things were working.

I'd be interested to learn more if you have any recollection after so long!

mattpolzin:
Ok, I just got done doing a run-through on my `aarch64-darwin` machine.

I found (as did the OP of this PR and Steve Dunham) that the changes on lines [52-58](https://github.com/idris-lang/Idris2/pull/2708/files#diff-09b140a43ebfdd8dbec31ce72cafffd15164d2860fd390692a030bcb932b54a0R52-R58) to INSTALL.md were very much needed prior to building Idris 2 and also prior to building a subsequent executable using the RefC backend. Those would be a great addition to the instructions.

I did not need to change the installation steps when I built ChezScheme, so I am not sure if the OP needed to change those steps in order to get things working or just changed them to simplify things. If it's the latter, I have not yet had the time to confirm that skipping many of the steps originally in the INSTALL.md file work on my machine and I would be tempted to say we should keep those instructions the same given that we don't have a second confirmation that the simplification works longterm (i.e. past the initial build & test of the compiler).

Finally, the other change in this PR is to put `$PREFIX/lib` into the `LD_LIBRARY_PATH` or `DYLD_LIBRARY_PATH` and this is not something that I found was needed when building the compiler or building a subsequent project using the RefC backend. Unless we get more details on what circumstances dictated requiring the change, I vote we do not add that bit to the instructions. If we do hear more about the circumstances requiring that change, I would be hopeful that we can make changes to the scripts or compiler themselves to obviate the need for the user explicitly adding the compiler's installed `lib` directory to their library path.

mattpolzin:
This PR has a very important additional instruction that I want to get merged so I went ahead and made the changes I had previously suggested in my review comments.

PR #2710, opened 2022-10-13
[ new ] Allow forward declarations of records

dunhamsteve:
This change allows users to forward declare `record`, like they can for `data`, by omitting the `where` clause. The implementation follows the pattern of same functionality for `data`.  After the change you can do:

```idris
record B
record A where
  b : B
record B where
  a : A
```
instead of
```idris
mutual
  record A where
    b : B
  record B where
    a : A
```

### DIFF
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -985,7 +985,13 @@ mutual
       isNamed Nothing = False
       isNamed (Just _) = True
 
-  desugarDecl ps (PRecord fc doc vis mbtot tn params opts conname_in fields)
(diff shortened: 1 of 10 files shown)


buzden:
Mutual records also have projections cross-available, when your approach, as far as I understand, forward-declares only data. Is it possible to forward-declare projections too?

dunhamsteve:
Do you have an example of the behavior you want? I believe happens with this patch should be exactly the same behavior as mutual blocks. (The code for desugaring the forward declaration was lifted from Mutual.idr.) My initial goal here was to enable not needing mutual (mentioned in the Yaffle docs), but perhaps there is room for improvement in both cases.

PR #2713, opened 2022-10-14
[ fix #2712 ] Don't consider `CyclicMeta` recoverable

### LINKED ISSUE #2712: Case-split generates non-covering pattern match
# Steps to Reproduce

Case-split on `cmd` in the following code:

```idris
data ISM : Type -> Nat -> Nat -> Type where
  CMD_INIT : (n : Nat) -> ISM () n n
  CMD_INCR : ISM () n (S n)

  Pure : (x : t) -> ISM t n n
  (>>=) :  ISM a from to
        -> ISM b to to2
        -> ISM b from to2

foo : ISM t st st -> ()
foo cmd = ?foo_rhs
```

(note that in the declaration of `foo` we've fixed the state `st` to be the same in both 'to' and 'from')

# Expected Behaviour

The code
```idris
foo (CMD_INIT n) = ?foo_rhs_0
foo (Pure x) = ?foo_rhs_1
foo (x >>= f) = ?foo_rhs_2
```
should be accepted as covering, since `CMD_INCR` changes the state.

# Observed Behaviour

Idris incorrectly complains that `foo` is not covering, claiming the `CMD_INCR` case is missing.

If one manually adds the case, it leads to a `Cycle detected in solution of metavariable` error. Adding the case followed by an `impossible` makes the coverage checker happy.

CodingCellist:
Cycle detection when resolving metavariables is considered valid `impossible` cases (see `src/TTImp/ProcessDef.idr:142`), so they should probably also be considered unrecoverable errors since, afaiu, "recoverable" errors are ones where you could reasonably fix the problem, which I don't think is the case for when you've introduced an infinite cycle in a metavariable resolution.


tests/idris2/interactive046/run
```diff
@@ -0,0 +1,4 @@
+rm -rf build
+
+$1 --no-color --console-width 0 --no-banner --check Issue2712.idr
```
gallais, line 3: Either
```suggestion
$1 --no-color --console-width 0 --no-banner Issue2712.idr < input
```
Or drop the `input` file.

tests/idris2/interactive046/input.idr
```diff
@@ -0,0 +1 @@
+
```
gallais, line 1: This file should probably not exist.

CodingCellist: Whoops!

gallais: You've kept it!

CodingCellist: Argh, sorry! Missed that there was also an `input.idr`

PR #2716, opened 2022-10-16
make the Int type explicit

tclfs:
Idris will treat a number as an `Integer` by default, it's required to make the `Int` explicit to ouput 94.


gallais:
Doesn't it make more sense to change `notId` to pattern-match on `Integer` instead?
This gives the wrong impression that functions dispatching on types need additional
annotations when it's an unrelated issue with defaulting.

tclfs:
@gallais Yes, it is more clear if  `Integer ` is pattern-matched.

PR #2718, opened 2022-10-16
[ fix ] remove --only option in Makefile when "only" is not specified

QDelta:
Run `make test` will not actually run any test after #2705.

It seems that when the option `--only` is followed by an empty string, the behavior changes from running all tests to filtering out all tests.


gallais:
Let's fix the lib instead, please

QDelta:
> Let's fix the lib instead, please

Ok, I'll create an issue.

PR #2720, opened 2022-10-16
[ fix #2719 ] `--only` behavior in Test.Golden

### LINKED ISSUE #2719: `make test` not actually run any test
Run `make test` will not actually run any test after #2705 

It seems that when the option `--only` is followed by an empty string, the behavior changes from running all tests to filtering out all tests.

QDelta:
Refer to #2718


QDelta:
The test `idris2/reflection001` fails

QDelta:
> The test `idris2/reflection001` fails

This test fails after #2715.

Issue #2087, opened 2021-11-02
Multiline strings issue when running tests on Github CI


ska80:
This is a follow-up issue to PR #2044 

The `idris2/pkg003` test always fails on Windows and NixOS when running tests on Github CI.  Running tests locally doesn't exhibit this problem.

# Steps to Reproduce

Re-run all CI workflows for PR #2044 (before the `4cb83984c` commit)

# Expected Behavior

All started CI workflows should finish successfully.

# Observed Behavior

The `idris2/pkg003` test always fails on Windows and NixOS (`windows-bootstrap-chez` and `nix-bootstrap-chez` workflows) with the following error:
```
idris2/pkg003: FAILURE                                           02.882s
Expected:
1/1: Building Main (Main.idr)
Not all command line options can be used to override package options.

Overridable options are:
    --quiet
    --verbose
    --timing
    --log <log level>
    --dumpcases <file>
    --dumplifted <file>
    --dumpvmcode <file>
    --debug-elab-check
    --codegen <cg>
    --directive <directive>
    --build-dir <dir>
    --output-dir <dir>

Packages must have an '.ipkg' extension: "malformed-package-name".
Uncaught error: File error (non-existent-package.ipkg): File Not Found

Given:
1/1: Building Main (Main.idr)
Not all command line options can be used to override package options.
Overridable options are:
    --quiet
    --verbose
    --timing
    --log <log level>
    --dumpcases <file>
    --dumplifted <file>
    --dumpvmcode <file>
    --debug-elab-check
    --codegen <cg>
    --directive <directive>
    --build-dir <dir>
    --output-dir <dir>

Packages must have an '.ipkg' extension: "malformed-package-name".
Uncaught error: File error (non-existent-package.ipkg): File Not Found
```

CodingCellist:
Copying gallais's message from the discord:

> We've disabled a test on the windows CI [for pr #2722] because it would fail in a really incomprehensible & irreproducible way. If someone can reproduce it & is willing to investigate, we'll happily accept PRs :slightly_smiling_face:

dunhamsteve:
I believe the underlying issue here is that nix is using a bootstrapped compiler to run the tests. I've reproduced this on nixos with the same command that CI uses:
```
nix-build -A checks.x86_64-linux
```
And it reproduces in `nix-shell` and a plain build on macos, if I just do a bootstrap followed by test, without rebuilding the compiler with the bootstrapped version.

andylokandy:
[...]

PR #2722, opened 2022-10-18
[ repl ] Add the ability to get detailed help, e.g. `:help :help`

CodingCellist:
This has been a thing that has bothered me and other users and helpers alike: REPL commands like `:set` and `:let` have subtle behaviour which is not easily discoverable. This adds the ability to actually get this info, by allowing `:help` to take a repl command as an argument. The general `:help` text (i.e. no arguments) containing the list of repl commands and their single lines of help is preserved.


CodingCellist:
(this should probably be squashed and merged, since it's just a bunch of WIP commits until I finally got it working as I intended it to)

CodingCellist:
Seems to unfortunately be suffering from #2087. I've used the same "fix" as with #2044 (which revealed the issue), but this really is an odd problem...

CodingCellist:
CI pls... :sob:

CodingCellist:
Dang, I was really hoping that'd work... :disappointed:

What I find super odd, is that commit [c879f41](https://github.com/idris-lang/Idris2/pull/2722/commits/c879f41684b9a743601b475bf35e155f158b2adb) fixes one of the tests but not the other. Which suggests it was never a multiline string vs `unlines` thing. But that's just even more confusing...

CodingCellist:
At this point I'm tempted to just comment out the test honestly. This is super weird behaviour, and if I understand #2087 correctly, an issue that **only** occurs on the CI machines; no "real" machines running Nix or Windows have observed the issue afaiu.

gallais:
I've added just enough of a feature to `Test.Golden` that we can stop running these tests on the CI
while still running them on people's machines. We'll see whether it is indeed the case that this only
manifests on CI machines.

CodingCellist:
@gallais, that's amazing, tysm!

Am I good to merge this in that case?  : )

gallais:
Go for it! 🚀

gallais:
Looks like there is a bug in the `Test.Golden` implementation: CI is not running any of the tests.

gqqnbig:
Thank you. Until I saw this, I'd never known how to use :let.

PR #2725, opened 2022-10-21
[ fix ] parsing of test lib's options

gallais:
Broken in #2722

### DIFF
--- a/libs/test/Test/Golden.idr
+++ b/libs/test/Test/Golden.idr
@@ -168,6 +168,9 @@ options args = case args of
 
   where
 
+    isFlag : String -> Bool
+    isFlag str = "--" `isPrefixOf` str
+
     go : List String -> Acc -> Options -> Maybe (Acc, Options)
     go rest acc opts = case rest of
       []                            => pure (acc, opts)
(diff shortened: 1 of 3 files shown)


libs/test/Test/Golden.idr
```diff
@@ -181,9 +184,17 @@ options args = case args of
...
+        ifThenElse (isFlag p)
+          (go (p :: xs) acc opts)
+          (go xs ({ onlyNames $= (words p ++) } acc) opts)
+      ("--except" :: p :: xs)       =>
+        ifThenElse (isFlag p)
+          (go (p :: xs) acc opts)
+          (go xs ({ exceptNames $= (words p ++) } acc) opts)
       ("--only-file" :: p :: xs)    => go xs ({ onlyFile := Just p } acc) opts
+      [p] => do guard (p `elem` ["--only", "--except"])
+                pure (acc, opts)
```
mattpolzin, line 197:
```suggestion
                pure (acc, opts)
                -- ^ we accept trailing only or except flags as unused (do not filter out any tests)
                -- for the convenience of populating these options from shell scripts. 
```
First read of this logic felt off to me until I realized why we accept these options to be specified with no arguments.

PR #2728, opened 2022-10-22
[ test ] Re-enable nix test and suppress repl005 test on nix.

dunhamsteve:
I don't fully understand the nix stuff, but this seems to resolve the issue.  This change re-enables the nix CI and passes `except="idris2/repl005"` to the make test command along with the existing `INTERACTIVE=`.

### DIFF
--- a/.github/workflows/ci-idris2.yml
+++ b/.github/workflows/ci-idris2.yml
@@ -231,10 +231,9 @@ jobs:
     needs: quick-check
     runs-on: ubuntu-latest
     if: |
-      false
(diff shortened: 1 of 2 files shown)


gallais:
The issue with this is that it means that users won't run the test either.
But we want to keep doing that because our hunch is that the error is
only happening on CI machines for some strange reason we don't know how to investigate.

dunhamsteve:
Yeah, I saw that but I wasn't thinking in terms of nix.  Exposing that parameter to the command line is beyond my nix skill.

PR #2730, opened 2022-10-24
Update nix build scripts to build Idris with the current version of Idris #2087

dunhamsteve:
This PR changes the nix build scripts to bootstrap idris and then use the bootstrapped compiler to build the final version. It also reenables CI for nix. This addresses #2087 for nix.

Previously, the nix scripts would test and install the copy of Idris built by the bootstrap compiler. The bootstrap compiler doesn't handle multiline strings properly, so the resulting compiler failed test `idris2/repl005`. 

To accomplish this, I duplicated `package.nix` as `bootstrap.nix` and changed `package.nix` to use the idris compiler in the result of `bootstrap.nix`.

I'm somewhat new to nix, so it would be good to have this reviewed by someone who knows idiomatic nix.

### DIFF
--- a/.github/workflows/ci-idris2.yml
+++ b/.github/workflows/ci-idris2.yml
@@ -231,10 +231,9 @@ jobs:
     needs: quick-check
     runs-on: ubuntu-latest
     if: |
-      false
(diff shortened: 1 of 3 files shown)


gallais:
Thanks!

PR #2731, opened 2022-10-25
[ ci ] Simplify bootstrap process in nix

nmccarty:
This PR:
- Combines the `package.nix` and `bootstrap.nix` files, using `idris2Bootstrap` as an optional argument, bootstrapping if it is not present
- Formats the `.nix` files with `nixfmt`
- Updates `flake.lock`

I would recommend setting up github actions to have some automation on updating `flake.lock`, it hadn't been touched in a while. I can follow up with a PR to do that, if there is interest.


nmccarty:
`/home/runner/work/_temp/77c03a0c-2ea1-4721-a67e-01e078103dfe.sh: line 1:  2178 Killed                  nix-build -A checks.x86_64-linux`

I think we just OOMed the CI.

 This can likely be resolved (and that CI step maybe substantially sped up, depending) by calling into the flake directly instead of using the flake-compat wrapper, something like:
`nix build .#checks.x86_64-linux.idris2Tests -L`

dunhamsteve:
Yeah,  it's worth a try.  The nix build is now failing everywhere with exit code 137, even without the change in this PR. 

I don't know how much memory nix uses, but that last step has both chez and the Idris compiler in memory.  I've had it OOM with normal builds in a too-small VM.

nmccarty:
Nix can end up using a lot of memory, and the `flake-compat` wrapper can make things even worse as its emulating flake functionality in nix instead of using the native support, and there's a lot going on under the hood there. I'll push up a new commit with tweaked CI

dunhamsteve:
That might have to be:
```
nix build --experimental-features nix-command --extra-experimental-features flakes .#checks.x86_64-linux.idris2Tests -L
```

nmccarty:
> That might have to be:
> 
> ```
> nix build --experimental-features nix-command --extra-experimental-features flakes .#checks.x86_64-linux.idris2Tests -L
> ```

I've accomplished the same by setting the `NIX_CONFIG` environment variable, which, like many things in nix is, uh, really hard to find documentation on and now I can't remember where I first read about it

nmccarty:
Taking out the `flake-compat` overhead seems to have a job, looks like that run completed fine

gallais:
Thanks

PR #2735, opened 2022-10-27
[ fix ] bump version of linear and papers packages

stefan-hoeck:
This might be an oversight. It leads to issues with pack, which uses the same version for all libraries from the Idris2 project. If this was left at 0.5.1 on purpose, feel free to close this and I'll fix pack instead.


gallais:
I'm guessing it's due to the fact it's not listed in the release checklist:

https://github.com/idris-lang/Idris2/blob/102d7ebc18a9e881021ed4b05186cccda5274cbe/Release/CHECKLIST#L3

Can you please add it in this PR too?

PR #2741, opened 2022-10-31
[ new ] deriving Show

gallais:
Includes bug fixing for other derivings too


gallais:
I can't think of new examples to handle so I'll merge this and wait for the inevitable bug reports to arrive.

Issue #2745, opened 2022-11-02
[ regression ] #2684 broke transitive dependency resolution


stefan-hoeck:
I get new build errors related to this when building pack's package collection. In particular, the [idris2-tls](https://github.com/octeep/idris2-tls) package fails to build due to a transitive dependency on the *linear* package, which Idris no longer resolves. I'll try and figure out the issue at hand.

stefan-hoeck:
OK, so the package resolution algorithm looks at the wrong place for the installed `.ipkg` files, which are now also in the directory with the ttc-version. Test `idris2/pkg014`, which thoroughly tests the transitive resolution algorithm, did not catch this because it uses a hand-written `depends` directory with preinstalled `.ipkg` files at the location the resolution algorithm expected.

The fix for this is trivial, however, I'm not yet sure how to adjust `idris2/pkg014` to always use the correct ttc-version in its `depends` directory. Any ideas?

PR #2746, opened 2022-11-02
[ fix ] issue 2745

stefan-hoeck:
This should fix issue #2745, but I'll have yet to adjust test `idris2/pkg014`.


stefan-hoeck:
I added a new command to the `idris2` executable: `--ttc-version` for printing exactly that. This allows us to properly set up the `depends` directory in the `pkg014` test.

gallais:
I tried to juggle the paths so that we would look into `ttc/TTC_VERSION` rather than `TTC_VERSION/ttc`
precisely so that users wouldn't need to know / worry about the `TTC_VERSION` and could simply point
at `ttc/`.

Do we actually need `TTC_VERSION` in that case if we're only looking up an `.ipkg` file? Can we not look
for it at the root of the directory?

stefan-hoeck:
> I tried to juggle the paths so that we would look into `ttc/TTC_VERSION` rather than `TTC_VERSION/ttc` precisely so that users wouldn't need to know / worry about the `TTC_VERSION` and could simply point to `ttc/` in their `depends` (cf. the `idris2/import008` test).
> 
> Do we actually need `TTC_VERSION` in that case if we're only looking up an `.ipkg` file? Can we not look for it at the root of the directory?

Doh. You are right of course, that's much simpler. I'll close this and come up with the (hopefully) straight forward fix. A command `--ttc-version` might still be useful for tools such as pack, but that should be for another package, if at all.

PR #2747, opened 2022-11-02
[ fix ] isue 2745

stefan-hoeck:
As discussed in #2746, this is a more straightforward fix for #2745. I added also a new test to make sure Idris can find the `.ipkg` files it installed.

### DIFF
--- a/tests/idris2/pkg016/run
+++ b/tests/idris2/pkg016/run
@@ -1,5 +1,3 @@
-IDRIS2_VERSION="idris2-0.6.0"
-
 $1 --install foo.ipkg > /dev/null
 $1 --install bar.ipkg > /dev/null
 $1 --install baz.ipkg > /dev/null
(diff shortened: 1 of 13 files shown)


tests/idris2/pkg016/run
```diff
@@ -0,0 +1,13 @@
+IDRIS2_VERSION="idris2-0.6.0"
+
+$1 --install foo.ipkg > /dev/null
+$1 --install bar.ipkg > /dev/null
+$1 --install baz.ipkg > /dev/null
+$1 --build test.ipkg
+
+build/exec/test
+
+rm -r "${IDRIS2_PREFIX}/${IDRIS2_VERSION}/foo-0"
```
gallais, line 10: Can't you write `rm -r "${IDRIS2_PREFIX}/idris2-*/foo-0"`?

Z-snails:
I don't know if this would simplify it at al, but you could probably specify set `IDRIS2_PREFIX` in the test (pkg016), and use `--no-prelude` so it doesn't complain about missing prelude. Then it definitely won't mess with any global install.

stefan-hoeck:
> I don't know if this would simplify it at al, but you could probably specify set `IDRIS2_PREFIX` in the test (pkg016), and use `--no-prelude` so it doesn't complain about missing prelude. Then it definitely won't mess with any global install.

Yeah, in that case I'd have to change how the test is being run (it's using a `main` function at the moment). Personally, I think the current solution is reasonably simple.

Z-snails:
Ok I'll merge it as is :)

PR #2748, opened 2022-11-02
[ fix ] dont prefix lib dirs with ttc version

stefan-hoeck:
As discussed on discord, this fixes another regression introduced by #2684: `.so` files installed into custom lib dirs are no longer found by Idris. This reverts the relevant section of `Compiler.Scheme.Chez.loadSO` to its version before #2684. I'm not sure about the `bdir <- ttcBuildDirectory` on line 221: Do we want the `ttc` prefix here or not?

@mattpolzin Could you please test this against pg-idris?

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -219,8 +219,7 @@ loadSO appdir "" = pure ""
 loadSO appdir mod
     = do d <- getDirs
          bdir <- ttcBuildDirectory
-         let allDirs = extra_dirs d ++ package_dirs d
-             fs = map (\p => p </> mod) (bdir :: allDirs)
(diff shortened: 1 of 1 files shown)


mattpolzin:
Thanks for this attempt at a fix. I've since discovered that `loadSO` is not responsible for copying the `so` files I've seen fail to copy, but rater `loadLib`. The culprit of this bug appears to actually be some code in #2684 that swaps out some lines that used to add directories to the list of "extra" dirs for lines that add  the same directories to the list of "package" dirs. The swap seems sensible since it is indeed a package directory where the `lib` folder pertaining to this bug lives.

I'll have an alternative fix PR up for this soon that tells `loadLib` about `lib` folders in `package_dirs`.

stefan-hoeck:
Perfect. I'll close this then in favor of #2750.

PR #2750, opened 2022-11-02
[fix] look for lib folders in package directories.

mattpolzin:
Prior to https://github.com/idris-lang/Idris2/pull/2684, any `lib` folder found in the `$(idris2 --libdir)/package-dir/` directory would be searched for `so` or `dylib` files that might need to be copied into the `_app` directory of an executable produced by the Chez Scheme backend.

After the TTC PR (2684), this stopped working because such directories were removed from the list of "extra" directories and added to the list of "package" directories. This move in isolation makes sense -- we are talking about a package directory, after all. So, this PR fixes the `lib` folder searching by adding these package directories to the list of searched paths in the `findLibraryFile` function that is ultimately responsible for locating the `so`/`dylib` files in question.

### DIFF
--- a/src/Core/Directory.idr
+++ b/src/Core/Directory.idr
@@ -149,20 +149,24 @@ readDataFile fname
             | Left err => throw (FileErr f err)
          pure d
 
--- Look for a library file required by a code generator. Look in the
--- library directories, and in the lib/ subdirectory of all the 'extra import'
--- directories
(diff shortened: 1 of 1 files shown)


mattpolzin:
This PR attempts a fix at the same problem https://github.com/idris-lang/Idris2/pull/2748 attempts to fix.

PR #2768, opened 2022-11-11
[ ide-mode ] Do not include whole error message in building file display result message.

keram:
**Why:**
Makes the error message being displayed twice in Emacs. 
This change aligns IDEMode/REPL.idr with Idris/REPL.idr which works as expected.
See https://github.com/idris-lang/Idris2/blob/main/src/Idris/REPL.idr#L1241-L1242

Before change:
async-minibuffer-before-change-idris2

After change:
async-minibuffer-after-change-idris2-idemode-repl

Relates to:
https://github.com/idris-hackers/idris-mode/pull/563


gallais:
This has been a really annoying bug. Thanks for the fix!
Do we need to do anything to https://github.com/idris-community/idris2-mode to take these changes into account?

jfdm:
@gallais

> Do we need to do anything to https://github.com/idris-community/idris2-mode to take these changes into account?

I believe no.

@keram has been doing a great job improving aspects of idris-mode. You may want to look at the recent PRs to incorporate into idris2-mode. I still need to do the opposite.

PR #2783, opened 2022-11-25
[ parser ] fix issue where indentation is not checked in record params

dunhamsteve:
This issue was reported by "Oxy" on discord.  

When parsing record params on forward record declarations, the Idris parser is not checking indentation and may interpret a subsequent top level line as part of the record declaration.  

So this valid code:
```idris
record Rec (a : Type)
b : Int
```
yields the following error:
```
1/1: Building Indent (Indent.idr)
Error: Expected 'where'.

Indent:2:3--2:4
 1 | record Rec (a : Type)
 2 | b : Int
       ^
```

This patch fixes the issue by adding an indentation check in front of each parameter.


gallais:
Thanks!

PR #2784, opened 2022-11-29
[ ci ] Replace deprecated feature with its functional equivalent

buzden:
According to [this](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), `set-output` feature will be deprecated. Now CI already prints warning during run. There is suggested change that does the same, but is not deprecated, thus let's use it.

### DIFF
--- a/.github/workflows/ci-bootstrap.yml
+++ b/.github/workflows/ci-bootstrap.yml
@@ -34,9 +34,9 @@ jobs:
         id: get_commit_message
         run: |
           if   [[ '${{ github.event_name }}' == 'push' ]]; then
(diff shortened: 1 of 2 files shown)


buzden:
Looks like GitHub Actions are having internal problems and jobs need to be restarted sometime. I encounter similar problems of inability of fetching from inside a job in other projects running Actions too (and restart helps).

gallais:
Yep @buzden I've been fighting with actions the whole afternoon, with `apt`
failing to get packages from azure. I'm happy to merge as-is given the job
you modified did get to run without any issue.

PR #2785, opened 2022-11-29
[ perf ] manually eta-expand unsaturated io_bind calls

gallais:
This gets us from
```
main = do
  putStrLn "What is your name?"
  nm <- getLine
  putStrLn "Hello \{nm}"
```
 to
```
(define Main-main
  (lambda (ext-0)
    (let ((act-1 (PreludeC-45IO-prim__putStr "What is your name?\xa;" ext-0)))
      (let ((act-2 (PreludeC-45IO-prim__getStr ext-0)))
        (PreludeC-45IO-prim__putStr (string-append (PreludeC-45Types-u--foldMap_Foldable_List (cons (lambda (arg-8326) (lambda (arg-8329) (string-append arg-8326 arg-8329))) "") (lambda (eta-0) eta-0) (cons "Hello " (cons act-2 '()))) "\xa;") ext-0)))))
```
thanks to a lot of inlining.

It'd be good to find a way to avoid the manual eta-expansion of an under-applied `io_bind` though.
A first test seems to suggest it's worth merging anyways though.

Runnig `time make` after `make clean` (only one run for each, so take that 10% with a pinch of salt):

Before: 7m4s
After: 6m23s


gallais:
That last commit means the inlining of `print` produces
```
 putStr @{io_HasIO} . show @{nat_Show}
 ```
rather than
```
 putStr @{fst (io_HasIO, nat_Show)} . show @{snd (io_HasIO, nat_Show)}
 ```
 which allows the transform rules for interfaces to kick in and we get a direct call
to the implementation of `show` for `Nat` rather than a call to the projection out
of the `Show` dictionary for `Nat`.

PR #2786, opened 2022-12-01
Let qualified do-notation apply to `pure` and to idiom brackets

Russoul:
This can come in handy when one deploys their own namespace with a subset of `pure`, `>>=`, `>>`, `<*>` overloaded in it and where one uses the qualified do-notation to disambiguate monad-style code.


gallais:
Can we please have test cases? It's not exactly clear to me what this is doing.

Russoul:
Added test cases :)

PR #2789, opened 2022-12-02
[ doc ] Add constructor docstrings

### LINKED ISSUE #2732: Inline documentation for constructor of record causes parse error
# Steps to Reproduce

Compile the following example from https://idris2.readthedocs.io/en/latest/reference/documenting.html

```idris
module Documenting

||| We can document records, including their fields and constructors
record Yummy where
  ||| Make a yummy
  constructor MkYummy
  ||| What to eat
  food : String
```

# Expected Behavior

Successful compilation.

# Observed Behavior

```
Error: Expected end of input.

Documenting:5:3--5:19
 1 | module Documenting
 2 | 
 3 | ||| We can document records, including their fields and constructors
 4 | record Yummy where
 5 |   ||| Make a yummy
```

# Info

$ idris2 --version

`Idris 2, version 0.5.1-bc8b6bb7a`

pinselimo:
This PR resolves #2732 by adding constructor docstrings for records and interfaces.

However, it also causes interface constructors to always be shown when asking for `:doc` in the REPL. This is consistent with the behavior for records, but these do not show up with their type if they were not user declared and I couldn't find out why.

Ran all tests and updated the golden test outputs where necessary.

So the remaining question is: Should the machine made constructors be shown in the docs? Should they show up with or without their type?


src/Idris/Desugar.idr
```diff
@@ -1022,7 +1022,9 @@ mutual
                                               map fst params) (mkNamespace recName))
                                fields
            let _ = the (List IField) fields'
-           let conname = maybe (mkConName tn) id conname_in
+           let conname = maybe (mkConName tn) snd conname_in
+           let condocs = maybe "" fst conname_in
```
gallais, line 1026: This is adding an empty docstring when there isn't any.
I'd rather we used `whenJust` and only add a docstring when there's actually one.

pinselimo: Right, thanks! Same in the interface elab, fixed both.

src/Idris/Doc/String.idr
```diff
@@ -347,9 +347,8 @@ getDocsForName fc n config
                 case !(traverse (pterm . map defaultKindedName) (parents iface)) of
                      [] => []
                      ps => [hsep (header "Constraints" :: punctuate comma (map (prettyBy Syntax) ps))]
-             let icon = case dropNS (iconstructor iface) of
-                          DN _ _ => [] -- machine inserted
-                          nm => [hsep [header "Constructor", dCon nm (prettyName nm)]]
+             icon <- do doc <- getDConDoc =<< toFullNames (iconstructor iface)
+                        pure $ [header "Constructor" <++> annotate Declarations doc]
```
gallais, line 351: I think that here we want to keep the logic that we're only showing the
constructors that have been user-declared, not the machine-introduced
ones.

Otherwise we may give users the wrong idea: these `__mkInterface` names
are not actually valid, cf.

```idris
interface Pointed a where
  point : a

failing "Undefined name mkPointed"

  nat : Pointed Nat
  nat = __mkPointed 0
```

Additionally, I don't think we should use: `getDConDoc` because that'll give
us the type too but here the type is not interesting: we are essentially defining
a record and so the type of the constructor is obvious from the type of the
methods IMO.

gallais:
Hum. wtf happened with the merge?

gallais:
Ah, I guess the issue with my changes is that we don't get to see the
documentation for interface constructors.  :thinking:

PR #2791, opened 2022-12-05
[ performance ] Implement weak memoisation of lazy values for chez and racket

buzden:
This PR implements a _weak_ memoisation of lazy values for the chez backend **as an codegen option (directive)**. Weakness is in the fact that no guarantees of no re-evaluation of a lazy value is given, and those memoised values can be garbage collected at any wish of the GC. I think, this is reasonable middling between current full re-evaluation and full memoisation behaviours, not requiring addition of separate type-constructor-like construction along with `Lazy` and `Inf`, as it was discussed in #1013.

Behaviour of other backends than chez and racket should be unaffected by this change.


buzden:
OMG, I've checked on one slow case in DepTyCheck after reworking on laziness, and gained from 30+ minutes to 1 minute. Surely, it depends on heaviness of laziness, but still.

gallais:
Testing this (with `main` merged in) vs `main` and I get a slight slowdown:

```
main:
  2m46s (real 2m52s)
  2m47s (real 2m54s)
  2m46s (real 2m53s)

branch:
  2m55s (real 3m02s)
  2m56s (real 3m02s)
  2m55s (real 3m02s)
```

buzden:
I didn't repeat my tests for a while, so I may guess that the bigger slowdown is induced by the last commit, when additional boxing is used.

Note around this commit. We've been using this PR's branch for a while in running algorithm that heavily relies on memoisation and realised very strange and flacky behaviour of chez complaining on invalid memory references (each time in different places and even not related to lazy values; looks like normal objects got freed accidentally). We've found out that additional boxing solves this (i.e. this does not happen when weak reference to memoised value is never a primitive value, always a box), which looks like a kind of a bug in GC of chez. However, I wasn't able to reproduce this behaviour in a small example, thus I couldn't add a test for it or report it to chez.

Anyway, this boxing led to a visible slowdown in my "worst-case-ish" example, which I talked about at IDM or Discord. But at the time, at least, together with independent improvements in performance since then, the total performance was no worse than the very original, when this change started to be developed.

So, it's hard to decide whether is it okay to pay such a slowdown price for an ability to gain performance in memoisation-relying algorithms. I'm definitely biased, since I use one of such algorithms, so I think it's worth doing.

buzden:
CI for MacOS fails with `failed to list vmcode: Cannot allocate memory` when tries to run `make test` before actually running the tests. Can anyone tell what is happening?

buzden:
Looks like those strange `Cannot allocate memory` error on the MacOS CI cured itself

buzden:
Now, a bug cisco/ChezScheme#714 in the garbage collector of Chez is fixed in the version 9.5.9, so it means that we can remove a workaround that lead to a visible slowdown from this PR.

buzden:
I brought the workaround back, because CI uses (and the current requirements allow) older buggy version of chez. As soon as we could require newer fixed version, we can remove the workaround, and have this functionality without those extra runtime performance penalty.

PR #2800, opened 2022-12-06
[ perf ] compile Core.CompileExpr.Pretty faster

gallais:
The auto search in the pretty instances is just too damn expensive.
It still takes 4s so we should be able to squeeze some more perf out of this.


gallais:
I had an additional slowdown due to being on battery when doing these experiments yesterday.
This morning it looks like `Core.CompileExpr.Pretty` is down to 2s so I'll merge as is.

PR #2803, opened 2022-12-07
Improve parser error messages when using reserved identifiers in decls

eayus:
## Problem

Currently, using a reserved identifier or keyword on the LHS of a type claim gives a generic error message.
```idris
-- Fails with "Couldn't parse declaration"
String : Type
```
This might be confusing for new users.

## Solution
I have updated the error messages:

```idris
Error: Cannot begin a declaration with a reserved identifier.

Test:4:1--4:7
 1 | module Test
 2 |
 3 |
 4 | String : Type
     ^^^^^^
```
```idris
Error: Keyword 'module' is not a valid start to a declaration.

Test:4:1--4:7
 1 | module Test
 2 |
 3 |
 4 | module : Type
     ^^^^^^
```

## Caveats

The error message is still poor when the reserved identifier is namespaced:
```idris
-- Fails with "Couldn't parse declaration"
X.String : Type
```

However I don't think this is a major problem as predeclaring names in other modules is an advanced use case, whereas this change is mainly benefiting new users.

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1779,7 +1779,11 @@ directiveDecl fname indents
 -- Declared at the top
 -- topDecl : OriginDesc -> IndentInfo -> Rule (List PDecl)
 topDecl fname indents
-    = do d <- dataDecl fname indents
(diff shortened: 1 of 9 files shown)


eayus:
@mattpolzin Ok that should be done now

PR #2804, opened 2022-12-08
Compile time strings

madman-bob:
This rephrases the `String` primitives in terms of `strCons` and `strUncons`, and add a couple of `believe_me`s about their behaviour. I've used `%transform` rules to leave the runtime behaviour of all primitives unchanged.

This allows us to write proofs about `String` functions. I've included a few proofs to demonstrate, and serve as tests for my primitive implementations.


mattpolzin:
Would you be opposed to moving the string properties to the bottom of base/Data/String.idr? I sometimes peruse modules like String in search of a useful operation I don’t know the name of and having the properties at the top seems like it makes that less convenient. 

I see them after a “properties” comment heading pretty commonly if I recall correctly.

gallais:
Sorry about that, got mixed up in my branches.

stefan-hoeck:
While it would certainly be great to be able to reason about strings, I have several thoughts/questions about this:

- How does this affect performance of compile-time string manipulation, for instance when using `(++)`, which now no longer uses the prim op?
- The core assumptions consist of calls to `believe_me Refl`? Are these critical in the same way as explained by @gallais in [`Data.Int.Order`](https://github.com/idris-lang/Idris2/blob/main/libs/contrib/Data/Int/Order.idr) and should be treated accordingly?
- I think, there should be some tests to make sure that the new transform rules actually apply.

madman-bob:
The compile-time performance of `++` will be worse now on statically known strings, yes. I'm not sure if there's a nice way around this. Do rewrite rules apply at compile-time for statically known values?

The `believe_me`s are neccessary as `String` is a primitive type. I think that's the same thing as what's going on in `Data.Int.Order`. Only three functions use them, which I believe is minimal.

Tests would be good. Where are the existing rewrite rule tests?

libs/prelude/Prelude/Types.idr
```diff
@@ -875,6 +882,67 @@ fastUnpack : String -> List Char
 -- always use 'fastPack' at run time
 %transform "fastUnpack" unpack = fastUnpack
 
+namespace String
+  public export
+  (++) : (x : String) -> (y : String) -> String
+  x ++ y = strInduction y (\c, s, res => strCons c res) x
+
+  %transform "fastAppend" String.(++) = prim__strAppend
```
gallais, line 890: Are we sure this rule will always fire? For all we know the compiler may decide to inline `++`
(the definition is not recursive & not pattern-matching) and then the transform rule will not
be deployed on the `strInduction`-headed expression.

Same comment for `reverse`, `substr`, etc.

PR #2807, opened 2022-12-09
[ performance ] replace `blodwen-lazy` with `delay` and `force` on Chez

stefan-hoeck:
As discussed on Discord, using `blodwen-lazy` to lazily compute and memoise top-level constants has a negative impact on performance. With a simple change, we can significantly speed up compiler performance on the Chez backend (other backends are not affected by this). If this would also apply to other scheme backends, I do not know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   Versions of the flags with the `IDRIS2_` prefix can also be used and take
   precedence.
 
+#### Chez
+
+* Use `delay` and `force` to lazily compute top-level constants.
+
 ### Compiler changes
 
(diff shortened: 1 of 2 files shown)


Z-snails:
Adding `(require racket/promise)` [here](https://github.com/idris-lang/Idris2/blob/main/src/Compiler/Scheme/Racket.idr#L42) should fix the issue with racket.
I'll check if gambit also needs an import

buzden:
By the way, it looks like `(define blodwen-lazy ...)` can be removed from all the scheme support files as a part of this change

stefan-hoeck:
Closing this in favor of #2817

PR #2816, opened 2022-12-14
[ fix ] elaboration of records' telescopes of parameters

gallais:
At the moment e.g.
```idris
record HasLength (xs : List a) (n : Nat) where
  constructor MkHasLength
  0 prf : length xs === n
```
leads to a truly horrendous error message when it's pretty clear we should just
add `{0 a : Type}` to the telescope of parameters. This patch fixes that.

# TODO

- [x] Fix unelaborating in a nested environment
- [x] Test bootstrapping
- [x] CHANGELOG

### DIFF
--- a/tests/base/deriving_foldable/DeriveFoldable.idr
+++ b/tests/base/deriving_foldable/DeriveFoldable.idr
@@ -57,7 +57,7 @@ namespace BigTree
 
 namespace Matrix
 
-  record Matrix m n a where
+  record Matrix (m, n : Nat) (a : Type) where
     constructor MkMatrix
     runMatrix : Vect m (Vect n a)
 
--- a/tests/base/deriving_foldable/DeriveFoldable.idr
+++ b/tests/base/deriving_foldable/DeriveFoldable.idr
@@ -57,7 +57,7 @@ namespace BigTree
 
 namespace Matrix
 
-  record Matrix (m, n : Nat) (a : Type) where
+  record Matrix m n a where
     constructor MkMatrix
     runMatrix : Vect m (Vect n a)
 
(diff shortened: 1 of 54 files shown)


gallais:
Grepping through the idris2 source file, it looks like we'll only need to change 10 out of 120+
record declaration to add a type annotation to a single argument every time.

In the libs, it looks to be  ~12 out of 90+.

tests/base/deriving_foldable/DeriveFoldable.idr
```diff
@@ -57,7 +57,7 @@ namespace BigTree
 
 namespace Matrix
 
-  record Matrix m n a where
+  record Matrix (m, n : Nat) (a : Type) where
```
buzden, line 60: Did you change the quantity of those `m`, `n` and `a` here?

gallais: No, untyped record arguments are assigned `top` in the parser.

buzden: OMG, rules for records and interfaces differ! Maybe, it's good that it can be (accidentally) removed by this PR

gallais:
Last problem (before attempting to bootstrap) shows up in `Search.CTL`:

```idris
parameters (Lbls, Sts : Type)
-- (...)
  Formula : Type
  Formula = (depth : Nat) -> (tree : CT) -> Type
-- (...)
  record DepthInv (f : Formula) where
    constructor DI
    prf : {n : Nat} -> {m : CT} -> f n m -> f (S n) m
```

Here we get back `(f : Formula Lbls Sts) -> Type` back as a type and unelab fails to strip off `Lbls`
and `Sts` leading to an `Formula` being over-applied & an error message complaining that
`Formula` has type `Type` and not `Type -> Type -> Type`.

I suspect it's just a matter of unelaborating the nested environment but I don't want to look at
it tonight.

gallais:
Almost there! I've [opened a PR](https://github.com/stefan-hoeck/idris2-elab-util/pull/58) updating elab-utils.

PR #2817, opened 2022-12-16
[ performance ] Compile non-recursive top-level constants to constants in Chez

stefan-hoeck:
This replaces #2807 : Non-recursive top-level constants are compiled to eagerly evaluated constants in Chez Scheme. In order to do this, they have to be added in the correct order. In addition, shared objects need to be loaded *before* the actual code, because they might already be needed when constants get eagerly evaluated.

This has a significant impact on performance: On my system, typechecking `idris2api.ipkg` goes from 3:40 minutes down to slightly below 3 minutes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@
 
 #### Chez
 
-* Use `delay` and `force` to lazily compute top-level constants.
+* Non-recursive top-level constants are compiled to eagerly evaluated
+  constants in Chez Scheme.
 
 ### Compiler changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   Versions of the flags with the `IDRIS2_` prefix can also be used and take
   precedence.
 
+#### Chez
+
+* Use `delay` and `force` to lazily compute top-level constants.
+
 ### Compiler changes
 
 * If `IAlternative` expression with `FirstSuccess` rule fails to typecheck,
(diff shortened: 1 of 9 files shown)


gallais:
Very strange, I got:

Current HEAD: 3m3s (twice in a row)
This PR: 3m4s (once), then after a second install 2m50s (twice in a row)

So either you need to go full self-host to reap the benefits or my install step between
the current HEAD experiments & testing this PR somehow failed. It's not as impressive
as your local results but it's still a good 7+% so I think we should merge.

If @edwinb has something better, we can always revert / override.

stefan-hoeck:
> Very strange, I got:
> 
> Current HEAD: 3m3s (twice in a row) This PR: 3m4s (once), then after a second install 2m50s (twice in a row)
> 
> So either you need to go full self-host to reap the benefits or my install step between the current HEAD experiments & testing this PR somehow failed. It's not as impressive as your local results but it's still a good 7+% so I think we should merge.
> 
> If @edwinb has something better, we can always revert / override.

My timings were collected over several days, so they may not be very accurate, sorry. But yes, you need to go self-host: This is an optimization that happens during code gen, so if you compile once, you still use the code generator from your currently installed compiler, which knows nothing about this: The freshly compiled compiler will still have calls to `blodwen-lazy` but it will itself no longer produce such calls during code-gen. After a second compilation step, you get a fully optimized version of the compiler.

gallais:
Ah yes of course!

PR #2818, opened 2022-12-16
[ performance ] use raw prim ops in IntMap

stefan-hoeck:
This shaves off another ~10 seconds when typechecking `idris2api.ipkg` on my system: Since `IntMap` is already hand-specialized, we might as well replace calls to `==` and `<=`, which go via an unnecessary conversion to `Bool`, by direct calls to the primops.

### DIFF
--- a/src/Libraries/Data/IntMap.idr
+++ b/src/Libraries/Data/IntMap.idr
@@ -76,53 +76,42 @@ merge3 (Branch3 a b c d e) f (Branch2 g h i) j k = branch6 a b c d e f g h i j k
 merge3 (Branch3 a b c d e) f (Branch3 g h i j k) l m = branch7 a b c d e f g h i j k l m
 
 treeLookup : Key -> Tree n v -> Maybe v
-treeLookup k (Leaf k' v) =
-  if k == k' then
-    Just v
-  else
-    Nothing
-treeLookup k (Branch2 t1 k' t2) =
-  if k <= k' then
-    treeLookup k t1
-  else
-    treeLookup k t2
-treeLookup k (Branch3 t1 k1 t2 k2 t3) =
-  if k <= k1 then
-    treeLookup k t1
-  else if k <= k2 then
-    treeLookup k t2
-  else
-    treeLookup k t3
+treeLookup k (Leaf k' v) = case prim__eq_Int k k' of
+  0 => Nothing
+  _ => Just v
+treeLookup k (Branch2 t1 k' t2) = case prim__lte_Int k k' of
+  0 => treeLookup k t2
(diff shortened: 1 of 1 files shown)


gallais:
Shouldn't this be happening automatically through automatically declared interface transform rules? 🤨

buzden:
You need to transform `case` expressions rather than a function. Can this be done via transform rules?

stefan-hoeck:
Here's the relevant part of the generated code for `treeLookup` in current HEAD:

```scheme
(let ((sc1 (PreludeC-45EqOrd-u--C-61C-61_Eq_Int arg-2 e-1))) (cond ((equal? sc1 1) (box e-2)) (else '())
```

With a call to `Eq_Int` implemented as follows:

```scheme
(define PreludeC-45EqOrd-u--C-61C-61_Eq_Int
  (lambda (arg-0 arg-1)
    (let ((sc0 (or (and (= arg-0 arg-1) 1) 0)))
      (cond ((equal? sc0 0) 0)(else 1)))))
```

And here's the same part with the optimizations in this PR:

```scheme
(let ((sc1 (or (and (= arg-2 e-1) 1) 0))) (cond ((equal? sc1 0) '())(else (box e-2))))
```

Z-snails:
would marking the `Eq Int` implementation as `%inline` also fix this?

stefan-hoeck:
> would marking the `Eq Int` implementation as `%inline` also fix this?

It would still go via an unnecessary conversion to `Bool`. Perhaps Chez would be able to optimize that away, though.

gallais:
I'm not saying any difference with current HEAD on

```shell
make src/IdrisPaths.idr && time idris2 --build idris2api.ipkg
```

Z-snails:
> I'm not seeing any difference with current HEAD on
> 
> ```shell
> make src/IdrisPaths.idr && time idris2 --build idris2api.ipkg
> ```
> 
> Edit: if anything, this branch is ~10s slower?!

Did you rebase the branch? the current main has a big performance improvement this branch most likely missing

gallais:
Yes I did merge `main` into this branch before testing

stefan-hoeck:
> Yes I did merge `main` into this branch before testing

I just rebased onto main and will have another look on my machine. Will report back soonish.

stefan-hoeck:
I'm getting only minor, non-significant improvements compared to current HEAD (self-hosted), so I guess this can be closed. I think, I'm going to profile this using [idris2-profiler](https://github.com/stefan-hoeck/idris2-profiler), as it might still be relevant to know, if our bool ops are as fast as they can get, but that's for another day.

Issue #1664, opened 2021-07-06
[RFC] Access line, file and function name from the source


andrevidela:
# Motivation

For purposes of logging and debugging it is useful to be able to extract the file, line and function name from which an error is thrown or a log statement appears. This proposal aims to bring this capability to Idris2.

# Example

```idris
-- filenameTest.idr
module Main

printFileName : IO ()
printFileName = putStrLn %filename

printLine : IO ()
printLine = putStrLn (show %line)

thisFunction : IO ()
thisFunction = putStrLn %function

main : IO ()
main = printFileName *> printLine *> thisFunction 
```

The previous program would print: 

```
filenameTest.idr
8
thisFunction
```

# Proposed design

Add `%line` `%file` and `%function` as pragmas to access the line, file and function name. Each one would have the following type:

```
%line : Nat
%file : String
%function : String
```

They can then be bound as variables and used normally. A small subtlety to take into account is what happens when they are used as implicit default arguments:

```idris
log : {default %filename file : String} 
   -> {default %line line : Nat} 
   -> {default %function func : String} 
   -> LogM ()
```
In this case the pragma take on the value at the call site of `log` and not their location at the definition of `log`

There is a question about typechecking, I suspect we can evaluate those pragma at the same time we do elaborator reflection but I am not familiar with how elaborator reflection is implemented, so we would need someone more knowledgeable to comment on that

mattpolzin:
This proposal seems useful and strikes me as familiar coming from other languages with similar pragmas.

👍

PR #2819, opened 2022-12-17
[ new ] __LOC__, __FILE__, __LINE__, __COL__ magic debug strings

gallais:
I grabbed the syntax from https://v2.ocaml.org/api/Stdlib.html#1_Debugging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Next version]
 
+### Language changes
+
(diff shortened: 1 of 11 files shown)


andrevidela:
Just noticed this, is this meant to fix #1664 ?

gallais:
#1664 has a complex semantics where `default` arguments are filled in at call sites.
It also provides things that require (at least a little bit of) scopechecking.
This is a purely syntactic transformation that happens directly in the parser.

PR #2826, opened 2022-12-21
[ docs ] Build IdrisDoc in a namespace.

jfdm:
When invoking `idris2 --mkdocs foobar.ipkg`, the current behaviour is for the compiler to dump the documentation in `${builddir}/docs`. The location `${builddir}/docs` is used regardless of the package name. Thus for projects that present multiple packages the `docs` build directory will become polluted and corrupt if the developer is not careful. This one change ensures that for package `x` the documentation is generated in the directory: `${builddir}/docs/x`.

Idris2 `Makefile` has been updated to reflect the change.

### DIFF
--- a/.github/scripts/katla.sh
+++ b/.github/scripts/katla.sh
@@ -20,7 +20,7 @@ rm tmp
 find "$prefix"/* -maxdepth 0 -type d >tmp
 while IFS= read -r libdirectory; do
     libname=$(echo "$libdirectory" | sed "s|$prefix/||")
-    cp -r "$prefix"/"$libname"/build/docs/* html/"$libname"
+    cp -r "$prefix"/"$libname"/build/docs/"$libname"/* html/"$libname"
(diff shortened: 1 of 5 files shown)


gallais:
IIUC this will require us to update the CI script that generates the docs automatically:

https://github.com/idris-lang/Idris2/blob/45fc038300979a9e3ff7799e8c22f3fc80b4014d/.github/scripts/katla.sh#L23


Something like:
```diff
-     cp -r "$prefix"/"$libname"/build/docs/* html/"$libname"
+     cp -r "$prefix"/"$libname"/build/docs/"$libname"/* html/"$libname"
```

jfdm:
Ah good catch! I thought I had updated everything that used the docs.

jfdm:
I will leave this open until Tuesday, when I will merge it.

gallais:
Abandoned.

PR #2833, opened 2022-12-26
[ prelude ] Add lacking implementation of `Traversable` for `Pair`

buzden:
`Either` is bitravesable, right-biased functor and right-biased traversable. `Pair` is bitravesable, right-biased functor, but is not a traversable. This PR fixes this surprising irregularity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,8 @@
 
 * Improved performance of functions `isNL`, `isSpace`, and `isHexDigit`.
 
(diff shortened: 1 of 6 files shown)


gallais:
I think @edwinb may object to that.

buzden:
> I think @edwinb may object to that.

Any clue why?

gallais:
Some people claim it's too confusing.

I can't remember whether Edwin is one of them (I'm not).
I'll ping him after the holidays (@-ing on github is fine as he does not read these notifications :smile:)

Issue #2834, opened 2022-12-26
Potential elaborator bug: Invalid @-pattern generation in with block


nmccarty:
It appears that the elaborator is generating some invalid @-patterns when you try to use a with block on the `SplitRec` view over `Vect`, after asking about this on the discord @gallais suggested I file a bug.  

# Steps to Reproduce

The following example fails to compile:

```idris
import Data.Vect
import Data.Vect.Views.Extra

mergeSort : Ord a => {n : _} -> Vect n a -> Vect n a
mergeSort input with (splitRec input)
  mergeSort [] | SplitRecNil = []
  mergeSort [x] | SplitRecOne = [x]
  mergeSort (xs ++ ys) | (SplitRecPair {xs} {ys} xs_rec ys_rec) =
    merge (mergeSort xs | xs_rec) (mergeSort ys | ys_rec)
```

# Expected Behavior

I would expect this to compile properly, like the following (hopefully) equivalent version over lists:

```idris
import Data.List
import Data.List.Views

mergeSort : Ord a => List a -> List a
mergeSort input with (splitRec input)
  mergeSort [] | SplitRecNil = []
  mergeSort [x] | (SplitRecOne x) = [x]
  mergeSort (lefts ++ rights) | (SplitRecPair lefts rights lrec rrec) =
    merge (mergeSort lefts | lrec) (mergeSort rights | rrec)
```

# Observed Behavior

The compiler says:
```
Welcome to Idris 2.  Enjoy yourself!
1/1: Building Failing (Failing.idr)
Error: While processing right hand side of with block in mergeSort. @-patterns only allowed in pattern clauses
```

gallais:
Fixed by #2837

PR #2837, opened 2022-12-28
[ fix ] at-pattern leak in recursive with blocks (#2834)

dunhamsteve:
Fix an issue where an @-pattern is leaked into the right side.  It occurs in a recursive call to a `with` clause if there are @ patterns and implicits. A typo in the function `getMatch` causes the left hand side to be used for the right hand side after an implicit is hit. 

**Should this change go in the CHANGELOG?**

This PR is just a bug fix, so it probably doesn't need a change log entry.  Let me know if it should have one.


gallais:
Any chance you can port the fix to yaffle too, @dunhamsteve?
Same issue show up there: https://github.com/edwinb/Yaffle/blob/main/src/TTImp/WithClause.idr#L86

PR #2838, opened 2022-12-29
[ performance ] String builder on all scheme backends

Z-snails:
this makes it about 50x faster at printing scheme code

I used this python script to generate a huge idris file:
```python
count = 100

out = "" 

out += "plus : Nat -> Nat = (+)\n"
out += "%noinline\n"

for i in range(count):
    out += "f" + str(i)
    if i < count - 1:
        out += ", "
out += " : Nat -> Nat\n"

pluses = "x"
for i in range(500):
    pluses = "plus x (" + pluses + ")"

for i in range(count):
    out += "f" + str(i) + " x = " + pluses + "\n"

out += "main : IO ()\n"
out += "main = do\n"
for i in range(count):
    out += "    printLn $ f" + str(i) + " 100\n"

print(out)
```

I added a call to `logTime` to this line: https://github.com/idris-lang/Idris2/blob/main/src/Compiler/Scheme/Chez.idr#L564

Previously it took 1.193s to pretty print the scheme code, with this change it takes 0.026s, a roughly 50x improvement

I also took the opportunity to clean up the code a little

Not necessary, it just makes things faster

### DIFF
--- a/idris2api.ipkg
+++ b/idris2api.ipkg
@@ -177,6 +177,7 @@ modules =
     Libraries.Data.Span,
     Libraries.Data.SortedMap,
     Libraries.Data.SortedSet,
+    Libraries.Data.String.Builder,
     Libraries.Data.String.Extra,
     Libraries.Data.String.Iterator,
     Libraries.Data.StringMap,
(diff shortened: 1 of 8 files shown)


stefan-hoeck:
Would it make sense to add this also to base (together with some explanation, why this is fast and when to use it)?

Z-snails:
> Would it make sense to add this also to base (together with some explanation, why this is fast and when to use it)?

Possibly, although I wouldn't want that to block this PR. It can be done later after a discussion about if it fits in base, or would be better as a library

PR #2842, opened 2023-01-07
[docs] javascript FFIs and packaging of JS or C support files

mattpolzin:
See https://github.com/idris-lang/Idris2/pull/2836 for context.

### DIFF
--- a/docs/source/reference/packages.rst
+++ b/docs/source/reference/packages.rst
@@ -172,6 +172,10 @@ Idris 2 treats specially when the package is depended upon by some other Idris p
 
 Those directories are ``lib`` and ``data``.
 
+Below we will refer to these directories as ``pkgname-<version>/lib`` and ``pkgname-<version>/data``
+and in all cases these folders may be located locally in the depending project within a ``depends``
+directory or globally (you can ask Idris 2 for the global install directory with ``idris2 --libdir``).
+
 Library files
 -------------
 
--- a/docs/source/reference/packages.rst
+++ b/docs/source/reference/packages.rst
@@ -162,3 +162,47 @@ package will then be located as follows:
 In each case, if more than one version satisfies the constraint, it will choose
 the one with the highest version number.
 If package versions are omitted in directory names, they are treated as the version ``0``.
+
+Support file install directories
+================================
+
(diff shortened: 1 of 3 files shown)


docs/source/reference/packages.rst
```diff
@@ -162,3 +162,47 @@ package will then be located as follows:
...
+and rely on a ``C`` support file. The FFI interface is documented elsewhere so we won't go into that
+here. Once you do have a support file written in ``C``, you can build an ``so`` file using your
+package's post-build hook. Then, install (copy) that ``so`` file into the ``lib`` subfolder where
+Idris has installed your package in your post-install hook. When an executable depends on your package,
+Idris will copy shared object files from the ``lib`` directory into the build folder for the exectuable.
+
+Data files
+----------
+
+Idris will look for data files at ``pkgname-<version>/data``. Data files are a bit open-ended
```
Z-snails, line 195: I think it would be good to make it clear this is refering to the global idris 2 install directory
(ie something like: idris will look for data files in idris's install directory, at <path>, where path is something along the lines of `$(idris2 --libdir)/...` (I can't remember the exact path right now))

mattpolzin: I think that information is covered by the preceding section ("Where does Idris look for packages?") pretty adequately so I attempted to make it clear I was referring to that directory at the top of the new section by saying 
>  noted above, packages can be installed globally or locally [...]. In either case, there are two directories within a package's install root that Idris 2 treats specially [...].

Do you think I should be clearer on that? Either way, it would not quite be accurate to say that the `pkgname-<version>/data` directory was under the global install location because it should work the same way when a dependency lives in the local `depends` directory.

Z-snails: Good point, I think would be useful to say where/how to figure out where that global directory is.
Although if you are planning to auto-install support files, this might not be so important

mattpolzin: Can't hurt, I'll add more detail to the top of the section (since the information applies to locating both the "data" and "lib" directories.

mattpolzin: Let me know if you think the new text adds enough clarity.

Z-snails: Looks good!

PR #2843, opened 2023-01-07
[new] Vect.Quantifiers.All QoL

mattpolzin:
I've included 3 separate small commits for QoL improvements on `Data.Vect.Quantifiers.All`. I am ok with these commits being individually approved/denied and will drop any commits that aren't deemed worthwhile.

1. Add a `Show` instance for `Data.Vect.Quantifiers.All`. Super minor, but definitely convenient. Precedent is set here by the `Show` instance for `Data.List.Quantifiers.All`.
2. Add an alias for `HVect` to `Data.Vect.Quantifiers.All`. There is a fully separate `HVect` type in the `contrib` library, but for 95% of operations I don't need the helpers defined against the contrib type (and if I do, they can be written against `All` anyway, so I just do that myself at that time). I've seen `All id` be recommended many times for an `HVect` type in various forums as well, so I believe the alias serves as a nice way to show people that `All id` _is_ `HVect` without them reaching out and asking about it on Discord (or similar). Precedent for this alias is also set by an `HList`  alias in `Data.List.Quantifiers.All`.
3. Add `head`, `tail`, `drop`, and `drop'` to `Data.Vect.Quantifiers.All` to make it more at home in listy computations. This is just something I recently needed so I figured I would pitch it.

~I'll update the CHANGELOG once I know which if any of these commits are going to be accepted.~ [EDIT] Done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,11 @@
   convert a `Bits8` to an `Int`), as their values are limited, as opposed to the
   assumption in `setByte` that the value is between 0 and 255.
 * Adds RefC support for 16- and 32-bit access in `Data.Buffer`.
+* Add `Show` instance to `Data.Vect.Quantifiers.All` and add a few helpers for listy
+  computations on the `All` type.
+* Add an alias for `HVect` to `All id` in `Data.Vect.Quantifiers.All`. This is the
+  approach to getting a heterogeneous Vect of elements that is generall preferred by
+  the community vs. a standalone type as seen in `contrib`.
 
 #### System
 
--- a/libs/base/Data/Vect/Quantifiers.idr
+++ b/libs/base/Data/Vect/Quantifiers.idr
@@ -125,3 +125,13 @@ namespace All
   forget : All (const p) {n} xs -> Vect n p
   forget []      = []
   forget (x::xs) = x :: forget xs
+
+  export
+  {0 xs : Vect n _} -> All Show (map p xs) => Show (All p xs) where
+    show pxs = "[" ++ show' "" pxs ++ "]"
+      where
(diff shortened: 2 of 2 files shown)


gallais (approved):

mattpolzin:
Whoops. Naturally, aliasing `HVect` _can_ create ambiguity, but it did so even in the elab-util library that we test against!

I kind of feel like it's worth heading in this direction even though it requires an update to elab-util but I realize this is an additional burden and the change is not a necessary one.

mattpolzin:
Ok, looking for your thoughts @stefan-hoeck, but I think what I discovered builds a strong case for making this change and updating elab-util along with it.

With only the following diff to the elab-util package that package builds & passes its own tests. That is, elab-util, like many many other use-cases, does not need anything out of `HVect` that isn't provided by `All id` anyway.

```diff
diff --git a/src/Language/Reflection/Pretty.idr b/src/Language/Reflection/Pretty.idr
index 8045d1a..cfe01b0 100644
--- a/src/Language/Reflection/Pretty.idr
+++ b/src/Language/Reflection/Pretty.idr
@@ -15,7 +15,8 @@
 |||   :exec putPretty `((1 index : Fin n) -> Vect n t -> t)
 module Language.Reflection.Pretty
 
-import public Data.HVect
+import public Data.Vect
+import public Data.Vect.Quantifiers
 import Data.String
 import public Language.Reflection
 import public Text.PrettyPrint.Prettyprinter
@@ -25,6 +26,10 @@ import Language.Reflection.Types
 
 %default total
 
+public export
+HVect : Vect n Type -> Type
+HVect = All id
+
 --------------------------------------------------------------------------------
 --          Utils
 --------------------------------------------------------------------------------
```

PR #2848, opened 2023-01-09
[ fix ] make color output toggling simpler and also more robust.

mattpolzin:
Closes https://github.com/idris-lang/Idris2/issues/1917.

I know that this works well in my testing of various Idris 2 CLI options and the REPL.

Would someone be willing to sanity check this works for IDE mode / emacs? It absolutely ought to, but seeing as how that was the express reason for checking `TERM` for "dumb" I'd like to be more confident than I am as a non-emacs user.

### DIFF
--- a/src/Idris/Env.idr
+++ b/src/Idris/Env.idr
@@ -43,7 +43,8 @@ envs = [
     MkEnvDesc "CPPFLAGS"             "RefC backend: C preprocessor flags (IDRIS2_CPPFLAGS takes precedence).",
     MkEnvDesc "LDFLAGS"              "RefC backend: C linker flags (IDRIS2_LDFLAGS takes precedence).",
     MkEnvDesc "NODE"                 "NodeJS backend: NodeJS executable.",
-    MkEnvDesc "PATH"                 "PATH variable is used to search for executables in certain codegens."]
+    MkEnvDesc "PATH"                 "PATH variable is used to search for executables in certain codegens.",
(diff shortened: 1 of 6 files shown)


src/Idris/Env.idr
```diff
@@ -43,7 +43,8 @@ envs = [
     MkEnvDesc "CPPFLAGS"             "RefC backend: C preprocessor flags (IDRIS2_CPPFLAGS takes precedence).",
     MkEnvDesc "LDFLAGS"              "RefC backend: C linker flags (IDRIS2_LDFLAGS takes precedence).",
     MkEnvDesc "NODE"                 "NodeJS backend: NodeJS executable.",
-    MkEnvDesc "PATH"                 "PATH variable is used to search for executables in certain codegens."]
+    MkEnvDesc "PATH"                 "PATH variable is used to search for executables in certain codegens.",
+    MkEnvDesc "NO_COLOR"             "Instruct Idris not to print color to stdout. Passing the --color/--colour option will supercede this env var."]
```
mattpolzin, line 47:
```suggestion
    MkEnvDesc "NO_COLOR"             "Instruct Idris not to print color to stdout. Passing the --color/--colour option will supersede this env var."]
```

mattpolzin:
> Would someone be willing to sanity check this works for IDE mode / emacs? It absolutely ought to, but seeing as how that was the express reason for checking TERM for "dumb" I'd like to be more confident than I am as a non-emacs user.

@dunhamsteve was kind enough to give me a sanity check on emacs support. I won't assume there's zero chance of this breaking something, but I feel good about moving it into the main branch and I am happy to address any problems that folks find over time.

PR #2852, opened 2023-01-13
Fix build on FreeBSD

mneumann:
gmp is located in `/usr/local/{include,lib}`.

With this patch, to build on FreeBSD:

```
$ sudo pkg ins chez-scheme gmp
$ cd Idris2 && gmake bootstrap SCHEME=chez-scheme && gmake install
```

- [ ] If this is a user-facing change or a compiler change, I have updated

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -36,6 +36,11 @@ else
 	SHLIB_SUFFIX := .so
 endif
 
+ifneq (, $(findstring freebsd, $(MACHINE)))
+	CFLAGS += -I/usr/local/include
+	LDFLAGS += -L/usr/local/lib
+endif
+
 ifneq ($(OS),windows)
 	CFLAGS += -fPIC
 else ifneq (, $(findstring NT-6.1,$(shell uname)))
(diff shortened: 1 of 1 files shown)


mattpolzin:
It is surprising that these pretty standard locations are not already being looked at, but I am seeing others talk about similar issues in various places online so since this change feels innocuous and you've made it quite targeted I am inclined to accept the PR. I'm just doing a tad more reading.

mattpolzin:
Would it be slightly more portable (within FreeBSD installs) to use `${LOCALBASE}` instead of `usr/local`?

mneumann:
@mattpolzin it's quite common that we have to add these paths on FreeBSD or DragonFlyBSD :). Usually, this takes place through patches in our ports system, but of course, it's easier when already done by the language authors.

Concerning `$LOCALBASE`: It is defined (on FreeBSD) in `/usr/ports/Mk/bsd.default-versions.mk` and cannot be easily used outside the ports system (it's a BSD makefile). It might not exists in case binary packages are used. But `/usr/local` is pretty much the standard location for any installed package (from FreeBSD ports collection). 

I've updated my PR and use `sysctl -n user.localbase`, which returns `/usr/local` on my system, but may return sth else when a different local base is used.

mattpolzin (approved):
Thanks! I'll merge once CI finishes.

PR #2854, opened 2023-01-17
[ base ] fix the definition of die

gallais:
We should not print error messages to stdout

### DIFF
--- a/libs/base/System.idr
+++ b/libs/base/System.idr
@@ -273,4 +273,12 @@ exitSuccess = exitWith ExitSuccess
 ||| Print the error message and call exitFailure
 export
 die : HasIO io => String -> io a
-die str = do putStrLn str; exitFailure
+die str
+  = do Right () <- fPutStrLn stderr str
+         | Left err => assert_total $ die
+           """
+           Error \{show err}
+           when printing the error
+           \{str}
+           """
+       exitFailure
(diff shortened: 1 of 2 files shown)


libs/base/System.idr
```diff
@@ -273,4 +273,12 @@ exitSuccess = exitWith ExitSuccess
 ||| Print the error message and call exitFailure
 export
 die : HasIO io => String -> io a
-die str = do putStrLn str; exitFailure
+die str
+  = do Right () <- fPutStrLn stderr str
+         | Left err => assert_total $ die
```
mattpolzin, line 278: This seems at least half-likely to loop forever given the unlikely event that it fails to print to stderr because I don't know how that could happen and it might be a problem that isn't intermittent.

mattpolzin (commented):
Not very likely, but in case someone is using die in a program and then wrapping it in a script that takes output and does something with it, this could be a breaking change of sorts. IMO would be a good candidate for the CHANGELOG.

libs/base/System.idr
buzden: As far as I know, the most likely reason for an error when printing to the `stderr` is the broken pipe error, which won't be cured any time, thus this code will always loop forever. I think, in this case it's better to lose an error message than to loop forever.

PR #2869, opened 2023-01-31
Create a separate Makefile for the Idris 2 support libraries.

mattpolzin:
This superficially just looks like moving stuff around, but putting all of the Makefile concerns for building the support libraries into a new Makefile is a step toward creating a Nix derivation that builds only the support libraries (a standing TODO item: https://github.com/idris-lang/Idris2/blob/main/nix/package.nix#L54).

I've started to put effort into the Nix derivation in question, but this refactor of the Makefile will be easier to review on its own since it is pretty straight forward.

The follow-up WIP branch can be seen [here](https://github.com/mattpolzin/Idris2/pull/2).

There is one thing I did in this PR that is strictly cleaning up: I removed the `cp` to a `lib` dir at [line 205 of the Makefile](https://github.com/idris-lang/Idris2/blob/main/Makefile#L205). This was an apparently vestigial* location for a `lib` directory because the directory copied at [lines 56-57 of the `c` Makefile](https://github.com/idris-lang/Idris2/blob/main/support/c/Makefile#L56..L57) is the directory actually added to Idris 2's `CG Library Directories` paths.

*ok, I've gone and tracked down where the extra lib install was added; apparently it was specifically for when the compiler is built with Racket... https://github.com/idris-lang/Idris2/commit/3eb67aebd87c5c0c55d9f513d0fde0f51ff5c4dc -- do we want this anymore if CI workflows that use Racket are succeeding?

[UPDATE] I've done more investigation. When the `${PREFIX}/lib` directory was added, the Racket build had not yet gained the now-ubiquitous executable wrapper that sets `LD_LIBRARY_PATH`. In the current Racket builds, this `lib` directory is not needed -- to check, I spun things up in a clean ubuntu environment and performed both bootstrap and self-host builds of the compiler using the Racket CG. I've added commits to this PR that tweak the INSTALL.md based on that experience.


CHANGELOG.md
```diff
@@ -116,6 +116,10 @@
   recognized as a "data" directory by Idris 2. See the
   [documentation on Packages](https://idris2.readthedocs.io/en/latest/reference/packages.html)
   for details.
+* The compiler no longer installs its own C support library into `${PREFIX}/lib`. This folder's
+  contents were always duplicates of files installed into `${PREFIX}/idris2-${IDRIS2_VERSION}/lib`. If you
+  need to adjust any tooling or scripts, point them to the latter location which still contains
+  these installed library files.
```
mattpolzin, line 122:
```suggestion
  these installed library files.
* Renamed `support-clean` Makefile target to `clean-support`. This is in line with most of the `install-<something>` and `clean-<something>` naming.
```

PR #2872, opened 2023-02-04
[ re #2871 ] Fix the resolvedN issue for good

gallais:


mattpolzin:
But, aren't I going to stop feeling like a human computer if I don't have to translate machine names while debugging? 😄

PR #2881, opened 2023-02-07
[ fix #2087 ] Fix multiline string with CRLF

### LINKED ISSUE #2087: Multiline strings issue when running tests on Github CI

andylokandy:
This patch resolves issue #2087 and enhances the handling of multiline string literals:

1. Recognize `\r\n` at the end of a line, which was recognized as `\n\n` for two lines.
2. Dectect padding error before unescaping the literal, therefore this case becomes invalid:
```idirs
"""
    A line
\n
    Another line
    """
```

Closes https://github.com/idris-lang/Idris2/issues/2087

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@
   the number of configured processors.
 
 #### Contrib
+
 * Remove Data.List.HasLength from contrib library but add it to the base library
   with the type signature from the compiler codebase and some of the naming
   from the contrib library. The type ended up being `HasLength n xs` rather than
@@ -112,6 +113,7 @@
   J. Garret Morris, and Aaron Stump
 
 ### Other Changes
+
 * The `data` subfolder of an installed or local dependency package is now automatically
   recognized as a "data" directory by Idris 2. See the
   [documentation on Packages](https://idris2.readthedocs.io/en/latest/reference/packages.html)
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -27,6 +27,7 @@ import Idris.Elab.Interface
 import Idris.Desugar.Mutual
 
 import Parser.Lexer.Source
+import Parser.Support
 
 import TTImp.BindImplicits
 import TTImp.Parser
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,14 @@
   substituted at parsing time with a string corresponding to the
   location, filename, line or column number associated to the
   magic constant's position.
+* Multiline string literals are now has new line (`\n`) for each line of the
```
gallais, line 11: Looks like you started writing one sentence then went for another but there are still remnants from the old one making it unintelligible (at least that's what happens to me when I produce similar sentences).

gallais (commented):
Does this mean it is now *impossible* to write a multiline string without a trailing newline?
If so I think that's a regression.

andylokandy:
>Does this mean it is now impossible to write a multiline string without a trailing newline?
If so I think that's a regression.

At the time of implementing Multiline String, the `lines` function did not insert a newline at the end of the last line. The multiline string was therefore designed to mirror this behavior. Later, a PR modified the lines function so that it included a newline at the end of all lines, including the final one. So in this PR, I attempted to follow this change. If we believe the original behavior of the multiline string was better suited, I would revert the modification.

mattpolzin:
I think you mean the `unlines` function always puts a newline at the end of its input, right? `lines` produces an empty list of lines from the empty string so it does not have a behavior of producing any newlines. 

I don’t really think of mutliline string literals as semantic equivalents to calling `unlines` on a list. I think of it as not needing to because I’ve got a multiline block of text in a literal to begin with. It also feels like putting a chunk of a file into the code and in my editor I’d expect to represent a trailing newline by a blank line at the end much the same as is the case with multi line strings prior to this PR.

andylokandy:
Thanks for the suggestions, I've reverted the behavior change on last new line.

PR #2884, opened 2023-02-11
[ contrib ] move SortedMap and SortedSet to base

stefan-hoeck:
So, here's my first take on disbanding parts of contrib (more will follow). I realize that there are different opinions on these container types, but I agree with @mattpolzin that we should use what we currently have, because every thing else might just never happen. On the other hand, starting with this is pretty safe, because it shouldn't break any libraries.


mattpolzin (approved):
Thanks!

I am going to leave this approved but unmerged for a brief time because it is definitely a big decision.

This move is of course in line with discussion from the [wiki proposal](https://github.com/idris-lang/Idris2/wiki/%5BLanguage%5D-Contrib-Organisation), but if the PR generates more discussion than the document did then we all still win by letting that discussion occur.

One thing that we might make room for is discussion around @andrevidela's preference from the wiki document
> [...] of including just the dependent version of SortedMap and introduce a non-dependent interface for other types of dictionaries (Hashmaps/non-dependent sorted map).

Personally, I am not clear on what this preference means, but if Andre has time to elaborate then I would be interested in hearing more.

gallais (approved):

mattpolzin:
To the comment that this PR
> shouldn't break any libraries

I'd agree except for the fact that the existing contrib version of these types will not be removed automatically so likely some will see ambiguity errors on their machines. I doubt that will be an issue for most if not all CI flows where libraries are installed from scratch. 

This is not an argument against the PR, just creating something for those who are troubleshooting to latch into. 

The solution is simply to delete the `Data/SortedMap.idr` and `Data/SortedMap` folder from the `${PREFIX}/idris2-0.6.0/contrib-0.6.0/` folder by hand. 

This is only an issue for development builds because the next Idris release will use a new versioned subfolder that won't collide anyway.

mattpolzin:
A note for future discussion is that we could avoid this problem in 100% of use-cases by versioning then library independently of the compiler (maybe contrib and base move to 0.6.2), but there may currently be people or scripts that are presumptive about library and compiler versions aligning (and reasonably so at this point since it's always been the case).

andrevidela:
@mattpolzin just for posterity: I don't remember what I meant with this comment and those changes look good to me. Sorry I didn't have the chance to answer earlier

PR #2886, opened 2023-02-12
[ contrib ] Performance improvement gcd in Data.Nat.Factor

Alex1005a:
Data.Nat.Factor.gcd function performance more than doubled (but still very slow).
Also added total and more faster gcdUnproven function, Factor properties related to modulo and one Division property.


Alex1005a:
It is possible to increase performance to the level of gcdUnproven, if you replace `CommonFactor p a b` with `(() -> CommonFactor p a b)` in GCD definition.
`
data GCD : Nat -> Nat -> Nat -> Type where
    MkGCD : {a, b, p : Nat} ->
        {auto notBothZero : NotBothZero a b} ->
        (() -> CommonFactor p a b) ->
        ((q : Nat) -> CommonFactor q a b -> Factor q p) ->
        GCD p a b
`
Thus, one can simply postpone CommonFactor calculation. 
I'm not sure if this is correct so I didn't do it.

gallais:
That's great! On

```idris
main : IO ()
main = print (fst $ gcd 1000000 2084 @{LeftIsNotZero})
```

I get an execution time from 10+s before your PR to less than 1s afterwards.
I'll add a perf test case and then merge.

PR #2894, opened 2023-02-24
[ refactor ] TTC Integer in terms of TTC Nat

gallais:
Implementing TTC Nat in terms of TTC Integer meant having a useless tag
(Nats are always >= 0). Let's do the opposite instead!


gallais:
Giving up on this for now: it does not make a measurable difference & using a better representation for limbs (`Bits32` instead of `Int`) requires changing the way we represent `Bits32` too (we're currently casting them on `Integer` and reusing the `Integer` primitive).

PR #2896, opened 2023-02-25
Fix 2895

### LINKED ISSUE #2895: Quantity checker gets confused by applications
In some complicated programs involving `0` or `1` the linearity checker sometimes throws 

> *complicated type* not a function type

The error message itself leaves a lot to be desired because the names aren't printed. More more importantly, this error is almost always a bug and workarounds are readily accessible, so either:
- the workarounds are wrong and it's a bug
- the error itself is a bug
- None of this is a bug, the error message should be changed.

Given how intuitive the workaround are, I suspect it's number 2 and the linearity checked should be updated to avoid this error entirely. I have two code samples,
one where the error is triggered by a pattern, and one where the error is triggered by a lambda

# Steps to Reproduce

#### Trigger with pattern

```haskell
record Ty where
  constructor MkTy
  0 obj : Type

record TyProj (d: Ty) where
  constructor MkTyProj
  proj : d.obj

record Id (x : Type) where
  constructor MkId
  getId : x

DepLensToDepOptic :
  {c1 : TyProj (MkTy (Nat -> Type))} ->
  Id (c1.proj 3) -> -- remove the Id works
  Type
DepLensToDepOptic (MkId b') = ?hu
```

#### Trigger with lambda

```haskell
record Ty where
  constructor MkTy
  0 obj : Type

record TyProj (d: Ty) where
  constructor MkTyProj
  proj : d.obj

DepLensToDepOptic :
  {c1 : TyProj (MkTy (Nat -> Type))} ->
  (c1.proj 3) ->
  Type
DepLensToDepOptic =
   (\(b) => ?mmm)
-- DepLensToDepOptic b = ?hole -- this works
```

# Expected Behavior

Should typecheck, specially given that:

- In the first case, removing `Id`  works.
- In the second case, putting `b` in the lhs works

# Observed Behavior

```
Linearity checking failed on (Main.TyProj.(.proj) [__] c1[0]) ([__] not a function type)
```
and
```
Linearity checking failed on (Main.TyProj.(.proj) [__] c1[1]) ([__] not a function type)
```
respectively

andrevidela:
fix #2895

### DIFF
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -2387,6 +2387,15 @@ addLogLevel Nothing  = update Ctxt { options->session->logEnabled := False, opti
 addLogLevel (Just l) = update Ctxt { options->session->logEnabled := True, options->session->logLevel $= insertLogLevel l }
 
 export
+setLogLevel : {auto c : Ref Ctxt Defs} ->
+              LogLevel -> Core ()
+setLogLevel = addLogLevel . Just
+
+export
+stopLogging : {auto c : Ref Ctxt Defs} -> Core ()
+stopLogging = addLogLevel Nothing
+
+export
 withLogLevel : {auto c : Ref Ctxt Defs} ->
                LogLevel -> Core a -> Core a
 withLogLevel l comp = do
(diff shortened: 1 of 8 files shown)


andrevidela:
any idea what this is? 

> Exception: attempt to reference unbound identifier PreludeC-45Uninhabited-__con-603 at line 2840, char 109 of /home/runner/work/Idris2/Idris2/tests/idris2/api001/build/exec/lazy-idris2_app/lazy-idris2.ss

I'm not able to reproduce it locally either

PR #2903, opened 2023-03-01
[ new ] Implement bit-rotation operators

CodingCellist:
Whereas `shiftR` and `shiftL` throw bits off the edge, the `rotR` and `rotL` operations wrap the bits around to the start of the bit-stream.


gallais:
For this kind of stuff I feel like we should really have alternative show functions that print the bit pattern.
I am willing to believe the test cases make sense but I am unfortunately not enough of a machine to check it myself
just by eyeballing the expected output.

CodingCellist:
> For this kind of stuff I feel like we should really have alternative show functions that print the bit pattern. I am willing to believe the test cases make sense but I am unfortunately not enough of a machine to check it myself just by eyeballing the expected output.

I entirely agree. I _did_ catch an implementation error via the tests though, so I am fairly confident that they now work correctly. But this is based on this by-hand check for a single `Bits8` case ^^;;
image

gallais:
Okay I got the bit pattern visualisation going. Looks nice.

PR #2904, opened 2023-03-01
[ new ] Couple of useful things for Vect

CodingCellist:
Adds the ability to split a `Vect (k * n)` into k blocks of n or vice-versa, as well as the ability to apply a permutation to a `Vect`.


CodingCellist:
These functions came up naturally as I was trying to do stuff with `Vect`s and indices. I also wanted to add:
```idris
||| All of the Fin elements as a Vect whose length is the number of elements.
|||
||| Useful, for example, when generating all the indices of a Vect.
public export
allFinsV : (n : Nat) -> Vect (S n) (Fin (S n))
allFinsV Z = FZ :: []
allFinsV (S n) = FZ :: map FS (allFinsV n)
```
But this requires importing `Data.Vect` into `Fin`, which creates a module import loop!
```
Uncaught error: Module imports form a cycle:
Decidable.Equality -> Data.DPair -> Data.Vect -> Data.Fin -> Data.List
  -> Data.Buffer -> System.File -> Control.App.FileIO -> Data.List
```
:scream:

gallais:
I think we should have the more general:

```
allFins : (n : Nat) -> Vect n (Fin n)
```

CodingCellist:
> I think we should have the more general:
> 
> ```
> allFins : (n : Nat) -> Vect n (Fin n)
> ```

Where should we place it though? Would putting it in `Data.Vect` avoid the cyclical import?

buzden:
> Where should we place it though? Would putting it in `Data.Vect` avoid the cyclical import?

It definitely should

CodingCellist:
Cool, I've added it to `Data.Vect`  : )

libs/base/Data/Vect.idr
```diff
@@ -862,9 +899,27 @@ zipWith3IndexLinear : (0 f : _) -> (xs, ys, zs : Vect n a) -> (i : Fin n) -> ind
...
+||| @ v the vector whose elements to rearrange
+||| @ p the permutation to apply
+|||
+||| ```idris example
+||| > permute ['a', 'b', 'c', 'd'] [0, 3, 2, 1]
+||| ['a', 'd' , 'c' ,'b']
+||| ```
+export
+permute : {len : _} -> (v : Vect len a) -> (p : Vect len (Fin len)) -> Vect len a
+permute v p = zipWith index p (replicate len v)
```
gallais, line 917: Isn't this `map (flip index v) p`? No need to keep a copy of `len` at runtime

gallais:
I ended up pushing & merging because I needed it for #2903

PR #2905, opened 2023-03-01
[ admin ] Update what should go in the changelog

CodingCellist:
There has been some stuff inconsistently slipping through the cracks, so it is probably best to explicitly include it.

(I'm being pedantic, I know, but I'd rather avoid having to re-crawl the git log à la #2703 :sweat_smile: )

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,7 @@
-# Description
+## Description
 
 
-# Should this change go in the CHANGELOG?
+## Should this change go in the CHANGELOG?
 
 <!-- Please delete this section if it doesn't apply -->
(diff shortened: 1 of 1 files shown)


mattpolzin (commented):
Isn't "an addition" too general and at that point you could just say "update the CHANGELOG for every change". I guess it rules out deletion of code that doesn't affect end users?

CodingCellist:
> Isn't "an addition" too general and at that point you could just say "update the CHANGELOG for every change". I guess it rules out deletion of code that doesn't affect end users?

Yeah I wasn't certain on that one. It was to catch the new paper additions/implementations, which used to go in the changelog. Guess I should just write that then ^^

PR #2906, opened 2023-03-02
Refactor Data.Nat to use preorder reasoning to improve readability

ohad:


buzden:
It's all about personal preferences, but I'd say that a single `rewrite` with a simple recursive call reads must simpler and understandable to me than one with preorder reasoning syntax and explicit `cong`.

ohad:
I'm happy not to merge this --- I was asked to produce this PR, and now I have. I'll let others discuss the merits and drawbacks.

CodingCellist:
I agree with buzden for the first 3-4 proofs, the single `rewrite` recursive call is clearer. But the monotone and mult proofs are definitely easier to understand now imo  : )

PR #2909, opened 2023-03-03
Add some useful OpenUnion functions

Alex1005a:
Also this will add a sublist relation.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,9 @@
 
 * Added new buffer (set/get) functions for built-in types `Bool`, `Nat`, `Integer`.
 
+* Added `swapInterleaving`, `nilInterleaving`, and `elemInterleaving` for Interleaving
+  in `Data.List.Quantifiers`.
+
 * Tightened the types of:
   `setInt16` (now takes an `Int16` instead of an `Int`),
   `setInt32` (now takes an `Int32` instead of an `Int`),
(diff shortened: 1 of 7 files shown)


Alex1005a:
The past definition was more universal, but difficult for proofs. Using OPE will be easier.

gallais:
Between 1) the breaking change of moving content between libraries 2) the use of auto and 3) the presence of computations in the return type (as opposed to using `Interleaving`), I am not convinced.

Alex1005a:
> the use of auto

Is this bad practice? Or did I use auto in the wrong place?


> the breaking change of moving content between libraries

Sorry I don't quite understand what you mean.

gallais:
> Is this bad practice? Or did I use auto in the wrong place?

It is bad practice IMO to use `auto` for argument whose value may not be unique.
`IsSublist (a :: []) (a :: a :: [])` has two possible proofs and `auto` will arbitrarily pick
one or the other. So the meaning of the program is not obvious from the source
code.

Note that you don't need to take all of `(xs, ys : List a)` and `IsSublist xs ys` explicitly:
idris can reconstruct `xs` and `ys` from `IsSublist xs ys` by unification.
And so it's usually better to program directly with the `IsSublist` value and let the `xs`
and `ys` be passed around implicitly.

> Sorry I don't quite understand what you mean.

When you move definitions from `contrib` to `papers` you are breaking code that
relies on these definitions being in `contrib`. I don't think this should be done without
good reason.


As for `Interleaving`, this is the type I would expect:
```diff
-decompSublist : {auto sublist : IsSublist ts xs} -> UnionF elt xs -> Either (UnionF elt (removeSublist sublist)) (UnionF elt ts)
+decompInterleaving : Interleaving xs ys zs -> UnionF elt zs -> Either (UnionF elt xs) (UnionF elt ys)
```
which is IMO a lot cleaner (and plays more nicely with unification).

Alex1005a:
Thanks a lot for the explanation!

gallais (commented):
There's still a lot of issues with the PR related to deep problems with inference, search, and performance.

I don't want to spend more time driving the changes as it's targeting a module
that is a priori not here for public consumption but rather showcasing the language /
some advanced ideas (that's the purpose of `libs/papers`).

I'm going to close this PR based on what we say in the CONTRIBUTING file:

> Please remember when making contributions via pull requests, though, that Idris is primarily a research project, and we are a small team, which limits the time we have available for reviewing and maintaining PRs. Nobody works full time on Idris - we're a team of academics and students at a university with other demands on our time such as teaching, and writing and presenting research papers, which will always take priority. This also means we have to be careful to make sure that we can commit to maintaining any new features which are contributed.

Feel free to experiment with building a 3rd party library on top of `Data.OpenUnion`.

PR #2914, opened 2023-03-05
[ base ] Support unbuffered reading of stdin

dunhamsteve:
# Description

Currently `getChar` doesn't return a character until the user presses return, because stdin defaults to "cooked" mode.  This PR adds `enableRawMode` and `disableRawMode` primitive functions to allow reading stdin character by character with `getChar`. 

This works on Linux and MacOS, but I was unable to find an acceptable solution for windows.

- [x] I have updated `CHANGELOG.md`

### DIFF
--- a/libs/base/System.idr
+++ b/libs/base/System.idr
@@ -98,6 +98,47 @@ prim__setEnv : String -> String -> Int -> PrimIO Int
          supportNode "unsetEnv"
 prim__unsetEnv : String -> PrimIO Int
 
+%foreign "C:idris2_enableRawMode, libidris2_support, idris_support.h"
+prim__enableRawMode : (1 x : %World) -> IORes Int
+
+%foreign "C:idris2_resetRawMode, libidris2_support, idris_support.h"
+prim__resetRawMode : (1 x : %World) -> IORes ()
+
+||| `enableRawMode` enables raw mode for stdin, allowing characters
+||| to be read one at a time, without buffering or echoing.
+||| If `enableRawMode` is used, the program should call `resetRawMode` before
+||| exiting. Consider using `withRawMode` instead to ensure the tty is reset.
+|||
+||| This is not supported on windows.
+export
+enableRawMode : HasIO io => io (Either FileError ())
+enableRawMode =
+  case !(primIO prim__enableRawMode) of
+    0 => pure $ Right ()
+    _ => returnError
+
+||| `resetRawMode` resets stdin raw mode to original state if
+||| `enableRawMode` had been previously called.
+export
+resetRawMode : HasIO io => io ()
+resetRawMode = primIO prim__resetRawMode
+
+||| `withRawMode` performs a given operation after setting stdin to raw mode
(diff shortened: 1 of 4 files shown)


Z-snails:
It could be useful to return a bool indicating if it succeeded (ie if it isn't on windows)?

dunhamsteve:
That seems prudent.  `tcsetattr` can return an error which we could also report.  One of the possibilities is `ENOTTY` (not a terminal), in which case you can't set raw mode, but you're getting character by character anyway.   But I suppose that should count as an error.  On OSX:

```
     Upon successful completion, the functions tcgetattr() and tcsetattr() return a value of 0.  Otherwise, they return -1 and the global variable errno is set to indicate the error, as follows:

     [EBADF]            The fildes argument to tcgetattr() or tcsetattr() was not a valid file descriptor.

     [EINTR]            The tcsetattr() function was interrupted by a signal.

     [EINVAL]           The optional_actions argument to the tcsetattr() function was not valid, or an attempt was made to change an attribute represented in the termios structure to an
                        unsupported value.

     [ENOTTY]           The file associated with the fildes argument to tcgetattr() or tcsetattr() is not a terminal.
```

libs/base/System.idr
```diff
@@ -98,6 +98,22 @@ prim__setEnv : String -> String -> Int -> PrimIO Int
          supportNode "unsetEnv"
 prim__unsetEnv : String -> PrimIO Int
 
+%foreign "C:idris2_enableRawMode, libidris2_support, idris_support.h"
+prim__enableRawMode : (1 x : %World) -> IORes ()
+
+%foreign "C:idris2_disableRawMode, libidris2_support, idris_support.h"
+prim__disableRawMode : (1 x : %World) -> IORes ()
+
+||| Enable raw mode for stdin (read unbuffered with no echo)
```
mattpolzin, line 107: Regardless of errors or no-Ops, I suggest we add the lack of windows support to the code docs.

gallais:
As I said on discord:

Is setvbuf not the portable solution? https://linux.die.net/man/3/setvbuf says it's C99 (C89 even)

dunhamsteve:
One other note - the sample that I cribbed this from was storing the original `termios` value and resetting it on exit via an `atexit()` hook. To keep things simple and because I'm not sure how kosher it is to use `atexit`, I didn't do that, but I did provide a means to manually restore the cooked mode.

Z-snails:
> One other note - the sample that I cribbed this from was storing the original `termios` value and resetting it on exit via an `atexit()` hook. To keep things simple and because I'm not sure how kosher it is to use `atexit`, I didn't do that, but I did provide a means to manually restore the cooked mode.

It is very inconvenient for the end user if a program doesn't set the terminal back to cooked mode, so maybe exposing a function that does cleanup for you, rather than the primitives would be good
eg
```idris
withRawMode : (Either Error () -> IO a) -> IO a
```

Using the c `atexit` function sounds like a bad idea, as the scheme runtime may have cleaned up after itself before that is run.

PR #2918, opened 2023-03-10
[ new ] Add generalisations of `min` and `max` for `StronglyConnex`

buzden:
# Description

We have a generalised order in `Control.Order`, but do not have appropriate generalised `min` and `max`, let's add it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,6 +159,8 @@
 * Implemented `Ord` for `Language.Reflection.TT.Name`, `Language.Reflection.TT.Namespace`
   and `Language.Reflection.TT.UserName`.
 
+* Adds `leftmost` and `rightmost` to `Control.Order`, a generalisation of `min` and `max`.
+
 #### System
 
 * Changes `getNProcessors` to return the number of online processors rather than
@@ -170,9 +172,9 @@
 
 * Adds `Data.List.Sufficient`, a small library defining a structurally inductive view of lists.
 
-* Remove Data.List.HasLength from contrib library but add it to the base library
+* Remove `Data.List.HasLength` from `contrib` library but add it to the `base` library
   with the type signature from the compiler codebase and some of the naming
-  from the contrib library. The type ended up being `HasLength n xs` rather than
+  from the `contrib` library. The type ended up being `HasLength n xs` rather than
   `HasLength xs n`.
 
 * Adds an implementation for `Functor Text.Lexer.Tokenizer.Tokenizer`.
--- a/libs/base/Control/Order.idr
+++ b/libs/base/Control/Order.idr
@@ -40,3 +40,75 @@ interface (PartialOrder ty rel, Connex ty rel) => LinearOrder ty rel where
 ||| Every equivalence relation is a preorder.
 public export
 [EP] Equivalence ty rel => Preorder ty rel where
+
+--- Derivaties of order-based stuff ---
+
+||| Gives the leftmost of a strongly connex relation among the given two elements, generalisation of `min`.
+|||
+||| That is, leftmost x y ~ x and leftmost x y ~ y, and `leftmost x y` is either `x` or `y`
+public export
+leftmost : (0 rel : _) -> StronglyConnex ty rel => ty -> ty -> ty
+leftmost rel x y = either (const x) (const y) $ order {rel} x y
+
+||| Gives the rightmost of a strongly connex relation among the given two elements, generalisation of `max`.
+|||
+||| That is, x ~ rightmost x y and y ~ rightmost x y, and `rightmost x y` is either `x` or `y`
+public export
+rightmost : (0 rel : _) -> StronglyConnex ty rel => ty -> ty -> ty
+rightmost rel x y = either (const y) (const x) $ order {rel} x y
+
+-- properties --
+
+export
+leftmostRelL : (0 rel : _) -> Reflexive ty rel => StronglyConnex ty rel => (x, y : ty) -> leftmost rel x y `rel` x
+leftmostRelL rel x y with (order {rel} x y)
+  _ | Left  _  = reflexive {rel}
+  _ | Right yx = yx
+
+export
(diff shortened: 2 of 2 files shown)


andrevidela:
Not sure what to do about this, looks nice but given the [contributing guidelines](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md#things-that-should-be-discussed-via-the-issue-tracker-first) I'm not quite sure this passes the threshold to be in `base`:
- it's not been discussed on the issue tracker first.
- There is no motivating use-case.

Can you provide some examples of use? If there isn't anything obvious it might be better to have this in `contrib` instead

buzden:
> If there isn't anything obvious it might be better to have this in contrib instead

I disagree about `contrib` since these notions are literally as basic as `min` and `max` for `Ord`, but for its typelevel-compatible nicely behaved algebraic brother from `Control.Order`. This is neither experimental nor controversial, I think. This is almost the first thing I needed when I used stuff from `Control.Order` to express various properties in a provable way.

> Can you provide some examples of use?

Consider a type indexed with an order, which is used to prove properties. Many operations do precisely weakening or strengthening this order. For example, I had a type `X (relation : r -> r -> Type) r a` with applicative-like operations `(<$>) : X rel m1 (a -> b) -> X rel m2 a -> X rel m3 b` where I needed to express that the resulting metric `m3` is actually a minimum of `m1` and `m2` according to the relation `rel`. I can't use `min` or `smaller` since relation `rel` can be, say, `GT` over `Nat` or something. So, I chose `leftmost` for preferring leftmost argument of `rel`, the same for `rightmost`.

@andrevidela is it now clearer, and do you still think this PR violates contribution guidelines?

andrevidela:
That answers why you would want `leftmost` or `rightmost` yes. However

- That does not motivate all the properties after line 60
- I'm happy with the scenario you gave, but I was expecting a link to a library or a project that makes use of this. Are you working on something that relies on this? Maybe even a tutorial of some sort?

Sorry to be a pest about this, but the guidelines are useless if we don't enforce them. And talking about additions to `base` on the issue tracker is the least we can do.

buzden:
> * I'm happy with the scenario you gave, but I was expecting a link to a library or a project that makes use of this. Are you working on something that relies on this? Maybe even a tutorial of some sort?

This scenario indeed raised in an experimental redesigning branch of DepTyCheck, where generators [are indexed](https://github.com/buzden/deptycheck/blob/50b5221e5de1400e0e421f46fa5e609e581539b3/src/Test/DepTyCheck/Gen.idr#L55-L69) by a data type which [happens](https://github.com/buzden/deptycheck/blob/50b5221e5de1400e0e421f46fa5e609e581539b3/src/Test/DepTyCheck/Gen/Emptiness.idr#L99-L108) to be a `StronglyConnex`. Generators implement various combinators which precise rules need to be expressed with those functions that I propose to be added to `base`. However, unfortunately, at the moment I removed those precise signatures in favor of simpler ones, which impose manual conversions from users, this is due to a hurry in delivery of this experimental branch to the users, so generally, I think the reasons are independent of this change anyway.

> * That does not motivate all the properties after line 60

You run into them immediately when you use your `StronglyConnex` relation of intermediate results, like the constructor parameter [here](https://github.com/buzden/deptycheck/blob/50b5221e5de1400e0e421f46fa5e609e581539b3/src/Test/DepTyCheck/Gen.idr#L63). They allow to build the relation value after you used `leftmost` and `rightmost` functions, and these values are expressed in terms of original arguments.

> Sorry to be a pest about this, but the guidelines are useless if we don't enforce them. And talking about additions to `base` on the issue tracker is the least we can do.

Well, :-(

Anyway, this is the meta-discussion unrelated to this particular PR, maybe there is a better place to discuss this. What I can say is that I personally try to filter myself in suggestions to the standard libraries a lot. And that's why, I've created a bunch of libraries, some of which could fit reasonably in `contrib` or even `base`, like [this one](https://github.com/buzden/idris2-if-unsolved-implicit), or [this](https://github.com/buzden/idris2-mtl-tuple-impls), or even [this](https://github.com/buzden/idris2-typelevel-emptiness-collections). But in those cases I haven't found arguments for creating PR that would be strong enough so that I was sure those PRs would be accepted. I thought that this PR was an obvious addition of basically a forgotten thing in the original design of `Control.Order` module. Now I see I was wrong.

Constructively, I would suggest the following. At the moment we have very good templates for language problem issues. It defines very important steps that are hard to be ignored: particular steps of reproduction, expected and observed behaviours. This defines a quality standard for a bug report. If you impose rules so much, maybe, we should do something similar to pull requests to `base`. Say, having separate templates for *PRs of issue fixes*, for *enhancements in the compiler*, for *changes in `prelude` or `base`* and for *changes in other libraries*. In these templates we can list requirements for the PR to be considered, say, motivation section or something, like the expected and observed behaviour sections in the language problem issue template.

Having said that, I personally think that Idris is still a too small language for a need of rules to be imposed very strictly. But if it generally disagreed by the official maintainers, I surely surrender, and, say, I will have no complaints if this PR is closed because of rules.

andrevidela:
Thank you for your time, I can see you spent a good amount of time on this and I appreciate the effort.

You've provided motivating examples and answered all the questions I had so I have no qualms about merging this.

About the meta-questions, I'm sure that you see that the problem isn't in your ability or judgement, but simply, as most human interaction goes, communication: What is obvious to you isn't obvious to everyone else.
Furthermore, the reason the guidelines have been introduced has not gone away: To limit the load in maintenance and reviewing, we ask contributors to discuss changes on the issue tracker first. Changing the the templates for PRs doesn't reduce the load in reviewing. Overall, the goal of the project is to see the language improve and for everyone to have a good time doing it. I'm sure you will agree that getting a PR stuck for 3 months because nobody knows what to do with it isn't an ideal outcome, the guidelines are aimed to minimize that effect.

andrevidela (approved):

gallais:
> This is neither experimental nor controversial, I think.

I never merged it because I don't like the names (but also could not think of anything to suggest).

Issue #1042, opened 2021-02-07
let-in in do notation fails to parse


mb64:
```idris
foo : IO ()
foo = do
  let x = 5 in printLn x
```
#### Steps to Reproduce

Use `let ... in ...` syntax in `do`-notation.

#### Expected Behavior

It should parse fine, and define a local variable for that statement in do-notation.

#### Observed Behavior
```
Error: Parse error at line 3:13:
Expected end of input (next tokens: [in, identifier printLn, identifier x, end of input])
```

buzden:
> It should parse fine, and define a local variable for that statement in do-notation.

It shouldn't since `do`-notation uses the syntax without `in`: https://idris2.readthedocs.io/en/latest/tutorial/interfaces.html?highlight=let#monads-and-do-notation

This typechecks fine as it should:
```idris
foo : IO ()
foo = do
  let x = 5
  printLn x
```

mb64:
Could you elaborate on why let-in should be disallowed in do-notation?  It seems somewhat arbitrary to me.  One [of the options](https://idris.readthedocs.io/en/latest/reference/syntax-reference.html#grammar-token-doblock) for a statement in a do-block is an expression, and `let x = 5 in printLn x` is a valid expression.

Your example is not the same -- it defines `x` as a local variable for the rest of the `do` block, not for that single expression.  For instance, this should print `5\n10\n`, since the second `x` temporarily shadows the first:
```idr
foo : IO ()
foo x = do
  let x = 10
  let x = 5 in printLn x
  printLn x
```
With your suggestion, `do { let x = 10 ; let x = 5 ; printLn x ; printLn x }` prints 5 twice.

Here's some things that work, and that print `10\n5\n`:

```hs
-- in Haskell
foo = do
  let x = 10
  let x = 5 in print x
  print x
```

```idr
-- in both Idris 1 and Idris 2
foo : IO ()
foo = do
  let x = 10
  (let x = 5 in printLn x)
  printLn x
```

gallais:
Is this what you are looking for?

```idris
foo : IO ()                                                                                                                     
foo = do
  let x = 10
  do let x = 5
     printLn x
  printLn x
```

buzden:
> Your example is not the same -- it defines `x` as a local variable for the rest of the `do` block, not for that single expression. For instance, this should print `5\n10\n`, since the second `x` temporarily shadows the first:

[...]

PR #2926, opened 2023-03-20
Add let-in syntax in do-notation

### LINKED ISSUE #2923: let ; in do notation fails to parse
# Steps to Reproduce

```
main: IO ()
main = do
  let x = 123; print x
```

This can be parsed well in idris1, but not in idris2.
And [idris2 document](https://idris2.readthedocs.io/en/latest/tutorial/interfaces.html#monads-and-do-notation) say it should be a valid expression.

# Expected Behavior
Compile success.

# Observed Behavior
Compile error.

Alex1005a:
I originally tried to fix #2923, but it seems to be easier to add the let-in syntax in do-notation requested in #1042. This partly solves the problem.

      `CONTRIBUTORS.md`).


gallais:
You can already write
```idris
main : IO ()
main = do
  x <- pure 10
  (let x = 5 in printLn x)
  printLn x
```
which IMO is less ambiguous.

Alex1005a:
This is actually more obvious. But it's probably not always obvious why to wrap let-in in brackets, especially for those who used let-in in do block in haskell.
And from the beginning, I was just trying to fix the situation when you can’t use let in a do block when its body is written in one line.
For example, it won't compile.
```idris
invalidDoBlock : IO ()
invalidDoBlock = do
  { let x = 123 ; y <- pure 123 ; printLn (x + y) }
```
And it works.
```idris
validDoBlock : IO ()
validDoBlock = do
  let x = 123 in y <- pure 123 ; printLn (x + y)
```
However, it may not be worth overloading the language with unnecessary constructions because of such minor problems. I don't know.

andrevidela:
It seems this has gone stale. I'm not particularly convinced that `let ... in ...` in do blocks expresses the intent as well as the parenthesized expression. Here is what I suggest:
- let's not have the feature
- reuse the parser to customise the error message when we detect "let .. in" with a message like "did you mean `(let ... in ...)`?"
- move the test into a existing category like `perror`
- close #1042

Alex1005a:
I think it would be right to do so.

andrevidela (approved):
LGTM

PR #2933, opened 2023-03-27
Add HasNamespaces Lifted* implementations

Adowrath:
# Description

This adds relevant `HasNamespaces` instances for the Lifted IR - structurally it's almost identical to the Named/Cases stage, minus Force/Delay/Lambdas and plus a split on how function/closure applications are represented.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,9 @@
 * Constructors mentioned on the left hand side of functions/case alternatives
   are now included in the `Refers to (runtime)` section of functions debug info.
 
+* The Lifted IR Representation now has a `HasNamespaces` implementation
+  in `Compiler.Separate` so Compilation Units at that stage can be generated.
+
 ### Library changes
 
 #### Prelude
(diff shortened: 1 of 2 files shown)


Adowrath:
Back in v0.4.0 when the `chez-sep` backend was introduced, to facilitate/support the output of multiple separate output files (https://github.com/idris-lang/Idris2/blob/4de7b2133ac4a21a951d317ff9afc78fa03a1e52/src/Compiler/Separate.idr)[Compiler.Separate] was added whose `getCompilationUnits` function traverses all passed definitions to find referred-to namespaces within, and then creates one Compilation Unit for each Strongly Connected Component of that Namespace-dependency graph.
At the time, only definitions for finding the namespaces mentioned within `NamedDef` and `VMDef` definitions were added, meaning only for those two stages could the CUs be extracted - this further adds `LiftedDef` so only `ANFDef` is missing (I could add those as well if it would round out the PR more).
Re: Message in Changelog, How about "`Compiler.Separate.getCompilationUnits` can now split definitions for the `Lifted` phase by a new `HasNamespaces` implementation as well", or something along those lines?

gallais:
That sounds good!

PR #2934, opened 2023-03-28
[ base ] Add extraction functions to `Data.Singleton`

buzden:
Without them it's not convenient to use `Singleton` in tacit expressions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,8 @@
 
 * Add `withRawMode`, `enableRawMode`, `resetRawMode` for character at a time input on stdin.
 
+* Adds extraction functions to `Data.Singleton`.
+
 #### System
 
 * Changes `getNProcessors` to return the number of online processors rather than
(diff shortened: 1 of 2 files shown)


gallais:
I don't think it's worth having 2 functions with slightly different names.

mattpolzin:
I’m not usually focused on this kind of thing, but this incidentally brings the type `a` into runtime since it doesn’t give it 0 quantity right?

buzden:
> I’m not usually focused on this kind of thing, but this incidentally brings the type `a` into runtime since it doesn’t give it 0 quantity right?

I don't understand what you are saying. Both `Val` and `unVal` have their `a` argument of quantity `0`. E.g., run
```
:module Data.Singleton
:ti Val
```
in REPL and you get
```idris
Data.Singleton.Val : {0 a : Type} -> (x : a) -> Singleton {a} x
```

mattpolzin:
Thanks. Bit early in the morning for me perhaps. I was reading the type of `unVal` as if it was semantically `{ a : _ } -> { x : a } -> Singleton {a} x -> a` but I see the difference now!

PR #2937, opened 2023-03-31
[ new ] %unsafe pragma for escape hatches

gallais:
# Description

      `CONTRIBUTORS.md`).


stefan-hoeck:
> No this is purely declarative. My motivation as a library writer is to have something like:
> 
> ```idris
> data Singleton : {0 a : Type} -> (x : a) -> Type where
>   MkSingleton : (x : a) -> Singleton x
> 
> %unsafe
> unsafeMkSingleton : (y : a) -> Singleton {a} x
> unsafeMkSingleton y = believe_me (MkSingleton y)
> ```
> 
> in case you cannot prove that `x` is equal to `y` but know it is the case.
> 
> If you were to use `believe_me (MkSingleton y)` directly then you wouldn't get a warning if the return type changed from e.g. `Singleton x` to `(Nat, Singleton x)`. With `unsafeMkSingleton` you'll instead get a type error.

OK, but the behavior of `unsafeMkSingleton` in your example would be exactly the same without the `%unsafe` pragma, yes?

stefan-hoeck:
Ah, that makes sense now. Thanks a lot for your explanations.

PR #2939, opened 2023-04-04
[ fix #2927 ] Do not eagerly evaluate expressions occurring inside `%delay` block

### LINKED ISSUE #2927: CSE optimisation is tough with `%inline` and lazy meeting in the same place
# Steps to Reproduce

Run the following code:

```idris
import Data.List.Lazy

import Debug.Trace

%default total

%inline
(::) : Maybe a -> Lazy (LazyList a) -> LazyList a
Nothing :: xs = xs
Just x  :: xs = x :: xs

fufu : a -> Maybe a
fufu = Just

g : LazyList Nat
g = trace "--- outmost ---"
  [ fufu $ trace "pure 6" 6
  , fufu $ trace "pure 5" 5
  ]

main : IO Unit
main = printLn g
```

# Expected Behavior

Prints

```
pure 6
--- outmost ---
pure 5
[6, 5]
```

# Observed Behavior

```
pure 5
pure 6
--- outmost ---
[6, 5]
```

i.e. the value in the tail of the lazy list computes eagerly.

# Analysis

Thanks to @AlgebraicWolf who realised that the reason for this is CSE, which puts internals of the lazy value (tail of the list) to a top-level declaration, and thus it is computed too early.

We've realised that this behaviour goes away
- if you remove `%inline` of `::`, or
- if function `::` does not pattern match (say, if we use `Identity` instead of `Maybe`), or
- if you inline manually function `fufu`.

So, this is the minimal case we could find reproducing the problem.

Pinging @stefan-hoeck as the original author of the CSE optimisation. Do you understand what is going on in this case?

### LINKED ISSUE #2928: Unification of linear and unrestricted function
# Steps to Reproduce
```idris
idLinear : (1 _ : a) -> a
idLinear x = x

id : a -> a
id = idLinear
```
# Expected Behavior
Given that this block of code works
```idris
idLinear : (1 _ : a) -> a
idLinear x = x

id : a -> a
id x = idLinear x
```
it seemed to me that you can omit the argument in lhs.
# Observed Behavior
```
Error: While processing right hand side of idd. When unifying:
    ?a -> ?a
and:
    (1 _ : a) -> a
Mismatch between: ?a -> ?a and (1 _ : a) -> a.
```

AlgebraicWolf:
# Description

This PR fixes #2927.

Thanks to @stefan-hoeck for his assistance in Discord. He suggested fixing the issue by using Scheme's `delay` and `force` to memoize constants, as in #2807 instead of #2817 that is currently used. However, this solution comes with a performance overhead.

To reduce the overhead, I modified CSE to track whether an expression occurs inside a `%delay` block. If that is the case, the expression is lifted as memoized lazy value using Scheme's `delay` and `force`. Otherwise, it becomes an eagerly evaluated top-level constant. Benchmarking the bootstrapped compiler on `idris2api.ipkg` showed 5% slowdown on my machine, which is an improvement compared to the 8.5% slowdown in #2807.  

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -88,6 +88,18 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 * Switch calling conventions based on the number of arguments to avoid limits on
   the number of arguments and to reduce stack usage.
 
+#### Chez
+
+* Fixed CSE soundness bug that caused delayed expressions to sometimes be eagerly
+  evaluated. Now when a delayed expression is lifted by CSE, it is compiled
+  using Scheme's `delay` and `force` to memoize them.
+
+#### Racket
+
+* Fixed CSE soundness bug that caused delayed expressions to sometimes be eagerly
+  evaluated. Now when a delayed expression is lifted by CSE, it is compiled
+  using Scheme's `delay` and `force` to memoize them.
+
 #### NodeJS Backend
 
 * The NodeJS executable output to `build/exec/` now has its executable bit set.
--- a/src/Compiler/Opts/CSE.idr
+++ b/src/Compiler/Opts/CSE.idr
@@ -44,10 +44,11 @@ import Libraries.Data.SortedMap
 
 ||| Maping from a pairing of closed terms together with
 ||| their size (for efficiency) to the number of
-||| occurences in toplevel definitions.
+||| occurences in toplevel definitions and flag for
+||| whether it was encountered in delayed subexpression.
 public export
 UsageMap : Type
-UsageMap = SortedMap (Integer, CExp []) (Name, Integer)
+UsageMap = SortedMap (Integer, CExp []) (Name, Integer, Bool)
 
 ||| Number of appearances of a closed expression.
 |||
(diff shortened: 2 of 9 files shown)


CodingCellist:
Is this related to #2791 ?  : )

AlgebraicWolf:
> Is this related to #2791 ? : )

No, not really. Moreover, this PR uses proper memoization instead of relying on #2791, as it yields better performance in this specific case

I guess the only way they are related is that both the need for #2791 and discovery of #2927 happened because of an attempt to manipulate large lazily-constructed data structures

gallais:
This makes sense to me however it would be nice to know what the cause
of the slowdown is. I don't think we're using many top level lazy constants
in the compiler.

AlgebraicWolf:
I have rebased this PR, all the tests still do pass. Perhaps it's worth re-measuring compilation time as well

mjustus:
Thanks @stefan-hoeck! I am merging this now and am happy to take the blame if something goes wrong.

andrevidela:
Software's not about blame! Don't feel pressure to mess up something.

PR #2942, opened 2023-04-07
[ perf ] Make Core.Normalise.Eval.evalDef stack safety

Alex1005a:
It seems that the original version of evalDef was not optimized due to the call to evalTree, in which it fully evaluated the case tree, including calling evalDef in itself:
https://github.com/idris-lang/Idris2/blob/3f7cd84ccb38df7d6ccc02b9e6a4ff425640b814/src/Core/Normalise/Eval.idr#L510
Because of this, evalDef was not a tail recursion.
This can be solved without doing a full evaluation, but by returning the term from the matched case, and only then evaluate.

Thanks to this, the call to the repl (Normalise everything mode) `facTail 100000`
```idris
facAux : Integer -> Integer -> Integer
facAux 0 r = r
facAux n r = facAux (n - 1) (r * n)

facTail : Integer -> Integer
facTail n = facAux n 1
```
almost does not consume memory, whereas before, the process started to take up a lot of memory.

### DIFF
--- a/src/Core/Normalise/Eval.idr
+++ b/src/Core/Normalise/Eval.idr
@@ -83,6 +83,12 @@ data CaseResult a
      | NoMatch -- case alternative didn't match anything
      | GotStuck -- alternative matched, but got stuck later
 
+record TermWithEnv (free : List Name) where
+    constructor MkTermEnv
+    varsEnv : List Name
+    locEnv : LocalEnv free varsEnv
+    term : Term $ varsEnv ++ free
+
 parameters (defs : Defs, topopts : EvalOpts)
   mutual
     eval : {auto c : Ref Ctxt Defs} ->
@@ -316,7 +322,7 @@ parameters (defs : Defs, topopts : EvalOpts)
     getCaseBound (arg :: args) []        loc = Nothing -- mismatched arg length
     getCaseBound (arg :: args) (n :: ns) loc = (arg ::) <$> getCaseBound args ns loc
 
-    evalConAlt : {auto c : Ref Ctxt Defs} ->
+    conAltTerm : {auto c : Ref Ctxt Defs} ->
                  {more, free : _} ->
                  Env Term free ->
                  LocalEnv free more -> EvalOpts -> FC ->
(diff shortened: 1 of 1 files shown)


CodingCellist (changes requested):
I'm not super sure about the renaming of the functions though. The old `evalXXXX` names seemed to clearly indicate what they did, so I'm unsure what the new names add. Perhaps reverting the names and adding some comments would be more appropriate?

Alex1005a:
Thanks for the review.
> The old `evalXXXX` names seemed to clearly indicate what they did, so I'm unsure what the new names add.

You're right. I just thought that a function that returns a `Term` and not `NF` is called `evalXXXX`. On the other hand, eval still happens in the function itself, so it would be correct to return the name and write a comment.

src/Core/Normalise/Eval.idr
```diff
@@ -394,26 +401,26 @@ parameters (defs : Defs, topopts : EvalOpts)
...
               LocalEnv free args -> EvalOpts -> FC ->
               Stack free -> NF free -> List (CaseAlt args) ->
-              Core (CaseResult (NF free))
+              Core (CaseResult (TermWithEnv free))
     findAlt env loc opts fc stk val [] = do
       log "eval.casetree.stuck" 2 "Ran out of alternatives"
       pure GotStuck
     findAlt env loc opts fc stk val (x :: xs)
-         = do Result val <- tryAlt env loc opts fc stk val x
+         = do (Result (MkTermEnv newLoc val)) <- tryAlt env loc opts fc stk val x
```
gallais, line 409: If you're not using the values and only reconstructing the closure in `pure` then I would suggest using `res@(Result {})` instead and then `pure res` later. Saves us from checking we are indeed returning the same thing.

gallais (commented):
Looks good to me. You will want to see how/whether to port this to github.com/edwinb/yaffle too

src/Core/Normalise/Eval.idr
Alex1005a: Thanks for the review. I made this mistake because I thought val was used (in fact, it can't be used in GotStuck branch).

Alex1005a:
> You will want to see how/whether to port this to github.com/edwinb/yaffle too

It seems there is no need to port this to yaffle since the case eval has changed a lot.

CodingCellist (approved):
LGTM! Thanks for the quick responses  : )

PR #2944, opened 2023-04-12
[ new ] Add support for bi-directional pipes on POSIX systems (resolves #2935)

### LINKED ISSUE #2935: Enhancing Idris' support for process management with BiDi pipes.
Currently the only means to have bi-directional IPC in Idris is to use the Networking Library and TCP/UNIX Domain sockets. The File API supports uni-directional pipes only.

It would be great to see basic exposure of UNIX pipes (`man pipe`), or a more comprehensive overlay _a la_ Python's subprocess.

dunhamsteve:
# Description

Adds a `popen2` function to `System.File.Process` which allows Idris to run subprocesses with access to both the input and output file streams.  It currently doesn't support windows, but updates are welcome if it's possible and somebody knows how to do it (and can verify that it works). This feature was requested by @jfdm in #2935, and I think @ohad was asking about it last year.

With this change you can do:
```idris
import System.File

main : IO ()
main = do
    Right (inFile, outFile) <- popen2 "idris2"
        | Left err => printLn err
    _ <- fPutStrLn outFile ":h\n"
    closeFile outFile
    Right result <- fRead inFile
        | Left err => printLn err
    putStrLn result
    putStrLn "done"
```

But I'd recommend writing be done on a separate thread from reading to avoid deadlocking if the subprocess is waiting for a reader while we're waiting to write.

Let me know if any changes are needed.

Yes, I added a note to the CHANGELOG.

### DIFF
--- a/libs/base/System/File/Process.idr
+++ b/libs/base/System/File/Process.idr
@@ -3,6 +3,7 @@ module System.File.Process
 import public System.Escape
 import public System.File.Error
 import public System.File.Mode
+import System.FFI
 import System.File.Support
 import public System.File.Types
 
@@ -14,6 +15,20 @@ prim__popen : String -> String -> PrimIO FilePtr
 %foreign supportC "idris2_pclose"
          supportNode "pclose"
 prim__pclose : FilePtr -> PrimIO Int
+
+data Popen2Result : Type where
+
+%foreign supportC "idris2_popen2"
+prim__popen2 : String -> PrimIO (Ptr Popen2Result)
+
+%foreign supportC "idris2_popen2ChildPid"
+prim__popen2ChildPid : Ptr Popen2Result -> Int
+
+%foreign supportC "idris2_popen2FileIn"
+prim__popen2FileIn : Ptr Popen2Result -> PrimIO FilePtr
+
+%foreign supportC "idris2_popen2FileOut"
+prim__popen2FileOut : Ptr Popen2Result -> PrimIO FilePtr
 
 ||| Force a write of all user-space buffered data for the given `File`.
 |||
@@ -53,3 +68,25 @@ namespace Escaped
 export
 pclose : HasIO io => (fh : File) -> io Int
 pclose (FHandle h) = primIO (prim__pclose h)
+
+||| Create a new bidirectional pipe by invoking the shell, which is passed the
+||| given command-string using the '-c' flag, in a new process. On success
+||| a pair of File is returned. The first File is used for reading from the
+||| process, and the second is used for writing to the process. Caller is
+||| responsible for closing them.
+|||
+||| IMPORTANT: You may deadlock if write to a process which is waiting to flush
+|||            its output.  It is recommended to read and write from separate threads.
+|||
+||| This function is not supported on windows or node at this time.
+export
+popen2 : HasIO io => (cmd : String) -> io (Either FileError (File,File))
+popen2 cmd = do
+  ptr <- primIO (prim__popen2 cmd)
+  if prim__nullPtr ptr /= 0
+    then returnError
+    else do
+      inFile <- primIO (prim__popen2FileIn ptr)
+      outFile <- primIO (prim__popen2FileOut ptr)
+      free (prim__forgetPtr ptr)
+      pure $ Right (FHandle inFile, FHandle outFile)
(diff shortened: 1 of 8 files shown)


jfdm:
Thanks for this! 

It was on my TODO list for this week, so thanks for checking this item off.

jfdm:
I have had a quick look at the code, it was more easier than I imagined. Kudos.

I do have a _minor_ quibble. 

I wonder if a small layer of encapsulation should be added. 
That is, one could:

1. add a record as the 'Right' return type for `popen2`, thus enabling differentiation between the two handles; and optionally;
2. extend the `API` for `popen` to support working with the returned record.

Then again that could be a bit OTT for such a low-level & untyped process API.

gallais:
:+1: I think it's a good idea to break the symmetry in the type.

The example was confusing to me as I understood in/out to be from the POV of the
callee but it was in fact from the POV of the caller.

libs/base/System/File/Process.idr
```diff
@@ -15,6 +16,20 @@ prim__popen : String -> String -> PrimIO FilePtr
          supportNode "pclose"
 prim__pclose : FilePtr -> PrimIO Int
 
+data Popen2Result : Type where
+
+%foreign supportC "idris2_popen2"
+prim__popen2 : String -> PrimIO (Ptr Popen2Result)
+
+%foreign supportC "idris2_popen2ChildPid"
+prim__popen2ChildPid : Ptr Popen2Result -> Int
```
mattpolzin, line 25: Did you intend to write an exposed API around this primitive?

support/c/idris_file.c
```diff
@@ -223,3 +223,79 @@ FILE *idris2_stdin() { return stdin; }
...
+    perror("execl");
+    exit(err);
+  }
+#endif
+}
+
+int idris2_popen2ChildPid(struct child_process *ptr) {
+  if (!ptr)
+    return 0;
+  return ptr->pid;
```
mattpolzin, line 288: This struct's pid doesn't currently get set that I can see.

mattpolzin (commented):
I'm still reading over the C code but I do think we should add a test that touches this new code prior to merging as well.

support/c/idris_file.c
```diff
@@ -223,3 +223,79 @@ FILE *idris2_stdin() { return stdin; }
...
+    return NULL;
+  }
+
+  err = pipe(&pipes[2]);
+  if (err) {
+    close(pipes[0]);
+    close(pipes[1]);
+    return NULL;
+  }
+  if ((pid = fork()) > 0) {
```
mattpolzin, line 254: I might need my morning coffee before reading this code, but I think this should be <= 0. Maybe the branches of the conditional got swapped by you at some point?

dunhamsteve: Line 254 is right - the first branch is the parent process which is the one that gets told the new pid (child process gets 0).  The "error checking" on 255-258 I added later, which is unreachable. I'll shift that up above line 254.

support/c/idris_file.c
dunhamsteve: Yeah, and I forgot the HasIO on the Idris side.  I didn't want to toss the information away.  I'm shifting to a struct, per earlier comments and was going to add the pid to that, but it doesn't look like we have a `wait` primitive either, so the pid might be useless at this point in time.

dunhamsteve:
I was thinking something along the lines of:
```idris
||| Result of a popen2 command, containing the 
public export
record SubProcess where
  constructor MkSubProcess
  ||| Process id of the spawned process
  pid : Int
  ||| The input stream of the spawned process
  input : File
  ||| The output stream of the spawned process
  output : File
```
Where `input` is the handle you write to on the Idris side and `output` is what you read from.  There doesn't appear to be a `wait` primitive - should we keep the `pid` here? Do we want to add `wait` and possibly `kill` to make use of that pid? The `output` should eof when the process is done, but we might want more control over the process itself.

The `popen` interface exposed above is higher level and the waiting is done internally to the C library when the special `pclose` is called.

mattpolzin:
RE having use of the PID, even without wait and kill we already have the ability to send signals to it exposed in the bare lib.

dunhamsteve:
I switched it over to a Record, added a test case, and got it working on windows.

mattpolzin (approved):
Nice. Thanks!

PR #2947, opened 2023-04-14
Public export TTImp reflection functions

madman-bob:
This allows using these functions in elaborator scripts.

I want to use `mapMTTImp` in a script, but people could plausibly want to use any of them.


madman-bob:
I already had `import public` before this.

Digging more deeply, the problem wasn't the script itself, but using the result of a script at the type-level.

Here's a simplified example of what I was trying to achieve, without even using a script:

```haskell
import Control.Monad.State

import Language.Reflection

addName : TTImp -> State (List String) TTImp
addName v@(IVar fc (UN (Basic nm))) = do
    modify (nm ::)
    pure v
addName s = pure s

names : TTImp -> List String
names s = execState [] $ mapMTTImp addName s

checkNames : names `(x * y) = ["y", "x", "*"]
checkNames = Refl
```

Issue #2865, opened 2023-01-26
Invalid program generated: erased argument is passed to string-append, when `(mod x 0)`.


andorp:
An invalid program was generated for this snippet.

# Steps to Reproduce
Compile and run the following module:
```
main : IO ()
main = printLn (mod 10 0)
```

# Expected Behavior
No string-append error.

# Observed Behavior
```
Exception in string-append: erased is not a string
```

# Idris version
```
Idris 2, version 0.6.0-dbb385348
```

andorp:
Cases: [main.cases.txt](https://github.com/idris-lang/Idris2/files/10513879/main.cases.txt)
```
prim__sub_Integer = [{arg:0}, {arg:1}]: (-Integer [!{arg:0}, !{arg:1}])
Main.main = []: (PrimIO.io_bind [(%lam {eta:0} (Day11.Part2c.simulation [(Day11.Part2c.oneItem [(%con [record] Prelude.Num.MkNum Just 0 [(%lam {arg:5756} (%lam {arg:5759} (+Integer [!{arg:5756}, !{arg:5759}]))), (%lam {arg:5766} (%lam {arg:5769} (*Integer [!{arg:5766}, !{arg:5769}]))), (%lam {arg:5776} !{arg:5776})])]), !{eta:0}])), (%lam _ (PrimIO.io_bind [(%lam {eta:0} (Day11.Part2c.part2m1 [(Day11.Part2c.oneItem [(%con [record] Prelude.Num.MkNum Just 0 [(%lam {arg:5756} (%lam {arg:5759} (Day11.Part2c.+ [!{arg:5756}, !{arg:5759}]))), (%lam {arg:5766} (%lam {arg:5769} (Day11.Part2c.* [!{arg:5766}, !{arg:5769}]))), (%lam {arg:5776} (Day11.Part2c.fromInteger [!{arg:5776}]))])]), !{eta:0}])), (%lam {_:0} (%lam {eta:0} (%con [unit] Builtin.MkUnit Just 0 [])))]))])
Day11.Part2c.fromInteger = [{ext:0}]: (Day11.Part2c.mkModulo [!{ext:0}])
Day11.Part2c.+ = [{arg:0}, {arg:1}]: (%let {eff:0} !{arg:0} (%let {eff:1} !{arg:1} 0))
Day11.Part2c.* = [{arg:0}, {arg:1}]: (%let {eff:0} !{arg:0} (%let {eff:1} !{arg:1} 0))
[...]

PR #2952, opened 2023-04-22
[ fix ] codegen issue when using partial case statements in prelude.

### LINKED ISSUE #2950: Bad code generation for Prelude.Num.mod
Idris sometimes generates `((Builtin-idris_crash 'erased) "...")` instead of `(Builtin-idris_crash "...")` for missing case branches. 

This issue was reported by @CodingCellist on discord. He got an `Exception in string-append: erased is not a string` error and narrowed it down to the `mod` function.

# Steps to Reproduce

Compile and run:

```idris
main : IO ()
main = printLn (mod 42 (the Int 0))
```

# Expected Behavior

The error:
```
Unhandled input for Prelude.Num.case block in mod at Prelude.Num:131:3--133:40
```

# Observed Behavior

Crashes without output for me.  If I run the .ss file directly, I get:
```
% scheme case.ss
Chez Scheme Version 9.9.9-pre-release.14
Copyright 1984-2022 Cisco Systems, Inc.

Exception in string-append: erased is not a string
```

The generated scheme code looks like:
```scheme
(define PreludeC-45Num-u--mod_Integral_Int 
  (lambda (arg-0 arg-1) 
          (let ((sc0 (PreludeC-45EqOrd-u--C-61C-61_Eq_Int arg-1 (blodwen-toSignedInt 0 63)))) 
            (cond ((equal? sc0 0) (blodwen-euclidMod arg-0 arg-1)) 
                  (else ((Builtin-idris_crash 'erased) "Unhandled input for Prelude.Num.case block in mod at Prelude.Num:131:3--133:40"))))))
```
Note that `Builtin-idris_crash` is being applied to `'erased` and the result is being applied the error message.  With node codegen, we get:
```js
function Prelude_Num_mod_Integral_Int($0, $1) {
 switch(Prelude_EqOrd_x3dx3d_Eq_Int($1, Number(_truncBigInt32(0n)))) {
  case 0: return _mod($0, $1);
  default: return Builtin_idris_crash(undefined)('Unhandled input for Prelude.Num.case block in mod at Prelude.Num:131:3--133:40');
 }
}
```
I haven't been able to reproduce this without using the `Prelude.Num` code, but it does seem to occur for all of the implementations of `div` and `mod`.  If you grep for `_crash 'erased` in `idris2.ss` you will see twenty instances.

dunhamsteve:
This pull request resolves #2950 and #2865 (which are the same issue).  There is a code generation problem in `Prelude.Num.mod` that causes it to pass an erased value to `idris_crash`.  

This error appears to happen because `Builtin` is not imported into `Prelude.Num` so `Builtin.idris_crash` is not in scope, but the code in `ProcessDef` is injecting a fallback case alt that references it.  I initially decided to detect this and report an error, but @Z-snails suggested we just call `prim__crash` directly. 

This problem affected all back-ends, but I'll discuss scheme below.

Previously the generated code looked like:
```scheme
(define PreludeC-45Num-u--mod_Integral_Int 
  (lambda (arg-0 arg-1) 
          (let ((sc0 (PreludeC-45EqOrd-u--C-61C-61_Eq_Int arg-1 (blodwen-toSignedInt 0 63)))) 
            (cond ((equal? sc0 0) (blodwen-euclidMod arg-0 arg-1)) 
                  (else ((Builtin-idris_crash 'erased) "Unhandled input for Prelude.Num.case block in mod at Prelude.Num:131:3--133:40"))))))
```
and, if you evaluated `mod 10 0` it would give the error:
```
Exception in string-append: erased is not a string
```

Now the generated scheme looks like
```scheme
(define PreludeC-45Num-u--mod_Integral_Int 
  (lambda (arg-0 arg-1) 
          (let ((sc0 (PreludeC-45EqOrd-u--C-61C-61_Eq_Int arg-1 (blodwen-toSignedInt 0 63)))) 
            (cond ((equal? sc0 0) (blodwen-euclidMod arg-0 arg-1)) 
                  (else (blodwen-error-quit (string-append "ERROR: " "Unhandled input for Prelude.Num.case block in mod at Prelude.Num:131:3--133:40")))))))
```
and the reported error is:
```
ERROR:  Unhandled input for Prelude.Num.case block in mod at Prelude.Num:131:3--133:40
```

### DIFF
--- a/src/TTImp/ProcessDef.idr
+++ b/src/TTImp/ProcessDef.idr
@@ -843,7 +843,7 @@ mkRunTime fc n
 
     mkCrash : {vars : _} -> String -> Term vars
     mkCrash msg
-       = apply fc (Ref fc Func (NS builtinNS (UN $ Basic "idris_crash")))
+       = apply fc (Ref fc Func (UN $ Basic "prim__crash"))
                [Erased fc Placeholder, PrimVal fc (Str msg)]
 
     matchAny : Term vars -> Term vars
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -89,7 +89,7 @@ idrisTestsError = MkTestPool "Error messages" [] Nothing
        "error011", "error012", "error013", "error014", "error015",
        "error016", "error017", "error018", "error019", "error020",
        "error021", "error022", "error023", "error024", "error025",
-       "error026",
+       "error026", "error027",
        -- Parse errors
        "perror001", "perror002", "perror003", "perror004", "perror005",
        "perror006", "perror007", "perror008", "perror009", "perror010",
--- a/tests/idris2/error027/Issue2950.idr
+++ b/tests/idris2/error027/Issue2950.idr
@@ -1,0 +1,2 @@
+main : IO ()
+main = printLn $ mod 10 0
--- a/tests/idris2/error027/expected
+++ b/tests/idris2/error027/expected
@@ -1,0 +1,1 @@
+ERROR: Unhandled input for Prelude.Num.case block in mod at Prelude.Num:94:3--96:44
--- a/tests/idris2/error027/run
+++ b/tests/idris2/error027/run
@@ -1,0 +1,3 @@
+rm -rf build
+
+$1 --no-color --console-width 0 --exec main Issue2950.idr


gallais:
Can we please add a test case so that it's not reverted when moving to yaffle?

ysangkok:
Thank you!

PR #2953, opened 2023-04-22
[ totality ] divergent function accepted as total

dunhamsteve:
The following function is accepted by Idris as total, but it diverges if the first argument is non-empty:
```idris
total
foo : List Char -> List Char -> ()
foo (c :: cs) _ = foo (c :: cs) cs
foo _ _ = ()
```
(This is a reduction of an issue reported by Monadic Cat on Discord.)

The size change info for the function is:
```
Size change: Main.foo: [Just (0, Same), Just (0, Smaller)]
```
The totality checker is correctly assessing that the second argument is a smaller copy of the first, but I believe we need an argument to be smaller than itself. That is the number in the tuple needs to be the same as the index in the array. 

I updated the code based on this assumption, and it resolved this issue and still passes all of the existing test cases.

### DIFF
--- a/src/Core/Termination/SizeChange.idr
+++ b/src/Core/Termination/SizeChange.idr
@@ -214,6 +214,16 @@ prefixPath pred g = go g []
           else
             go f ((l, g) :: path)
 
+-- returns `Just a.path` iff there is no self-decreasing edge,
+-- i.e. there is no argument `n` with `index n a.change === (n, Smaller)`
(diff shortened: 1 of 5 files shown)


mjustus (approved):
Good catch! Both diagnosis and fix are correct. Here's the corresponding bit from the paper:

termination

I took the liberty of adding two comments and renaming the confusingly named `checkDesc`.

PR #2960, opened 2023-05-05
Add symmetric and transitive closure relations

madman-bob:
Add `SymClosure` and `TransClosure` relations, which take the symmetric and transitive closures of a given relation.

### DIFF
--- a/libs/base/Control/Relation.idr
+++ b/libs/base/Control/Relation.idr
@@ -127,3 +127,59 @@ PartialEquivalence ty Equal where
 
 public export
 Equivalence ty Equal where
+
+----------------------------------------
+
+public export
+data SymClosure : Rel ty -> ty -> ty -> Type where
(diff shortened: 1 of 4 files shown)


libs/base/Control/Relation.idr
```diff
@@ -127,3 +127,59 @@ PartialEquivalence ty Equal where
...
+Symmetric ty (SymClosure rel) where
+  symmetric (Fwd xy) = Bwd xy
+  symmetric (Bwd yx) = Fwd yx
+
+----------------------------------------
+
+public export
+data TransClosure : Rel ty -> ty -> ty -> Type where
+  Nil : TransClosure rel x x
+  (::) : {x, y, z : ty} -> rel x y -> TransClosure rel y z -> TransClosure rel x z
```
gallais, line 152: I think you only need `y : ty`, the others can be runtime irrelevant because they're determined by the endpoints which can be made runtime-relevant if need be.

PR #2961, opened 2023-05-06
[ codegen ] get rid of trivial let statements

stefan-hoeck:
# Description

Idris sometimes introduces trivial and unnecessary let expressions which prevent the identity optimizer from performing its task. Consider the following example:

```idris
module Id

import Data.Nat

%default total

record Small where
  constructor S
  value : Nat
  {auto 0 prf : LTE value 20}

Show Small where show = show . value

record Smaller where
  constructor XS
  value : Nat
  {auto 0 prf : LTE value 10}

-- This is the identity function
toSmall : Smaller -> Small
toSmall (XS v @{prf}) = S v @{transitive prf %search}

-- This is again the identity function
manyToSmall : List Smaller -> List Small
manyToSmall [] = []
manyToSmall (x::xs) = toSmall x :: manyToSmall xs

main : IO ()
main = printLn $ manyToSmall [XS 1, XS 2, XS 3]
```

`toSmall` is the identity function as is `manyToSmall`. However, as can be seen in the generated JS code, `toSmall` is not optimized away due to a trivial `const` statement.

```javascript
/* Id.toSmall : Smaller -> Small */
function Id_toSmall($0) {
 const $1 = $0;
 return $1;
}

/* Id.manyToSmall : List Smaller -> List Small */
function Id_manyToSmall($0) {
 switch($0.h) {
  case 0: /* nil */ return {h: 0};
  case undefined: /* cons */ return {a1: Id_toSmall($0.a1), a2: Id_manyToSmall($0.a2)};
 }
}
```
Luckily, it is straight forward to get rid of these as part of constant folding, and this is the content of this PR.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@
   `TTImp.ProcessDef.genRunTime`. This allows us to track when incomplete `case`
   blocks get the runtime error added.
 
+* Constant folding of trivial let statements.
+
 ### Library changes
 
 #### Prelude
--- a/src/Compiler/Opts/ConstantFold.idr
+++ b/src/Compiler/Opts/ConstantFold.idr
@@ -69,6 +69,11 @@ lookup fc (MkVar p) rho = case go p rho of
       Z => Left (MkVar First)
       S i' => bimap later weaken (go (dropLater p) (Wk ws' rho))
 
+replace : CExp vars -> Bool
+replace (CLocal _ _)   = True
+replace (CPrimVal _ _) = True
+replace _              = False
+
 -- constant folding of primitive functions
 -- if a primitive function is applied to only constant
 -- then replace with the result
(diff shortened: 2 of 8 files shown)


gallais:
I had similar issues with functions on `Singleton` not being optimised away. I'm excited to test this!

gallais:
So this did successfully eliminate `getSingleton` and generally generates much nicer code,
however I can still see a call to `unsafeMkSingleton`

https://github.com/gallais/potpourri/blob/f7ff862bba1b0b69bccad127798066381e4d3d5e/idris/serialised/src/idris/Data/Singleton.idr#L10-L11

which is compiled to:

```
/* Data.Singleton.unsafeMkSingleton : a -> Singleton x */
function Data_Singleton_unsafeMkSingleton($0) {
 return $0;
}
```

Is this because the `believe_me` gets in the way of idris seeing it's the identity function?

stefan-hoeck:
> So this did successfully eliminate `getSingleton` and generally generates much nicer code, however I can still see a call to `unsafeMkSingleton`
> 
> https://github.com/gallais/potpourri/blob/f7ff862bba1b0b69bccad127798066381e4d3d5e/idris/serialised/src/idris/Data/Singleton.idr#L10-L11
> 
> which is compiled to:
> 
> ```
> /* Data.Singleton.unsafeMkSingleton : a -> Singleton x */
> function Data_Singleton_unsafeMkSingleton($0) {
>  return $0;
> }
> ```
> 
> Is this because the `believe_me` gets in the way of idris seeing it's the identity function?

Yes, I think so. `believe_me` is implemented via a call to `prim__believe_me`, which is a prim op and thus ignored by the identity optimizer. It would be trivial to get rid of it during constant folding as well. I can give this a go either as part of this PR or as part of a new one (I'd prefer coming up with a new PR).

gallais:
Ok let's merge this in and leave `prim_believe_me` to a separate one!
Does it need to be ported to yaffle?

PR #2964, opened 2023-05-10
[ re #2960 ] Move defs of closures data types to a separate module

buzden:
PR #2960 adds couple of data structures, one of which breaks some existing code due to introduced ambiguity.

I suggest these data structures to be in a separate module so that those who explicitly want these to use would bite their ambiguities and those who do not (while still using `Control.Relation` module) would be unaffected.

I also reverted some consequences of this ambiguity from the tests.

Pinging @madman-bob as the original author of the change. What do you think?


madman-bob:
I have no complaint about this.

I originally thought this would cause issues with Frex, due to name collisions. But I note that the Frex versions are in `Data`, not `Control`.

PR #2974, opened 2023-05-15
[ papers ] Port the first part of "Deferring the details [...]" by Liam O'Connor

CodingCellist:
# Description

This ports the proof delay part of the paper as `papers/Data/ProofDelay.idr`.

I did start an implementation of Trip, but it's quite a bit more work than I anticipated, so I'm putting that on ice for now. If anyone else thinks it would be fun to do, by all means go for it!  : )

      `CONTRIBUTORS.md`).


gallais:
`HList` is just `All id` so we should probably be using that. If some of the lemmas are missing then we should just add them. I'll try to take a look soonish if no one does it.

Issue #2979, opened 2023-05-17
Build Idris2 compiler for NodeJS


freddi301:
I'm trying to build the idris compiler to a js executable for nodejs.
 
 The ultimate goal is:
 - build idris2 for nodejs and publish a package on npmjs so users can install js artifact with `npm install idris2`
 - once the node build is done, use browserify to make a idris2 compiler build for the browser
 - include a idris2 web editor in the idris site eventually
 
I'm trying with this command:

 ```idris2 --codegen node --build idris2.ipkg```
 
 [related issue](https://github.com/idris-lang/Idris2/issues/1349)

freddi301:
Till now i get this error

```
Uncaught error: INTERNAL ERROR: No supported backend found in the definition of System.File.ReadWrite.prim__readChar. Supported backends: "node","javascript". Backends in definition: "C".
```

And it keeps popping up even after i added

```
%foreign "C:fgetc,libc 6"
		"node:lambda:() => { throw new Error() }"
prim__readChar : FilePtr -> PrimIO Int
```

mattpolzin:
The compilation of Idris will use currently installed libraries to build itself before then using the new build of itself to rebuild all libraries. If you want to make a change to a library to test how it will impact building the compiler, you need to explicitly rebuild just that library and then install it and then rerun a build of the compiler. 

While admirable, you'll likely hit a number of issues (some show-stopping, if my vague recollection is correct) with building the compiler with the Node backend. To date, no one has done the work to build the compiler with any non-scheme backends, though at one point RefC was close.

Here's a likely partial list of missing Node FFI functions: https://github.com/idris-lang/Idris2/issues/2207

You may also run into the lacking support for a scheme interpreter, but I don't recall if that interpreter support ever got used by the compiler code itself.

dunhamsteve:
I suspect stack or memory size might be an issue, too.  But I've entertained similar ideas, and it wouldn't hurt to at least catalog where the process gets stuck.

I have gotten TinyIdris and  an earlier version of Yaffle working in the browser by compiling for node and stubbing out the node api. But they couldn't do interaction in the browser, because of the async thing.    (I do wonder if we could compile IO to async functions, but I haven't explored the idea.)

The other thing that might be worth exploring is writing a WASM back end. It's only my long list of ideas.

freddi301:
[...]

Issue #1349, opened 2021-04-28
JS codegen uses ridiculous amounts of memory when compiling the compiler


Gaelan:
If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!

# Steps to Reproduce

1. Build and install Idris normally from [this branch](https://github.com/Gaelan/Idris2/tree/self-hosting-node), which implements the primitives used by the Idris compiler that didn't have Node implementations yet (and stubs out the socket-related code to avoid having to port networking stuff for now).
    * You might be able to skip the install step, as long as you can set up paths so the resulting libraries are used for the next step
2. `idris2 --cg node --build idris2.ipkg`

# Expected Behavior

It compiles successfully, or at least fails in a useful fashion.

# Observed Behavior

The compiler allocates memory until it runs out, following this pattern:
1. Gradually allocate a large amount of memory, initially around 12 GB
2. Free most of it, in the span of a few seconds
3. Repeat, but reaching a higher peak

I made it to around 40 GB before hitting the OOM killer. (Did you know macOS had an OOM killer? I didn't!)

PR #2981, opened 2023-05-18
Idris2 build for nodejs

freddi301:
# Description

It builds working nodejs idris2 executable!

Ultimate goal is to build idris2 compiler for various javascript targets.
It will enable things like:
- `npm install idris2`
- use idris compiler in the browser (i would love to see "Try Idris 2 right here in the browser" in the doc site)

### DIFF
--- a/support/js/support_system_file.js
+++ b/support/js/support_system_file.js
@@ -58,6 +58,27 @@ function support_system_file_readLine (file_ptr) {
   return line
 }
 
+function support_system_file_readChars (length, file_ptr) {
+  const readBuf = Buffer.alloc(length)
+  const bytesRead = support_system_file_fs.readSync(file_ptr.fd, readBuf, 0, length, null)
+  if (bytesRead === 0) {
+    file_ptr.eof = true
+  }
+  const chars = file_ptr.buffer.slice(0, bytesRead + 1).toString('utf-8')
+  return chars
+}
+
+function support_system_file_readChar (file_ptr) {
+  const readBuf = Buffer.alloc(1);
+  if (support_system_file_fs.readSync(file_ptr.fd, readBuf, 0, 1) === 0) {
+    file_ptr.eof = true
+    // No bytes read, getChar from libc returns -1 in this case.
+    return String.fromCharCode(-1)
+  } else {
+    return readBuf.toString('utf-8')
+  }
+}
+
 function support_system_file_getStr () {
   return support_system_file_readLine({ fd: 0, buffer: Buffer.alloc(0), name: '<stdin>', eof: false })
 }
--- a/support/js/support_system_file.js
+++ b/support/js/support_system_file.js
@@ -79,6 +79,17 @@ function support_system_file_readChar (file_ptr) {
   }
 }
 
+function support_system_file_flush (file_ptr) {
+  try{
+    support_system_file_fs.fdatasyncSync(file_ptr.fd)
+    return 0
+  }catch(e){
+    process.__lasterr = e
+    file_ptr.eof = true
+    return String.fromCharCode(-1)
+  }
+}
+
 function support_system_file_getStr () {
   return support_system_file_readLine({ fd: 0, buffer: Buffer.alloc(0), name: '<stdin>', eof: false })
 }
--- a/Dumb.idr
+++ b/Dumb.idr
@@ -1,3 +1,5 @@
 
-x : Integer
-x = 2
+main : IO ()
+main = do
+  putStrLn "Hello"
+  pure ()
--- a/Dumb.idr
+++ b/Dumb.idr
@@ -1,0 +1,3 @@
+
+x : Integer
+x = 2
--- a/NODE_BUILD.md
+++ b/NODE_BUILD.md
@@ -1,2 +1,10 @@
-`make distclean && make bootstrap=chezscheme && make install && idris2 --build idris2.node.ipkg && cp build/exec/idris2.js ~/.idris2/bin/idris.js && node ~/.idris2/bin/idris.js --codegen node Dumb.idr -o Dumb
+`
+make distclean && \
+make bootstrap SCHEME=chezscheme && \
+make install && \
+mkdir -p build/ttc/2023033100 && \
+cp -r ~/.idris2/idris2-0.6.0/*/2023033100/* build/ttc/2023033100 && \
+~/.idris2/bin/idris2 --build idris2.node.ipkg && \
+node build/exec/idris2.js --codegen node Dumb.idr -o Dumb.js && \
+node build/exec/idris2.js
 `
--- a/NODE_BUILD.md
+++ b/NODE_BUILD.md
@@ -1,2 +1,2 @@
-`make distclean && make bootstrap=chezscheme && make install && idris2 --build idris2.node.ipkg
+`make distclean && make bootstrap=chezscheme && make install && idris2 --build idris2.node.ipkg && cp build/exec/idris2.js ~/.idris2/bin/idris.js && node ~/.idris2/bin/idris.js --codegen node Dumb.idr -o Dumb
 `
--- a/NODE_BUILD.md
+++ b/NODE_BUILD.md
@@ -1,0 +1,2 @@
+`make distclean && make bootstrap=chezscheme && make install && idris2 --build idris2.node.ipkg
+`
--- a/idris2.node.ipkg
+++ b/idris2.node.ipkg
@@ -1,0 +1,9 @@
+package idris2app
+
+depends = network
+
+sourcedir = "src"
+
+main = Idris.Main
+executable = "idris2.js"
+opts = "--codegen node"
--- a/libs/base/System.idr
+++ b/libs/base/System.idr
@@ -31,10 +31,12 @@ libc fn = "C:" ++ fn ++ ", libc 6"
 
 %foreign "scheme,racket:blodwen-sleep"
          supportC "idris2_sleep"
+         "node:lambda:() => { throw new Error('TODO prim__sleep') }"
 prim__sleep : Int -> PrimIO ()
 
 %foreign "scheme,racket:blodwen-usleep"
          supportC "idris2_usleep"
+         "node:lambda:() => { throw new Error('TODO prim__usleep') }"
 prim__usleep : Int -> PrimIO ()
 
 ||| Sleep for the specified number of seconds or, if signals are supported,
@@ -90,6 +92,7 @@ getArgs = do
 prim__getEnv : String -> PrimIO (Ptr String)
 
 %foreign supportC "idris2_getEnvPair"
+         "node:lambda:() => { throw new Error('TODO prim__getEnvPair') }"
 prim__getEnvPair : Int -> PrimIO (Ptr String)
 %foreign supportC "idris2_setenv"
          supportNode "setEnv"
--- a/libs/base/System/Errno.idr
+++ b/libs/base/System/Errno.idr
@@ -4,6 +4,7 @@ module System.Errno
 %default total
 
 %foreign "C:idris2_getErrno, libidris2_support, idris_support.h"
+         "node:lambda:() => { throw new Error('TODO prim__getErrno') }"
 prim__getErrno : PrimIO Int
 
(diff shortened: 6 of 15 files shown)


freddi301:
I'm still very new to the codebase so be forgiving.

Right now i managed to get a working nodejs idri2 executable.

A this point i need to figure out how to hook in the compiling process correctly.

When eveything works smoothly i will do final commit changes for bike-shedding, aligment to code style etc.

Clues and any help is really appreciated. I'm really looking forward to make this PR land!

freddi301:
related and ispiring:

- https://github.com/idris-lang/Idris2/issues/2979
- https://github.com/idris-lang/Idris2/issues/2207
- https://github.com/idris-lang/Idris2/issues/1349

mattpolzin:
I can appreciate that this allows you to make progress toward getting the compiler to build with the NodeJS backend, but it also turns tons of compile-time errors over missing NodeJS FFI functions into runtime errors. I don't like the idea of that tradeoff.

freddi301:
Here is preliminary package that should work.
Unpack, cd into the directory and run `idris2.node.js`
[idris2-node-js.zip](https://github.com/idris-lang/Idris2/files/11512282/idris2-node-js.zip)

freddi301:
> I can appreciate that this allows you to make progress toward getting the compiler to build with the NodeJS backend, but it also turns tons of compile-time errors over missing NodeJS FFI functions into runtime errors. I don't like the idea of that tradeoff.

As of now i think i can implement all the missing FFI so the runtime erros should be gone.

Except for `Libraries.Utils.Scheme` that maybe is not needed. I can make a script to fill in those missing implementations with runtime erros at build time only. Anyway, as far as you know, is that module used?

freddi301:
Almost there for the in browser version https://stackblitz.com/edit/node-nxhv96

support/js/support_system_file.js
```diff
@@ -58,6 +58,38 @@ function support_system_file_readLine (file_ptr) {
   return line
 }
 
+function support_system_file_readChars (length, file_ptr) {
+  const readBuf = Buffer.alloc(length)
+  const bytesRead = support_system_file_fs.readSync(file_ptr.fd, readBuf, 0, length, null)
+  if (bytesRead === 0) {
+    file_ptr.eof = true
+  }
+  const chars = file_ptr.buffer.slice(0, bytesRead + 1).toString('utf-8')
```
dunhamsteve, line 67: This looks like it's off by one.

support/js/support_system_file.js
```diff
@@ -58,6 +58,38 @@ function support_system_file_readLine (file_ptr) {
...
+}
+
+function support_system_file_flush (file_ptr) {
+  try{
+    support_system_file_fs.fdatasyncSync(file_ptr.fd)
+    return 0
+  }catch(e){
+    process.__lasterr = e
+    file_ptr.eof = true
+    return String.fromCharCode(-1)
```
dunhamsteve, line 89: This looks like it's returning a number (0)  in one case and a string in the other case.

CodingCellist:
> > I can appreciate that this allows you to make progress toward getting the compiler to build with the NodeJS backend, but it also turns tons of compile-time errors over missing NodeJS FFI functions into runtime errors. I don't like the idea of that tradeoff.
> 
> As of now i think i can implement all the missing FFI so the runtime erros should be gone.

I think it would be best to convert this PR to a draft until then. This would still let us track and discuss progress, while preventing an accidental merge  : )

CodingCellist:
> Except for `Libraries.Utils.Scheme` that maybe is not needed. I can make a script to fill in those missing implementations with runtime erros at build time only. Anyway, as far as you know, is that module used?

Based on a quick look at the file itself, it _seems_ the file is just FFI-bindings for some of the functions in the Chez and Racket support files ([`support/{chez,racket}`](https://github.com/idris-lang/Idris2/tree/main/support)). So assuming it's only used when we're targeting Scheme, it might be fine to leave unimplemented. A quick ripgrep returns the following files as importing it:
```
$ rg 'Libraries.Utils.Scheme'
src/Libraries/Utils/Scheme.idr
1:module Libraries.Utils.Scheme

src/Core/Context.idr
15:import Libraries.Utils.Scheme

src/Core/SchemeEval/Builtins.idr
8:import Libraries.Utils.Scheme

src/Core/SchemeEval/ToScheme.idr
4:import Libraries.Utils.Scheme

src/Core/SchemeEval/Evaluate.idr
12:import Libraries.Utils.Scheme

src/Core/SchemeEval/Compile.idr
24:import Libraries.Utils.Scheme

src/Core/TTC.idr
20:import Libraries.Utils.Scheme

src/Core/Context/Context.idr
18:import Libraries.Utils.Scheme

idris2api.ipkg
219:    Libraries.Utils.Scheme,

src/Compiler/Common.idr
20:import Libraries.Utils.Scheme
```
Try not implementing it and see what happens I guess ^^

freddi301:
> > > I can appreciate that this allows you to make progress toward getting the compiler to build with the NodeJS backend, but it also turns tons of compile-time errors over missing NodeJS FFI functions into runtime errors. I don't like the idea of that tradeoff.
> > 
> > 
> > As of now i think i can implement all the missing FFI so the runtime erros should be gone.
> 
> I think it would be best to convert this PR to a draft until then. This would still let us track and discuss progress, while preventing an accidental merge : )

What it means to convert to a draft? Still learning how to contribute.

support/js/support_system_file.js
freddi301: adressed

dunhamsteve:
Tail call optimization is already implemented for the javascript backend in `TailRec.idr`.  Any mutually recursive subgraph of functions are rewritten to use a trampoline for tail calls.  If you're seeing stack issues, they're from non-tail calls.  The Parser, for example, is almost tail-recursive, but the `mergeWith` calls break that. 

With a scheme backend, Idris uses > 2G of memory to compile itself. I'm not sure a node Idris would be successful at compiling itself, but it might be useful for smaller programs.

I would like to see more FFI implemented, I think network would be generally useful for writing servers in node.  I'm not a fan of stubbing the all of the unimplemented FFI out with throw.  I'd propose an _optional_ backend flag that replaces them with throws and emits a warning during codegen seems acceptable. A flag is already used to turn on compact mode.  I think that could be implemented in `CodeGen.idr:foreignDecl` turning missing FFI into something like `idris_crash` if the flag is present. The flag would have to be propagated down in ESSt.

For the browser use-case, it might be worth considering a separate driver / project that uses the `idris2api` package - this is what typically is done for new backends. They will instantiate the main driver of Idris with the new backend in the list of backends.  I think that would allow you to strip out unnecessary stuff, like the scheme and RefC backends, and editor protocol, slimming down the resulting code and reducing FFI references. (Idris does dead code elimination, if you do need something you can mark it as extern.)  The browser will likely require a reimagining / simplification of the driver anyway - maybe acquiring the Core monad and calling externed functions directly from javascript as interaction occurs.

freddi301:
> Tail call optimization is already implemented for the javascript backend in TailRec.idr. Any mutually recursive subgraph of functions are rewritten to use a trampoline for tail calls. If you're seeing stack issues, they're from non-tail calls. The Parser, for example, is almost tail-recursive, but the mergeWith calls break that.

Super useful to know!

> With a scheme backend, Idris uses > 2G of memory to compile itself. I'm not sure a node Idris would be successful at compiling itself, but it might be useful for smaller programs.

I was thinking the same, node js idris2 would be a limited version for smaller programs as you said.

> I would like to see more FFI implemented, I think network would be generally useful for writing servers in node. I'm not a fan of stubbing the all of the unimplemented FFI out with throw. I'd propose an optional backend flag that replaces them with throws and emits a warning during codegen seems acceptable. A flag is already used to turn on compact mode. I think that could be implemented in CodeGen.idr:foreignDecl turning missing FFI into something like idris_crash if the flag is present. The flag would have to be propagated down in ESSt.

The throw stubs are there just to get something running, neither i would like to let them stay.
I too though about an annotation that applies FFI declaration conditionally. I do not understand what you said about implementing it yet.

> For the browser use-case, it might be worth considering a separate driver / project that uses the idris2api package - this is what typically is done for new backends. They will instantiate the main driver of Idris with the new backend in the list of backends. I think that would allow you to strip out unnecessary stuff, like the scheme and RefC backends, and editor protocol, slimming down the resulting code and reducing FFI references. (Idris does dead code elimination, if you do need something you can mark it as extern.) The browser will likely require a reimagining / simplification of the driver anyway - maybe acquiring the Core monad and calling externed functions directly from javascript as interaction occurs.

I thought it the same way, maybe wrapping idirs2api with a js api.

I really appreciate all the support 🙏

dunhamsteve:
>  do not understand what you said about implementing it yet.

It's maybe moot if we're implementing the FFI anyway, but here are more details if you're curious:

Here the javascript backend checks for a "minimal" directive (which you can set on the command line of `idris2`) and saves it as `mode` in the ESS object on line 807. 

https://github.com/idris-lang/Idris2/blob/2739c3a389a853836762b642ce5e11c18d4e8a05/src/Compiler/ES/Codegen.idr#L799-L804

We could add a new directive (e.g. `stub-ffi`) and also store whether it is enabled in the ESS object (call the field `stubFFI : Bool` for the sake ofyea discussion).

Then in `foreignDecl`, in the `Left` branch, we could check if that flag is set `shouldStubFFI <- stubFFI <$> get ESs`. And if it is set, print a warning and return something like `makeForeign n "lambda:() => { throw new Error(\"{n} not implemented\") }"` (and do the errorConcat if it is not set).

https://github.com/idris-lang/Idris2/blob/2739c3a389a853836762b642ce5e11c18d4e8a05/src/Compiler/ES/Codegen.idr#L537-L552

This should turn unimplemented FFI into a warning + runtime error, but you'd have to opt in when you invoke the compiler.

CodingCellist:
> > > > I can appreciate that this allows you to make progress toward getting the compiler to build with the NodeJS backend, but it also turns tons of compile-time errors over missing NodeJS FFI functions into runtime errors. I don't like the idea of that tradeoff.
> > > 
> > > 
> > > As of now i think i can implement all the missing FFI so the runtime erros should be gone.
> > 
> > 
> > I think it would be best to convert this PR to a draft until then. This would still let us track and discuss progress, while preventing an accidental merge : )
> 
> What it means to convert to a draft? Still learning how to contribute.

It's mostly a cosmetic thing: it marks the PR as being a work-in-progress, but nothing else changes UI-wise. Except, it also greys out the `merge` buttons, preventing someone with merge access to accidentally add the PR to main  : )

You should be able to change it by going to the top of the PR page, on the right hand side, and clicking `Convert to draft`. Nothing will change, no discussions or commits will be lost, it'll just be an extra safeguard. You can then change that when the PR is fully ready to be merged  : )
image

freddi301:
Updates:
Boostrap done (with some manual help) with nodejs, figuring out what is going wrong in the build process, any pointers are highly appreciated. See [notes](https://github.com/idris-lang/Idris2/pull/2981/files#diff-f5bbc38db94b1c0bd09546b50e01ba17770f0899e7da6d24b54a9eee57768598)

gallais:
I am assuming this draft has been abandoned.

dunhamsteve:
A note for anyone who happens across this PR:

I now have a patched idris that compiles to node and can compile itself (to compile itself takes 3GB of RAM and `node --stack-size=8000`). The repository applies a patch to disable scheme eval, disable ide-mode, and make a few things tail recursive. I've mainlined the primitive implementations. (I haven't looked at ide-mode, it may be easy to get it working.)  There are pre-built copies of Idris as releases of that project, found here: https://github.com/dunhamsteve/idris2-js  

The goal of that project was to enable Idris use on systems too small to compile Idris, like some chrome books.

I've also made and Idris2 web playground that runs entirely in the browser. That uses the node build of Idris from the idris2-js repository and an emulation of parts of the node api. I will soon be adding repl support and type at point in the editor. I'll move it over to codemirror next, which supports mobile better than monaco. This repository lives here: https://github.com/dunhamsteve/idris2-playground and web playground is https://dunhamsteve.github.io/idris2-playground

The goal there was to run Idris2 on my phone.

Issue #490, opened 2020-07-24
@-Patterns forcing bound implicits to linearity 0


donovancrichton:
If possible, please include complete self contained source files (for example by uploading to https://gist.github.com) that exhibit the issue in addition to quoting from them here. The smaller the example the better!

# Steps to Reproduce
Take the following definition of an ordinary cons-list where we specifically want 'don't care' linearity:
```idris
data MyList : Type -> Type where
  Nil : {a : Type} -> MyList a
  (::) : {a : Type} -> a -> MyList a -> MyList a
```
and the following function:
```idris
g : MyList a -> ()
g [] = ()
g l@(x :: xs) = ?hole
```
# Expected Behavior
```
:t hole
   a : Type
   x : a
   xs : MyList a
   l : MyList a
-------------------------------------
```
# Observed Behavior
```
:t hole
 0 a : Type
   x : a
   xs : MyList a
   l : MyList a
-------------------------------------
```

gallais:
One problem I can see: what should the multiplicity of the as-pattern be
when the variable is linear?

For zero and unlimited, we can pick zero & unlimited. Otherwise should we
also default to zero? Or try something more subtle: if it's not bound in
a user-written argument then one else zero?

donovancrichton:
I'm not very across the subtleties of linear types yet to be honest, so please bear with me while I clarify my understanding (and please point out anything I get wrong!) :)
As I understand it we have the following:

- Linearity 0 means we cannot match on the type. So we cannot use this on the LHS or the RHS.
- Linearity 1 means we must match on the type(?). We must use this on the RHS.
- Linearity Infinity means we may or may not match on the type, and we may or may not use this on the RHS. 

Let if me know if I got this right to start with, and then we can think about where to go from here. :)

andrevidela:
I think you got it but just for completeness here is what's going on with linearity 1

There are two modes:
1. Match on the constructor and use any variables bound by the matching on the rhs
2. Do not match but use the bound variable on the rhs

Here is an example of the first case

```
incNotZero : (1 _ : Nat) -> Nat
incNotZero Z = Z -- no values bound, nothing to use
incNotZero (S n) = S (S n) -- we bind the predecessor to `n` so it must be used
```

Here is an example of the second case

```
increment : (1 _ : Nat) -> Nat
increment n = S n -- we don't match but we bind `n` so it must be used
```

Here is an example mixing the two

```
plus : (1 a, b : Nat) -> Nat
[...]

PR #2977, opened 2023-05-16
[ fix #1878 ] Programmer-provided terms should be `alwaysReduce`

AlgebraicWolf:
# Description

This PR ports commit edwinb/Yaffle@b852a6f84b2f8db6e3c06152c2eb9fdd1a795aa1 back to Idris.

After experimenting with Yaffle, I've noticed that #1878 is not reproducible in Yaffle after the aforementioned commit. Porting the change back to Idris seems to fix the issue. 

## Should this change go in the CHANGELOG?

- [x] If this is a fix, user-facing change, a compiler change, or a new paper
      implementation, I have updated `CHANGELOG.md` (and potentially also
      `CONTRIBUTORS.md`).

PR #2984, opened 2023-05-22
Ignore `UseSide` annotation in non-linear as-patterns

mjustus:
This should hopefully unblock PR #2977.

### DIFF
--- a/src/TTImp/ProcessDef.idr
+++ b/src/TTImp/ProcessDef.idr
@@ -285,10 +285,12 @@ findLinear top bound rig tm
       findLinArg : {vars : _} ->
                    RigCount -> NF [] -> List (Term vars) ->
                    Core (List (Name, RigCount))
-      findLinArg rig ty (As fc UseLeft _ p :: as)
-          = findLinArg rig ty (p :: as)
-      findLinArg rig ty (As fc UseRight p _ :: as)
-          = findLinArg rig ty (p :: as)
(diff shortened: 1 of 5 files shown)


AlgebraicWolf:
It looks like it not just unblocks my PR, but also fixes #490 along the way : )

AlgebraicWolf:
As far as I understand, this PR does not affect the way `@`-pattern works on linear arguments, only behaviour of non-linear ones is changed

buzden:
We definitely need to add both now working test for #490 and tests that linear arguments with `@`-pattern behave correctly (i.e., that we can't reuse matched linear parts) if we don't have them already.

mjustus:
> We definitely need to add both now working test for #490 [...]

The example provided by @donovancrichton in #490 is almost identical to the test that I provide: both use an unrestricted pattern variable bound on the RHS of an as-pattern that also happens to appear in an erased position elsewhere in the LHS of the same clause. But it also uses implicit arguments which obscure the binding site of the variable in question.

> [...] and tests that linear arguments with `@`-pattern behave correctly (i.e., that we can't reuse matched linear parts) if we don't have them already.

We already have a [test for that](https://github.com/idris-lang/Idris2/blob/04c5531d9bf95f2a02cb4d74b6594390d565b6b9/tests/idris2/linear009/qtt.idr)!

gallais (approved):

PR #2985, opened 2023-05-22
[ doc ] Add Windows WSL usage

freddi301:
# Description

Add step by step instruction to use Idris 2 on windows easily.

### DIFF
--- a/docs/source/tutorial/starting.rst
+++ b/docs/source/tutorial/starting.rst
@@ -82,7 +82,7 @@ Open PowerShell and run:
 	cd Idris2 # move to the source code directory
 	make bootstrap SCHEME=chezscheme # builds the executable
 	make install # installs the executable
-	ln -s /root/.idris2/bin/idris2 /usr/local/bin/idris2 # add the binary to executable path
+	ln -s ~/.idris2/bin/idris2 /usr/local/bin/idris2 # add the binary to executable path
 	exit # exit the linux terminal 
 
 Open PowerShell in your source code directory and run:
@@ -90,6 +90,8 @@ Open PowerShell in your source code directory and run:
 .. code-block:: PowerShell
 
 	wsl -d Ubuntu idris2 # to run idris2 command in the current Windows directory
+
+Or in alternative you can use vscode with wls and idris extensione combined.
 
 
 Installing from a Package Manager
--- a/docs/source/tutorial/starting.rst
+++ b/docs/source/tutorial/starting.rst
@@ -56,6 +56,42 @@ install with the command ``make install``.  This will, by default, install into
 
     IDRIS2_PREFIX ?= /usr/local
 
+On Windows
+-------------------------
+
+On Windows systems you can build and use Idris 2 from sources using WSL (Windows Subsystem for Linux).
+
+Open Windows PowerShell as administrator and run:
+
+.. code-block:: PowerShell
+
+	wsl --instal # enable Window to run a linux system 
+	wsl --install -d Ubuntu # install a specific linux distro
+
+A restart may be required.
+
+Open PowerShell and run:
+
+.. code-block:: PowerShell
+
+	wsl -d Ubuntu # open a linux terminal
+	apt update # it is necessary to install dependencies
+	apt install make gcc chezscheme libgmp3-dev # install dependencies required for build from source
+	cd ~ # by default wsl will open in the current Windows directory, so we move to a directory inside the linux filesystem
+	git clone https://github.com/idris-lang/Idris2.git # download source code from github
+	cd Idris2 # move to the source code directory
+	make bootstrap SCHEME=chezscheme # builds the executable
+	make install # installs the executable
+	ln -s /root/.idris2/bin/idris2 /usr/local/bin/idris2 # add the binary to executable path
+	exit # exit the linux terminal 
+
+Open PowerShell in your source code directory and run:
+
+.. code-block:: PowerShell
+
+	wsl -d Ubuntu idris2 # to run idris2 command in the current Windows directory
+
+
 Installing from a Package Manager
 =================================
 


docs/source/tutorial/starting.rst
```diff
@@ -56,6 +56,44 @@ install with the command ``make install``.  This will, by default, install into
...
+
+A restart may be required.
+
+Open PowerShell and run:
+
+.. code-block:: PowerShell
+
+	wsl -d Ubuntu # open a linux terminal
+	apt update # it is necessary to install dependencies
+	apt install make gcc chezscheme libgmp3-dev # install dependencies required for build from source
```
CodingCellist, line 79: * You need to use `sudo` for package installations.
* There is a `build-essential` meta-package which installs all the relevant tools (`make`, `gcc`, etc.)
  - Chez and GMP will still need a separate listing to be installed.
* Probably good to ask them to install the man-pages (documentation) as well: `manpages-dev` ?

docs/source/tutorial/starting.rst
```diff
@@ -56,6 +56,44 @@ install with the command ``make install``.  This will, by default, install into
...
+	cd ~ # by default wsl will open in the current Windows directory, so we move to a directory inside the linux filesystem
+	git clone https://github.com/idris-lang/Idris2.git # download source code from github
+	cd Idris2 # move to the source code directory
+	make bootstrap SCHEME=chezscheme # builds the executable
+	make install # installs the executable
+	ln -s ~/.idris2/bin/idris2 /usr/local/bin/idris2 # add the binary to executable path
```
CodingCellist, line 85: You'd need `sudo` (the destination is owned by `root` by default), **however** symlinks are almost always a bad idea (it is very easy to get the order of the arguments wrong, and it creates unnecessary complications in terms of where things are located in this case).

It would be better to instruct users to add the following line at the end of their `~/.bashrc` file and restart their terminal:

```bash
export PATH="$HOME/.idris2/bin:$PATH"
```

(This adds the default Idris2 install location to the magic variable that the shell searches for executables when you try to run a command.)

CodingCellist (changes requested):
I think this is definitely something we ought to have, thank you for writing it up! I have left some comments and suggested edits  : )

docs/source/tutorial/starting.rst
```diff
@@ -56,6 +56,44 @@ install with the command ``make install``.  This will, by default, install into
...
+
+Open Windows PowerShell as administrator and run:
+
+.. code-block:: PowerShell
+
+	wsl --instal # enable Window to run a linux system 
+	wsl --install -d Ubuntu # install a specific linux distro
+
+A restart may be required.
+
```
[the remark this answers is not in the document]
freddi301: I thought about that, but it's too detailed. I like guides that are a copy pasteable single script, nay time i can search for more detailed docs. I can think of some alterantives:

- leave only the link
- leave the commands + link
- leave only the commands

What do you think is best?

docs/source/tutorial/starting.rst
```diff
@@ -56,6 +56,44 @@ install with the command ``make install``.  This will, by default, install into
...
+	ln -s ~/.idris2/bin/idris2 /usr/local/bin/idris2 # add the binary to executable path
+	exit # exit the linux terminal 
+
+Open PowerShell in your source code directory and run:
+
+.. code-block:: PowerShell
+
+	wsl -d Ubuntu idris2 # to run idris2 command in the current Windows directory
+
+Or in alternative you can use vscode with wls and idris extensione combined.
```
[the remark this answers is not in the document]
freddi301: There are several extensions, which one do you reccomend? I think to link the "best" one as an example is a good idea

docs/source/tutorial/starting.rst
```diff
@@ -56,6 +56,44 @@ install with the command ``make install``.  This will, by default, install into
...
+	wsl --instal # enable Window to run a linux system 
+	wsl --install -d Ubuntu # install a specific linux distro
+
+A restart may be required.
+
+Open PowerShell and run:
+
+.. code-block:: PowerShell
+
+	wsl -d Ubuntu # open a linux terminal
```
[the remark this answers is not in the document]
freddi301: I can understand that, the existing docs are very flexible about how to isntall and build things.
I come from web development where to install or build there is only one command to type, nothing to choose, just plain convention. I wrote these docs considering my experience trying to make it work, there were many things to choose, details to hunt down.
Ultimately my intention here is boost adoption, anybody within hours could get how to build idris in windows under wls, my intention is to provide a nobrainer.
At this point i think a good compromise would be a disclaimer on top like "This is a oneshot copypaste command to build, for more customization you find all the stuff in the docs"

docs/source/tutorial/starting.rst
```diff
@@ -56,6 +56,44 @@ install with the command ``make install``.  This will, by default, install into
...
+	apt update # it is necessary to install dependencies
```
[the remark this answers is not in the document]
freddi301: I didn't know that, i thought the semver resolution was enough and also maybe we could save the user to upgrade the whole system. Anyway will change it as you suggested.

docs/source/tutorial/starting.rst
freddi301: default wls installation has root user only (at least from what i know). But good idea to prepend sudo just in case the user created other linux users. Ok, i will update to build-essentials. What would the user need the man-pages for?
Thank you for all the precious feedback! (I'm learning a lot)

docs/source/tutorial/starting.rst
freddi301: I though about that, but is seems that commands like "wsl -D ubuntu idris2" do not pick up the `~/.bashrc` i will test

freddi301:
@CodingCellist maybe i want to rethink the guide. Something along the lines: "here is an opinionated setup on windows"
- build from source using chez on ubuntu under wls
- install pack and idris2-lsp (add to bashrc)
- use vscode with wls container extension and bamboo.idris2-lsp idris

In the form of a copy pasteable script, and leaving out all other details.

What do you think?

freddi301:
I thought more about your feedback, we could include both opinionated setup and general instructions. In the form of:
```markdown
Example enviroment setup on windows:
This is a step-by-step guide for an opinionated setup. It provides general instructions for each step and copy pasteble commands that should work in most cases for the hasty.
- install windows wsl
  - see instructions [here](https://learn.microsoft.com/en-us/windows/wsl/install)
  - or `wsl --install`
- install a linux distribution if you do not have already one
  - see guide [here](other link)
  - or `wsl --install -d Ubuntu`
- and so on
```

I wait for feedback before rewriting.

Thank you very much for the precious info, it is very appreciated.

docs/source/tutorial/starting.rst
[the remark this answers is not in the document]
CodingCellist: Based off the number of downloads, this one? [https://marketplace.visualstudio.com/items?itemName=bamboo.idris2-lsp](https://marketplace.visualstudio.com/items?itemName=bamboo.idris2-lsp)

It also seems to be the recommended one from the other popular extension:

> For Idris 2, the language-server based extension provides additional features, such as semantic highlighting. If you are using the latest version of Idris 2, it may provide a better experience.

\- [maraymond2/idris-vscode:README](https://github.com/meraymond2/idris-vscode#which-extension-should-i-use)

docs/source/tutorial/starting.rst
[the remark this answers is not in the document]
CodingCellist: Hmm, yeah it's a question of balance... I definitely get where you're coming from, I am not particularly fond of "guides" that require you to be 5 tabs side-tracked every other step, so I think commands + link (with an "would you like to know more/advanced instructions" note) is the compromise to go with  : )

docs/source/tutorial/starting.rst
CodingCellist: The [WSL documentation](https://learn.microsoft.com/en-us/windows/wsl/setup/environment#set-up-your-linux-username-and-password) seems to suggest otherwise:
> Once the process of installing your Linux distribution with WSL is complete, open the distribution (Ubuntu by default) using the Start menu. You will be asked to create a User Name and Password for your Linux distribution.

The man-pages would mostly be for trouble-shooting on the user's end. I guess they're not strictly required, but a lot of guides, StackOverflow answers, and similar refer to them as a reference point and/or further reading. So I figured it'd be best to include them  : )

docs/source/tutorial/starting.rst
CodingCellist: Any luck? I'm not familiar with that command, I've always just opened up a WSL terminal, and then done everything as I would on a regular linux terminal  : )

freddi301: Yes, at least onmy machine a cannot make it work with bashrc. Anyway i think this feature is not worth the trouble

andrevidela:
Any progress @freddi301 ?

freddi301:
Lacking fre time lately. I will eventually complete the PR

freddi301:
For my personal usage i settled on installing idris 2 with [pack](https://github.com/stefan-hoeck/idris2-pack) inside a WLS and editing with VSCode + idris2-lsp extension

PR #2986, opened 2023-05-22
Add Pack to list of package managers

l-monnier:
Hello,

It would probably be fair to also mention Pack as package manager currently in development.

### DIFF
--- a/docs/source/libraries/index.rst
+++ b/docs/source/libraries/index.rst
@@ -6,8 +6,9 @@ https://github.com/idris-lang/Idris2/wiki/Third-party-Libraries
 
 Please feel free to contribute your own libraries there! Eventually, we aim to
 have a package manager for managing libraries and dependencies. We do not yet
-have an official one, but there are (at least) two in development:
+have an official one, but there are (at least) three in development:
 
 * Inigo: https://github.com/idris-community/inigo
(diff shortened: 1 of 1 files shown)


stefan-hoeck:
The development of inigo has stalled (last commit was two years ago). Personally, I think it should be removed from this list.

docs/source/libraries/index.rst
```diff
@@ -6,8 +6,9 @@ https://github.com/idris-lang/Idris2/wiki/Third-party-Libraries
 
 Please feel free to contribute your own libraries there! Eventually, we aim to
 have a package manager for managing libraries and dependencies. We do not yet
-have an official one, but there are (at least) two in development:
+have an official one, but there are (at least) three in development:
 
 * Inigo: https://github.com/idris-community/inigo
+* Pack: https://github.com/stefan-hoeck/idris2-pack
 * sae: https://github.com/DoctorRyner/sae
```
gallais, line 13:
```suggestion
have an official one, but there are (at least) one actively developed:

* Pack: https://github.com/stefan-hoeck/idris2-pack
```

l-monnier:
Thank you for the suggestion. I believe this is indeed a very good idea to mention only Pack. It would make the life of beginners like me much easier as it would push them to one good place.

Issue #2599, opened 2022-07-25
Refactor the lazy keywords to be actual keywords


CodingCellist:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

Prompted by a question from locria/locriacyber and spotted by gallais, the keywords for lazy things (e.g. `Inf` and `Lazy`) are actually not keywords but magic strings in the parser! This means we cannot do nice things like output `:doc` for them, leading to confusion for what they mean.

The immediately relevant parts seem to be:
- https://github.com/idris-lang/Idris2/blob/main/src/Idris/Parser.idr#L922-L935=
- https://github.com/idris-lang/Idris2/blob/main/src/TTImp/Parser.idr#L438-L459=

Although there'll probably be other bits that need changing as well.

## Motivation

`:doc` support and having keywords be properly+consistently marked as such.

locriacyber:
Is it a good idea to change `Lazy a` to `() -> a` and add syntax magic on it? Totality checker seem to work.

```
data LazyList : Type -> Type where
    Nil : LazyList a
    (::) : a -> (Unit -> LazyList a) -> LazyList a

total
all42 : LazyList Nat
all42 = 42 :: (\_ => all42)
```

```
1/1: Building reclazy (reclazy.idr)
Error: all42 is not total, possibly not terminating due to recursive path Main.all42 -> Main.all42

reclazy:5:1--6:21
 5 | total
 6 | all42 : LazyList Nat

Error(s) building file reclazy.idr
```

locriacyber:
There is also `Delay`, which is kind of like runtime hint?
```
Main> :let d = \_ => delay 3
Main> d
\_ => Delay 3
```

CodingCellist:
Will actually be closed if we merge #2992. Although #2987 adds `:doc`-support, it _doesn't_ refactor the laziness primitives to be reserved (which is what the later PR does).

CodingCellist:
Closing this as there seems to be opposition to #2992 , and all I mostly cared about was `:doc` support. If we discover that we absolutely _need_ the keywords to be reserved, or someone has a strong case for it, feel free to reopen and have a look at #2992  : )

PR #2987, opened 2023-05-23
[ new ] Implement `:doc` for laziness annotations

CodingCellist:
# Description

Long requested, this implements REPL `:doc` for laziness annotations (`Lazy`, `Inf`, `Delay`, and `Force`). I'm mostly confident I understand these, but if I've misunderstood anything about how these work, or there are better ways to describe it, please let me know! : )

There is a custom parser for parsing the doc command, since the default `lazy` one requires a `simpleExpr` immediately after the laziness annotation and so would've been awkward to use (e.g. `:doc (Lazy ?huh)` vs simply `:doc Lazy`).

~~I have also modified the lexer+parser to have a `LazyPrim`. It turned out not to be necessary for the docs, but I thought it was slightly tidier so I left it in. I don't _think_ it breaks anything, all tests seem to pass. But if someone more familiar with the parser internals thinks having it is a bad idea, I'll happily remove it.­~~ EDIT: This requires further thought+discussion (see comments further down, and PR #2992).

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,10 @@
 
 * Constant folding of trivial let statements and `believe_me`.
 
+* Moved `Lazy`, `Inf`, `Delay`, and `Force` from magic strings to reserved
+  compiler primitives. This may mean that certian namespaces and libararies need
+  renaming (see, for example, the `Prelude.Interfaces` change below).
+
 ### Library changes
 
 #### Prelude
@@ -101,6 +105,9 @@
 * Added an `Interpolation` implementation for `Void`.
 
 * Added `Compose` instances for `Bifunctor`, `Bifoldable` and `Bitraversable`.
+
+* Rename the `Lazy` namespace in `Interfaces` to `ILazy` due to `Lazy` now being
+  a reserved compiler primitive.
 
 #### Base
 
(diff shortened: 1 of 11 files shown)


gallais (approved):
Looks great!

buzden:
> The refactor means we've made `Lazy` & friends proper reserved keywords which I'm not opposed to
> but we'll have to fix the libs.

I also have modules and namespaces called just `Lazy` in my libs, and that was really handy for lazy variants of different data structures. I don't feel disallowing to be good. What do you suggest to use instead?

buzden:
> Personally, I think this is the opportunity to think about a feature to allow reserved words for any identifier

Indeed, but it means that we shouldn't do exactly the opposite now.

CodingCellist:
N.B. As of the latest commit to this PR, it **no longer closes #2599**. But it doesn't seem like GH has a function for me to remove that link...

CodingCellist:
@buzden I've moved the breaking changes to a separate PR. It seems that part probably needs further discussion, but I still think we want to be able to `:doc` laziness primitives, which is now all this PR does (the way it should've been, in hindsight)  : )

Hopefully this makes merging this PR uncontroversial. And the other concerns can now be discussed in #2992.

Issue #2936, opened 2023-03-31
Unexpected delayed evaluation of higher-order functions


ohad:
# Steps to Reproduce
```idris
versionA : Bool -> Nat -> Nat
versionA True  k0 = k0
versionA False k0 = 0

versionB : Bool -> Nat -> Nat
versionB True  = \k0 => k0
versionB False = \k0 => 0

-- This works as expected
Ex1 : Nat -> Nat
Ex1 = versionB True

-- This does not
Ex2 : Bool -> (Nat, Nat, Nat) -> (Nat, Nat, Nat)
Ex2 b = let f = versionB b
        in \(x,y,z) => (f x, f y, f z)
```
and then `:di Ex1` and `:di Ex2` in the REPL.

# Expected Behavior
```
λΠ> :di Ex1
Compiled: \ {ext:0} => {ext:0}
λΠ> :di Ex2
Compiled: \ {arg:0} => let f = case {arg:0} of
  { 1 => \{clam:0} => {clam:0}
  ; 0 => \{clam:0} => 0
  } in
\{lamc:0}=> case {lamc:0} of
  { Builtin.MkPair {tag = 0} [cons] {e:2} {e:3} => case {e:3} of  { Builtin.MkPair {tag = 0} [cons] {e:6} {e:7} => Builtin.MkPair {tag = 0} [cons] (f {e:2}) (Builtin.MkPair {tag = 0} [cons] (f {e:6}) (f {e:7}))}
  }

```
# Observed Behavior
```
λΠ> :di Ex2
Compiled: \ {arg:0} => let f = \{clam:0}=> case {arg:0} of
  { 1 => {clam:0}
  ; 0 => 0
  } in
\{lamc:0}=> case {lamc:0} of
  { Builtin.MkPair {tag = 0} [cons] {e:2} {e:3} => case {e:3} of  { Builtin.MkPair {tag = 0} [cons] {e:6} {e:7} => Builtin.MkPair {tag = 0} [cons] (f {e:2}) (Builtin.MkPair {tag = 0} [cons] (f {e:6}) (f {e:7}))}
  }
```

Rationale: I'm trying to use DIY-partial evaluation using higher-order-functions. By doing the case-splitting ahead of time, I return the correct function so that when I call this function 3 times at runtime, I'm not repeating the case-splitting.

Z-snails:
I imagine this is caused by the following optimisation
[...]

Issue #1066, opened 2021-02-14
Top-level lazy functions don't typecheck


mb64:
#### Steps to Reproduce

Make a top-level declaration with a lazy type that can't be inferred:

```idris
foo : Inf (Unit -> Unit)
foo = \x => x

bar : Inf (Unit -> Unit)
bar = Delay (\x => x)
```

#### Expected Behavior

They should type check fine.

#### Observed Behavior

```
Error: Not a function name in pattern LHS

t.idr:2:1--2:14
   |
 2 | foo = \x => x
   | ^^^^^^^^^^^^^

Error: While processing right hand side of bar. Can't infer delay type

t.idr:5:7--5:20
   |
 5 | bar = Delay \x => x
   |       ^^^^^^^^^^^^^
```

gallais:
Note that `Lazy` and `Inf` are not the same thing.

mb64:
That is true.  In this case, changing `Inf` to `Lazy` gives exactly the same output, hence the lowercase-L "lazy" -- sorry for not making that clear.

locriacyber:
> Note that `Lazy` and `Inf` are not the same thing.

@gallais Can you elaborate? I was recently testing `Inf` but discovered that it can do the same thing as `Lazy` in the following code:

```idris2
record Foo where
    constructor MkFoo
    bar : Lazy Foo -- could be Inf Foo
    baz : Nat


print_rec : Foo -> Nat -> IO ()
print_rec ff Z = pure ()
print_rec ff (S remaining) = do
    putStrLn (show ff.baz)
    print_rec ff.bar remaining


aa : Foo
bb : Foo
aa = MkFoo bb 1
bb = MkFoo aa 2
selfrec = MkFoo selfrec 3


main : IO ()
main =
    do
        let cc = { baz := 42 } aa
        let dd = { baz := 20 } bb
        print_rec cc 10
        print_rec dd 10
        print_rec selfrec 4

```

I was also confused by `Delay` in REPL. Sometimes `Delay a` show up where `Lazy a` should in REPL.

```
Main> :ti aa.bar
aa .bar : Inf Foo
Main> :ti Foo.bar
Main.Foo.bar : Foo -> Inf Foo
Main> aa
MkFoo (Delay bb) 1

# now change  Lazy Foo  to  Inf Foo

Main> :r
1/1: Building inf (inf.idr)
Loaded file inf.idr
Main> aa
MkFoo (Delay bb) 1
[...]

PR #2988, opened 2023-05-26
Rudimentary support for lazy functions (addressing issue #1066)

mjustus:
This PR makes it possible to define lazy/delayed functions. It is not a full solution: neither implicit functions nor definition by pattern matching are supported and will fail with an error message indicating as much but this PR should already be enough to enable @ohad's [workaround](https://github.com/idris-lang/Idris2/issues/2936#issuecomment-1521955870) for issue #2936. 

      `CONTRIBUTORS.md`).


gallais (approved):

andrevidela:
I'm assuming this should be merged since it's approved?

PR #2992, opened 2023-05-29
[ breaking ] Make laziness + codata reserved prims

### LINKED ISSUE #2599: Refactor the lazy keywords to be actual keywords

CodingCellist:
# Description

Changes `Lazy`, `Inf`, `Delay`, and `Force` to be reserved primitives instead of magic `exactIdent` strings.

This is moved from PR #2987 , which added `:doc` REPL interaction for these. This change is A) separate, and B) more controversial, since it potentially requires refactoring of libraries and tools outwith the Idris2 upstream.

Closes #2599 .

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,11 @@
 
 * Constant folding of trivial let statements and `believe_me`.
 
+* Made `Lazy`, `Inf`, `Delay`, and `Force` reserved compiler primitives, rather
+  than being magic string identifiers. This may mean that certain namespaces and
+  libraries need renaming (see, for example, the `Prelude.Interfaces` change
+  below).
+
 ### Library changes
 
 #### Prelude
@@ -99,6 +104,9 @@
 * Added an `Interpolation` implementation for `Void`.
 
 * Added `Compose` instances for `Bifunctor`, `Bifoldable` and `Bitraversable`.
+
+* Renamed the `Lazy` namespace in `Interfaces` to `ILazy` due to `Lazy` now
+  being a reserved compiler primitive.
 
 #### Base
 
--- a/libs/prelude/Prelude/Interfaces.idr
+++ b/libs/prelude/Prelude/Interfaces.idr
@@ -449,13 +449,13 @@ public export
 [MonoidApplicative] Applicative f => Monoid a => Monoid (f a) using SemigroupApplicative where
   neutral = pure neutral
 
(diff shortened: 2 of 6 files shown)


CodingCellist:
## Context (copied from #2987)

@buzden  commented [May 24, 2023](https://github.com/idris-lang/Idris2/pull/2987#issuecomment-1560630859)

> The refactor means we've made Lazy & friends proper reserved keywords which I'm not opposed to
> but we'll have to fix the libs.

I also have modules and namespaces called just Lazy in my libs, and that was really handy for lazy variants of different data structures. I don't feel disallowing to be good. What do you suggest to use instead?


-----

@andrevidela commented [May 25, 2023](https://github.com/idris-lang/Idris2/pull/2987#issuecomment-1562632642)

@buzden Personally, I think this is the opportunity to think about a feature to allow reserved words for any identifier (proof is one that trips me up all the time). Now the question is to think about what would be a good solution.

Exercise left for the reader

In the meantime, my suggestion would be to replace `Data.Lazy.List` by `Data.LazyList`


-----

@buzden commented [May 27, 2023](https://github.com/idris-lang/Idris2/pull/2987#issuecomment-1565298514)


> Personally, I think this is the opportunity to think about a feature to allow reserved words for any identifier

Indeed, but it means that we shouldn't do exactly the opposite now.

CodingCellist:
Don't really have a strong preference on this; I mostly cared about `:doc` support and that's been added. Since this has been open for a good while now, and there _is_ a strong preference against it, I'm just closing this. We can always add it if need really be; it's not a complicated change  : )

PR #2997, opened 2023-06-09
Infix namespace hiding

andrevidela:
# Description

- Allow to hide fixities using `%hide` and a namespaced fixity identifier
- Remove old fixity warning replace with a more explanatory one
- Update docs
- Update the printing of namespaced operators to match how we parse them

      `CONTRIBUTORS.md`).


src/Core/Options.idr
```diff
@@ -220,7 +221,7 @@ defaultSession : Session
 defaultSession = MkSessionOpts False CoveringOnly False False Chez [] 1000 False False
                                defaultLogLevel Nothing False Nothing Nothing
                                Nothing Nothing False 1 False True
-                               False [] False False
+                               False False [] False False
```
ohad, line 224: I wonder if the default should be `True`?

ohad: The rationale: if you don't know that what's happening is a fixity conflict, it can be quite frustrating to diagnose.
If you get the warning, it's straightforward to suppress with the option.

andrevidela: Ah, indeed, that was my original intent, however this creates some issues with bootstrapping and some tests relying on -Werror

gallais: Can we easily fix the warnings that make the errors break?
(e.g. by passing a `--no-prelude`?)

andrevidela: I don't think we can during bootstrap because the old version does not parse `%hide Namespace.infix.(op)`

andrevidela: looks like bootstraping is fine actually, only frex stops CI now, I'll remove the compiler option since there is no need to let people rely on buggy fixities

andrevidela:
I've disabled the test for frex since the new warning was making them fail.

andrevidela:
@gallais ready for merging?

ohad:
I can fix the frex tests, I think it's mostly just correcting the output?

andrevidela:
@ohad there are some duplicate fixity definitions according to this run: https://github.com/idris-lang/Idris2/actions/runs/5242173961/jobs/9466044829

The way to fix this right now is to remove the duplicates

andrevidela:
looking more closely, it seems like the frex fixities are consistent, even though they are definied in multiple places. In the current state I can detect _multiple non-conflicting_ fixities so I'll do that and not report the warning for them.

src/Core/Name.idr
```diff
@@ -78,6 +112,7 @@ export
 userNameRoot : Name -> Maybe UserName
 userNameRoot (NS _ n) = userNameRoot n
 userNameRoot (UN n) = Just n
+userNameRoot (OP _ n) = Just (Basic n) -- operator and fixities are user names
```
gallais, line 115: I am really uneasy about this. This mean that everywhere we used to match
on `UN` we should now match on both `UN` and `OP`. How do we know we
are not missing cases that are now handled by catchalls?

src/Idris/Desugar.idr
```diff
@@ -82,72 +97,80 @@ extendSyn newsyn
...
             PTerm -> Core (List (Tok OpStr PTerm))
 toTokList (POp fc opFC opn l r)
     = do syn <- get Syn
          let op = nameRoot opn
-         case lookup op (infixes syn) of
-              Nothing =>
+         let foundFixities = lookupName (UN (Basic op)) (infixes syn)
+         let ignoreCompatible = nubBy (\(_, _, pr1), (_, _, pr2) =>
+                                    pr1 == pr2) foundFixities
+         case ignoreCompatible of
```
gallais, line 135: This only compares infixes to infixes (and the later one only prefixes to prefixes)
whereas the previous approach compared all (in/pre)fixes and thus detected strictly
more conflicts.

src/Core/Name.idr
andrevidela: Currently only fixity declarations use the `OP` constructor. I started porting the rest of the parser to use `OP` as well (also because it was mentionned in TODO in comments) but that was too much work, so right now all the machinery uses `UN` as before and `OP` never appears.

This suggests that `OP` should probably be `Fixity -> String -> Name` rather than `Maybe fixity`.

src/Idris/Desugar.idr
andrevidela: Will fix.

This also somewhat suggests that we can get rid of the `prefix`/`infix` maps in `SyntaxInfo` and handle the `-` special case as an intended conflict

src/Core/Name.idr
andrevidela: Ah yes, this version bypasses that in two ways:
- when declaring fixities you are not allowed to use namespaced backticked names
- When looking up fixity conflicts we always use the nameRoot so it does not matter if the operator used is a namespaced backticked name or a plain operator

Here is an example where this is handled: https://github.com/idris-lang/Idris2/pull/2997/files#diff-6a3fe63c93fc56d8ac342e2910c5e4957fbcda5a72480912648a24b499b79880R317

andrevidela: I've removed `OP` and I'm now relying on querying the namespace. I'm not sure this is ideal either, this is probably worth thinking about

andrevidela:
ready to review again

docs/source/reference/pragmas.rst
```diff
@@ -99,3 +99,28 @@ definition.
 
    %name Foo foo,bar
 
+``%hide``
+====================
+
+Hide a definition from imports. This is particularly useful when you are re-definiing functions or types from
```
gallais, line 105:
```suggestion
Hide a definition from imports. This is particularly useful when you are re-defining functions or types from
```

src/Idris/Parser.idr
```diff
@@ -1306,7 +1306,7 @@ logLevel fname
 directive : OriginDesc -> IndentInfo -> Rule Directive
 directive fname indents
     = do decoratedPragma fname "hide"
-         n <- name
+         n <- (fixityNS <|> name)
```
gallais, line 1309: This really feels like a hack: why not have a
```idris
data HidingDirective = HideName Name | HideFixity Fixity Name
```
and have `Hide` take a `HidingDirective` rather than a `Name`
instead of manufacturing an illegal namespace in `fixityNS` and then looking for it
in `isOperatorFixityNS`?

andrevidela: Should be fixed

src/Idris/Desugar.idr
```diff
@@ -82,72 +82,92 @@ extendSyn newsyn
...
                   syn)
 
+-- Check if a namespace is pointing toward a fixity
+-- An namespace can end in an operator fixity if it's in a %hide
+-- directive in order to remove the fixity from the module in order
+-- to better control parsing in the file.
+isOperatorFixityNS : Name -> Bool
+isOperatorFixityNS (NS ns _) = any (`isInPathOf` ns) fixityKeywords
+isOperatorFixityNS _ = False
+
```
gallais, line 102: I think we should be able to garbage collect that definition now.

```suggestion
```

andrevidela: Fixed, I also noticed `checkFixity` wasn't doing anything anymore so it's gone

gallais (approved):
Let's ship this great feature!

PR #3001, opened 2023-06-23
Update docs copyright notice from 2020 to 2023

noamraph:
# Description

The pages at https://idris2.readthedocs.io/ show "© Copyright 2020, The Idris Community". This actually made me think that the project is dead. It turns out that the year is hardcoded in a file and should be updated.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -26,7 +26,7 @@ import sphinx_rtd_theme
 
 # General information about the project.
 project = 'Idris2'
-copyright = '2020, The Idris Community'
+copyright = '2023, The Idris Community'
 author = 'The Idris Community'
 
 # The short X.Y version.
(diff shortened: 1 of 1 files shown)


buzden:
According to my knowledge, not a single yeah should finally be present, but a period of time, where copyright holder was active. In this case, I'd say, it should be `2020-2023`, not just `2023`.

noamraph:
I have no problem with `2020-2023`.

I think that just `2023` would be fine, since at the bottom of this page I see "© 2023 GitHub, Inc.", and I guess they have some lawyers there.

CodingCellist:
Thanks for spotting this, I'll merge it as soon as the linter is happy  : )
(Also, wow! You're the person behind tqdm! That's really cool! It's such a great tool that I've used in many projects, thanks for making and maintaining it!  : D)

PR #3002, opened 2023-06-24
Treat unit types as erased in constructors

Z-snails:
# Description

Credit to @andrevidela for the suggestion

This doesn't pass tests on my PC, maybe somehow it will in CI?

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,10 @@
 * `getChar` is now supported on Node.js and `putChar` is supported on both JavaScript backends.
 * Integer-indexed arrays are now supported.
 
+#### Optimizations
+
+* Unit types can now be erased by the optimiser in data types.
+
 ### Compiler changes
 
 * If `IAlternative` expression with `FirstSuccess` rule fails to typecheck,
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -29,7 +29,7 @@ import public Libraries.Utils.Binary
 ||| version number if you're changing the version more than once in the same day.
 export
 ttcVersion : Int
-ttcVersion = 20230331 * 100 + 0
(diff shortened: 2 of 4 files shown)


dunhamsteve:
chez033, chez035, and building prelude (self-hosted) are failing. Not sure exactly what's going on here, but it seems to fail with `No Clauses` (in casebuilder) when incremental codegen is turned on. 

One option is CompileExpr.idr:802 where `eraseArgs` is not called for holes, but I'm not sure that applies to the test case.

The other possibility is Inline.idr.  Lines 585 and 549 consult `incrementalCGs` and behave differently from normal compiles.

Z-snails:
I think I've figured out the problem, and that is chez scheme evaluates zero-argument functions before they are called - but this is only a problem in incremental mode, because otherwise the function doesn't even get compiled (in this case it's one of the uninhabited instances)

I can fix this by disabling eager evaluation of constants on the incremental backend. Let me know if you can think of a better way to fix this

Z-snails:
Turns out this bug is also present on idris2 main, even without incremental mode:

```idris
bad : (0 _ : Z = S Z) -> Void
bad _ impossible

data BadPrf = Ok | No (Z = S Z)

Show BadPrf where
    show Ok = "Ok"
    show (No prf) = absurd $ bad prf

main : IO ()
main = printLn Ok
```

dunhamsteve:
I was curious what effect this had on performance, so I ran `time idris2 --build idris2api.ipkg` with a clean build directory using an idris built with this branch (with my patches) and also with the main branch of idris. (M1 mac, racket chez at the racket-v8.9 tag.)

The times for the main branch were 1:56.53 and 1:56.57, the times for this branch were 1:55.48 and 1:55.59.  So it looks like about a 1% reduction in time.

PR #3005, opened 2023-06-30
Generalize Prelude proof helpers

madman-bob:
Erase the proof arguments in `cong` and `cong2`. This is a breaking change as we don't have quantity subtyping, but the fix is generally simple. This brings `cong` and `cong2` inline with `sym` and `trans`, to help avoid excessive use of `irrelevantEq` when doing lots of proofs.

Also generalizes `irrelevantEq` to use heterogeneous equality. I think this is technically a breaking change, as there might be cases where we need the type of one side to infer the type of the other, but no such cases occur in this repository.


buzden:
I recently had an idea to define several standard functions that are good to be quantity-polymorphic through `IAlternative`. That has pros of being backward-compatible while solving that problem. Cons are a cluttery code, especially for more-than-one argument functions like `(.)`, if this is used directly, rather than generated through an elaboration script.

PR #3008, opened 2023-07-07
Revert "Treat unit types as erased in constructors" (#3002)

CodingCellist:
Reverts idris-lang/Idris2#3002 which currently breaks some packages on [pack-db](https://github.com/stefan-hoeck/idris2-pack-db), e.g. [TyRE](https://github.com/kasiamarek/tyre). Further discussion on the Discord in the `package-manager` channel, starting [here (n.b. requires Discord access)](https://discord.com/channels/827106007712661524/841274390481600562/1126865667233161268).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,10 +44,6 @@
 * `getChar` is now supported on Node.js and `putChar` is supported on both JavaScript backends.
 * Integer-indexed arrays are now supported.
 
-#### Optimizations
-
-* Unit types can now be erased by the optimiser in data types.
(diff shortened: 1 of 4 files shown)


gallais:
Summary:

>  TyRE also no longer typechecks, and the compiler seems to hang on module TyRE.StringRE for a couple of minutes before it dies,

(...)

> The argument that diverges is the implicit `IsJust` in:
> `toRE : (str : String) -> {auto isJust : IsJust (rAux str)} -> RE`

PR #3009, opened 2023-07-09
[ base ] Add getTermCols and getTermLines to base library and fix pri…

dunhamsteve:
# Description

Adds the`getTermCols` and `getTermLines` functions from the compiler source to the base library.   It also adds error handling to the primitives and makes them work successfully in more cases.

I also changed the C code to return 0 in case of error.  The Idris code expects that, but on macos random numbers are returned instead (possibly unitialized memory). I also changed it to try stdout if stdin fails. So a program that reads from a pipe and writes to the terminal can get the screen size.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,6 +169,9 @@
 
 * Adds `leftmost` and `rightmost` to `Control.Order`, a generalisation of `min` and `max`.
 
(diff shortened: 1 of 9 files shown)


mattpolzin:
Seems like a good thing to have in base to me. The changes look good so far. Can you add a note to the compiler's Term module to remove that module once the next version is released?

mattpolzin (approved):

Issue #2998, opened 2023-06-09
Make fixity declaration consistent with module syntax


andrevidela:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.
	
## Summary

Fixity declarations are important to facilitate code readability, they allow the user to inherit existing notions of syntax from other programming language (like `++` and `>>=`), from domain-specific knowledge (like `/` for file paths), or from mathematics (like `|>` for $\rhd$ or `|-` for turnstile).

However they suffer from being a second-class citizen in the language since they have no proper presence in the module system. As part of program parsing they are more akin to syntactic _macros_ than to proper functions. This aims to outline a plan to introduce fixity as a proper construction that's consistent and integrated with the rest of the feature set from the language.

## Motivation

There has been multiple issues and problem related to fixity and operators in the past, in recent memory #2832 and #2889 come to mind. But other languages have have their own share of issues like Haskell or Swift. One of the strength of Idris that comes with dependent types is the ability to write embedded DSL with strong type guarantees. Those DSL make heavy use of operators with custom fixities, for example [preorder reasoning]() [URIs]() and [records]().

However, the current implementation exposes some unexpected behaviors:
1. All fixities are always exported
2. Parsing depends on the import order
3. Conflicting fixities are unresolvable without messing with the import order

### Fixities are always exported

Here is a proper example of this phenomenon:

```
module Lib

infixl 3 !>
infixl 2 #>

-- run with aribtrary scheduler
(!>) : (s : Scheduler t m) => (() -> m a) -> m a

-- run with default scheduler
public
(#>) : (() -> a) -> Promise a
(#>) = (!>) {s = defaultScheduler}

... -- code using !> for different purposes
```
Now if a user imports this library, they can write:
```
module User

import Lib

-- can use this with fixity infixl 3
(!>) : (a -> b) -> (b -> c) -> a -> c 
```

## The proposal

In a nutshell, the proposal is to integrate the syntax for modules with fixity declaration. 

### Fixities are private by default

This is the biggest breaking change from this proposal. Just like any other top-level definitions, fixities should be private by default. This means writing:

```
infixl 5 +
```

Only exposes the fixity at a file-level, to export the fixity for users the following is required:

```
public
infixl 5 +
```

This fixes problem (1).

[...]

PR #3011, opened 2023-07-11
Fixity access modifier

andrevidela:
# Description

This is part of #2998 

- Allow export modifiers in front of a fixity declaration
- Update the parsing error message for "public" instead of "public export" 
- import order does not influence fixity parsing anymore

      `CONTRIBUTORS.md`).


buzden:
I can see it from the code, but maybe it should be documented somewhere explicitly, that by default fixity declaration is exported, as before.

Z-snails:
As `public export` and `export` mean the same thing for fixity declarations, I don't think `public export` should be allowed. Not allowing it leads to a more consistent style, and there's no reason I can think of to prefer the longer way.
Alternatively this could be a warning (something like "`public` does not have any effect here, consider removing it")

andrevidela:
@Z-snails I think your suggestion goes against the goal of #2998 which is to enable more functionality without introducing a new bespoke export/module system for fixities. It already is consistent with the fact that you can `public export` a definition with no implementation and that will effectively do the same as `export` since the computation will be stuck.

Adding a warning could be an interesting feature but I'm not sure it's strictly necessary. Until now, we've always used warnings  when we're unable to detect an error but we suspect there is one. The difference between `public export` and `export` isn't a source of error. Regardless of what we decide, we should probably discuss this on the #2998 issue itself.

andrevidela:
@gallais all good to merge?

src/Idris/Resugar.idr
```diff
@@ -47,7 +47,7 @@ mkOp tm@(PApp fc (PApp _ (PRef opFC kn) x) y)
        -- that the fixity context contains the name `(++)`
        let rootName = UN (Basic (nameRoot raw))
        let asOp = POp fc opFC kn (unbracketApp x) (unbracketApp y)
-       if not (null (lookupName rootName (infixes syn)))
+       if not (null (lookupName rootName (fixities syn)))
```
gallais, line 50: What if we have a (private prefix) (export definition) in one module and a similar
(export infix) (export definition) in another? When resugaring aren't we going to
resugar the prefix op as an infix one here? That would be really quite confusing.

andrevidela: Good point, let me see if I can create a test case for that

andrevidela: I can confirm prefix `!!` in `!! a` is desugared as `a !!`. In the previous version, it was desugared as `(!!) a`. I'll work on a fix

andrevidela: didn't have the time to look at this until now but using `infixes` instead of `fixities` solved it

Issue #3016, opened 2023-07-15
Infer type  for local definitions inside let/where block


scarf005:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

Allow inferring type for definitions inside let/where block similar to global type inference for top level declarations.

[from discussion in idris server](https://discord.com/channels/827106007712661524/827106088343175180/1129643731813933138)

## Motivation

Some pain point of refactoring functions into small chunks with let/where blocks were having to manually declare types. With help from @Russoul i found out that holes could be also be used without any names `: ?` to explicitly infer types, however it'd be even better if the type inference could be done without explicit annotation.

### Examples

before:
```idris
-- global type inference, can be omitted
getLength : ? -> ?
-- inferred type is getLength : String -> Nat
getLength str = len where 
  -- local type inference, cannot be omitted
  len : ?
  -- inferred type is len : Nat
  len = length str
```

after:
```idris
getLength str = len where 
  len = length str
```

<details><summary>More examples</summary>

```idris
test str = len + totalLen
  where
    xs : ?
    xs = words str
    len : ?
    len = length xs
    totalLen : ?
    totalLen = sum . (map length) $ xs
```
```idris
test str = len + totalLen
  where
    xs = words str
    len = length xs
    totalLen = sum . (map length) $ xs
```

</details> 

### Technical implementation

As inferring types for local declaration is already possible (using explicit `: ?` notation), It may be achievable by implicitly marking local definitions as holes.

## Alternatives considered

Maybe introduce another syntax sugar like `?=` but i don't think this is the right solution.
```idris
getLength str = len where 
  len ?= length str -- same as len : ?; len = length str
```

## Conclusion

For small definitions holding intermediate computation (like `words xs`), being able to infer them would be immensely useful. As global type inference already works, I'd love to see it be done in local definitions too.

gallais:
Related: #1952

See the related issues mentioned there: being lax enough to allow
declaration-free definitions leads to poorer error messages.

andrevidela:
I wonder how much of this can be handled with better error reporting and how much of this needs to change.

For example, assume we have a `where`-clause `f` which does not have a type declaration. We could synthesize a hole for it as `f : ?`. But in addition to that, in any error reporting involving `f`, we could say 

[...]

Issue #2593, opened 2022-07-15
confusing error message for unannotated definition in `where` block attached to polymorphic function


rntz:
```idris
two: Nat
two = 2

bar: a -> a
bar x = x where
 wat = two
```

yields the error message:

```
     While processing right hand side of bar. While processing right hand side of wat. When unifying:
         Nat
     and:
         (0 a : Type) -> (x : a) -> ?_
     Mismatch between: Nat and (0 a : Type) -> (x : a) -> ?_.
     
     error-wtftype:6:8--6:11
      2 | two = 2
      3 | 
      4 | bar: a -> a
      5 | bar x = x where
      6 |  wat = two
                 ^^^
```

which is quite confusing; the type `(0 a : Type) -> (x : a) -> ?_` seems to have nothing to do with what I wrote, namely `two`.

rntz:
this may be related to #2592

Issue #2592, opened 2022-07-15
using `where` twice with the same name and no type annotations produces a name collision


rntz:
```idris
bar = 3
 where wat = 2
baz = 3
 where wat = 2
```

Produces the error:
```
     While processing right hand side of baz. Main.wat is already defined.
     
     error-double-where:4:8--4:15
      1 | bar = 3
      2 |  where wat = 2
      3 | baz = 3
      4 |  where wat = 2
                 ^^^^^^^
```

If I add a type annotation to either or both `wat`s the problem disappears. Adding annotations to `bar` and `baz` has no effect.

PR #3017, opened 2023-07-16
[ fix ] Ensure local defs with no claim are local 

### LINKED ISSUE #2782: let parameters within a function specified in a certain way leak into main
In the following code, `y` within `foo` becomes defined as `Main.y` and is accessible from the other function, `fee`:

https://gist.github.com/redfish64/975ab9f95fb508021ba2948781db59a0

```
foo : Integer
foo = 
  let z : Int
      z = 1
      
      y = 1
  in y
    

fee : Integer
fee = y
```

This only happens if I give `z` a type. If I comment out `z : Int` it errors apprpriately.

# Steps to Reproduce

Typecheck

# Expected Behavior

```
- + Errors (1)
 `-- ParamBug.idr line 12 col 6:
     While processing right hand side of fee. Undefined name y. 
     
     ParamBug:12:7--12:8
      08 |   in y
      09 |     
      10 | 
      11 | fee : Integer
      12 | fee = y
                 ^
```     

# Observed Behavior

OK!

dunhamsteve:
# Description

PR #1952 introduced top level aliases - definitions without a type declaration.  This had a number of side effects, including local function names being exposed at top level and strange type errors for local function definitions without type declarations.

It turns out that the root cause was the function names not being added to `nested`.  This PR addresses that by looking at `IDef` in addition to `IClaim` when determining the list of locally defined names.  It fixes #2782, #2592, and #2593, which are included in the test case. It also addresses a case raised on discord.

I believe it addresses #3016, but note that, as with top level declarations, the definitions have to have simple bind variables on the LHS. (A constraint imposed by the #1952 code.)

### DIFF
--- a/src/TTImp/TTImp.idr
+++ b/src/TTImp/TTImp.idr
@@ -13,6 +13,8 @@ import Data.List
 import Data.List1
 import Data.Maybe
 
+import Libraries.Data.SortedSet
+
 %default covering
 
 -- Information about names in nested blocks
@@ -772,7 +774,7 @@ export
 definedInBlock : Namespace -> -- namespace to resolve names
                  List ImpDecl -> List Name
 definedInBlock ns decls =
-    concatMap (defName ns) decls
+    SortedSet.toList $ foldl (defName ns) empty decls
   where
     getName : ImpTy -> Name
     getName (MkImpTy _ _ n _) = n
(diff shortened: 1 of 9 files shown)


src/TTImp/TTImp.idr
```diff
@@ -790,6 +790,7 @@ definedInBlock ns decls =
 
     defName : Namespace -> ImpDecl -> List Name
```
gallais, line 791: Should this be a `Set` to avoid too many duplicates?

dunhamsteve: Yeah, it should.  I got here via `Local.idr`, which calls `nub` on the result, but I see it is called from two other locations that do not do that.  I don't see `nubOrd` in the library, is `(SortedSet.toList . SortedSet.fromList) $` ok as a makeshift `nubOrd` or should I thread a `SortedSet` through the function?  (I could also call `nub`, but it's O(n^2).)

gallais: My instinct would be to thread a `Set` and then keep `definedInBlock`'s type unchanged so
that the call sites do not need to be modified (or only modified to remove calls to `nub`).

gallais (approved):

PR #3018, opened 2023-07-16
docs: typo in list difference documentation

scarf005:
# Description

https://github.com/idris-lang/Idris2/blob/6729fa8c89792c8d75b58daa86ad2e70c438120f/libs/base/Data/List.idr#L273-L287

A trivial typo fix for `\\` in `Data.List` where the example had a typo showing it as `//`

I guess probably not.


mattpolzin (approved):
Thanks! That typo has been bugging me for a while.

Issue #3022, opened 2023-07-19
Bootstrapped Idris2 fails to resolve paths when `IDRIS2_PREFIX` is manually set


scarf005:
# Summary

```
# ...logs
/home/scarf/repo/Idris2/bootstrap-build/bin/idris2 --build idris2.ipkg
Uncaught error: EmptyFC:Failed to resolve the dependencies for idris2app:
  required network any but no matching version is installed

make[1]: *** [Makefile:67: /home/scarf/repo/Idris2/build/exec/idris2] Error 1
make[1]: Leaving directory '/home/scarf/repo/Idris2'
make: *** [Makefile:249: bootstrap] Error 2
```

Idris2 bootstrapped from source (commit 23694c7e5af11c263d4dba7f40c550d9bc37509c) with custom environment variables cannot resolve dependency paths.

## environment

environment variables (followed from [official docs](https://idris2.readthedocs.io/en/latest/reference/envvars.html))

IDRIS2_PREFIX: /home/scarf/.local/share/idris2
PACK_DIR: /home/scarf/.local/share/pack
PATH: /home/scarf/.idris2/bin /home/scarf/.pack/bin # ...

scheme version: 9.5.9 (built from source with threads enabled)
OS: Kubuntu 23.04 x86_64
Shell: fish 3.6.0

```
$ SCHEME=scheme ./bootstrap-build/bin/idris2 --paths
+ Working Directory      :: "/home/scarf/repo/Idris2"
+ Source Directory       :: Nothing
+ Build Directory        :: "build"
+ Local Depend Directory :: "depends"
+ Output Directory       :: "build/exec"
+ Installation Prefix    :: "/home/scarf/.local/share/idris2"
+ Extra Directories      :: ["."]
+ Package Directories    :: []
+ CG Library Directories :: ["/home/scarf/.local/share/idris2/idris2-0.6.0/lib", "/home/scarf/repo/Idris2"]
+ Data Directories       :: ["/home/scarf/.local/share/idris2/idris2-0.6.0/support"]
```

## build log

[build.3.log](https://github.com/idris-lang/Idris2/files/12095912/build.3.log)

CodingCellist:
> PATH: /home/scarf/.local/share/idris2 /home/scarf/.local/share/pack # ...

Might it be that PATH should point to `[...]/bin` for both of those? I.e.

```
/home/scarf/.local/share/idris2/bin
/home/scarf/.local/share/pack/bin
```

?  : )

[...]

PR #3024, opened 2023-07-20
[ fix ] Only set `IDRIS2_PREFIX` if it is unset (fixes Issue 3022)

CodingCellist:
# Description

Discovered by scarf005 in #3022, if the `IDRIS2_PREFIX` envvar is set, bootstrapping fails. This happens regardless of `PREFIX` being set or not, and _only_ happens if `IDRIS2_PREFIX` is set.

Changing the syntax in the `Makefile` to use `?=` (assign-if-unset) instead of `:=` (assign) seems to fix the issue.

I also updated the `envvars.rst` docs, because they were outdated and didn't mention `PREFIX` being a thing (something which could've avoided the issue, but also would've left it buried for someone else to unearth). There were several changes, e.g. #2649, which weren't included.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,8 @@
 
 * Generated JavaScript files now include a shebang when using the Node.js backend
 * NodeJS now supports `popen`/`pclose` for the `Read` mode.
-* `getChar` is now supported on Node.js and `putChar` is supported on both JavaScript backends.
+* `getChar` is now supported on Node.js and `putChar` is supported on both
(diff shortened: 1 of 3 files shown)


dunhamsteve:
Previously `envvars.rst` described the environment variables consulted by Idris at runtime, and `PREFIX` is only used by the `Makefile`.  Should we list build-time environment variables separately?

gallais (approved):

PR #3029, opened 2023-07-23
Create CITATION.cff

pawelsawicz:
# Description

Hi team

I needed a citation for my MSc Project, and I found that GH supports citation files, so I thought I could create a PR for this so other people could easily get citations for the repo.

I would need your help with the content :)

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,14 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+authors:
+- family-names: "YOUR_NAME_HERE"
+  given-names: "YOUR_NAME_HERE"
+  orcid: "https://orcid.org/0000-0000-0000-0000"
+- family-names: "Lisa"
+  given-names: "Mona"
+  orcid: "https://orcid.org/0000-0000-0000-0000"
+title: "Idris2"
+version: 1.0.0
+doi: 10.5281/zenodo.1234
+date-released: 2023-07-23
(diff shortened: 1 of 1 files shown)


gallais:
The current standard practice AFAIK is to cite https://dblp.org/rec/conf/ecoop/Brady21.html

pawelsawicz:
Thanks, I cite that too, and I wanted to cite some specific fragments of code; hence I thought that I could raise a PR

jfdm:
Idris1's [citation file was very straightforward](https://github.com/idris-lang/Idris-dev/blob/master/CITATION.md) and pointed to the journal article.

Looking at the [documentation for citation files](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files), however, we can add a preferred citation that then points to the academic journal. Pointing to preferred citations is what [other projects do](https://github.com/haskell/haskell-language-server/blob/master/CITATION.cff) when there is an academic output to cite.

Hopefully the citation.cff can also produce a reference to the software itself.

The only issue is keeping the list of contributors up-to-date for the software citation. Perhaps defaulting to _The Idris Community_ with a note to see the contributors file would be sufficient?

pawelsawicz:
Thanks both for your input, this is very helpful. I will make adjustment accordingly

gallais:
I suppose we could create zotero entries for the releases.
I am concerned about making the production of releases even more work though
given we've gone more than a year without one, presumably because it's already
too much work.

Issue #3030, opened 2023-07-24
Projections out of a negative record are sometimes seen as total


91khr:
# Steps to Reproduce

Sorry for the long example and illustration, since I can't find something simpler to reproduce the error.

``` idris
module Test

namespace NS
  public export
  data Neg : Type where
    Nt : (Neg -> Neg) -> Neg

record RF infot where
  constructor MkRF
  info : infot
  op : Neg -> Neg

Functor RF where
  map f (MkRF info op) = MkRF (f info) op

Applicative RF where
  pure a = MkRF a id
  f <*> a = MkRF (f.info a.info) (f.op . a.op)

Monad RF where
  a >>= f = let res = f a.info in res
  join fs = ?fs
```

This would produce the following error:

``` text
Error: >>= is not total, possibly not terminating due to call to Test.RF

Test:21:1--23:16
 21 | Monad RF where
 22 |   a >>= f = let res = f a.info in res
 23 |   join fs = ?fs
```

Meanwhile, the following modifications would *not* produce such error:

- Define an interface with the same signature as Monad, and implement that interface for it:

  ``` idris
  interface Applicative m => SM m where
    (>>=) : m a -> (a -> m b) -> m b

  SM RF where
    a >>= f = let res = f a.info in res
  ```

  The `Applicative` constraint can be removed, and there's no error produced too.

- Define `Neg` in the same namespace as `RF`:

  ``` idris
  module Test

  data Neg : Type where
    Nt : (Neg -> Neg) -> Neg

  record RF infot where
    constructor MkRF
    info : infot
    op : Neg -> Neg

  -- The same as the full example above
  ```

- Remove the `let` expression:

  ``` idris
    Monad RF where
      a >>= f = f a.info
      join fs = ?fs
  ```

# Expected Behavior

Termination check either fails for all variants above, or successes for all of them.

gallais:
Did you mean to start your file with `%default total`? Without it,
definitions will not be rejected if they are not terminating.

> Define an interface with the same signature as Monad

Note that in `prelude` the definition of `Monad` *demands*
that `(>>=)` is total which you are not doing here.

91khr:
>Did you mean to start your file with %default total? Without it,
>definitions will not be rejected if they are not terminating.

No, this is the whole file.

>Note that in prelude the definition of Monad demands
>that (>>=) is total which you are not doing here.

Sorry for not noticing this. However, the other two variants may be expected to have the same behavior as the original one.

[...]

PR #3032, opened 2023-07-27
[ fix ] Don't include type of binder for let in addRefs

dunhamsteve:
This PR addresses an issue raised in #3030 where a `let` binder can cause false negatives in the totality checker. The root cause is that the type of the `let` binder is considered in the function's references.  If the `let` were substutited in, those references would not be there, so I believe they should not be counted.

Ideally, this PR would be double checked by someone who knows the totality checker.

### Analysis

In the following code, `test` and `test3` are currently rejected as not total, while `test2` is accepted as total. The reason given for failure is a reference to `Main.RF`. Only `Main.RF.(.info)` is referenced, and `RF` has another field with a negative occurrence of itself. 

```idris
record RF infot where
  constructor MkRF
  info : infot
  op : RF infot -> RF infot

total -- fails check
test : RF a -> (a -> RF b) -> RF b
test a f = let foo = f a.info in foo

total
test2 : RF a -> (a -> RF b) -> RF b
test2 a f = f a.info

total -- fails check
test3 : RF a -> (a -> RF b) -> RF b
test3 a f = let foo = a in f foo.info
```

For `test` and `test3`, we have:
```
Refers to: Main.RF.(.info), Main.RF
Refers to (runtime): Main.MkRF
```
And for `test2`:
```
Refers to: Main.RF.(.info)
Refers to (runtime): Main.MkRF
```

Upon investigating, I found that the `Main.RF` reference was being picked up from the type of the `let` binder. The fix drops the type of the `let` binder from consideration when collecting references.

      implementation, I have updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,9 @@
 
 * Constant folding of trivial let statements and `believe_me`.
 
+* Fixed an issue that caused total functions containing a `let` to be rejected
+  as not total.
+
 * Fixed a bug that caused holes to appear unexpectedly during quotation of dependent pairs.
 
 ### Library changes
--- a/src/Core/TT.idr
+++ b/src/Core/TT.idr
@@ -1843,7 +1843,7 @@ addRefs ua at ns (Meta fc n i xs)
     addRefsArgs ns [] = ns
     addRefsArgs ns (t :: ts) = addRefsArgs (addRefs ua at ns t) ts
 addRefs ua at ns (Bind fc x (Let _ c val ty) scope)
-    = addRefs ua at (addRefs ua at (addRefs ua at ns val) ty) scope
+    = addRefs ua at (addRefs ua at ns val) scope
 addRefs ua at ns (Bind fc x b scope)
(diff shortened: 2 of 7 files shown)


gallais:
I think the opposite is true:  `info` and `op` should always be considered
partial because they're projections out of a non-strictly positive type.
The issue is that the file is not declared as `%default total` and so `op`
is (incorrectly) not labelled as partial when `RF` is being elaborated.

E.g. the following is accepted

```
record RF infot where
  constructor MkRF
  info : infot
  op : RF infot -> RF infot

total
op' : RF i -> RF i -> RF i
op' = op
```

when adding `%default total` correctly rejects the definition.

gallais:
Worse, this can be exploited to prove false:

```idris
record Oops where
  constructor MkOops
  runOops : Not Oops

notOops : Not Oops
notOops x = runOops x x

oops : Oops
oops = MkOops notOops

total
boom : Void
boom = runOops oops oops
```

PR #3033, opened 2023-08-01
[ fix ] Fixes build broken in #3021

dunhamsteve:
Fixes the broken build


scarf005:
oh no, sorry i overlooked that. could you give a pointer on why did the CI pipe failed to find issues on #3021 ?

dunhamsteve:
One issue was that it the import of Data.Integral failed - this was something I ran into myself recently. When a module is added to a library/package, it needs to be declared in the corresponding `.ipkg` file or it will not be installed with the package.

The second issue was that the `Integral` implementation for `Nat` was not found. I addressed this by adding the appropriate import to the test case.  I'm not sure why that test case worked locally for you.  Maybe you had a local commit that didn't get pushed up to github?  

I let the CI run through on my account long enough to get through the tests before submitting the PR, so this should address the issue.

dunhamsteve:
If I'm reading the github screens right, I don't think CI was run on that PR before merging.  Probably waiting for approval to run.

scarf005:
I see. that would explain the build breakage.

PR #3034, opened 2023-08-01
[ elab ] Make `%macro`-function be callable without the `ElabReflection` extension

buzden:
# Description

Macro-functions are a great way of hiding some complex thing in a library without showing this to a user. A pretty example was shown as a motivating example in PR #2930.

But since currently call to a macro function, generally, is just a `%runElab` of an appropriate expression, each use of such function means to users that they need to turn on the language extension flag `ElabReflection`, despite they may not feel why.

I mean, imagine a library that does some hard stuff in a macro function (like enabling some syntax in PR mentioned above, or some DSL-y thing), and user just imports this library and uses its functions and gets "`%language ElabReflection not enabled`" all the time. I understand this requirement in explicit `%runElab` expressions and declarations -- here a user explicitly has chosen to run an elaboration script and we hope they know what they are doing, thus, this is their responsibility to say explicitly that they want this extension.

But `%macro` function seem to fill the different use case, when we want to hide this machinery from the user, i.e. to put all the responsibility on the effects to the library's author. But the current setting makes such library's user still to enable an extension, which, probably, they don't understand. The current setting also unables library authors to change existing function definition to a `%macro` without additional requirements from the users to enable the extension.

So, my suggestion is to enable *usage* of `%macro`-functions without the language extension, with no change in the behaviour of explicit `%runElab` calls. This is fully a backward-compatible change. This PR does this, among with couple of tiny textual improvements.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,9 @@
   [#1066](https://github.com/idris-lang/idris2/issues/1066)).
 * `%hide` directives can now hide conflicting fixities from other modules.
 * Fixity declarations can now be kept private with export modifiers.
-* New fromTTImp, fromName, and fromDecls names for custom TTImp, Name, and Decls
-  literals.
+* New `fromTTImp`, `fromName`, and `fromDecls` names for custom `TTImp`,
+  `Name`, and `Decls` literals.
+* Call to `%macro`-functions do not require the `ElabReflection` extension.
 
 ### REPL changes
 
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -25,11 +25,11 @@ import public Libraries.Utils.Binary
 %default covering
 
 ||| TTC files can only be compatible if the version number is the same
-||| Update with the current date in YYYYMMDD format, or bump the auxiliary
+||| Update with the current date in YYYY_MM_DD_00 format, or bump the auxiliary
 ||| version number if you're changing the version more than once in the same day.
 export
 ttcVersion : Int
-ttcVersion = 20230829 * 100 + 0
+ttcVersion = 2023_09_04_00
 
 export
 checkTTCVersion : String -> Int -> Int -> Core ()
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -372,7 +372,7 @@ mutual
   desugarB side ps (PUnquote fc tm)
       = pure $ IUnquote fc !(desugarB side ps tm)
   desugarB side ps (PRunElab fc tm)
-      = pure $ IRunElab fc !(desugarB side ps tm)
+      = pure $ IRunElab fc True !(desugarB side ps tm)
   desugarB side ps (PHole fc br holename)
       = do when br $ update Syn { bracketholes $= ((UN (Basic holename)) ::) }
            pure $ IHole fc holename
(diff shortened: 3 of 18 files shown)


GunpowderGuy:
So to reiterate, this would allow modules without the elab extension turned on, to depend on libraries that use macro functions internally, without exposing that implementation detail to the end user?

buzden:
> So to reiterate, this would allow modules without the elab extension turned on, to depend on libraries that use macro functions internally, if they dont expose that implementation detail to the end user?

It depends on what you are calling the implementation detail. E.g., if there is a library that internally uses some macro function, but this is completely not exposed to the end user, it is okay now. E.g.:

```idris
module Lib1

import public Language.Reflection

export %macro
deriveNiceThing : Params -> Elab NiceThing
...
```

```idris
module Lib2

import Lib1

%language ElabReflection

export
aNiceThing : NiceThing
aNiceThing = deriveNiceThing niceThingParams
```

```idris
module SomeApp

import Lib2

usageOfNiceThing : Nat
usageOfNiceThing = someComputationOf aNiceThing
```

In this case a library `Lib2` uses internally a macro function `deriveNiceThing`, but the end-used of the library `Lib2` is not aware of it. And this type of code would compile by the current compiler.

But if we look at the `Lib2` as the end-user of the library `Lib1`, then we'll see that `Lib1` exposes a `%macro`-function, and `Lib2` which uses this function as if it is a regular function of type `Params -> NiceThing`, needs to declare `%language Elab.Reflection`, which I propose to not to require to be declared.

GunpowderGuy:
"Lib2 which uses this function as if it is a regular function of type Params -> NiceThing, needs to declare %language Elab.Reflection"
So macros only allow things that could theoretically be done by functions with normal ( type level ) metaprogramming. However macros may be easier to use or the compiler may need to be expanded before said usage can be represented?

buzden:
> So macros only allow things that could theoretically be done by functions with normal ( type level ) metaprogramming?

No, `%macro` functions in Idris are just functions which when you call, are prepended with `%runElab` with respect to arguments. For example, if function `fun` is declared as `%macro`, the call `fun a b c` is elaborated to `%runElab fun' a b c`, or to `(%runElab fun' a b) c`, or to `(%runElab fun' a) b c`, or to `(%runElab fun') a b c`, whichever typechecks first. I used `fun'` in the example above to denote an access to `fun` __as if there was no `%macro` in it__, i.e., elaboration script of the function `fun` itself.

So, answering shortly, `%macro` functions have exactly the same abilities, as usual elaboration script, but the fact that there is an elaboration script underlying, is hidden. That makes then easier to use by the end user, especially when those macro functions are used as part of nice syntax or eDSL.

GunpowderGuy:
While I support the idea of making macros more accessible by detaching them from the ElabReflection extension, I think it's crucial that they still require a specific extension to activate. Metaprogramming features have a history of being revised or updated, either for better design or to prevent unintended leakage of internal language details.

Take Rust as an example. It started with compiler plugins and pattern-matching macros. Later, more advanced procedural macros were introduced. These were initially intended to only replace compiler plugins but ended up largely replacing the first macro system as well due to its limitations and flaws. Eventually rust attempted to replace its pattern-matching macros with a new system, but this turned out to be a much bigger undertaking than anticipated, leaving the situation unresolved.

This situation would have been even more complicated if those features hadnt been feature gated, Idris 2 should keep current macros gated as well, potentially under a dedicated extension. This approach minimizes the risk of complications as the language evolves

gallais:
Probably worth talking about it during one of the weekly meetings to get @edwinb's opinion.

GunpowderGuy:
Looks like next week's meeting was cancelled. We should coordinate on discord whether we can all discuss this the week after that

gallais (approved):

Issue #3036, opened 2023-08-03
Misapplication of `Delay` causes indentation error instead of type mismatch


CodingCellist:
If `Delay` is applied to multiple arguments (an easy mistake to make when constructing codata), the parser reports an indentation error rather than a type mismatch (like when you've accidentally used a type-name instead of its constructor) or otherwise more appropriate error.

## Steps to Reproduce

Self-contained example:

```idris
module DelayParse

public export
data MyStream : Type where
  (::) :  (anA : a)
       -> (nextA : a -> a)
       -> Inf MyStream
       -> MyStream

public export
go : a -> (a -> a) -> MyStream
go initA fn =
  MyStream.(::) initA fn (Delay (fn initA) fn)
  -- the above should be `Delay (go (fn initA) fn)`
```

## Expected Behavior

An error along the lines of "`Delay` expects a `MyStream` not an [insert cryptic `?ty1 -> ?ty2 -> Type` here]".

## Observed Behavior

Indentation error:

```
Error: Not the end of a block entry, check indentation.

DelayParse:13:24--13:25
 09 |
 10 | public export
 11 | go : a -> (a -> a) -> MyStream
 12 | go bar fn =
 13 |   MyStream.(::) bar fn (Delay (bar fn) fn)
                             ^
```

gallais:
It's a parse error because `Delay` is not a constructor like e.g. `S`, it's a language
construct that expects exactly one argument.
We could perhaps tweak the parser to give back a better error message.

PR #3037, opened 2023-08-03
[ error ] Improve error messages for Delay &co

dunhamsteve:
# Description

This PR improves error messages in `Delay` statements to address #3036. It adds a check for an additional `simpleExpr` after the statement. It also introduces a `mustFailBecause` helper function that will fail with a message if a given parser succeeds at the current point.

Prior to this change, the following code:
```idris
module DelayParse

public export
data MyStream : Type where
  (::) :  (anA : a)
       -> (nextA : a -> a)
       -> Inf MyStream
       -> MyStream

public export
go : a -> (a -> a) -> MyStream
go initA fn =
  MyStream.(::) initA fn (Delay (fn initA) fn)
```
would fail with:
```
1/1: Building tests.idris2.perror029.DelayParse (tests/idris2/perror029/DelayParse.idr)
Error: Not the end of a block entry, check indentation.

tests.idris2.perror029.DelayParse:13:26--13:27
 09 |
 10 | public export
 11 | go : a -> (a -> a) -> MyStream
 12 | go initA fn =
 13 |   MyStream.(::) initA fn (Delay (fn initA) fn)
                               ^
```
and now we get:
```
1/1: Building DelayParse (DelayParse.idr)
Error: Delay only takes one argument.

DelayParse:13:44--13:46
 09 | 
 10 | public export
 11 | go : a -> (a -> a) -> MyStream
 12 | go initA fn =
 13 |   MyStream.(::) initA fn (Delay (fn initA) fn)
                                                 ^^
```

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -946,16 +946,20 @@ mutual
   lazy : OriginDesc -> IndentInfo -> Rule PTerm
   lazy fname indents
       = do tm <- bounds (decorate fname Typ (exactIdent "Lazy")
-                         *> simpleExpr fname indents)
+                         *> simpleExpr fname indents
+                         <* mustFailBecause "Lazy only takes one argument" (continue indents >> simpleExpr fname indents))
            pure (PDelayed (boundToFC fname tm) LLazy tm.val)
     <|> do tm <- bounds (decorate fname Typ (exactIdent "Inf")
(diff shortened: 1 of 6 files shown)


gallais:
Sorry about the missclick. I wanted to say that I suspected this PR
would arrive when I mentioned the fact the parser was the problem. :)

PR #3042, opened 2023-08-12
[ ttc ] Relax existing second-precision time requirements

buzden:
Currently source modification time must be strictly less than TTC file modification time. Since, time is measured in seconds, it can be read that source modification time must be at least one second less than the TTC file modification time. In some obscure situations this gives to extra rebuilds.

I propose to relax this requirement and allow source modification time to be in the same second as TTC file modification time. I couldn't imagine situation where this can break something comparing to the current behaviour. WDYT?

### DIFF
--- a/src/Idris/ProcessIdr.idr
+++ b/src/Idris/ProcessIdr.idr
@@ -258,9 +258,9 @@ isTTCOutdated ttcFile sourceFiles
        srcTimes <- traverse modTime sourceFiles
        log "module.hash" 20 $
          unlines $
-           "Checking whether source code mod times are newer than \{show ttcTime}; src times:"
+           "Checking whether source code mod times are not older than \{show ttcTime}; src times:"
            :: zipWith (\ src, tm => "\{src} : \{show tm}") sourceFiles srcTimes
-       pure $ any (>= ttcTime) srcTimes
+       pure $ any (> ttcTime) srcTimes
 
 ||| If the source files hash hasn't changed
 export
--- a/tests/Main.idr
+++ b/tests/Main.idr
@@ -43,6 +43,7 @@ idrisTestsBasic = MkTestPool "Fundamental language features" [] Nothing
        "basic056", "basic057", "basic058", "basic059", "basic060",
        "basic061", "basic062", "basic063", "basic064", "basic065",
        "basic066", "basic067", "basic068", "basic069", "basic070",
+       "basic071",
        "idiom001",
        "dotted001",
        "rewrite001",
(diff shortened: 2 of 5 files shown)


mattpolzin:
Can you say more about the obscure situations where extra rebuilds occur? My fear is that it's just as likely that there are cases where we do want to rebuild when the times are the same and it is safer to err on the side of rebuilding.

dunhamsteve:
This seems a little dangerous to me as in some case it won’t rebuild source files that have changed after ttc files. This is happening in the `interactive043` test.

Are the extra builds happening within a single idris invocation or in separate invocations?

AlgebraicWolf:
Since `stat` returns modification time up to nanoseconds, I guess it would be safer to increase the precision rather than relax the comparison. I've implemented a proof-of-concept at https://github.com/AlgebraicWolf/Idris2/commit/c8e2b7c6141e93c4783ab6640a30c69e6337b6ff

buzden:
> This seems a little dangerous to me as in some case it won’t rebuild source files that have changed after ttc files. This is happening in the `interactive043` test.

Right, I haven't thought of this case when I was proposing this.

> My fear is that it's just as likely that there are cases where we do want to rebuild when the times are the same and it is safer to err on the side of rebuilding.

As it turned out, this looks so.

> Can you say more about the obscure situations where extra rebuilds occur?

Yeah, we have different parts of a big project build, when different modules are built on different machines and/or built modules are stored in a repo or intermediate CI storage because they take a too long time to build (probably, you've heard of my hours-long-running elaboration scripts). So, when we restore the state of built modules, usually their original build time is either lost, or inconsistent with modification time of source files. Thus we tried solution of `touch`ing those modules, and it worked for some time. Then, we started to see some flacky behaviour of rebuilding already built modules, which for some reason became stable in rebuilding always. Addition of waiting for a single second completely fixes this behaviour, after which we saw that the current check has a precision of a second, and thus when correction of modification time and building is happening in a script, this leads to those rebuilds. Thus, the original idea of the PR has born.

Now I can see that this was not so good idea, because there are cases that I didn't originally thought of.

> Since `stat` returns modification time up to nanoseconds, I guess it would be safer to increase the precision rather than relax the comparison.

Yes, I guess, this is strictly better solution than mine. It should not beat safety comparing to the current approach, even if the underlying OS does not support nanoseconds precision and always returns zeros, since in this case behaviour turns to the current one.

buzden:
I'll close this PR, then, since this is not a good solution, and there seems to be better one anyway

PR #3044, opened 2023-08-14
[ fix ] Use freshly built `libidris2_support` in compiler

AlgebraicWolf:
Currently, when the compiler executable is built, `libidris2_support.so` from the previous compiler version is shipped with the new one. This makes it impossible to add new foreign functions to the compiler without an intermediate step adding new functions to `libidris2_support.so`  but not using them yet. Additionally, it makes it impossible for commits that use new foreign functions to pass CI, because the compiler is first built with the last release version, which does not contain the new foreign functions in its support files.

This PR addresses the issue by replacing the `libidris2_support.so` file with a freshly built one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,12 +59,13 @@ export SCHEME
 
 .PHONY: all idris2-exec libdocs testenv testenv-clean support clean-support clean FORCE
 
-all: support ${TARGET} libs
+all: ${TARGET} libs
 
 idris2-exec: ${TARGET}
 
-${TARGET}: src/IdrisPaths.idr
+${TARGET}: support src/IdrisPaths.idr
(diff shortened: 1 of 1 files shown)


gallais:
This looks like a dangerous step towards breaking bootstrapping.

AlgebraicWolf:
I don't think I understand why that could break bootstrapping. Bootstrapping had already used the freshly built support for compiler shipped with the repository, and this PR doesn't change that.

Moreover, since freshly built support was used in bootstrapping, this fresh support was shipped with the compiler executable. Hence, this change would make the result of compiling Idris2 with previous Idris2 installation more consistent with the result of bootstrapping.

AlgebraicWolf:
Closing this as it has effectively been merged in alongside #3046

PR #3047, opened 2023-08-15
[ prelude ] Add some lacking `%tcinline`s

buzden:
Without this you may get weird wrong totality checker errors when using simple functions like `<$>`, you can see an example in the added test.

### DIFF
--- a/libs/prelude/Prelude/Interfaces.idr
+++ b/libs/prelude/Prelude/Interfaces.idr
@@ -87,7 +87,7 @@ interface Functor f where
 ||| in a parameterised type.
 ||| @ f the parameterised type
 ||| @ func the function to apply
-%inline public export
+%inline %tcinline public export
 (<$>) : Functor f => (func : a -> b) -> f a -> f b
 (<$>) func x = map func x
 
@@ -95,22 +95,22 @@ interface Functor f where
 ||| everything of type 'a' in a parameterised type.
 ||| @ f the parameterised type
 ||| @ func the function to apply
-%inline public export
+%inline %tcinline public export
 (<&>) : Functor f => f a -> (func : a -> b) -> f b
 (<&>) x func = map func x
 
 ||| Run something for effects, replacing the return value with a given parameter.
-%inline public export
+%inline %tcinline public export
 (<$) : Functor f => b -> f a -> f b
 (<$) b = map (const b)
 
 ||| Flipped version of `<$`.
-%inline public export
+%inline %tcinline public export
 ($>) : Functor f => f a -> b -> f b
 ($>) fa b = map (const b) fa
 
 ||| Run something for effects, throwing away the return value.
-%inline public export
+%inline %tcinline public export
 ignore : Functor f => f a -> f ()
 ignore = map (const ())
 
@@ -156,7 +156,7 @@ interface Bifunctor f where
   mapSnd : (b -> d) -> f a b -> f a d
   mapSnd = bimap id
 
-public export
+public export %tcinline
 mapHom : Bifunctor f => (a -> b) -> f a a -> f b b
 mapHom f = bimap f f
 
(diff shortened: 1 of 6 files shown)


libs/prelude/Prelude/Interfaces.idr
```diff
@@ -310,13 +310,13 @@ interface Foldable t where
   foldMap f = foldr ((<+>) . f) neutral
 
 ||| Combine each element of a structure into a monoid.
-%inline public export
+%inline %tcinline public export
```
gallais, line 313: I can see tcinline making a difference on simple functions, does it actually
make a difference on these (potentially) recursive functions?

buzden: Criteria was a function that is implemented through an interface function, so that if one uses such inlined function in an interface implementation, they won't get an unexpected termination checker error.

In particular case of `concat`, I've added and pushed an example of when code typechecks with added `%tcinline`s and doesn't without. It's arguable whether it is realistic to meet something like this in real code, but I think it's possible. Anyway, as far as I understand, it is safe in this particular case anyway.

gallais: Thanks for the explanation!

dunhamsteve:
In case anyone is wondering why it is happening without the `%tcinline` -  findSC sees:
```
Prelude.Interfaces.MkFunctor [Main.X, \{0 b : Type} => \{0 a : Type} => \(func : ?Main.{_:849}_[b[1], a[0]]) => \({arg:8747} : (Main.X a[1])) => (Main.map b[3] a[2] func[1] {arg:8747}[0])]
```
It finds is a call to `Main.map` with unknown arguments and flags the loop. With the `<$>` inlined, I think this is reduced out to a function call with known arguments.

buzden:
> It finds is a call to `Main.map` with unknown arguments and flags the loop

Look like the current behaviour is strange. I thought it just doesn't look inside unless it's `%tcinline`d, but it looks inside, but with strange consequences.

dunhamsteve:
You probably know most of this already, but my understanding of the situation:

When it's checking totality, it goes through each of the functions on the graph and calculates the size changes. For example, if it is looking at `map`, then for each function `map` calls, it will emit a pair of the function name and a list of size changes of the arguments passed into the function. So it would say which argument of `map` each the argument to `<$>` came from (or `Nothing`) and whether it is smaller than value that it came from.

When it looks at `map`, the metas are already solved, and it does an additional normalization with a flag to inline anything marked `%tcinline` (in `CallGraph.idr:findCalls`).

Below I'm using a simplified test case (to remove some extra variables):
```idris
%default total
data X : Type -> Type where
  R : X b -> X c -> X c

Functor X where
  map f (R x y) = R x (f <$> y)
```

When processing `Main.map`, it doesn't look inside `<$>` (it would later process `<$>` separately and then analyze the graph for loops).  The input to `findCalls` looks like (`lhs` and `rhs_in`in `findCalls`):
```idris
Main.map b' a f (R a {b} x y) =
  R b' {b} x ((<$>) b' a X "Main.Functor implementation at..." f y)
```
After normaization with tcinline turned on (`lhs` and `rhs`):
```idris
Main.map b' a f (R a {b} x y) =
  R b' {b} x ((<$>) b' a X (MkFunctor X \{0 b'' : Type} -> \{0 a' : Type} => \(func : _) => \(arg : X a') => Main.map b'' a' func arg) f y)
```
So the definition of the `Functor X` instance is marked `%tcinline`.

Now it goes into the mechanism at `findSC`, which drills down under the lambdas and sees a call to `Main.map` where none of the arguments line up with the left hand side, and flags it as not total.

If you tcline the `<$>`, the normalization substitutes it in and β-reduces the lambdas out, leaving a call to `Main.map` with the arguments from the LHS.

I considered tweaking `findSC` to not look under lambdas (or pi's) - and it passes all of the tests - but I constructed a case where it fails to mark a non-total function as total:
```idris
%default total
data Foo = MkFoo Nat (Nat -> Nat)

bar : Foo -> Nat
bar (MkFoo k f) = f (k + 2)

foo, baz : Nat -> Nat

baz k = bar (MkFoo k (\x => foo x))

foo (S k) = baz k
foo Z = Z
```

dunhamsteve:
One further note: I think the implementation is marked `TcInline` only long enough to process the declaration. This is done near the bottom of `elabImplementation` in `Idris/Elab/Implementation.idr`. (It sets the flag, processes the decl and then clears it.)

PR #3053, opened 2023-08-20
implement DESTDIR support for distros

juhp:
# Description
Prefix `${PREFIX}` with `${DESTDIR}` allowing installation into a staging directory for packaging Idris2.

Also take DESTDIR into account with `idris2 --install` etc.

      `CONTRIBUTORS.md`).

I tested this a little on Fedora Linux.

When DESTDIR is not set, installation should behave exactly as now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -237,6 +237,8 @@
 * Fixes an error in the `Makefile` where setting `IDRIS2_PREFIX` caused
   bootstrapping to fail.
 * Updates the docs for `envvars` to match the changes introduced in #2649.
+* Both `make install` and `idris2 --install...` now respect `DESTDIR` which
+  can be set to install into a staging directory for distro packaging.
 
 ## v0.6.0
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -30,6 +30,7 @@ Hiroki Hattori
 Ilya Rezvov
 Jan de Muijnck-Hughes
 Jeetu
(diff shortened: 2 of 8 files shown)


juhp:
I would appreciate if someone could just have a glance at my simple idris [code](https://github.com/idris-lang/Idris2/pull/3053/files#diff-279507ff32414ec2812f1582dde07581ec1dd5617935867852caa3591c9eb403)

gallais (commented):
Should PREFIX be computed from DESTDIR rather than prepending DESTDIR everywhere (and risking to forget it in places)?

mattpolzin:
> Should PREFIX be computed from DESTDIR 

That might be more confusing than not since `DESTDIR` is documented as both "only applied in install steps" and "never showing up in built files" (like IdrisPaths.idr). So, for example, we would need to be careful to track where `PREFIX` gets used very carefully and also not redefine it prior to setting `IDRIS2_PREFIX` based on it. 

OTOH, I think I'd be ok with either the current approach or a new variable like `INSTALL_PREFIX`. Would still perhaps benefit from explicit comment that such a variable is only to be used to offset installation and not to impact any build operations.

juhp:
Right, this patch tries to do the minimal simplest thing, so that it should be obvious it does no harm.
DESTDIR support has been missing for a long time.
Perhaps further down the road it could be integrated more deeply.

juhp:
Thank you @gallais for reviewing :yay:

PR #3060, opened 2023-08-28
Refactor S-Exp protocols to depend on fewer Idris modules

ohad:
# Description

Remove dependency on a lot of the compiler source-code.

This means specialising some of the S-Expr parser from the general Idris2 source-code parsing code to a separate copy for the S-Expr parser.

It's difficult to share source-code between the two parsers because the S-exp parser doesn't need the full token structure an Idris2 source-code requires.

- [X ] If this is a fix, user-facing change, a compiler change, or a new paper
      `CONTRIBUTORS.md`).


idris2protocols.ipkg
buzden, line 1: Doesn't this renaming break existing LSP? I mean, when there are several `ipkg` files, LSP looks for any one. Currently there were `idris2.ipkg` and `idris2api.ipkg`, but they have almost equal set of modules, thus didn't matter which `ipkg` file was found by LSP. With this renaming now LSP can find this `ipkg` first, thus it won't show correctly modules that are not listed in this file.

buzden: I was bitten with this behaviour in my projects where I had `docs.ipkg` for documentation which didn't list any modules from main source of the lib. As a solution, I moved this `ipkg` up into a dir, so that this `ipkg` file can't be found by `LSP` when looking from inside `src` dir. Maybe, this file could be moved from the root too to ensure old LSP behaviour?

ohad: Ah, thanks for flagging it.
This is really a bug in LSP, but of course I don't want to break it!
Do you know what would be involved in fixing the bug in LSP?
If it's not too complicated, I could delay the creation of the `.ipkg` file for some time until the LSP bug is fixed, but I don't think I should delay it indefinitely.

buzden: > Do you know what would be involved in fixing the bug in LSP?

It was discussed somewhere a lot and I remember that main problem: single module can belong to several `ipkg` files and may compile and not compile depending on the choice of particular `ipkg`. Also, functionality like "jump to definition" and etc. can also depend on the choice of particular `ipkg`. Determining correct `ipkg` can be difficult, since some `ipkg`s (like `idris2.ipkg`) does not list used modules at all, thus we cannot easily determine which `ipkg` to use just by looking into them.

ohad: Usually the safest way to deal with ambiguity isn't to guess, but to provide the user with a way to specify the correct choice.
I think the same applies here --- users should be able to specify which .ipkg to use for a given session.
Is implementing something like this difficult for LSP?

gallais: The obvious (to me) solution is to ask the user which of the `.ipkg` files they want to use and potentially remember that choice in a hidden file somewhere in the directory hierarchy.

buzden: This all sounds reasonable, but I'm not competent enough to judge if it's difficult for LSP server

mattpolzin: Just speaking strictly to feasibility, I know that LSP servers do sometimes stash their own information in hidden files in the current project directory; I don't know if this is part of the specification or even best practice, but I've seen LSes do it.

gallais (approved):

idris2protocols.ipkg
ohad: I moved it to a sub-directory (`ipkg`), but we ought to resolve this issue one way or another. I'll open an issue

PR #3062, opened 2023-08-29
[ new ] function options for case blocks

gallais:
We can now use function options on the (anonymous) functions introduced by a `case` block.
In particular this means we can specify that a `case` should be inlined during code-generation.


src/TTImp/TTImp/Traversals.idr
```diff
@@ -94,8 +94,8 @@ parameters (f : RawImp' nm -> RawImp' nm)
     = f $ ILam fc rig (mapPiInfo pinfo) x (mapTTImp argTy) (mapTTImp lamTy)
   mapTTImp (ILet fc lhsFC rig n nTy nVal scope)
     = f $ ILet fc lhsFC rig n (mapTTImp nTy) (mapTTImp nVal) (mapTTImp scope)
-  mapTTImp (ICase fc t ty cls)
-    = f $ ICase fc (mapTTImp t) (mapTTImp ty) (assert_total $ map mapImpClause cls)
+  mapTTImp (ICase fc opts t ty cls)
+    = f $ ICase fc opts (mapTTImp t) (mapTTImp ty) (assert_total $ map mapImpClause cls)
```
buzden, line 98: Shouldn't `libs/base/Language/Reflection/TTImp.idr` be changed accordingly?

gallais: Ah, yes. I always forget about that. Thanks!

Issue #1236, opened 2021-03-26
Access modifiers of definitions overwrite forward definitions


ohad:
# Steps to Reproduce
```idris
$ cat Mutual.idr
module Mutual
private
data Foo : Type 

public export
data Foo : Type where
------------------------------
$ cat Import.idr
import Mutual

x : Foo

```
# Expected Behavior
Type-checking `Mutual.idr` should report an error because the access modifiers are incompatible.

# Observed Behavior

The latest definition overwrites the access modifier, and `Import.idr` type-checks fine.

gallais:
I'm going to go ahead and tag this as good first issue because I assume it's something silly
like overwriting the existing value with a record update instead of making sure it's not set yet.

gallais:
If only we had a gadget that allowed us to express that something is the default
without conflating it with the default value it takes. :thinking: 

http://gallais.github.io/blog/ceci-pas-default

gallais:
Yeah I think it'd be nice to complain if there are conflicting declarations which means changing
the merge strategy to take into account lack of value vs. value that happens to equal the default.

emdash:
Is this still an issue? It seems like we get correctly warned now...

PR #3063, opened 2023-08-29
Add a warning for incompatible visibilities on forward decls and definitions.

Adowrath:
# Description

This addresses #1236 by introducing a warning when a forward declaration and actual definition of a `data` or `record` type have differing visibility modifiers, unless the definition has none at all - thus it allows code like the following:

```idris
public export
data Foo : Type

data Foo : Type where
  MkFoo : Foo
```

This makes `Foo` be publicly exported, yet either `export` or `private` would be a warning as suggested by @andrevidela so it can be turned into an error after an upcoming release.

Current behaviour is to take the maximum of the two visibility modifiers, so this is technically a breaking change (though taking the max felt like a buggy behaviour). 

The special-casing of an absent visibility modifier required the addition of `Libraries.Data.WithDefault`, adapted from @gallais's [post on explicit defaults](https://gallais.github.io/blog/ceci-pas-default). I had to therefore add to the TTC format since this field is stored in the `IData` and `IRecord` constructors, and also modiy the TTImp datatypes used in `Reflection`, making this truly a breaking change.

      `CONTRIBUTORS.md`).


src/TTImp/ProcessData.idr
```diff
@@ -462,17 +463,25 @@ processData {vars} eopts nest env fc vis mbtot (MkImpData dfc n_in mty_raw opts
...
+                    Just fullty => pure ([], collapseDefault def_vis, fullty)
+                  Just ndef => do
+                    vis <- the (Core Visibility) $ case def_vis of
+                      Default => pure ndef.visibility
+                      Value newVis =>
+                        if newVis == ndef.visibility
+                        then pure newVis
+                        -- TODO : In a later release, replace this with an error.
+                        else do recordWarning (IncompatibleVisibility fc ndef.visibility newVis n)
+                                pure (max ndef.visibility newVis)
```
gallais, line 483: I don't think that's the correct logic here and that's why you had to add some `public export`s to
some of the `libs/papers` modules.

What you want is something like the following that attempts to merge two withdefaults and:

* Fails if they're incompatible values
* Happily overwrites any `Default

```
mergeValues : Eq a => {def : a} -> WithDefault a def -> WithDefault a def ->
			Either (a, a) (WithDefault a def)
mergeValues (Value x) (Value y) = if x == y then Left (x, y) else Right (Value x)
mergeValues (Value x) Default = Right (Value x)
mergeValues Default w = Right w
```

Once you have that you can check whether `def_vis` and the `ndef`'s withdefault are compatible.

src/TTImp/ProcessData.idr
```diff
@@ -424,7 +425,7 @@ processData {vars} eopts nest env fc vis mbtot (MkImpLater dfc n_in ty_raw)
          arity <- getArity defs [] fullty
 
          -- Add the type constructor as a placeholder
-         tidx <- addDef n (newDef fc n linear vars fullty vis
+         tidx <- addDef n (newDef fc n linear vars fullty (collapseDefault def_vis)
```
gallais, line 428: I think that collapses too early and you're now forgetting what the visibility setting was.

gallais (changes requested):
Nice. I think we're close!

src/TTImp/ProcessData.idr
Adowrath: So you mean to allow
```idris
public export
data Foo : Type
data Foo : Type where
```
and
```idris
data Foo : Type
public export
data Foo : Type where
```
to both mean the same? In that case, I can add that in tomorrow, will just result in a bigger overall change once I modify `GlobalDef` as well.

gallais: Yes I think so. It's indeed a bit more invasive but gives users more freedom.

Adowrath: Added! I left `IClaim`s alone/not carrying the default-info since we don't have a "IClaimLater/IClaimImp" distinction like with records and data declarations.

src/Compiler/Opts/InlineHeuristics.idr
```diff
@@ -55,7 +56,7 @@ inlineHeuristics fn = do
     -- intervention by this function.
     -- We could lift this public restriction if we checked that the source def was _either
     -- public OR the destination was the same module as the source_.
-    let Public = gdef.visibility
+    let Value Public = gdef.visibility
```
gallais, line 59: I don't think you should ever explicitly match on `Value` as that breaks the abstraction
that we should default to the default value.

I think it makes sense to have `WithDefault`  just `export`ed (rather than `public export`ed)
and a smart constructor `value` defined to prevent these kind of matches.

src/Core/Context.idr
```diff
@@ -1118,13 +1121,13 @@ getFieldNames ctxt recNS
 -- Find similar looking names in the context
 export
 getSimilarNames : {auto c : Ref Ctxt Defs} ->
-                   Name -> Core (Maybe (String, List (Name, Visibility, Nat)))
+                   Name -> Core (Maybe (String, List (Name, WithDefault Visibility Private, Nat)))
```
gallais, line 1124: I think these can be collapsed eagerly. We will never care about anything but the actualy visibility here.

src/Core/Context.idr
```diff
@@ -1659,7 +1664,7 @@ getSizeChange loc n
 
 export
 setVisibility : {auto c : Ref Ctxt Defs} ->
-                FC -> Name -> Visibility -> Core ()
+                FC -> Name -> WithDefault Visibility Private -> Core ()
```
gallais, line 1667: I think this should still be ` Visibility`  as you are not going to be explicitly setting `Default`, are you?

src/Compiler/Opts/InlineHeuristics.idr
Adowrath: Makes sense, yeah. I'll add `value : a -> WithDefault a def` and `defaultValue : WithDefault a def`.

Adowrath:
@gallais Encapsulated `WithDefault` with the two smart constructors, as well as adding a `maybe`-like catamorphism in `onWithDefault` for when you still want to reduce the two cases differently - mainly used in the `TTC` and `Reflect` implementations because otherwise it'd look e.g. like this
```idris
{def : a} -> TTC a => TTC (WithDefault a def) where
  toBuf b def = if isDefault def
                then tag 0
                else do tag 1
                        toBuf b $ collapseDefault def
```
`onWithDefault` makes sure you don't mix up the two cases but also in this case allows removing the implicit `{def : a}` from the constraints. If we passed along the default value to the `defaultHandler` (`(tag 0)` in the above example) it could still lead to ambiguity for a user, which is why I chose not to do that.

Adowrath:
Regarding the failing CI test, it appears to be ubuntu-test-elab, which pulls in [Z-snail's prettier](https://github.com/Z-snails/prettier) and tries to build it, failing because it needs to be updated with regards to the `WithDefault` change. What should I do about this? I can't exactly introduce such a change in their repository yet I think, no?

gallais:
You can open PRs to these repos to get the fixes merged upstream before this PR gets merge.
Cf. what I did for "3062

Adowrath:
Some more changes to accomodate the LSP package on `pack` and such are required, and I will be busy for this week, so I'm going to mark this as a draft for now. 
@gallais Also a general question, would it be okay to rebase this later into more coherent commits or should I leave the history intact, which would fit the commit history after merge better?

andrevidela:
Is this just stuck ebcause of `elab-utils`?

buzden:
> Is this just stuck ebcause of `elab-utils`?

No, not really, `elab-util`'s PR is ready to go in a sync with this PR. But there was decided that a small change to this PR should be done + LSP breakage fix should be suggested.

andrevidela:
I'm confused as to where this is blocking now, is it because of LSP?

Adowrath:
> I'm confused as to where this is blocking now, is it because of LSP?

I'm sorry, I lost track of this. Currently blocked because of the LSP, but potentially other stuff in `pack` too, plus the names are a little too general/may cause confusion or other conflicts so I'll change those into more explicit ones too. I'll look into seeing what it does with the libraries in `pack` by the end of this week and finish up.

gallais (approved):

gallais:
I think we should move ahead with this PR.

Adowrath:
> I think we should move ahead with this PR. It's good work and it'd be a shame to see it disappear because it'll force downstream projects to make some minor updates.

Hey Guillaume, thanks for the kind words. I actually renamed a couple of functions & the data constructors due to feedback (`value` was very generic, it changed to `specified` and `defaulted` respectively) and will push that change in a bit. I'm also trying to test pack's db build at the same time to verify/at least find the bugs, and will push the rename fix to elab-util as well, but that will be a bit later. Just look at the minor changes in the commit that comes in a few minutes please, that will be the final in Idris' codebase itself. 👍

buzden:
`elab-util` has been checked upon this PR and appropriate PR to `elab-util` is ready to be merged as soon as this PR is decided to be merged.

Adowrath:
Idris2-lsp also received a pending PR. Some other libraries in pack currently don't compile, but I'll work on PRs for those after this.

Adowrath:
Actually needed one small change to let `pretty-show` derive its `PrettyVal` instance (stefan-hoeck/idris2-pretty-show#30), `WithDefault` needs to be `public` in the user-facing version (even though the helper constructors are to be preferred).

gallais:
I'm happy to merge this today.

Adowrath:
> I'm happy to merge this today.

Should be good, PRs to libraries in Pack have been submitted and I can check the collection in a docker container with those modified dependencies, and all compile (some tests are failing but that's neither a blocker nor caused by this change, necessarily).

buzden:
> I'm happy to merge this today.

As far as I understand, LSP's PR won't be merged until this PR is merged because they use this repo as a submodule. I'm ready to merge the accodring update to `elab-util` now. Are we good to go with red CI because of LSP?

gallais:
Sure, let's go.

buzden:
> Sure, let's go.

`elab-util` is ready

Adowrath:
LSP has been updated with the commit of the new main. 👍

Adowrath:
With [v0.7.0](https://github.com/idris-lang/Idris2/releases/tag/v0.7.0) out, should I already create a patch changing this into a hard error that we can merge later, or can that wait?

PR #3064, opened 2023-08-31
bugfix: change order of pattern-matching to avoid dead-code bug in IDE parser

ohad:
The coverage checker doesn't issue a warning because the scrutinee disambiguating the match is a string literal


andrevidela:
This is marked as draft, is it because it's not done or is it because of something else?

gallais:
There are many more bugs which I had started tracking down by trying to prove that
deserialisation and serialisation were inverses of each other on: https://github.com/gallais/Idris2/commits/bugfix-ide-protocol

PR #3067, opened 2023-09-01
Test pack and lsp during CI

CodingCellist:
# Description

Occasionally, changes to `main` have broken [pack](https://github.com/stefan-hoeck/idris2-pack) and/or the [LSP](https://github.com/idris-community/idris2-lsp). Given that these are by now widely used, and fairly critical, pieces of machinery, it seems appropriate to include them in the CI we already run.

This PR also separates out the external library tests (frex, collie, katla, pack, and lsp) from the Ubuntu self-hosting test.

Nah.


.github/workflows/ci-extern-libs.yaml
```diff
@@ -0,0 +1,269 @@
...
+          name: ubuntu-installed-idris2-${{ env.IDRIS2_MINIMUM_COMPAT_VERSION }}-chez
+          path: ~/.idris2/
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y chezscheme
+          echo "$HOME/.idris2/bin" >> "$GITHUB_PATH"
+          chmod +x "$HOME/.idris2/bin/idris2" "$HOME/.idris2/bin/idris2_app/"*
+      - name: Build self-hosted from previous version
+        run: make all IDRIS2_BOOT="idris2 -Werror" && make install
```
gallais: The annoying thing with the decoupling is that we are duplicating work by re-building idris here.

CodingCellist: If we accept `pack` as a dependency, we could just move it all to depend on building `pack`?...  : )

gallais: What's wrong with the monolithic file?

buzden: We can save built and installed state of the compiler between jobs. I do it all the time in [my stuff](https://github.com/buzden/deptycheck/blob/7c3a4884b794e60e1edb9a1f052cbbd304421c67/.github/workflows/ci-deptycheck.yml#L100-L106)

gallais: Doesn't that mean we'd run an old version of the compiler?

buzden: > Doesn't that mean we'd run an old version of the compiler?

If we save the state after building and installing the new version, we would restore new version too. So, no, not really

gallais: But the point is that because we're not building in a monolithic file anymore then
we're not ensuring that this build happens before the library testing, are we?

buzden: > But the point is that because we're not building in a monolithic file anymore then we're not ensuring that this build happens before the library testing, are we?

If I understand your complaint correctly, we can assure this with several files too, by setting dependency to main build jobs and by requiring a built artifact from them.

But if the question is in the count of files, I don't see the **need** for several ones

CodingCellist: The reason to separate things out into a different file was for clarity: currently, if `ci-idris2` fails, that (to me) means that the Idris2 build failed. But it turns out it could be a number of non-idris2-build related things like katla or collie not compiling. Hence, creating a new CI job called "external-libs".

The work duplication is true, but these jobs (afaik) run in parallel so I'd only consider it semi-wasted work. I'd rather not have this job depend on `ci-idris2` since that would defeat the purpose of concurrency, from my point of view.  
I copied the quick-check and self-host-previous because they seemed like sensible checks before building the external libs. However, from a search through the CI documentation, it seems it is possible to change this such that the `ci-idris2` job cancels this one if it fails, thereby removing the duplicate work while maintaining the separate files.

Thoughts?  : )

gallais: > since that would defeat the purpose of concurrency, from my point of view.

You can see the dependency graph on any build e.g. https://github.com/idris-lang/Idris2/actions/runs/6089530244?pr=3063

We are already building the external libs concurrently with the ones testing the compiler can self-host with various backends.

The point of having `quick-check` as a universal dependency is that we do not try to build everything else unless we are sure the basic build already works. This helps with concurrency across PRs as we are rate limited in the number of builds we can have running at the same time.

buzden: > I would rather not have duplicated work at all.

I personally fully agree. Even for the sake of concurrency, I think, it's better to depend on a job rather than to repeat the prefix work

CodingCellist: Fair enough, I have merged the files back into one to avoid work duplication. I can revert the rename if need be?  : )

.github/workflows/ci-idris2-and-libs.yml
```diff
@@ -578,12 +593,86 @@ jobs:
...
+        with:
+          repository: 'stefan-hoeck/idris2-pack'
+      # by default, pack uses the main idris2 head, not the current job's one
+      - name: Toml setup
+        run: |
+          echo "[idris2]" > pack.toml
+          echo "url = \"https://github.com/${GITHUB_REPOSITORY}\"" >> pack.toml
+          echo "commit = \"latest:${GITHUB_REF_NAME}\"" >> pack.toml
+          echo "bootstrap  = true" >> pack.toml
+      - name: Build idris2-pack
```
buzden, line 626: I think, after setting `pack.toml`, one needs to call `pack fetch` to make sure `pack` has fetched the correct commit of the branch mentioned in `latest:`

CodingCellist: The local testing I did with this seemed to correctly get the commit, but I'll @stefan-hoeck just to be sure; afaiu, `pack install pack` is clever enough to fetch the latest/specified Idris2-commit before rebuilding and reinstalling itself  : )

buzden: My knowledge would be that in this case the latest commit at the moment of creation of the docker image would be used instead (without calling `pack fetch`, as I suggest), but right, let's wait for Stefan's answer

CodingCellist: Did some testing using the docker image rather than my local `pack` install; it seems `pack fetch` is indeed redundant:

Before, i.e. having just launched a fresh copy of the image:
```
# idris2 --version
Idris 2, version 0.6.0-72270962f
```

After creating a matching `pack.toml` and running `pack install pack`:
```
# idris2 --version
Idris 2, version 0.6.0-2a96c36fd
```

Which matches the latest commit on this branch  : )

buzden: `git fetch` is needed when there were some commits into the `main` branch between the docker image creation and running the CI. Since there were no commits into `main` recently, you cannot reproduce this situation

.github/workflows/ci-idris2-and-libs.yml
```diff
@@ -634,7 +634,7 @@ jobs:
   ######################################################################
 
   ub-pack-test-lsp:
-    needs: ubuntu-self-host-previous-version
+    needs: [initialise, ubuntu-self-host-previous-version]
```
gallais, line 638: Shouldn't this dependency be transitively implied anyways?

PR #3070, opened 2023-09-05
[ fix ] consider nest when guessing scrutinee

### LINKED ISSUE #2048: imported constructor preferred over whatever under `where` clause
Let's say you want to compile the following code taken directly from [the tutorial](https://idris2.readthedocs.io/en/latest/tutorial/typesfuns.html#where-clauses)
```
module Main

foo : Int -> Int
foo x = case isLT of
            Yes => x*2
            No => x*4
   where
      data MyLT = Yes | No

      isLT : MyLT
      isLT = if x < 20 then Yes else No
```

It produces a very weird error:
```
Error: While processing right hand side of foo. When unifying:
    MyLT x
and:
    Dec ?_
Mismatch between: MyLT x and Dec ?_.

test:4:14--4:18
 1 | module Main
 2 | 
 3 | foo : Int -> Int
 4 | foo x = case isLT of
```

The issue is that the first reference of `Yes` is to `Prelude.Yes`. The error message is unhelpful.

By moving `MyLT` before `foo`, the program will compile:
```
foo : Int -> Int
data MyLT = Yes | No
foo x = case isLT of
            Yes => x*2
            No => x*4
   where

      isLT : MyLT
      isLT = if x < 20 then Yes else No
```

This means that the semantic of a function can change depending on what's imported.

dunhamsteve:
This patch fixes #2048 - When trying to guess the scrutinee type, Idris was looking in the context for the case alt constructors, but it did not look in `nest`.  In the example below (from the Idris tutorial), a `Yes` was found in prelude and Idris decided the type was `Dec a`, not knowing there was another `Yes` in the `where` clause.

I add any matching names from `nest` to the list of possibilities. In the case below, this yields an ambiguity and Idris places a hole for the type, which it resolves later. In other cases it would see the constructor from the where clause and infer the type.  If we know that Idris always replaces the name with the one from nest when processing the case alts,  an alternative, simpler approach would be to replace the name with the one from nest when present.

Before the change, this code from the tutorial:
```idris
module Main

foo : Int -> Int
foo x = case isLT of
            Yes => x*2
            No => x*4
   where
      data MyLT = Yes | No

      isLT : MyLT
      isLT = if x < 20 then Yes else No
```
would produce the error:
```
Error: While processing right hand side of foo. When unifying:
    MyLT x
and:
    Dec ?_
Mismatch between: MyLT x and Dec ?_.

test:4:14--4:18
 1 | module Main
 2 | 
 3 | foo : Int -> Int
 4 | foo x = case isLT of
```
With this patch, Idris accepts the file.


src/TTImp/Elab/Case.idr
```diff
@@ -460,7 +460,9 @@ checkCase rig elabinfo nest env fc opts scr scrty_in alts exp
         = case getFn x of
                IVar _ n =>
                   do defs <- get Ctxt
-                     [(n', (_, ty))] <- lookupTyName n (gamma defs)
+                     [(n', (_, ty))] <- do nested <- maybe (pure []) (\n => lookupTyName n (gamma defs))
+                                                      (join $ Builtin.fst <$> List.lookup n nest.names)
+                                           (nested ++) <$> lookupTyName n (gamma defs)
```
gallais, line 465: Is it worth lifting this pattern to a top-level definition
```
lookupLocalTyNames : {auto c : Ref Ctxt Defs} -> NestedNames vars -> Name -> Core (List (Name, NF []))
```
? Surely this pattern occurs in other places too?

gallais: Yes please. Whatever the common pattern is should be isolated so that it is used uniformly!

PR #3075, opened 2023-09-15
[ ci ] re #3067: fix CI on main branch

CodingCellist:
# Description

The extra CI jobs introduced in #3067 work fine as long as 'main' is not the checked out branch. This is due to the fetch to a new branch, which git does (reasonably) not allow when you're trying to fetch 'main' into a new branch that's also called 'main'. In this case, we should just `git pull origin main`, which is what the script now (hopefully) does. (source: https://stackoverflow.com/questions/2236743/git-refusing-to-fetch-into-current-branch)

### DIFF
--- a/.github/workflows/ci-idris2-and-libs.yml
+++ b/.github/workflows/ci-idris2-and-libs.yml
@@ -627,7 +627,14 @@ jobs:
       - name: Build idris2-pack
         run: |
           git config --global --add safe.directory "${PWD}"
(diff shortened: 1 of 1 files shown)


CodingCellist:
(merging this as soon as the linting CIs have passed, since CI off a PR is working fine, it's CI on main which is broken)

CodingCellist:
Seems to pass when merged to `main` on my fork, so I'm merging this now  : )

PR #3076, opened 2023-09-15
[ fix ] Make `traverse` and friends lazy for `LazyList`

buzden:
# Description

This change is like #2412, but for old long-existing functions. Since `Applicative` interface is inherently not lazy, the current implementation was not lazy by the effect, even if the underlying monad's effect could be short-cutting. One can see an example in the added test.

Since, `LazyList` is still `Foldable`, old behaviour is reachable by converting the `LazyList` to `List`.

It may seem that this change relies on some exploiting of the breakage of monadic laws, but it is fundamental for an eager language since monad interface is inherently lazy anyway.

      `CONTRIBUTORS.md`).

### DIFF
--- a/libs/contrib/Data/List/Lazy.idr
+++ b/libs/contrib/Data/List/Lazy.idr
@@ -124,16 +124,16 @@ Monad LazyList where
 -- The result of a traversal will be a non-lazy list in general
 -- (you can't delay the "effect" of an applicative functor).
 public export
-traverse : Applicative f => (a -> f b) -> LazyList a -> f (List b)
+traverse : Monad f => (a -> f b) -> LazyList a -> f (List b)
 traverse g [] = pure []
-traverse g (x :: xs) = [| g x :: traverse g xs |]
+traverse g (x::xs) = pure $ !(g x) :: !(traverse g xs)
 
 public export %inline
-for : Applicative f => LazyList a -> (a -> f b) -> f (List b)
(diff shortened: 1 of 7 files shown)


libs/contrib/Data/List/Lazy.idr
```diff
@@ -124,16 +124,16 @@ Monad LazyList where
 -- The result of a traversal will be a non-lazy list in general
 -- (you can't delay the "effect" of an applicative functor).
 public export
-traverse : Applicative f => (a -> f b) -> LazyList a -> f (List b)
+traverse : Monad f => (a -> f b) -> LazyList a -> f (List b)
 traverse g [] = pure []
-traverse g (x :: xs) = [| g x :: traverse g xs |]
+traverse g (x::xs) = pure $ !(g x) :: !(traverse g xs)
```
gallais, line 129: How is this any lazier? `!` lifts the expression to the surrounding `do` block so we do run `traverse` before emitting `(::)`.

buzden: if gives `g x >>= \y => traverse g xs >>= pure (y :: ys)`, which means that `traverse g xs` is computed only lazily demanded by `>>=`. When we use `<*>` for this, it is computed eagerly. By monadic laws, these runs give the same result, but monadic laws do not talk about laziness.

buzden: This is the same difference as between non-lazy applicative `*>` and lazy monadic `>>`

PR #3079, opened 2023-09-21
[ elab ] Print script's FC in the bad elaboration script error

buzden:
# Description

When you are writing (and running) elaborator scripts, occasionally you can get the `Bad elaborator script` error. Pretty long time ago I suggested a PR that added this bad elaborator script to the error output, and it helped a lot, but required some custom and thought on what's going on. Today, I had such an error with a mystical output, and I thought that printing the position of where the compiler thinks there is a problem is good.

So, I suggest to print not only `FC` of the call to the script, but also an `FC` of a position in a script, which cause the `Bad elaborator script` error itself (it it is present, surely).

One can see in the added test how the error can look like with this change.


gallais (approved):
This looks great!

PR #3081, opened 2023-09-29
Several small changes each not worth its own PR

buzden:
# Description

Each commit contains description of a change

## Should this change go in the CHANGELOG?

- [x] If this is a fix, user-facing change, a compiler change, or a new paper
      implementation, I have updated `CHANGELOG.md` (and potentially also
      `CONTRIBUTORS.md`).

PR #3085, opened 2023-10-03
[ test ] fix pkgs test failures

### LINKED ISSUE #3084: pkg006 test fails
See title. Happens on both my machine (Arch Linux with AUR idris2-git package) and in a Debian docker container when building the latest commit. I haven't bisected yet to find where this problem originated, but I believe that commit cf9a73f86 is clean. All other tests are fine.

The diff when accepting it as the new golden value is
```diff
diff --git a/tests/idris2/pkg/pkg006/expected b/tests/idris2/pkg/pkg006/expected
index e96b65d49..da1f9c5ba 100644
--- a/tests/idris2/pkg/pkg006/expected
+++ b/tests/idris2/pkg/pkg006/expected
@@ -1,7 +1,4 @@
 Uncaught error: EmptyFC:Failed to resolve the dependencies for test3:
   required foo >= 0.4 && < 0.5 but no matching version is installed

-Uncaught error: EmptyFC:Failed to resolve the dependencies for test4:
-  required baz any but no matching version is installed
-
 Warning: Deprecation warning: version numbers must now be of the form x.y.z
```
I can submit a PR with this if desired, although looking at the test I feel like the current expected value is correct.

dunhamsteve:
The tests `pkg006` and `pkg015` are randomly failing. It looks like this is happening because `pkg016` is installing (and later removing) packages named `foo`, `bar`, and `baz`. These packages are seen by the other tests if they happen to run at the same time. This PR mitigates the issue by renaming the packages in the `pkg016` test.

This should fix #3084 (the issue was also noted in #3081).

### DIFF
--- a/tests/idris2/pkg/pkg016/bar.ipkg
+++ b/tests/idris2/pkg/pkg016/bar.ipkg
@@ -1,4 +1,4 @@
-package bar
+package two
 sourcedir = "bar"
 modules = Bar
-depends = foo
+depends = one
--- a/tests/idris2/pkg/pkg016/baz.ipkg
+++ b/tests/idris2/pkg/pkg016/baz.ipkg
@@ -1,4 +1,4 @@
-package baz
+package three
 sourcedir = "baz"
 modules = Baz
-depends = bar
+depends = two
--- a/tests/idris2/pkg/pkg016/foo.ipkg
+++ b/tests/idris2/pkg/pkg016/foo.ipkg
@@ -1,3 +1,3 @@
-package foo
+package one
 sourcedir = "foo"
 modules = Foo
--- a/tests/idris2/pkg/pkg016/run
+++ b/tests/idris2/pkg/pkg016/run
@@ -7,6 +7,6 @@ idris2 --build test.ipkg
 
 build/exec/test
 
-rm -r "${IDRIS2_PREFIX}"/idris2-*/foo-0
-rm -r "${IDRIS2_PREFIX}"/idris2-*/bar-0
-rm -r "${IDRIS2_PREFIX}"/idris2-*/baz-0
+rm -r "${IDRIS2_PREFIX}"/idris2-*/one-0
+rm -r "${IDRIS2_PREFIX}"/idris2-*/two-0
+rm -r "${IDRIS2_PREFIX}"/idris2-*/three-0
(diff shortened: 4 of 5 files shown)


buzden:
Would it be more stable to set installation directory to a local one via environment variable to avoid this effect at all? It would be also more stable for potential future tests, because it's hard to remember this all the time and avoid to do it again.

buzden:
By the way, we could set `IDRIS2_PREFIX` in `testutils.sh`, which now is universally used in the tests, so that we won't need to think about this each time we create new package test

dunhamsteve:
The test requires installation and then finding the installed packages in a subsequent build. I originally wanted a separate prefix, but thought it would break locating `prelude` and `base` and didn't try it.  

It turns out that `support` is also lost if we relocate `IDRIS2_PREFIX`.  I think I'd have to put a prefix directory inside the test (e.g. at `prefix` in the test directory) _and_ symlink those directories in there. I also need to figure out the string `idris-0.6.0`.  The test `pkg017` is using a custom prefix, but doesn't reference prelude.

If we want this approach, here is what I have so far. I was only planning on calling it for the tests that install packages.

```sh
# creates a local prefix directory for package install tests
# and symlinks needed files
setup_prefix_dir() {
    prefix=$("$idris2" --prefix)
    base=$(basename "$prefix"/idris2-*)
    export IDRIS2_PREFIX="${test_dir}/prefix"
    # cleanup from previous tests
    rm -rf "$IDRIS2_PREFIX"
    mkdir -p "$IDRIS2_PREFIX/$base"
    ln -s "$prefix/$base"/* "$IDRIS2_PREFIX/$base"
}
```

buzden:
> I originally wanted a separate prefix, but thought it would break locating `prelude` and `base` and didn't try it.
> 
> It turns out that `support` is also lost if we relocate `IDRIS2_PREFIX`

This is all correct, and that's why I suggest to do [this](https://github.com/idris-lang/Idris2/pull/3088/commits/ec88d2e57735e8830b6ed6fd9992f19e354cc5b6#diff-b531f997153ae22980b8fa55908f3cf53c15e63df4b8915ae8c4b86657777267R49-R51) here #3088

dunhamsteve:
I prefer #3088 to what I described above, so I'm going to close this PR.

PR #3092, opened 2023-10-05
[ fix #3083 ] Fix record update with implicit args

### LINKED ISSUE #3083: Updates of records with auto implicits result in unsolved holes
# Steps to Reproduce

Create `Foo.idr` with contents
```idris
module Foo

record WithShow (ty : Type) where
  constructor MkWS
  {auto hasShow : Show ty}
  name : String

foo : WithShow ty -> WithShow ty
foo ws = { name := "meh" } ws
```
and try loading it into repl.

[gist for convenience](https://gist.github.com/0xd34df00d/4b6fe75e21cffb8f2ce786ce151bc9e5)

# Expected Behavior

Successful typecheck. Or, at least, a more telling hole name, but successful load of the module is more expected since the record update doesn't touch the type and the implicit `hasShow` field.

# Observed Behavior

```
Error: Unsolved holes:
Foo.{_:905} introduced at: 
Foo:9:10--9:30
 5 |   {auto hasShow : Show ty}
 6 |   name : String
 7 | 
 8 | foo : WithShow ty -> WithShow ty
 9 | foo ws = { name := "meh" } ws
              ^^^^^^^^^^^^^^^^^^^^
```

Alex1005a:
# Description

This PR fix #3083.
Now update of record leaves a hole only for implicit type variable.
Moreover, would it be correct to have a different behavior for auto arguments to make the following code correct?
```idris
record WithShow (ty : Type) where
  constructor MkWS
  {auto hasShow : Show ty}
  name : ty

foo : WithShow ty -> WithShow String
foo ws = { name := "meh" } ws
```

### DIFF
--- a/src/TTImp/Elab/Record.idr
+++ b/src/TTImp/Elab/Record.idr
@@ -26,6 +26,10 @@ getRecordType : Env Term vars -> NF vars -> Maybe Name
 getRecordType env (NTCon _ n _ _ _) = Just n
 getRecordType env _ = Nothing
 
+isType : NF vars -> Bool
+isType (NType {}) = True
+isType _ = False
+
 data Rec : Type where
      Field : Maybe Name -> -- implicit argument name, if any
              String -> RawImp -> Rec -- field name on left, value on right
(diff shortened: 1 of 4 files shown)


Alex1005a:
Why is the first argument needed in the `getRecordType` function?

PR #3098, opened 2023-10-11
[ parser ] Fix issues parsing %logging followed by named impls - fix #3097

### LINKED ISSUE #3097: Strange parser error when named implementation with an implicit argument goes after `%logging`
# Steps to Reproduce

Consider the following code:

```idris
import Data.Vect

[Named] {n : Nat} -> Show (Vect n a) where
  show x = "whatever"
```

It typechecks well.

Now, add a `%logging` directive:

```idris
import Data.Vect

%logging "elab" 100

[Named] {n : Nat} -> Show (Vect n a) where
  show x = "whatever"
```

# Expected Behavior

Typechecks well, too

# Observed Behavior

```console
Error: Expected '}'.

BugImplImplLogging:5:12--5:13
 1 | import Data.Vect
 2 | 
 3 | %logging "elab" 100
 4 | 
 5 | [Named] {n : Nat} -> Show (Vect n a) where
                ^
```

dunhamsteve:
In issue #3097, a `%logging` pragma followed by a named implementation will produce an odd parse error.  The root cause for this is a second, inline parser for `%logging` inside `simplerExpr` that is being hit while `implDecl` attempts to parse the `%logging` line.  I addressed the issue by moving the `directiveDecl` parser before the `impDecl` parser in the alt.

But this change broke `%runElab` statements. It turns out that there is an old parser for `%runElab` inside `directiveDecl` that was no longer being used because `runElabDecl` occurred first.  It parsed `%runElab` to `PReflect`, which desugars to a "Reflection not implemented yet" error. So I removed that.

Since `PReflect` is not used (the real parser for `%runElab` produces a `PRunElabDecl`), I also removed `PReflect`.

Previously:
```idris
import Data.Vect
%logging "elab" 100
[Named] {n : Nat} -> Show (Vect n a) where
  show x = "whatever"
```
produced this error:
```
Error: Expected '}'.

BugImplImplLogging:5:12--5:13
 1 | import Data.Vect
 2 | 
 3 | %logging "elab" 100
 4 | 
 5 | [Named] {n : Nat} -> Show (Vect n a) where
                ^
```
And now it compiles without error.

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1848,6 +1848,8 @@ topDecl fname indents
          pure [d]
   <|> do ds <- claims fname indents
          pure (forget ds)
+  <|> do d <- directiveDecl fname indents
+         pure [d]
   <|> do d <- implDecl fname indents
          pure [d]
   <|> do d <- definition fname indents
@@ -1872,8 +1874,6 @@ topDecl fname indents
   <|> do d <- runElabDecl fname indents
          pure [d]
   <|> do d <- transformDecl fname indents
-         pure [d]
-  <|> do d <- directiveDecl fname indents
          pure [d]
   <|> do dstr <- bounds (terminal "Expected CG directive"
                           (\case
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1828,13 +1828,9 @@ fixDecl fname indents
 
 directiveDecl : OriginDesc -> IndentInfo -> Rule PDecl
 directiveDecl fname indents
-    = do b <- bounds ((do d <- directive fname indents
-                          pure (\fc : FC => PDirective fc d))
-                     <|>
-                      (do decoratedPragma fname "runElab"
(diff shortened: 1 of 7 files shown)


src/Idris/Parser.idr
```diff
@@ -1828,13 +1828,9 @@ fixDecl fname indents
...
-                          pure (\fc : FC => PDirective fc d))
-                     <|>
-                      (do decoratedPragma fname "runElab"
-                          tm <- expr pdef fname indents
-                          atEnd indents
-                          pure (\fc : FC => PReflect fc tm)))
+    = do b <- bounds (do d <- directive fname indents
+                         pure (\fc : FC => PDirective fc d))
+
          pure (b.val (boundToFC fname b))
```
gallais, line 1834:
```suggestion
    = do b <- bounds (directive fname indents)
         pure (PDirective (boundToFC fname b) b.val)
```

gallais (commented):
Looks good. Just a tiny clean up suggested.

PR #3101, opened 2023-10-15
[ fix ] Prevent relative path traversal in elaborator scripts

AlgebraicWolf:
# Description

Compile-time file operations introduced in #3099 attempt to check file paths and prevent exiting the directory specified by `LookupDir`. This safeguard, however, can be easily fooled by the use of `./` as part of the path. For instance, the following call would result in the creation or  modification of a file outside the project directory:

```idris
%runElab writeFile ProjectDir "./../outerFile" "uh oh"
```

This occurs because the introduced function `pathDoesNotEscape` performs the check by tracking how deep the provided relative path goes down in the directory tree, making sure it doesn't go up more than it goes down. It, however, treats "." like a regular directory, allowing the user to increase the depth perceived by the function "for free". This PR adds special handling of the "." directory in the path to prevent escaping via this mechanism.


gallais (approved):
Haha, nice one!

PR #3109, opened 2023-10-17
[ derive ] Refine a constructor view to support implicit args

buzden:
# Description

The current constructor view supports only explicit parameters, this PR changes it to support implicit ones too. Existing code that relies on constructor view is changed accordingly, and, to the level of my knowledge, without any change in the semantics.

      `CONTRIBUTORS.md`).


gallais:
Is there any chance we can add tests showing what kind of support this adds?

buzden:
> Is there any chance we can add tests showing what kind of support this adds?

I originally wanted something like in a test that I've just added. But then I realised that 1) `isType` is much more limiting than I originally thought 2) order of nats inside the constructor view can hypotetically differ depending on whether `Ty {impl} expr` and `Ty expr {impl}` elaborate to the same expression or not.

So, now I see that what I meant to be possible with the changes constructor view from `Deriving.Common`, maybe, cannot be done with preserving the same semantics.

PR #3111, opened 2023-10-19
Switch from 'fast' string functions to normal string functions

### LINKED ISSUE #791: JSON parser not working in repl

# Steps to Reproduce
```
idris2 -p contrib
Main> :module Language.JSON
Imported module Language.JSON
Main> parse "{}"
>>= (case let (ts, (l, (c, str'))) =
                case case >>= (scan (Pred isSpace) [] (fastUnpack "{}")) (\{_:0} => let (tok', rest) = _ in assert_total (scan Force r2 tok' rest)) of
                     {Just (tok, rest) => let line' = line + cast (countNLs a str tmap acc col line pred tok)
                                          in let col' = getCols a str tmap acc col line pred tok col in Just (MkBounded (fn (fastPack (reverse tok))) False line col line' col', (line', (col', rest)));
                     Nothing => getFirstToken a str tmap acc col line pred ts str} of
                {Just (tok, (line', (col', rest))) => if pred tok then (reverse acc, (line, (col, []))) else assert_total (tokenise pred line' col' (tok :: acc) tmap rest);
                Nothing => (reverse acc, (line, (col, str)))}
          in (ts, (l, (c, fastPack str'))) of
     {(tokens, (_, (_, ""))) => Just (map (tok ?b) tokens);
     _ => Nothing}) (\{bind:0} =>
                       case case case case doParse False (Terminal "Unrecognised input" (\t =>
                                                                                           if == (.kind (.val t)) (JTPunct (Curly Open))
                                                                                             then Just (tokValue k (.text (.val t)))
                                                                                             else Nothing)) (filter (\x => not (ignored x)) bind) of
                                      {Failure com fatal msg ts => Failure com fatal msg ts;
                                      Res com v xs => case assert_total (doParse com (Force next (.val v)) xs) of
                                                      {Failure com' fatal msg ts => Failure com' fatal msg ts;
                                                      Res com' v' xs => Res com' (mergeBounds v v') xs}} of
                                 {Failure com' fatal msg ts => if com' || Delay fatal then Failure com fatal msg ts else assert_total (doParse False Force y xs);
                                 Res _ val xs => Res com val xs} of
                            {Failure _ _ msg ts => Left (Error msg ts);
                            Res _ v rest => Right (.val v, rest)} of
                       {Right (j, []) => Just j;
                       _ => Nothing})
Main>
```

# Expected Behavior
should output `"{}"`

# Observed Behavior
Weird error above

Additionally if I create a program with similar code and execute it I get:
```
Exception: invalid memory reference.  Some debugging context lost
```

rvs314:
# Description

This switches the string functions used in the Lexer and Parser libraries from their `fast` variants to the normal variants. This allows them to be used at compile-time, but doesn't introduce performance regression, as they're translated to their faster versions at compile-time anyways. This should fix #791

I don't think so

### DIFF
--- a/libs/contrib/Text/Distance/Levenshtein.idr
+++ b/libs/contrib/Text/Distance/Levenshtein.idr
@@ -10,7 +10,7 @@ import Data.List.Extra
 
 ||| Self-evidently correct but O(3 ^ (min mn)) complexity
 spec : String -> String -> Nat
-spec a b = loop (fastUnpack a) (fastUnpack b) where
+spec a b = loop (unpack a) (unpack b) where
 
   loop : List Char -> List Char -> Nat
   loop [] ys = length ys -- deletions
--- a/libs/contrib/Text/Lexer/Core.idr
+++ b/libs/contrib/Text/Lexer/Core.idr
@@ -159,7 +159,7 @@ tokenise pred line col acc tmap str
                Just (tok, rest) =>
                  let line' = line + cast (countNLs tok)
                      col' = getCols tok col in
-                     Just (MkBounded (fn (fastPack (reverse tok))) False (MkBounds line col line' col'),
+                     Just (MkBounded (fn (pack (reverse tok))) False (MkBounds line col line' col'),
                            line', col', rest)
                Nothing => getFirstToken ts str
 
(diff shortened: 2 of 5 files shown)


buzden:
Maybe a compile-time parsing test would be a good addition to this change

PR #3116, opened 2023-10-23
[ base ] Add bindings for ieee `Double` number consts

### LINKED ISSUE #29: Double IEEE constants: infinity, NaN, epsilon ... where are they?
<a href="https://github.com/nicolabotta"></img></a> **Issue by [nicolabotta](https://github.com/nicolabotta)**
_Tuesday Oct 29, 2019 at 07:48 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/146_

----

Is there a way of accessing these values in Idris 2? Something like `Numeric.IEEE`in Haskell? At lest `epsilon`, `infinity` and `NaN` are badly needed, e.g., to implement interval arithmetics! Thanks, Nicola

CodingCellist:
MVP (afaict) for having these constants available to those who want it, along with some tests that I think check they behave as expected

Closes #29

      `CONTRIBUTORS.md`).


gallais (commented):
I think our standard approach has been to call the naked foreign primitive
something like `prim__xxx`  and then defining an `xxx`  toplevel function.
We can attach documentation to that `xxx`.

Then again this may be because we typically have `PrimIO` involved and we
use these toplevel definitions to wrap the call into a nicer `IO`...

CodingCellist:
> I think our standard approach has been to call the naked foreign primitive something like `prim__xxx` and then defining an `xxx` toplevel function. We can attach documentation to that `xxx`.
> 
> Then again this may be because we typically have `PrimIO` involved and we use these toplevel definitions to wrap the call into a nicer `IO`...

That was almost exactly my rationale; don't wrap the names with `prim` because they're not going via `PrimIO`. Then again, if it makes documentation and/or workings clearer, should there be a redirect anyway?

CodingCellist:
(Copying from Discord so it doesn't get lost in the aether):

Issue 1 is that the new bindings require `math/flonum` from the Racket libraries in order to use `epsilon.0`. And for _some reason_ using `epsilon.0` anywhere in `support.rkt`, even if it's never linked to anything in Idris and just hanging out as unused code in the support file, trips the totality checker.

Now **_why_** that causes the totality checker of all things to trip, and why it only trips it on papers' `Data/Description/Regular.idr` I have no clue. It makes no sense to me, and I don't know where to begin trying to figure out what/how it happens. I thought it was because `bootstrap/idris2_app/idris2.rkt` does not require the library in its header. But I reran the bootstrap with `(require math/flonum)` manually added to the bootstrap file and it still crashed

Issue #3113, opened 2023-10-21
`typebind` and `autobind` modifiers for operator fixity


andrevidela:
## Summary

This proposal is for adding two new modifiers for fixity declarations that transform the behavior of operators to allow them to bind their left argument to a name such that it can be used on its right-hand-side. This feature already exists when desugaring dependent pairs with `(**)` this proposal makes this feature user-definable: `typebind infixr 7 **`.

A prototype of this feature is available here: https://github.com/andrevidela/idris2/tree/autobind

## Motivation

One of the lost features between idris1 and idris2 is arbitrary rebindable syntax blocks. While powerful, they were slow, a source of errors, and increase the complexity of the language, for little benefit in real-world programs. In simple terms: they did not pull their own weight.

The motivation for arbitrary rebindable syntax hasn't gone away: the ability to write Domain Specific Languages (DSL) that rely on existing notation conventions to reduce the cognitive load of learning a language within a language. `typebind` and `autobind` modifiers aim to address this original motivation, without introducing the reasons why arbitrary syntax blocks were abandoned.

## Proposal

Augment operator fixity/precedence declaration with a modifier `typebind`/`autobind` telling the parser to desugar this operator as a _binding operator_.

### `typebind` keyword

A binding operator behaves like `(**) : (x : Type) -> (x -> Type) -> Type`, the type-constructor for dependent pairs. It allows to bind its left argument to a name and use the name directly on the right side without needing to write a lambda. 

This means that one can write the type-constructor for dependent pairs like this:

```idris
typebind infixr 7 **

record (**) (x : Type) (y : x -> Type) where
  constructor MkDPair
  fst : x
  snd : y fst
```

and can be used the same way we expect:

```idris
pair : (n : Nat ** Fin n)
pair = MkDPair 3 FZ
```

Because of their nature, binding operators can only be `infixr` and must be used in conjunction with a function that binds a name in its second argument.
Those restrictions make the following definition erroneous:

```
autobind infixr 4 /\

record /\ (a, b : Type) where
  constructor And
  fst : a
  snd : b

pair : (x : Int) /\ String
pair = 3 `And` "four"
```

Unfortunately, because operator fixity definitions and function definitions are separate, we cannot detect this misuse at the time of declaration. Only when the function is used by the parser do we notice that the function does not bind any names in its second argument. See [[#Error messages]].

Because `typebind` operators cannot be confused with regular operators, they require the use of parenthesis to be distinguished:

```
[...]

PR #3120, opened 2023-10-26
Draft: Typebind & autobind using Pi-like syntax

andrevidela:
# Description

This implements half of #3113 by allowing pi-like syntax for typebind and autobind operators.

This is a draft to show off what the feature can do and in which ways it affects (mostly doesn't affect) programs

Todo:
- [x] Add `typebind` and `autobind` keywords
- [x] Desugar `typebind` and `autobind` appropriately
- [x] Add documentation for the keywords
- [x] Display appropriate error messages when syntax is misused
- [ ]  Fix error message changes (one of them got worse)
- [ ] Add online documentation
- [ ] update changelog

      `CONTRIBUTORS.md`).


docs/source/reference/operators.rst
```diff
@@ -0,0 +1,340 @@
...
+   linearSingleton = ...
+
+The above syntax is allowed if the operator is declared as ``typebind``. For
+this, simply add the ``typebind`` keyword in front of the fixity declaration.
+
+.. code-block:: idris
+
+   typebind infixr 0 =@
+
+``typebind`` is a mofidier like ``export`` and both can be used at the same time.
```
LiamGoodacre, line 276: (typo `mofidier`)

andrevidela: thnaks

tests/idris2/operators/operators001/Test.idr
```diff
@@ -0,0 +1,80 @@
+
+module Test
+
+import Data.Vect
+
+typebind infixr 0 =@
+infixr 0 -@
```
gallais, line 7: Is there no hope to get a single operator for the binding & non-binding cases?
It seems really noisy to have to introduce 2 names for each function space.

andrevidela: I think there is, but I wasn't ready to introduce more human-level parsing ambiguity. Basically the problem is that if we see `x -!> y`, the parser can already know if it's going to be a binding operator or not. If it is then we could make it behave like `(-!>) x (\_ => y)`. If it's not it would behave like `(-!>) x y`. This is perfectly reasonable and achievable today. But the humans cannot tell the difference between the two, specially since the fixity declaration might come from an import.

So what I suggest is that, because this is a niche feature anyways, we leave the duplication. And once enough people in the community get used to it, and start complaining that it should work both ways, we introduce the ability to with both `(x : ty) -@ y` and `x -@ y`, which should be backward compatible.

src/Idris/Desugar.idr
```diff
@@ -714,35 +756,59 @@ mutual
...
+           desugaredRHS <- desugarTree side ps r
+           pure $ IApp loc (IApp loc (IVar opFC op) desugaredLHS)
+                           (ILam loc top Explicit (Just name.val) desugaredLHS desugaredRHS)
+  -- (x := exp ** f x) ==>> (**) exp (\x : ? => f x)
+  desugarTree side ps (Infix loc opFC (op, Just (BindExpr name lhs)) _ r)
+      = do desugaredLHS <- desugarB side ps lhs
+           desugaredRHS <- desugarTree side ps r
+           pure $ IApp loc (IApp loc (IVar opFC op) desugaredLHS)
+                           (ILam loc top Explicit (Just name.val) (Implicit opFC False) desugaredRHS)
+  -- (x : ty := exp ** f x) ==>> (**) exp (\x : ty => f x)
```
andrevidela, line 795: Just noticed those comments are out of date, I'll update them

src/Core/TT.idr
```diff
@@ -103,6 +104,128 @@ Ord Visibility where
...
+        && x.precedence == y.precedence
+
+-- Left-hand-side information for operators, carries autobind information
+-- an operator can either be
+-- - not autobind, a regular operator
+-- - binding types, such that (nm : ty ** fn nm) desugars into (ty ** \(nm : ty) => fn nm)
+-- - binding expressing with an inferred type such that
+--   (nm := exp ** fn nm) desugars into (exp ** \(nm : ?) => fn nm)
+-- - binding both types and expression such that
+--   (nm : ty := exp ** fn nm) desugars into (exp ** \(nm : ty) => fn nm)
```
andrevidela, line 180: Same here, inaccurate documentation

tests/idris2/operators/operators001/Test.idr
gallais: Given that `->` already works like that, I think having a uniform behaviour
for all function spaces would be less confusing.

andrevidela: This is not just for function spaces, specially when you start using `autobind`. Maybe it would make sense for `typebind` to allow the overloading but I still think it's too early to tell if disallowing the overloading is a usability regression or usability improvement. I would still insist on disallowing the overloading for three reasons:
- allowing it is a non-breaking change, we can always add it later
- Allowing binding operators is already a usability improvement, even if it comes with overzealous safety nets
- If the overloading was allowed, I would like to see proper error messaging diagnosing the fact that switching from/to binding syntax is a valid way to resolve the problem. But doing it naively would simply result in an obscure unification error.

gallais: I think I've already encountered people puzzled by `(x : a) -@ b` not working in the current state of affairs.

andrevidela: Indeed, and those users are still getting benefits:
- now they have an alternative with `=@` (assuming we add it to `linear`).
- they get an error message that explains why `-@` does not work.

Like I said in my first response, I am not inherently against implementing the overloading. However I don't think we have enough data about the user experience to make the call right now. Finally, this PR implement #3113 which did not make any claims about overloading, quite the opposite:

> this proposal errs on the side of caution  by making conservative choices with regard to overloading and syntax disambiguation

I'd be happy to discuss the details and implementation of overloading binding feature in a dedicated channel but I think it's out of scope for this PR.

gallais: I don't like `=@` because I think it should be to `-@` what `=>` is to `->`, again for design uniformity.
But that's a separate bikeshedding discussion.

I am concerned that when we attempt to implement the overloading we will actually discover that
it cannot be done because it makes the grammar too ambiguous and then we'll be stuck with the
current design. Additionally, I find the "not enough data" argument dubious when the choice is between
1. a consistent design across the board
2. a design with more corner cases

Finally, this PR bundles together `typebind` (which is not limited to types IIUC) and
`autobind` (which has nothing to do with auto but is more like an infix notion of let-binding
IIUC) but AFAIU these are orthogonal features. Should we have separate PRs and separate
discussions?

andrevidela: > I don't like =@ because I think it should be to -@ what => is to ->, again for design uniformity.
But that's a separate bikeshedding discussion.

Indeed, this PR is about the `autobind` and `typebind` feature, not about `=@` which is not in this diff.

> I am concerned that when we attempt to implement the overloading we will actually discover that
it cannot be done

The error messages can detect when you're misusing a regular operator as a binding operator and vice-versa. Lifting the restriction means changing this error message into the proper desugaring.

> I find the "not enough data" argument dubious 

It comes from the fact that I've been using this diff for my own code for 4 weeks and haven't felt the need to add the overload. Maybe someone else will and I'm waiting for this to happen. In the meantime, I ask that people hold their judgement until they've tried things. As for the two points, I think the design is consistent with the rationale given, and the corner cases are properly handled with helpful error messages.

> AFAIU these are orthogonal features. Should we have separate PRs and separate
discussions

Possibly, but this PR is the result of a proposal written in october 21st, almost 2 months ago. I would have appreciated if those concerned were raised on the appropriate channel during the last 6 weeks. I know we don't have a set timeline for proposal review and implementation (maybe it's time to set one up), but bringing up design concerns, that question the existence of the diff itself, during code review feels bad.

ps: I left 2 design flaws in this PR that I was hoping to get to in the next couple weeks. I'm happy they haven't been noticed yet, but I'm bummed that the alternative (getting the feature scraped) is worse

gallais: >  I would have appreciated if those concerned were raised on the appropriate channel during the last 6 weeks

I understand that and I am sorry that I did not react earlier but then again
we are all busy and investing time into thoroughly discussing a 4 thousand
words proposal is also a big ask.

I can easily read the test cases using the typebind operators and, although
I still think the design is broken, they behave as I expect. But I must admit
the autobind ones are still pretty confusing / unreadable.

andrevidela: I understand this is a big undertaking and I'm happy to provide additional info and context. But the review process is not the right place. We also share an office so I suggest we meet there instead.

> But I must admit the autobind ones are still pretty confusing / unreadable.

I entirely agree with this and this is why I'm am resistant to adding the overloading syntax at the same time. This is already confusing enough as it is. 

I suggest we put this back as a draft, I'll add the missing test case for regular used as typebind, fix the two flaws in the backlog (allow `_` as a name and use levenshtein distance to suggest other operators in diagnostics)

mattpolzin:
At some point it will come up that this PR's changelog entry may get merged under the v0.7.0 release unintentionally if git is left to its own devices.

andrevidela:
Any idea how to fix the self-hosting tests?

andrevidela:
I've added support for inline patterns, alternative suggestions and underscore identifier to ignore the binder

andrevidela:
Wth is going on with windows???

docs/source/reference/operators.rst
```diff
@@ -0,0 +1,340 @@
...
+    sig = VPi VStar                 (\fstTy -> VPi
+          (VPi fstTy (const VStar)) (\sndTy -> VPi
+          fstTy                     (\val1 -> VPi
+          (sndTy `vapp` val1)       (\val2 ->
+          VSigma fstTy sndTy)))))
+
+We would like to use a custom operator to build values using ``VPi``, but its
+signature does not follow the pattern that ``typebind`` uses. Instead, we use
+``autobind`` to tell the compiler that the type of the lambda is not given
+by the first argument. For this we use ``:=`` instead of ``:``:
```
andrevidela, line 313:
```suggestion
``autobind`` to tell the compiler that the type of the lambda must be inferred. 
For this we use ``:=`` instead of ``:``:
```

src/Idris/Desugar.idr
```diff
@@ -714,35 +787,55 @@ mutual
...
-  desugarTree side ps (Infix loc _ (UN $ Basic "$") l r) -- special case since '$' is special syntax
+           pure (PApp loc l' r')
+  -- normal operators
+  desugarTree side ps (Infix loc opFC (op, Just (NoBinder lhs)) l r)
       = do l' <- desugarTree side ps l
            r' <- desugarTree side ps r
-           pure (IApp loc l' r')
-  desugarTree side ps (Infix loc opFC op l r)
+           pure (PApp loc (PApp loc (PRef opFC op) l') r')
+  -- (x : ty) =@ f x ==>> (=@) exp (\x : ty => f x)
```
andrevidela, line 810:
```suggestion
  -- (x : ty) =@ f x ==>> (=@) ty (\x : ty => f x)
```

src/Idris/Error.idr
```diff
@@ -619,6 +620,89 @@ perrorRaw (BadRunElab fc env script desc)
...
+       <+> line <+> line
+       <+> "Possible solutions:" <+> line
+       <+> indent 1 (vsep (map ("-" <++>)
+           (expressionDiagnositc ++ [fixityDiagnostic, moduleDiagnostic] ++ spellingCandidates)))
+    where
+      spellingCandidates : List (Doc IdrisAnn)
+      spellingCandidates = case candidates of
+                              [] => []
+                              [x] => ["Did you mean" <++> enclose "'" "'" (pretty0 x) <++> "?"]
+                              xs => ["Did you mean one of:" <++> hcat (punctuate ", "
```
andrevidela, line 647:
```suggestion
                              xs => ["Did you mean either of:" <++> hcat (punctuate ", "
```

andrevidela (commented):
Did a self-review, if anyone else wants to chime in, feel free to do so.

~~Still no clue why the windows test fails~~

I figured it out and I'm actually thankful for the test not passing since it was an error on my part to edit the test in the first place.

src/Idris/Syntax/Traversals.idr
```diff
@@ -99,9 +99,29 @@ mapPTermM f = goPTerm where
...
+      POp fc opFC
+          <$> (NoBinder <$> goPTerm left)
+          <*> pure op
+          <*> goPTerm right
+      >>= f
+    goPTerm (POp fc opFC (BindType nm left) op right) =
+      POp fc opFC
+          <$> (BindType nm <$> goPTerm left)
+          <*> pure op
+          <*> goPTerm right
```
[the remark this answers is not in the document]
andrevidela: yes

src/Libraries/Utils/Shunting.idr
```diff
@@ -87,23 +92,25 @@ isLAssoc _ = False
...
+         then throw (GenericMsgSol loc ("Operator \{opx} is non-associative")
+                                       ["Add brackets around \{opy}"
+                                       , "Change the fixity of \{show opy} to `infixl` or `infixr`"])
          else pure (x > getPrec oy)
 higher loc opx ox opy (NonAssoc y)
     = if getPrec ox == y
-         then throw (GenericMsg loc ("Operator '" ++ show opy ++
-                                     "' is non-associative"))
+         then throw (GenericMsgSol loc ("Operator \{opy} is non-associative")
+                                       ["Add brackets around \{opy}"
```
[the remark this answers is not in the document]
andrevidela: too hard, that would require reworking the shunting algorithm

andrevidela:
Sorry I was a bit overwhelmed with deadlines this week. Indeed you can define a `typebind` or `autobind` fixity and hide it like any other fixity. I've added a test for it. There is one issue though and that's we don't check if there are conflicting fixities in the same module when you declare them, I planned to fix this as part of stage 3 of #2998 so it's not in this PR

andrevidela:
quite tempted to merge this now that it's been 4 months since the proposal, nobody cares enough to review it, and I've been using this for 3 months and I have nothing wrong to report about it

mattpolzin:
I think that’s reasonable at this point.

andrevidela:
I would be a lot more hesitant if I didn't have @edwinb's blessing but we talked about it during WITS (and on discord) and he seemed on board with the idea. In it goes

Issue #3068, opened 2023-09-02
First docstring in a file can never belong to a top level declaration


blackberryshortcake:
# Steps to Reproduce
```
||| This should document f.
f : Nat -> Nat
```
Save this file, load it in the REPL and run `:doc f`. It doesn't matter that `f` is a function, it just needs to be a top level declaration.

# Expected Behavior
We should see
```
Main> :doc f
Main.f : Nat -> Nat
  This should document f.
```

# Observed Behavior
The top level declaration `f` is missing its docstring.
```
     ____    __     _         ___
    /  _/___/ /____(_)____   |__ \
    / // __  / ___/ / ___/   __/ /     Version 0.6.0-870bc8243
  _/ // /_/ / /  / (__  )   / __/      https://www.idris-lang.org
 /___/\__,_/_/  /_/____/   /____/      Type :? for help

Welcome to Idris 2.  Enjoy yourself!
Main>       :l "FirstDocstringAlwaysBelongsToFile.idr"
1/1: Building FirstDocstringAlwaysBelongsToFile (FirstDocstringAlwaysBelongsToFile.idr)
Loaded file FirstDocstringAlwaysBelongsToFile.idr
Main>       :doc f
Main.f : Nat -> Nat
```

I think (but haven't verified) that the issue comes down to
https://github.com/idris-lang/Idris2/blob/870bc824371d504a03af937f326216302210a875/src/Idris/Parser.idr#L1902
because `doc    <- optDocumentation fname` will always run before documentation of top level declarations such as `f` is parsed.

gallais:
Nameless modules indeed have this intrinsic ambiguity.

But tbh if you're going to document a module, you should probably start by naming it.

gallais:
Yeah I mean if you're adding documentation to a module then it should probably be a named module.
Anonymous ones are only really useful as throwaways to test some code quickly.

So it is true that the grammar is ambiguous but I don't think it's worth fixing (not even sure how
we would go about doing that).

dunhamsteve:
The issue could be handled by moving the `optDocumentation` into the `option` for the module (both `prog` and `progHdr` need to match though).  The problem with that is that it then fails to parse files with no module declaration, and a docstring before an `import` statement. This happens in a half dozen tests, and I'm afraid it might affect other code (although most code should have a module declaration). 

So we can either make it an error if there is a docstring before import or paste together the docstrings before module and between module and import.

[...]

PR #3122, opened 2023-10-27
[ cleanup ] remove duplicate copy of header parsing

dunhamsteve:
This PR removes a duplicate copy of the header parser. (This was discussed in  #3068)

There are two copies of the parser code for the file header in `Parser.idr`.  One is in `progHdr`, which is used to parse the header for building the module tree and deciding whether to rebuild a file or not.  The other is in `prog` which is used to parse the entire `.idr` file.  Because there are two copies, they could accidentally diverge from each other.  This has already happened, but currently the only difference is a `Module` decoration.

Here I keep the copy in `progHdr`, adding the decoration, and have `prog` call `progHdr` and then add the declarations.

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1930,31 +1930,24 @@ import_ fname indents
          pure (MkImport (boundToFC fname b) reexp ns nsAs)
 
 export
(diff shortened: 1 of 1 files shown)


andrevidela:
omg yesss this has bugged me for years

PR #3125, opened 2023-10-27
[ contrib ] Support options parsing errors in `getOpt`

buzden:
# Description

**TLDR**: This PR adds a backward-compatible ability to parse particular option's bodies in the current `getOpt` with errors being recorded into the current `Result` type, rather than external aftermath errors parsing.

Existing simple options parsing facility is really nice, but since usually not every "body" of an option is acceptable, it is common to have the result of the option parsing to be `Either <error-type> <desired-result-type>`, and after the parsing to analyse those `Either`s. I've seen this pattern somewhere (i.e. in `pack`) and even implemented it in my code. After a while I thought of a helper function that would help with this pattern (and I implemented it in the first commit of this PR).

But this attempt showed that different errors of options parsing are present differently in the result -- the `Result` type already has an ability to represent some errors, but some other errors which are origined in those `Either`s are passes differently (and that's why, there were two functions suggested in the first commit).

But this seemed very strange to me, since these type of errors (basically, *bad argument*) do not seem to be more important than other (like *lack of an arbitrary argument*). So, I tried to integrate parsing checks into the main machinery.

This module is known to be used in several apps, so it seemed to be good to make a non-breaking change. Non-failing parsing functions are present in three places in the original module, in three data constructors. What I've done to embed parsing errors ability is that I added another three constructors (named with the original + the prime sign) that included `String -> Either String a` instead of original `String -> a` functions.

I thought of custom error data type to be supported by the parsing function, but this leads to a non-compatible change, so it was rejected internally.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
 
 * `Data.List.Lazy` was moved from `contrib` to `base`.
 
+* Existing `System.Console.GetOpt` was extended to support errors during options
+  parsing in a backward-compatible way.
+
 ## v0.7.0
 
 ### Language changes
--- a/libs/contrib/System/Console/GetOpt.idr
+++ b/libs/contrib/System/Console/GetOpt.idr
@@ -7,6 +7,7 @@ module System.Console.GetOpt
 import Data.List
 import Data.List1
 import Data.String
+import Data.These
 
 %default total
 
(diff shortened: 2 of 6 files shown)


buzden:
> I think that for once I would be in favour of a test case for some contrib content.

I'm not sure I understand the working correctly, but I did add a test (both for new and existing functionality), and rebased upon the current `main`.

mattpolzin (approved):

mattpolzin:
As caught by you on another PR, this PR's changelog entry may end up in the wrong spot now that v0.7.0 was released.

buzden:
> As caught by you on another PR, this PR's changelog entry may end up in the wrong spot now that v0.7.0 was released.

Corrected and rebased

Issue #2152, opened 2021-11-25
jobserver unavailable


xgqt:
Please check INSTALL.md and README.md to ensure you have all the required dependencies. In particular, you need a version of Chez Scheme compiled with threading support (this is the default in most distributions) in order to run the tests successfully.

Some common possible solutions:

* In `make bootstrap`, make sure you have the right exectuable name for `SCHEME`
* Stale `.ttc` files from an earlier version, or an out of data `IdrisPaths.idr` might be in the way. Try removing these with `make distclean`
* Removing all trace of Idris 2 from your installation directory might help. By default this is `$HOME/.idris2` - if you have a particularly out of date version, or you have used a broken build at some point (as much as we try to avoid this) then deleting this might help.

Otherwise, please describe the problem, including any relevant parts of build logs, and describing as much about your environment as possible.

if executed with: `make SCHEME=chez -j8 bootstrap `

the make process will complain:
```
/bin/sh ./bootstrap-stage1-chez.sh
Bootstrapping SCHEME=chezscheme IDRIS2_VERSION=0.5.1
Building idris2-boot from idris2-boot.ss
compiling idris2_app/idris2-boot.ss with output to idris2_app/idris2-boot.so
IDRIS2_CG="chez" /bin/sh ./bootstrap-stage2.sh
/home/xy/source/public/github.com/xgqt/Idris2/bootstrap-build
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
```

xgqt:
I suspect this is because bootstrap shell script calls `${MAKE}`.

attila-lendvai:
note that PR https://github.com/idris-lang/Idris2/pull/1990 should also fix this.

benhormann:
Parallel make isn't supported (currently): #864

The only eligible jobs are `network` / `test-lib`, and the sub-jobs of `support`. Since these are the smallest jobs it wouldn't be a big gain.

It'd likely be more beneficial to make the Idris2 compiler threaded. It could compile leaf nodes of the dependency graph in parallel, in theory.

Issue #864, opened 2020-12-19
Bootstrap doesn't work with parallel make


lambda-fairy:
When I use the `-j` (parallel) flag with `make bootstrap`, it fails with the error below. It succeeds when I remove it, so it's not that big of a deal, but I think it would still be nice to mention that caveat in the docs.

I'm using Arch Linux x64 with Chez Scheme.

```
$ git describe
v0.2.1-325-g60e9cf4
$ make -j4 bootstrap SCHEME=scheme
make test INTERACTIVE='' IDRIS2_PATH="/home/chris/.local/Idris2/libs/prelude/build/ttc:/home/chris/.local/Idris2/libs/base/build/ttc:/home/chris/.local/Idris2/libs/contrib/build/ttc:/home/chris/.local/Idris2/libs/network/build/ttc" IDRIS2_DATA=/home/chris/.local/Idris2/bootstrap/idris2-0.2.1/support IDRIS2_LIBS=/home/chris/.local/Idris2/bootstrap/idris2-0.2.1/lib
make[1]: Entering directory '/home/chris/.local/Idris2/support/c'
make[1]: Entering directory '/home/chris/.local/Idris2'
make[2]: Entering directory '/home/chris/.local/Idris2/tests'
./build/exec/runtests ../../../build/exec/idris2  --only 
make[2]: ./build/exec/runtests: No such file or directory
make[2]: *** [Makefile:6: test] Error 127
make[2]: Leaving directory '/home/chris/.local/Idris2/tests'
make[1]: *** [Makefile:86: test] Error 2
make[1]: Leaving directory '/home/chris/.local/Idris2'
make: *** [Makefile:172: bootstrap-test] Error 2
make: *** Waiting for unfinished jobs....
cc -Wall  -fPIC -O2   -c -o getline.o getline.c
cc -Wall  -fPIC -O2   -c -o idris_buffer.o idris_buffer.c
cc -Wall  -fPIC -O2   -c -o idris_directory.o idris_directory.c
cc -Wall  -fPIC -O2   -c -o idris_file.o idris_file.c
cc -Wall  -fPIC -O2   -c -o idris_net.o idris_net.c
cc -Wall  -fPIC -O2   -c -o idris_support.o idris_support.c
cc -Wall  -fPIC -O2   -c -o idris_term.o idris_term.c
ar rc libidris2_support.a getline.o idris_buffer.o idris_directory.o idris_file.o idris_net.o idris_support.o idris_term.o
[...]

PR #3135, opened 2023-11-07
Support "make -j"

depressed-pho:
# Description

There are two cases where `make -j` fails: `make bootstrap -j` and `make all -j`. I fixed them both in two separate commits. 

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -301,6 +301,7 @@
   can be set to install into a staging directory for distro packaging.
 * Updates the docs for `envvars` to categorise when environment variables are
   used (runtime, build-time, or both).
+* Fixed build failure occuring when `make -j` is in effect.
 
 ## v0.6.0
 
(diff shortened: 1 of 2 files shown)


CodingCellist:
I'm slightly cautious of this: parallel builds with `-j` are unsupported by Idris2 in general [1, 2] and that hasn't changed (afaik)...  
If it turns out we actually _can_ build in parallel that's of course fantastic! But I'd like to know if this speeds anything up or if it just still runs a single thread  : )

[1] https://github.com/idris-lang/Idris2/issues/864#issuecomment-949947832
[2] https://github.com/idris-lang/Idris2/issues/2152#issuecomment-986307777

depressed-pho:
With this patch you can at least build `libs/network` and `libs/contrib` parallelly. Not a huge win, but is still better than nothing (I believe).

dunhamsteve:
I wouldn't expect it to do much, since the build process is largely serialized (aside from tests, which we already parallelize), but it does seem to shave about 7 % off the 1h 15m CI task.  Either way, I think it's worth it if it keeps the build from failing if a user decides to pass the `-j` option to make.

PR #3139, opened 2023-11-10
[ base ] Add `mapPropertyRelevant`, `tabulate` and `(++)` for `Vect`'s `All`

0xd34df00d:
# Description

Adding some more functions that are useful in practice and might be not exactly trivial to implement (like `tabulate`'s erased `xs` but unerased `n` to allow for pattern-matching on the shape of `xs`).

      `CONTRIBUTORS.md`).

### DIFF
--- a/libs/base/Data/Vect/Quantifiers.idr
+++ b/libs/base/Data/Vect/Quantifiers.idr
@@ -126,6 +126,16 @@ namespace All
   mapProperty f [] = []
   mapProperty f (p::pl) = f p :: mapProperty f pl
 
+  ||| A variant of `mapProperty` that also allows accessing
+  ||| the values of `xs` that the corresponding `ps` prove `p` about.
+  export
+  mapPropertyRelevant : {xs : Vect n a} ->
+                        (f : (x : a) -> p x -> q x) ->
+                        (ps : All p xs) ->
+                        All q xs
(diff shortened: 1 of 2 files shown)


libs/base/Data/Vect/Quantifiers.idr
```diff
@@ -195,6 +205,19 @@ namespace All
...
+  public export
+  tabulate : {n : _} ->
+             {0 xs : Vect n _} ->
+             (f : (ix : Fin n) -> p (ix `index` xs)) ->
+             All p {n} xs
+  tabulate {xs = []} f = []
+  tabulate {xs = _ :: _} f = f FZ :: tabulate (\ix => f (FS ix))
+
+  public export
+  (++) : (axs : All p xs) -> (ays : All p ys) -> All p (xs ++ ys)
```
buzden, line 217: Why not adding the same functions to `Data.List.Quantifiers.All`?

0xd34df00d: You mean just these two or all of what this PR adds?

buzden: I meant at least these two, but I think all of them worth adding

gallais:
Merging as is.
If there is motivation to actually push the other ones too then open another PR.

PR #3140, opened 2023-11-11
[ gambit ] fix pack / unpack segfault with gambit scheme

dunhamsteve:
This PR fixes a segfault involving `unpack` in gambit and syntax errors in `support.scm` when using the latest gambit. 

The following code segfaults when compiled with the gambit backend:
```idris
main : IO ()
main = printLn "hello, world"
```
The root cause was that `pack` / `unpack` expect a different list representation than Idris currently uses. I suspect it was written before Idris mapped lists to native scheme lists. 

I also discovered that gambit 4.9.5 reports syntax errors for `support.scm`. It doesn't like the use of `[]` instead of `()`.  Gambit 4.9.3 on a different platform is happy with the brackets. I couldn't track down when this changed, but did find that R7RS reserves `[]` and `{}` for future use.

I've addressed both of these issues in this PR.  The code has been tested locally with:
```idris
main : IO ()
main = printLn $ fastConcat [ "Hello, ", "world"]
```
which exercises `pack`, `unpack`, and `fastConcat`.

Other issues remain - the FFI doesn't support passing `Buffer` and a lot of the `blodwen-` functions are missing (including buffer and threading functions). I briefly investigated, but decided to punt.

### DIFF
--- a/support/gambit/support.scm
+++ b/support/gambit/support.scm
@@ -2,10 +2,10 @@
 ;; https://github.com/gambit/gambit/blob/master/gsc/_t-x86.scm#L1106 #L1160
 (define (blodwen-os)
   (cond
-    [(memq (cadr (system-type)) '(apple)) "darwin"]
-    [(memq (caddr (system-type)) '(linux-gnu)) "unix"]
-    [(memq (caddr (system-type)) '(mingw32 mingw64)) "windows"]
-    [else "unknown"]))
+    ((memq (cadr (system-type)) '(apple)) "darwin")
+    ((memq (caddr (system-type)) '(linux-gnu)) "unix")
+    ((memq (caddr (system-type)) '(mingw32 mingw64)) "windows")
+    (else "unknown")))
 
 ;; TODO Convert to macro
 (define (blodwen-read-args desc)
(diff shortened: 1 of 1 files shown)


gallais:
Do we even want to keep the gambit backend? We turned the CI for it off because it would just diverge on some (simple) numerical tests.

buzden:
As far as I remember, originally it was the only truly compilable-to-native portable standard backend, but now we have refc one, so, maybe the main reason behind gambit has gone

dunhamsteve:
Yeah, gambit seems slow, too. I tried building idris with it and gave up after 3.5 hours and 12 GB of memory. It seems like RefC is our best path to C. 

It looks like the build results are smaller than other schemes, but there are a lot of holes in the FFI and blodwen function coverage, so I’m not sure how useful it would be for actual programs. RefC is probably a better fit for that use case.

I originally looked into this because I wanted to test #3137 on all of the scheme backends. I didn’t know Gambit backend was essentially abandoned until later. Leaving it around may incur a small maintenance burden.

andrevidela:
We should probably remove it at some point, maybe move it into another repo as a library

andrevidela:
This looks good to me, I suggest we merge it and start thinking about moving gambit out of the main repo. Anyone objects?

PR #3141, opened 2023-11-11
[ error ] Improve error messages when reading ipkg files

dunhamsteve:
The parser for `.ipkg` files reports any error as "Expected end of input". The root cause was a "many" and an "alt" that eat the error messages. Here I've added `mustWork` where appropriate, a better error message for when it hits an unknown keyword, and added the original source to the error message.

Previously:
```
Uncaught error: Parse errors ([("bad.ipkg":3:1--3:7, "Expected end of input.")])
```
Now it says:
```
Uncaught error: Error: Unrecognised property "depend".

"bad.ipkg":3:1--3:7
 1 | package idris2app
 2 |
 3 | depend = network
     ^^^^^^
```
The other test case reports this instead of "Expected end of input":
```
Uncaught error: Error: Expected string.

"bad2.ipkg":2:13--2:16
 1 | package bad
 2 | sourcedir = src
                 ^^^
```

### Should this change go in the CHANGELOG?

Probably not significant enough to warrant a change log entry.

### DIFF
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -34,6 +34,7 @@ import Libraries.Text.PrettyPrint.Prettyprinter.Render.String
 import Idris.CommandLine
 import Idris.Doc.HTML
 import Idris.Doc.String
+import Idris.Error
 import Idris.ModTree
 import Idris.Pretty
 import Idris.ProcessIdr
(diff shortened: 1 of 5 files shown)


gallais (approved):
Great stuff!

Issue #3072, opened 2023-09-07
Document `%unsafe`


andrevidela:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## What needs to be documented

We have added `%unsafe` in but there is not documentation for it.

## Current status

| Documented  | Area |
| ------------- | ------------- |
| N/A | REPL |
| <ul><li>[ ] </li></ul> | Online  |
| N/A  | Source code  |
| N/A | Code samples |

dunhamsteve:
If we want to go further, here are all of the pragmas I found in `Parser.idr` and the documentation files they appear in, if any.  There is a second list of pragmas in `Pragmas.idr` - I believe it is only used for auto-completion.

| pragma | documents |
| ------ | --------- |
| MkWorld | backend-cookbook.rst, multiplicities.rst |
| TTImpLit |
| World | backend-cookbook.rst, multiplicities.rst |
| allow_overloads |
| ambiguity_depth | updates.rst |
| auto_implicit_depth |
| auto_lazy |
| builtin | builtins.rst, pragmas.rst |
| charLit |
| declsLit |
| default | updates.rst, parsing.rst, theorems.rst |
| defaulthint |
| deprecate | pragmas.rst |
| doubleLit |
| export |
| extern |
| foreign | ffi.rst, readline.rst, refc.rst, backend-cookbook.rst, chez.rst, gambit.rst, libraries.rst, javascript.rst, racket.rst |
| globalhint |
| hide | pragmas.rst |
| hint | updates.rst, miscellany.rst |
| inline | pragmas.rst |
| integerLit | overloadedLit.rst |
| language | updates.rst |
| logging | debugging.rst |
| macro |
| name | pragmas.rst |
| nameLit |
| nf_metavar_threshold |
| noinline | pragmas.rst |
| nomangle | libraries.rst (this is deprecated, use export) |
| pair |
| prefix_record_projections | records.rst, typesfuns.rst |
| rewrite |
| runElab |
| search |
| search_timeout |
| spec |
| start |
| stringLit | overloadedLit.rst |
| syntactic |
| tcinline |
| transform | builtins.rst |
| unbound_implicits | overview.rst |
| unhide |
| unsafe |

Also, `%auto_implicits` is mentioned in `faq.rst`, but it does not exist. I believe `%unbound_implicits` was intended.

[...]

PR #3148, opened 2023-11-22
[ doc ] Add missing pragmas to documentation

dunhamsteve:
# Description

This PR adds documentation for all of the pragmas in Idris. It is based off of the list in the comments of #3072.

There are 43 of them, so I broke them into sections:

- Global pragmas
- Pragmas on declarations
- Internal
- Expressions

Let me know if there is a better way to organize them.

It still needs some clarification on how `%globalhint` differs from `%defaulthint`. I copied the description for `%syntactic` from a comment in the source code, but I don't fully understand what it does.

I found that `%allow_overloads` is no longer used and documented that. It updates the configuration, but nothing reads that data. It was disabled here:

https://github.com/idris-lang/Idris2/blob/d6aa70dbe16005fbbf68e7525be3784525f4f2ec/src/TTImp/Elab/Ambiguity.idr#L294-L298

We don't seem to have any documentation on elab reflection in `docs`. It'd be nice to add that someday.

At some point, `libraries.rst` should be rewritten or removed.  It currently references a deprecated pragma, and the first two examples now give errors.

No, it's just updates to documentation.

### DIFF
--- a/docs/source/reference/pragmas.rst
+++ b/docs/source/reference/pragmas.rst
@@ -14,17 +14,127 @@ small niche of pragmas apply directly to one or more arguments instead of the co
     This page is a work in progress. If you know about a pragma that is not described yet, please consider
     submitting a pull request!
 
+Global pragmas
+====================
+
+``%language``
+--------------------
+
+Enable language extensions.  Currently, the only extension is ``ElabReflection``.
+
+.. code-block:: idris
+
+   %language ElabReflection
+
+``%default``
+--------------------
+
+Set the default totality requirement for functions.  The initial value is `covering`, which
+required case statements to cover, but does not require totality. The other options are
+`total` and `partial`.
 
 ``%builtin``
-====================
+--------------------
 
 The ``%builtin Natural`` pragma converts recursive/unary representations of natural numbers
 into primitive ``Integer`` representations.
 
 This pragma is explained in detail on its own page. For more, see :ref:`builtins`.
 
+
+``%name``
+--------------------
+
+Give the compiler some suggested names to use for a particular type when it is asked to generate names for values.
+You can specify any number of suggested names; they will be used in-order when more than one is needed for a single
+definition.
+
+.. code-block:: idris
+
+   data Foo = X | Y
+
+   %name Foo foo,bar
+
+
+``%ambiguity_depth``
+--------------------
+
+Set how deep nested ambiguous names can be before Idris gives up. The default is 3, increashing this
+will effect compiler performance. For more details, see :ref:`ambiguous-name-resolution`.
+
+``%auto_implicit_depth``
+------------------------
+
+Set the search depth for ``auto`` implicits. The default value is 50.
+
+``%logging``
+--------------------
+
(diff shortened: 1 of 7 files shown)


docs/source/reference/pragmas.rst
```diff
@@ -14,17 +14,127 @@ small niche of pragmas apply directly to one or more arguments instead of the co
...
+Enable language extensions.  Currently, the only extension is ``ElabReflection``.
+
+.. code-block:: idris
+
+   %language ElabReflection
+
+``%default``
+--------------------
+
+Set the default totality requirement for functions.  The initial value is `covering`, which
```
buzden, line 32: Maybe, it's worth adding that `total` can be an initial value if an appropriate flag is passed to the compiler

docs/source/reference/pragmas.rst
```diff
@@ -124,3 +226,247 @@ You can also hide fixity declarations if you need to redefine your own.
...
+``%hint``
+--------------------
+
+Mark a function to be used for ``auto`` search (see :ref:`auto-implicits` for more).
+
+
+``%defaulthint``
+--------------------
+
+Mark a default for functions like ``fromString`` in cases where the compiler cannot
```
buzden, line 292: This seems to give a wrong to impression. I would say that this is like a `%hint` which is tried when none of the proper `%hint`s succeeded.

docs/source/reference/pragmas.rst
```diff
@@ -124,3 +226,247 @@ You can also hide fixity declarations if you need to redefine your own.
...
+``%nomangle``
+--------------------
+
+This pragma is deprecated.  Instead use ``%export`` to expose functions to the backend.
+
+
+``%hint``
+--------------------
+
+Mark a function to be used for ``auto`` search (see :ref:`auto-implicits` for more).
```
buzden, line 286: Maybe, it's important to add that every non-named implementation receives this pragma automatically. Also, it may be added that every data constructor effectively receives this unless `[noHints]` is present in the declaration of the data type.

docs/source/reference/pragmas.rst
```diff
@@ -124,3 +226,247 @@ You can also hide fixity declarations if you need to redefine your own.
...
+``%defaulthint``
+--------------------
+
+Mark a default for functions like ``fromString`` in cases where the compiler cannot
+determine which type the user wants.
+
+``%globalhint``
+--------------------
+
+This pragma is similar to ``%defaulthint``.
```
buzden, line 298: As far as I know, this is similar to `%hint`, not `%defaulthint`, but the rules of search are different. `%hint`s have determining type arguments, but `%globalhint`s are tried anyway.

dunhamsteve: I see. The `%globalhint` are stored with the `%defaulthint`, because neither relies on the type., but they're used very differently.  It looks like `%globalhint` wins over `%hint`, too.

Thanks for the feedback.  I'll try to get an update out tomorrow.

andrevidela:
Seems like all the feedback has been addressed and it looks good to me. 

Maybe we can start thinking about removing the deprecated feature soon

PR #3150, opened 2023-11-23
[ error ] Fix error message for %defaulthint on a local function

dunhamsteve:
# Description

If `%deaulthint` is used in a local definition, Idris complains "%globalhint is not valid in local definitions`.  This PR changes it to say "%defaulthint is not valid in local definitions" in that case.

```idris
foo : Nat -> Nat
foo x = x
    where
        %defaulthint
        bar : Show String
```
gives the error message:
```
Error: While processing right hand side of foo. %globalhint is not valid in local definitions
```
With the fix, it now says:
```
Error: While processing right hand side of foo. %defaulthint is not valid in local definitions
```


andrevidela:
Looks good to me

PR #3151, opened 2023-11-28
[ refactor ] Split Core.TT

gallais:


src/Compiler/LambdaLift.idr
```diff
@@ -417,12 +362,12 @@ mutual
            -- Find out which variables aren't used in the new definition, and
            -- do not abstract over them in the new definition.
            let scUsedL = usedVars initUsed scl
-               unusedContracted = contractUsedMany {remove=bound} scUsedL
-               unused = getUnused unusedContracted
-               scl' = dropUnused {outer=bound} unused scl
+               unusedContracted = ?scUsedL
+               us = ?zgh -- getUnused unusedContracted
```
gallais, line 366: oops

src/Core/Name/Scoped.idr
```diff
@@ -0,0 +1,264 @@
...
+keeps : (args : List a) -> Thin xs ys -> Thin (args ++ xs) (args ++ ys)
+keeps [] th = th
+keeps (x :: xs) th = keep (keeps xs th)
+
+namespace Thin
+  -- At runtime, Thin's `Refl` does not carry any additional
+  -- information. So this is safe!
+  export
+  embed : Thin xs ys -> Thin (xs ++ outer) (ys ++ outer)
+  embed = believe_me
```
Z-snails, line 127: The following definition:
```haskell

embed : Thin xs ys -> Thin (xs ++ outer) (ys ++ outer)
embed Refl = Refl
embed (Drop x) = Drop (embed x)
embed (Keep x) = Keep (embed x)
```
compiles to the identity function and avoids using `believe_me`.
On the other hand using `believe_me` is less fragile in case that optimisation is removed. (The optimisation here is the same one used to optimise `toInteger` and `fromInteger` for arbitrary `Nat` shaped types)

gallais: Yep. It's not even used in the codebase AFAICT so I may as well just drop it.

PR #3152, opened 2023-11-28
[ cli ] option to disable CSE

AlgebraicWolf:
# Description

Sometimes, when debugging code generation, it seems useful to perform `diff` on two versions of generated code. However, because of common subexpression elimination, sometimes small changes may lead to vastly different resulting scheme files. 

This patch adds a new compiler flag, `--no-cse`, that disables CSE during code generation. 

      `CONTRIBUTORS.md`).


AlgebraicWolf:
I've figured out what I was doing wrong, now it works as intended.

PR #3153, opened 2023-11-28
[ fix ] Fix docs build

dunhamsteve:
It looks like `sphinx-ci.yml` is just running sphinx to see if it succeeds.  It doesn't seem to do anything with the results.  On this assumption, I've replaced the `ammaraskar/sphinx-action` action with a direct invocation of sphinx in the ubuntu container.

### DIFF
--- a/.github/workflows/ci-sphinx.yml
+++ b/.github/workflows/ci-sphinx.yml
@@ -21,6 +21,6 @@ jobs:
     - name: Build docs
       run: |
         sudo apt-get install -y python3-sphinx
-        pip3 install sphinx-rtd-theme
+        pip3 install -r requirements.txt
         cd docs
         make html
--- a/.github/workflows/ci-sphinx.yml
+++ b/.github/workflows/ci-sphinx.yml
@@ -21,6 +21,6 @@ jobs:
     - name: Build docs
       run: |
         sudo apt-get install -y python3-sphinx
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-sphinx.yml
```diff
@@ -15,7 +15,12 @@ jobs:
...
-    - uses: ammaraskar/sphinx-action@master
-      with:
-        docs-folder: "docs/"
+
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Build docs
+      run: |
+        sudo apt-get install -y python3-sphinx
+        pip3 install sphinx-rtd-theme
```
buzden, line 24: We should install there everything from `docs/requirements.txt` in order to be compatible with how RTD builds the docs (for potential future changes)

.github/workflows/ci-sphinx.yml
```diff
@@ -21,6 +21,6 @@ jobs:
     - name: Build docs
       run: |
         sudo apt-get install -y python3-sphinx
-        pip3 install sphinx-rtd-theme
+        pip3 install -r requirements.txt
```
gallais, line 24:
```suggestion
-        pip3 install -r requirements.txt
+        pip3 install -r docs/requirements.txt
```

PR #3159, opened 2023-12-07
[ doc ] Improve docs for let and :=

CodingCellist:
# Description

Prompted by discussion on the Discord: there was confusion as to when computations would be re-evaluated, whether type annotations made a difference (and when), the difference between `let` with `=` and with `:=`.

The string for `:doc (:=)` was not very helpful; only a single line mentioning what the symbol could be used for. This has now been expanded, and the expanded version lifted out to its own string so as to be (re)used in both `:=` and `let`'s documentation.

      `CONTRIBUTORS.md`).


src/Idris/Doc/Keywords.idr
```diff
@@ -505,31 +505,66 @@ withabstraction = vcat $
...
+    power4 : Nat -> Nat
+    power4 n = let square := n * n
+               in square * square
+
+    power4T : Nat -> Nat
+    power4T n = let square : Nat := n * n
+                in square * square
+    ```
+    Here, `square` will only be computed once instead of every time (which might
+    have happened had we used `=` instead of `:=`).
```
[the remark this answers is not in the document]
CodingCellist: Ah no that's absolutely my own misunderstanding; I thought `=` and `:=` were always different, not just when used with a type annotation. I'll correct that  : )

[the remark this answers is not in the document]
CodingCellist: I think I've now correctly described the behaviour?  : )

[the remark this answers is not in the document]
buzden: > I thought `=` and `:=` were always different, not just when used with a type annotation

Historically, as far as I remember, `:=`-syntax was added to be able to disambiguate expressions like `let x : a = b = <equality-providing-expresion> in ...`

src/Idris/Doc/Keywords.idr
```diff
@@ -505,31 +505,69 @@ withabstraction = vcat $
...
+    power4T n = let square : Nat := n * n
+                in square * square
+    ```
+    Without the type annotation, `:=` behaves the same as `=` and gets
+    elaborated to the lambda expression `(\square => square * square) n * n`.
+    With the type annotation, `square` is the value of a computation, not its
+    generalisation, and so will only be computed once instead of potentially on
+    every reference.
+    Let bindings will not unfold in the type of subsequent terms so may not be
+    appropriate in all cases.
```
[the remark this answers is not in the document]
emdash: I agree with that. I have run into this myself, and did not understand what was happening.

emdash:
Thanks for doing this. I am not officially a contributor here, but it gets my vote.

CodingCellist:
@emdash @buzden, thank you both for looking this over! I think I've incorporated the feedback, let me know what yous think (after the holidays if you're away; please don't take time away from a break to look at this)  : )

src/Idris/Doc/Keywords.idr
```diff
@@ -505,31 +505,71 @@ withabstraction = vcat $
...
+    power4T : Nat -> Nat
+    power4T n = let square : Nat := n * n
+                in square * square
+    ```
+    Without the type annotation, `:=` behaves the same as `=` and gets
+    elaborated to the lambda expression `(\square => square * square) (n * n)`.
+
+    With the type annotation, `square` is the value of a computation, not its
+    generalisation, and so will only be computed once instead of potentially on
+    every reference.
```
gallais, line 529: I don't understand these paragraphs

src/Idris/Doc/Keywords.idr
```diff
@@ -699,6 +739,24 @@ recordUpdate = vcat $ header "Record updates" :: ""
   """
   ]
 
+letOrRecord : Doc IdrisDocAnn
+letOrRecord = vcat $
+  header "Let binding or record assignment"
```
gallais, line 744:
```suggestion
  header "Let binding or record update"
```

src/Idris/Doc/Keywords.idr
```diff
@@ -699,6 +739,24 @@ recordUpdate = vcat $ header "Record updates" :: ""
...
+  :: ""
+  :: map (indent 2) [
+    """
+    \{letBinding}
+
+
+    * Record assignment uses `:=` to assign individual fields to a new value.
```
gallais, line 751:
```suggestion
    * Record updates use `:=` to set a field's value.
```

mattpolzin:
The changelog entry on this PR will need to be updated now that v0.7.0 was released.

src/Idris/Doc/Keywords.idr
CodingCellist: I've attempted a clarification in the latest commit. Does that help?  : )

emdash: It makes sense to me. I'd be inclined to ship this, as it's better than the current help text. But it's not up to me.

One thing I would like more clarification on is how the two forms might affect type inference. I.e. certain things work with one form or the other, and that can be quite surprising. I still don't have a good intuition for this.

src/Idris/Doc/Keywords.idr
```diff
@@ -505,31 +505,71 @@ withabstraction = vcat $
...
+    power4T : Nat -> Nat
+    power4T n = let square : Nat := n * n
+                in square * square
+    ```
+    In the first example, the type annotation is omitted and so `:=` gets
+    elaborated to the lambda expression `(\square => square * square) (n * n)`.
+
+    The second example does have a type annotation and as a result, `square`
+    gets computed rather than lambda-abstracted, meaning the result will be
+    reused on reference, instead of potentially being recomputed every time.
```
gallais, line 529: I don't think that's true. The code for parsing let bindings is as follows:

```
    letBinder : Rule LetBinder
    letBinder = do s <- bounds (MkPair <$> multiplicity fname <*> expr plhs fname indents)
                   (rig, pat) <- pure s.val
                   ty <- option (PImplicit (virtualiseFC $ boundToFC fname s))
                                (decoratedSymbol fname ":" *> typeExpr (pnoeq pdef) fname indents)
                   (decoratedSymbol fname "=" <|> decoratedSymbol fname ":=")
                   val <- typeExpr pnowith fname indents
                   alts <- block (patAlt fname)
                   pure (MkLetBinder rig pat ty val alts)
```

As you can see if there is no type annotation, we replace it with a hole.

And so computationally the two expressions are the same.

gallais (changes requested):

PR #3162, opened 2023-12-12
[fix] correctly pass environment vars to the RefC compiler

AdamHarries:
This patch fixes some oversights in `CC.idr`, and passes the contents of the `CFLAGS` and `LDFLAGS` environment variables down to the RefC C compiler. 

Additionally, it introduces a simple test exercising this behaviour, by building a simple shared library, compiling it, and then correctly using it from Idris2. See the comment in `tests/RefC/ccompilerArgs/run` for more details.

      `CONTRIBUTORS.md`).


AdamHarries:
It looks like the newly added test is failing on Ubuntu/Nix (but not windows?) which is interesting. I'll try and dig in when I have time.

src/Compiler/RefC/CC.idr
```diff
@@ -41,6 +43,14 @@ findLDFLAGS
          | Just ldflags => pure ldflags
        pure ""
 
+findLDLIBS : IO String
+findLDLIBS
+  = do Nothing <- getEnv "IDRIS2_LDLIBS"
```
buzden, line 48: It should be `idrisGetEnv` instead of `getEnv` in the whole file, because this function checks for the presence of used environment variables in the help description.

tests/refc/ccompilerArgs/library/Makefile
```diff
@@ -0,0 +1,10 @@
+all: libexternalc.so
+
+externalc.o: externalc.c externalc.h
+	cc -c -fPIC $< -o $@
```
mattpolzin, line 4: Although it may never happen to be necessary, I suggest `$(CC)` here (and on line 7 below) for portability.
```suggestion
	$(CC) -c -fPIC $< -o $@
```

AdamHarries: Agreed. I had something like this in a previous version of this patch, but had some difficulties with my setup. I'll give it another shot, as I agree that it is the way we *should* be doing it.

AdamHarries: Also, it may be that this is the cause of the failing makefile on NIX & Ubuntu.

src/Compiler/RefC/CC.idr
AdamHarries: Done. It does introduce a dependency on `Idris.Env`, but I think that's acceptable.

mattpolzin:
Interestingly, I cannot reproduce the above failure when I check this branch out and build it via Nix locally. You may have to add a commit where you don't dump all the build output to `/dev/null` to see if anything in there is enlightening!

AdamHarries:
> Interestingly, I cannot reproduce the above failure when I check this branch out and build it via Nix locally. You may have to add a commit where you don't dump all the build output to `/dev/null` to see if anything in there is enlightening!

Ah, I was hoping to avoid that! But given that you can't reproduce it separately it seems like it's the only reasonable course of action. At least we should get some sort of error!

AdamHarries:
I think the issue is with the -pie flag to the linker, which should be vaguely harmless, but it's possible that I'm wrong about that. I've replaced it with a different "harmless" flag, that should be even more harmless!

mattpolzin:
Looks like you got it!

mattpolzin:
This looks pretty good to me, but I'll give @madman-bob a ping and some time to check in on this.

src/Compiler/RefC/CC.idr
```diff
@@ -68,6 +79,8 @@ compileCObjectFile {asLibrary} sourceFile objectFile =
               "-o", objectFile,
               "-I" ++ refcDir,
               "-I" ++ cDir]
+              ++ (words cppFlags)
+              ++ (words cFlags)
```
madman-bob, line 83: Using `words` here will cause issues with quoted spaces, such as in filenames. This is one of the few cases where we *don't* want to escape a string for a command. It should be concatenated on as a string, after `escapeCmd` has run.

eg. `escapeCmd (...) ++ " " ++ cppFlags ++ " " ++ cFlags`

AdamHarries: Ah, that's an alternative approach that I hadn't considered! I'll give it a shot and see if it still has the correct behaviour.

AdamHarries: From a local test, it looks like that also has the behaviour that we want, so I've made the change and I'm running it through CI now.

mattpolzin (approved):

buzden (approved):

PR #3163, opened 2023-12-12
[doc] add note regarding C header inclusion to FFI documentation

AdamHarries:
A small addition to the FFI documentation showing an example of including a header file in the generated C code.


mattpolzin:
Nice addition! Thanks.

mattpolzin (approved):

mattpolzin:
My only suggestion would be that it could be moved under the "C Sidenote" below which could be renamed "C Details" but I would merge this as is, too.

AdamHarries:
> My only suggestion would be that it could be moved under the "C Sidenote" below which could be renamed "C Details" but I would merge this as is, too.

I considered doing this, but was unsure as to whether the "C Sidenote" was a sidenote to the Scheme section, or about RefC in general. I think it might be worth a follow-up patch to document each backend a little bit more clearly.

PR #3164, opened 2023-12-14
[ refactor ] Move lazy lists from `contrib` to `base`

buzden:
# Description

We have `List`, `SnocList` and `Stream` in `prelude`, we have `Colist` and `Colist1` in `base`, but (for historical reasons) still have `LazyList` in `contrib`, despite this is even used in the compiler itself. I suggest to finally move `LazyList` to `base` from `contrib`, I think it should be near its closest relatives, like `Colist`. Moreover, I know at least several libraries that currently depend on `contrib` solely because of lazy lists.

This should be a non-breaking change, because nothing changes at the use site except for dropped need for depending on `contrib` (which can still stay with no harm).

      `CONTRIBUTORS.md`).


mattpolzin:
Those seem like good arguments for moving this into base. I think the compiler argument perhaps resonates with me the most since otherwise I try to be in the mindset that base was originally intended to be quite small and yet the number of modules in it grows quite a lot each release. 

Do you mind taking a quick pass at this now that 0.7.0 is out to make sure the changlelog entry ends up in the right place and the note about when to remove the compilers copy of lazy list is accurate and clear?

buzden:
> Do you mind taking a quick pass at this now that 0.7.0 is out to make sure the changlelog entry ends up in the right place and the note about when to remove the compilers copy of lazy list is accurate and clear?

Sure, I did it!

mattpolzin (approved):
Did you double check that the compiler and base (previously contrib) files for lazy list were the same?

buzden:
> Did you double check that the compiler and base (previously contrib) files for lazy list were the same?

They were not, because `Data.List.Lazy` evolved over time, including some fixes, but `Libraries.List.Lazy` stayed untouched last years. Do you mean, we should update `Libraries.List.Lazy` in this PR in order to make removal to be guaranteedly safe after the next release?

mattpolzin:
I was more concerned about the compiler code being incompatible than I was with it lacking some fixes and such. I don’t think you need to align the two in this PR as long as it looks to you like it’ll be painless enough to switch implementations eventually.

buzden:
Well, I now checked that one dependent module indeed needed to be synchronised with the `List.Lazy` module, so, it's better to do in one hop. So, I added an adiitonal commit for the sync.

mattpolzin (commented):
I might be missing something, but did you more or less bring things into the `src/Libraries` version of `LazyList` simply because they were in the `contrib` version?

I think I would avoid adding unused definitions to the `src/Libraries` version; sure, those will exist in the `base` version once the compiler switches over, but no need to add unneeded definitions in the meantime. I could see an argument for making implementation changes where appropriate now because that will give us more time to catch bugs in the compiler based on implementation details of the `base` version.

The other kind of change that I think is good is adding things to `base` now if they will need to be there in order for the `src/Libraries` version to be removed anyway.

buzden:
> I might be missing something, but did you more or less bring things into the `src/Libraries` version of `LazyList` simply because they were in the `contrib` version?

The main reason why I did this is to make sure the current state of the compiler source is ready for switching to `LazyList` from `base` and will remain so till the next release (including potential future ambiguities or whatever). Do you think it is not important? If so, we can remove the last commit.

mattpolzin:
Well, I'm not sure anymore. I had hoped they were already nearly the same because that would be easiest. My second choice would be making as few changes as possible to make the transition to `base` almost just a file deletion. The only reason I mention that it appears you've copied some definitions into src/Libraries that aren't used is because if we do find a compiler regression we need to fix it would be nice to have a small diff to investigate the problem.

buzden:
> The only reason I mention that it appears you've copied some definitions into src/Libraries that aren't used is because if we do find a compiler regression we need to fix it would be nice to have a small diff to investigate the problem.

In the end, I don't understand if you insist on removing unused functions from `src/Libraries` or not. My position it that it's good for them by be in sync despite some functions being unused, but I don't understand your point from your wording.

But nevertheless, regarding to actually used functions, I think we should stick to implementations copied from `contrib`, because
1) some implementations were changed because the original ones used to break laziness;
2) if they would cause any regressions, it's better to find them between releases, not right after the new one, that's why they should go into `src/Libraries` before their delition on the next release.

mattpolzin:
I’m not insisting on one way or the other, just talking through my thoughts on process for this kind of move. Either way the change is being made between releases, we are currently just after this release and otherwise we’d be just after the next one. Still, I was trying to say that I agree it would be nice to make implementation changes to src/Libraries now as you just did; not because it is a better time for it (between releases) but because it gives a more tangible git changeset to look at if something breaks. The line by line diff now is easier than a whole file deletion later. 

You can take or leave my point about not adding definitions that aren’t used at this point in time. Not sure I can explain my reasoning there much better: fewer git diff lines makes it easier to find regressions if they occur.

PR #3165, opened 2023-12-14
[ fix ] Correctly manage the empty file case while reading a file

buzden:
Currently, if you try to `fGetChars` on an empty file, you get a `FileError`, which being printed gives, strangely, `Success`. I digged into the implementation an realised that `fread` that is used internally is *specified* to not to distinguish between a file error an empty file, thus requiring to analyse the situation using `ferror`. So, I changed the implementation to do so.

Also, some memory leaked in the case of empty file, to I fixed this. A test showing the difference in the behaviour is added (I used not only empty files in the test, just in case).


libs/base/System/File/ReadWrite.idr
```diff
@@ -54,10 +54,10 @@ fSeekLine (FHandle f)
 
 ||| Get a garbage collected String from a Ptr String and and free the original
 export
-getStringAndFree : HasIO io => Ptr String -> io (Either FileError String)
-getStringAndFree res
+getStringAndFree : HasIO io => File -> Ptr String -> io (Either FileError String)
```
mattpolzin, line 57: I'm not really sure why this is `export`ed from the `File.ReadWrite` module, but since it is, can you add a note about this breaking change to the CHANGELOG?

mattpolzin (changes requested):

libs/base/System/File/ReadWrite.idr
buzden: Right. I also have no idea why this is exported (at least, it is not used anywhere in the standard libs), but I did add this change into the changelog. I also realised that we have a separate `System` subsection in the `Base` section of the changelog, and I moved couple already exitstent items there.

CHANGELOG.md
```diff
@@ -254,17 +252,22 @@
...
+* `System`'s `die` now prints the error message on stderr rather than stdout
+
 * Adds `popen2` to run a subprocess with bi-directional pipes.
 
+* A function `popen2Wait` was added to wait for the process started with `popen2`
+  function and clean up all system resources (to not to leave zombie processes in
+  particular).
+
+* Function `getStringAndFree` from `System.File.ReadWrite` was given an extra
+  argument of type `File` to return an empty string is no error happened.
```
mattpolzin, line 269:
```suggestion
  argument of type `File` to return an empty string if no error happened.
```

PR #3166, opened 2023-12-14
[ cleanup ] Clean up stuff that needs to be cleaned after 0.6.0 release

buzden:
We have a lot of stuff postponed till the next release because we have a requiement that the current Idris has to be able to be built with the previous release version. But sometimes, these postponed stuff is forgotten to be done. I've found several things that should have been done after the current release has been released more than a year ago. So, let's do this.

Also, I've changed some comments to contain the word 'release' to ease search for the future cleanups, and changed comments of style "after release after version 0.6.0" to comments of style "after 0.7.0 is released" because we had comments in both styles, and I personally find the first style a bit confusing.


dunhamsteve:
There is also this, but I'm not sure what needs to be done - maybe a question for @Z-snails:

https://github.com/idris-lang/Idris2/blob/4cb45daba4b6b1c32f2f0dd7195daebe531cb335/src/Compiler/CompileExpr.idr#L196-L210

buzden:
Right, I did what I understand, this is not an obvious one

Z-snails:
> There is also this, but I'm not sure what needs to be done - maybe a question for @Z-snails:
> 
> https://github.com/idris-lang/Idris2/blob/4cb45daba4b6b1c32f2f0dd7195daebe531cb335/src/Compiler/CompileExpr.idr#L196-L210
> 
> 

Those can't be removed yet, at least not until next release, since there is some code in prelude that needs changing. Hopefully I will have a little time after Christmas to do this.

buzden:
Well, it means this particular change worths its own PR

CodingCellist:
@buzden is this good to be merged? Seems like updates that would be great to have before #3155 gets merged, properly creating `0.7.0` -- I just want to be sure the comments are correct with the rc's version number being updated from `0.6.9` to `0.7.0`?  : )

buzden:
Yes, this PR is good to be merged. It contains only things that should have been done quite a long time ago

mattpolzin (approved):
finally skimmed it. naturally, no objections.

PR #3169, opened 2023-12-18
[ fix #3168, re #3063 ] Fix forgotten reflection values

### LINKED ISSUE #3168: Declaration quotes of type declarations causes error
It seems that, since __commit ea093ff__, the declaration quote `` `[] `` on type declarations will cause an __undefined name error__.

# Steps to Reproduce
Trying to load the following code into REPL:
```idris
import Language.Reflection

%language ElabReflection

data_decl : List Decl
data_decl = `[data T = A | B]

record_decl: List Decl
record_decl = `[record R where]
```

# Expected Behavior
The code can be loaded, and we can inspect the value of `data_decl` and `record_decl` (confirmed behaviour in the __commit 305604d__).

# Observed Behavior
Following errors present:
```
- + Errors (2)
 |-- test.idr line 6 col 12:
 |   While processing right hand side of data_decl. Undefined name Language.Reflection.TTImp.Default. 
 |   
 |   test:6:13--6:30
 |    2 | 
 |    3 | %language ElabReflection
 |    4 | 
 |    5 | data_decl : List Decl
 |    6 | data_decl = `[data T = A | B]
 |                    ^^^^^^^^^^^^^^^^^
 |   
 `-- test.idr line 9 col 14:
     While processing right hand side of record_decl. Undefined name Language.Reflection.TTImp.Default. 
     
     test:9:15--9:32
      5 | data_decl : List Decl
      6 | data_decl = `[data T = A | B]
      7 | 
      8 | record_decl: List Decl
      9 | record_decl = `[record R where]
                        ^^^^^^^^^^^^^^^^^
```

buzden:
Fixes #3168

### DIFF
--- a/src/Core/Reflect.idr
+++ b/src/Core/Reflect.idr
@@ -330,9 +330,9 @@ export
 Reflect a => Reflect (WithDefault a def) where
   reflect fc defs lhs env def
     = onWithDefault
-        (appCon fc defs (reflectionttimp "Default") [Erased fc Placeholder, Erased fc Placeholder])
+        (appCon fc defs (reflectionttimp "DefaultedValue") [Erased fc Placeholder, Erased fc Placeholder])
         (\x => do x' <- reflect fc defs lhs env x
-                  appCon fc defs (reflectionttimp "Value") [Erased fc Placeholder, Erased fc Placeholder, x'])
+                  appCon fc defs (reflectionttimp "SpecifiedValue") [Erased fc Placeholder, Erased fc Placeholder, x'])
         def
 
 export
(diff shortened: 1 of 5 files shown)


CodingCellist:
Could we add #3168 (or something similar) as a test so that we don't accidentally re-break this in the future?  : )

buzden:
> Could we add #3168 (or something similar) as a test so that we don't accidentally re-break this in the future? : )

Haven't I done this originally? I even extended this to test the alternative case of non-default visibility. Or do you mean something else?

CodingCellist:
> > Could we add #3168 (or something similar) as a test so that we don't accidentally re-break this in the future? : )
> 
> Haven't I done this originally? I even extended this to test the alternative case of non-default visibility. Or do you mean something else, @CodingCellist?

Sorry! I'm just being an idiot; went off the PR's message rather than checking the modified files... :facepalm:

PR #3170, opened 2023-12-19
[ re #2944, fix ] Make `popen2` be able to not to leak system resources

buzden:
# Description

Current implementation of `popen2` function introduces in #2944 has slightly different behaviour for POSIX and Windows systems, and, more importantly, it leaves zombie processes on POSIX systems, which, when being run regularly, leads to exhaustiveness of PIDs in a system.

Anyway, leaky and unstable behaviour is not to nice, so I propose the simplest solution I came with. I suggest to make behaviour similar in POSIX and WIndows AND requiring a cleanup call `popen2Wait`. Roughtly speaking, this is not a much stretch, because for usual `popen` we already have a pair function `pclose` with the similar meaning. We can't have direct "closing" analogue for `popen2`, because moment and order of closing of the file handlers depends on a particular use (unlike `popen`), so, full cleanup has to be in a series of closings of file handlers and then calling to the newly added function `popen2Wait`.

Behaviour of the existing (leaky) Idris code that currently uses existing `popen2` function remains unchanged with this PR (being run on POSIX systems). So this PR shouldn't break anything.

Alternative solution of the leaking problem was considered (and even mosly implemented in a separate branch). I produces fully detached process which does not produce a zombie in the end, disallowing, however, to get the exit code of this process (at least, easily). But this solution (suddenly) happened to be much more complicated, and, after several attempts to design nicely, I realised that this alternative solution should contain the solution from this PR, so we should first do this before thinking whether we need an ability of detached processes or not.

Pinging @dunhamsteve and @mattpolzin as the original's PR author and main reviewer.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -254,6 +254,10 @@
 * `Ref` interface from `Data.Ref` inherits `Monad` and was extended by a function
   for value modification implemented through reading and writing by default.
 
+* A function `popen2Wait` was added to wait for the process started with `popen2`
+  function and clean up all system resources (to not to leave zombie processes in
+  particular).
+
 #### System
 
 * Changes `getNProcessors` to return the number of online processors rather than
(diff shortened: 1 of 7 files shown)


dunhamsteve:
Looks good, thanks for taking care of this.

PR #3171, opened 2023-12-20
[ interpolate ] Implement `Interpolation` for primitive numeric types

buzden:
# Description

Interfaces `Interpolation` and `Show` have different meanings, and polymorphic implementations of `Interpolation` tend to require `Interpolation` for inner types rathen than `Show`, which is logical. Primitive numeric types have almost single meaningful canonical string representation, so I suggest to add `Interpolation` implementaitons for them (and for `Nat` too).

      `CONTRIBUTORS.md`).


stefan-hoeck:
Personally, I don't think this is a good idea. I'm not opposed to adding default implementations for `Interpolation`, but in my opinion, they should not be auto-imported via the Prelude. There are quite a few use cases, where custom implementations would be useful (displaying numbers in a different base or adding separators for readability), and those use cases will now require newtype wrappers or explicit function calling to still work correctly.

I'd prefer it, if these default implementations would go into a module in base, that needs to be imported explicitely.

buzden:
Well, that sounds reasonable, let's move it to `base`. At least, this change indeed will be backwards compatible with custom user declarations

gallais (approved):

PR #3172, opened 2023-12-20
Separate support derivation (and small related tweaks to the Makefile)

mattpolzin:
# Description

I aimed to not make any drastic or unneeded changes with this PR. Hopefully others agree that the small Makefile changes are all net positive in any context even if they were motivated in this case by a Nix refactor.

I tried to tell the story with the commit messages, but to enumerate the changes briefly:
1. Instead of each of the two Chez bootstrap stages reading the `SCHEME` env var separately, take advantage of the fact that this variable is known in stage 1 and carry it forward to stage 2 automatically.
2. Export needed library paths at the top of bootstrap stage 2 instead of inlining the environment variables. This reduces repetition which is even more significant given that I added two additional variables (one of them, `DYLD_LIBRARY_PATH`, just for good measure).
3. Retain test isolation while still supporting flexibility with what support library to use by introducing optional overrides with `TEST_` prefixes for library paths.
4. Refactor Makefile slightly to specify the path for support files as a variable and at the same time build the support library based on a target that refers to the file name as is idiomatic to avoid rebuilding the support library if it is already there.
5. Tackle a TODO in the Nix configs to build the support library as its own derivation and then use that in the Idris 2 build process.

      `CONTRIBUTORS.md`).

### DIFF
--- a/bootstrap-stage1-chez.sh
+++ b/bootstrap-stage1-chez.sh
@@ -25,7 +25,7 @@ mkdir -p ../build/exec/idris2_app
 # but with the SCHEME var that we already know at this time
 # baked into it.
 echo '#!/bin/sh' >../build/exec/idris2
-echo "SCHEME=${SCHEME}" >>../build/exec/idris2
+echo "SCHEME=\"${SCHEME}\"" >>../build/exec/idris2
 cat ../bootstrap/idris2-boot.sh >>../build/exec/idris2
 chmod +x ../build/exec/idris2
 
--- a/bootstrap-stage1-chez.sh
+++ b/bootstrap-stage1-chez.sh
@@ -20,5 +20,14 @@ ${SCHEME} --script ../bootstrap/compile.ss
 # Put the result in the usual place where the target goes
 mkdir -p ../build/exec
 mkdir -p ../build/exec/idris2_app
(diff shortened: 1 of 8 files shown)


bootstrap-stage1-chez.sh
```diff
@@ -20,5 +20,14 @@ ${SCHEME} --script ../bootstrap/compile.ss
 # Put the result in the usual place where the target goes
 mkdir -p ../build/exec
 mkdir -p ../build/exec/idris2_app
-install ../bootstrap/idris2-boot.sh ../build/exec/idris2
+
+# we "install" bootstrap/idris2-boot.sh as build/exec/idris2
+# but with the SCHEME var that we already know at this time
+# baked into it.
+echo '#!/bin/sh' >../build/exec/idris2
+echo "SCHEME=${SCHEME}" >>../build/exec/idris2
```
gallais, line 28: Will this work if there is a space or a character that needs escaping in the `SCHEME`  path?

mattpolzin: I think not. Good catch.

mattpolzin:
I'll merge this and keep an eye out for subtle build related problems that might come up in Discord. If anyone wants to suggest changes post-merge I am still happy to discuss them here and then follow-up with another PR to address them.

PR #3176, opened 2023-12-28
- [RefC] delete unnecessary conversion to keep code simply.

seagull-kamome:
- [RefC] delete unnecessary conversion to keep code simply.
- [RefC] rename some C functions to confliction safe.
- [RefC] make some C function inline for speed.
- [RefC] reuse some objects to reduce allocation cost.
-
# Description

Just for cleanup and safety.

no. I think this is transparent.

### DIFF
--- a/src/Compiler/RefC/RefC.idr
+++ b/src/Compiler/RefC/RefC.idr
@@ -223,53 +223,6 @@ cOp Crash         [_, msg]  = "idris2_crash(" ++ msg ++ ");"
 cOp fn args = plainOp (show fn) (toList args)
 
 
-data ExtPrim = NewIORef | ReadIORef | WriteIORef
-             | NewArray | ArrayGet | ArraySet
-             | GetField | SetField
-             | VoidElim
-             | SysOS | SysCodegen
-             | OnCollect
-             | OnCollectAny
-             | Unknown Name
-
-export
-Show ExtPrim where
-  show NewIORef = "newIORef"
-  show ReadIORef = "readIORef"
-  show WriteIORef = "writeIORef"
-  show NewArray = "newArray"
-  show ArrayGet = "arrayGet"
-  show ArraySet = "arraySet"
-  show GetField = "getField"
-  show SetField = "setField"
-  show VoidElim = "voidElim"
-  show SysOS = "sysOS"
-  show SysCodegen = "sysCodegen"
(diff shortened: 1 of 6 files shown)


mattpolzin:
I think I lean in the favorable direction on the removal of the `ExtPrim` type and `toPrim` function. I am hesitant to approve since I am not the person most familiar with the RefC backend on the project, especially given that this PR seems like fairly superficial clean up -- unless you have numbers showing that this PR has a measurable impact on performance. Still, if this PR doesn't get much attention over the next week or so I will likely merge it.

seagull-kamome:
This is purely removing overhead, confusion and stress on the reader. As far as I understand it, performance can be improved, but not degraded.

I have further plans for static object initialization and optimization or conditional compilation of backend dependent constants. Hopefully it will be merged before my New Year's holidays are over.

src/Compiler/RefC/RefC.idr
```diff
@@ -689,7 +640,7 @@ mutual
         pure $ MkRS opStatement opStatement
     cStatementsFromANF (AExtPrim fc _ p args) _ = do
         emit fc $ "// call to external primitive " ++ cName p
-        let returnLine = (cCleanString (show (toPrim p)) ++ "("++ showSep ", " (map varName args) ++")")
+        let returnLine = "idris2_" ++ (cName p) ++ "("++ showSep ", " (map varName args) ++")"
```
gallais, line 643: Here we are losing the logic of a hard fail in case the primitive is not supported.
We should at least keep the check in.

tests/refc/prims/Test.idr
```diff
@@ -0,0 +1,24 @@
+module Test
+
+import Data.IOArray
+import Data.IORef
+import System.Info
+
+main : IO ()
+main = do
+  do
```
mattpolzin, line 9: The `do`s here are a bit unusual (nested and used as a visual separator I gather), but I'm not about to turn down a PR based on that especially in a test case.

mattpolzin (approved):

tests/refc/prims/Test.idr
seagull-kamome: Thanks.

I am not aiming for visual purposes, but to limit the scope and lifetime of the variable (like C's bracket).

I wish there was a more concise and beautiful way to write this.

PR #3177, opened 2023-12-29
[RefC] Suppress arglist wrapper

seagull-kamome:
# Description

* Suppress code generation of _arglist wrappers to reduce code size and compilation time.
* Removed Value_Arglist to reduce Closure's allocation overhead and make code simple.
* Cleanup RefC code generator and support codes to keep code simple.

      `CONTRIBUTORS.md`).


support/refc/runtime.c
```diff
@@ -7,64 +7,83 @@ void missing_ffi() {
...
+  case 13:
+    return (*f)(xs[0], xs[1], xs[2], xs[3], xs[4], xs[5], xs[6], xs[7], xs[8],
+                xs[9], xs[10], xs[11], xs[12]);
+  case 14:
+    return (*f)(xs[0], xs[1], xs[2], xs[3], xs[4], xs[5], xs[6], xs[7], xs[8],
+                xs[9], xs[10], xs[11], xs[12], xs[13]);
+  case 15:
+    return (*f)(xs[0], xs[1], xs[2], xs[3], xs[4], xs[5], xs[6], xs[7], xs[8],
+                xs[9], xs[10], xs[11], xs[12], xs[13], xs[14]);
+  case 16:
```
madman-bob, line 57: Does this limit us to functions with at most 16 arguments?

madman-bob (commented):
If I understand correctly, this inlines the notion of `ArgList` into `Closure`s. For the most part, this seems reasonable, except for the limit on the number of arguments.

A constructor for a particularly complicated data-type could easily overcome any finite limit on the number of arguments.

support/refc/runtime.c
seagull-kamome: I may be misremembering, but I believe the old C compilers would only accept about 31 arguments, 31 for C98, maybe even 100 for C99.
Also, considering the stack consumption, I don't think it's a good idea to put everything on the stack. Perhaps some larger functions should be improved to accept arguments as Value* arrays.

seagull-kamome:
The previous commit changed the large argument list to be passed as an array. This may be overkill, but I think it avoids the limitation on the number of arguments.

support/refc/stringOps.c
```diff
@@ -180,30 +177,24 @@ Value *onCollectStringIterator(Value_Pointer *ptr, void *null) {
...
-  }
+  if (c == '\0')
+    return NULL;
 
   it->pos++; // Ok to do this as StringIterator linear
 
-  Value_Constructor *retVal = idris2_newConstructor(2, 1);
-  retVal->args[0] = (Value *)idris2_mkChar(c);
+  Value_Constructor *retVal = (Value_Constructor *)idris2_newConstructor(2, 0);
+  retVal->args[0] = idris2_mkChar(c);
```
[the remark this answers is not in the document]
seagull-kamome: I edited the tag just in case.

PR #3178, opened 2023-12-29
[ re #3171 ] Correct mistakenly forgotten to be updated changelog

buzden:
Unfortunately, I wasn't fast enough to notice this, and this is an examplary case for future (and the only reason why I do this as a separate PR, not as a part of some future bigger one): after an addition to the PR template which mentions the appropriate changelog update, lots of PRs indeed update it, but they update the place relevant to the point of PR creation, which not has changed after the recent release. A lot of PRs need updates in precisely this part.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 * `Data.List.Lazy` was moved from `contrib` to `base`.
 
+* Added an `Interpolation` implementation for primitive decimal numeric types and `Nat`.
+
 #### Contrib
 
 * `Data.List.Lazy` was moved from `contrib` to `base`.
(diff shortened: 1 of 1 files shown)


mattpolzin (approved):

mattpolzin:
Do you have a recommendation for avoiding this? I haven't thought on it at all, but it certainly is easy to make this mistake and there are at least other still open PRs as of now with the same problem ahead of them.

buzden:
> Do you have a recommendation for avoiding this? I haven't thought on it at all, but it certainly is easy to make this mistake and there are at least other still open PRs as of now with the same problem ahead of them.

I've seen in the repository of the Iosevka font, that they have two separate changelog files for the current not yet released version, and the main changelog of all released versions, and they simply copy the first to the second and clear the first during the release process. This could be a helpful practise.

mattpolzin:
Yeah, that’s a nice low tech solve.

gallais:
In agda/agda-stdlib we have the running CHANGELOG.md and a separate CHANGELOG directory
with a file for each released versions. During release the running one gets renamed and moved to
the directory while a new empty one is created.

It does create pretty annoying merge conflicts when we release and some PRs were targeting the old
version of the file. But then again if we don't release that often... 😁

PR #3174, opened 2023-12-24
[ cleanup ] Post v0.7.0 cleanup

buzden:
0.7.0 has been released, so it's time to clean up the stuff that was waiting for this

PR #3179, opened 2023-12-30
Add `(++)` for `List` variation of `All`

joelberkeley:
# Description

This exists in `Data` for `SnocList` and `Vect` but not for `List` for some reason. I copied it verbatim from [src](https://github.com/idris-lang/Idris2/blob/2f2102c4134c8579a0406894b0beb563673255f0/src/Libraries/Data/List/Quantifiers/Extra.idr#L17-L20)

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Libraries/Data/List/Quantifiers/Extra.idr
+++ b/src/Libraries/Data/List/Quantifiers/Extra.idr
@@ -14,6 +14,8 @@ lookup v (px :: pxs)
       No _ => lookup v pxs
       Yes Refl => Just px
 
+-- this function can be deleted once we release the next compiler version
+-- (after 0.7.0) as the function has been replicated in base
 export
 (++) : All p xs -> All p ys -> All p (xs ++ ys)
 [] ++ pys = pys
--- a/src/Libraries/Data/List/Quantifiers/Extra.idr
+++ b/src/Libraries/Data/List/Quantifiers/Extra.idr
@@ -14,8 +14,8 @@ lookup v (px :: pxs)
       No _ => lookup v pxs
       Yes Refl => Just px
 
--- TODO: delete this function once we release the next compiler version
--- (after 0.7.0) as it has been replicated in base's Data.List.Quantifiers
+-- TODO: delete this function after 0.7.1 is released
+-- as it now exists in base's Data.List.Quantifiers
 export
 (++) : All p xs -> All p ys -> All p (xs ++ ys)
 [] ++ pys = pys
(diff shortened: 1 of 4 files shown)


joelberkeley:
I don't understand why this is failing. It's unable to disambiguate between the version in src and that I've added here in base. But src isn't imported in base

mattpolzin:
Correct that `src` isn't imported into base, but rather the other way around: `base` is imported into the compiler source code -- specifically on a module by module basis, but the module of `base` that you added the new function to is imported upstream of the locations where ambiguity is occurring.

This is a good PR IMO, but you'll need to explicitly use the library definition of `(++)` in the source code to disambiguate for now. Once we release the next version of the compiler, it will be safe to bump the minimum required compiler version and remove the `src` version of the function in favor of the `base` version.

mattpolzin:
I'd leave a note in the compiler source code about the future desire to remove the `src` version of the function (after the next release).

joelberkeley:
thank @mattpolzin, is the current compiler version 0.7.0?

mattpolzin:
>  is the current compiler version 0.7.0?

Yeah. And for reference, this is the sort of TODO that I'd leave for future us: https://github.com/idris-lang/Idris2/pull/3174/files#diff-458f49aa263db28859e440611f0e9e49e4cbbef7393c4215580029d938980c24L5

joelberkeley:
all done

mattpolzin:
You actually won't need to disambiguate where you did in the latest changes. Within `base`, this is the only `(++)` for the `List.Quantifiers.All` scope. You'll need to make a similar change within `src` (https://github.com/idris-lang/Idris2/blob/main/src/Libraries/Data/List/Quantifiers/Extra.idr#L20), though. That's where you need to indicate that the compiler code should be using `Extra.(++)` for now.

src/Libraries/Data/List/Quantifiers/Extra.idr
```diff
@@ -14,10 +14,12 @@ lookup v (px :: pxs)
       No _ => lookup v pxs
       Yes Refl => Just px
 
+-- TODO: delete this function once we release the next compiler version
+-- (after 0.7.0) as it has been replicated in base's Data.List.Quantifiers
```
buzden, line 18: Could you please change the wording to something like "delete this function after 0.7.1 is released" because wording "next release after 0.7.0" proved to be really confusing (well, at least for me, doing cleanups). I think wording mentioning `0.7.1` is okay even if next release would be `0.8.0`.

joelberkeley: done. All good?

mattpolzin (approved):
Thanks!

PR #3180, opened 2023-12-30
Add a new CHANGELOG_NEXT.md file and update documentation around CHANGELOG entries.

mattpolzin:
# Description

We have a small problem that when releasing a new version, all existing PRs with entries in the CHANGELOG need to be updated so they do not mistakenly add entries to the already-released version. This PR adopts a strategy that was [suggested twice](https://github.com/idris-lang/Idris2/pull/3178) (as in use by other projects). It won't always make post-release management of open PRs effortless, but it should make it so that we have no risk of accidentally modifying the CHANGELOG for an already-released version.

      `CONTRIBUTORS.md`).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,6 +5,6 @@
 
 <!-- Please delete this section if it doesn't apply -->
 - [ ] If this is a fix, user-facing change, a compiler change, or a new paper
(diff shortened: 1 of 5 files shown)


mattpolzin:
If we merge this, I will go through open PRs and push changes to adopt this strategy for any PR that allows maintainers to make edits.

gallais (approved):

PR #3181, opened 2023-12-31
[RefC] Unbox small integers.

seagull-kamome:
# Description

This patch unboxes some small types by packing them into pointers. We hope this will improve efficiency.
Additionally, this renames some C functions to protect them from collisions.

      `CONTRIBUTORS.md`).


buzden:
Correct me if I'm wrong, but it seems that your change presupposes pointer type to be at least 32 bits, which is not true for all architectures (even among capable to run Idris). So, 32-bit case should be put under some ifdef and fire only in appropriate archs.

Moreover, the size of `uniptr_t` is not required to follow, say, size of `int` by C99 standard, so we need some rigorous way to check that it's size is enough.

seagull-kamome:
Actually I expect the pointer to be 64 bit. Actually I expect the pointer to be 64 bit.

uintptr_t is at least the same but larger type as a pointer, so a pointer is sufficient to accommodate a 32-bit type on a 64-bit CPU.

As you mentioned, my code has problems when compiling for 32-bit CPUs.

support/refc/_datatypes.h
```diff
@@ -50,15 +45,31 @@ typedef struct {
...
+#define idris2_vp_is_unboxed(p) ((uintptr_t)(p)&3)
+
+#define idris2_vp_int_shift                                                    \
+  ((sizeof(uintptr_t) >= 8 && sizeof(Value *) >= 8) ? 32 : 16)
+
+#define idris2_vp_to_Bits64(p) (((Value_Bits64 *)(p))->ui64)
+#define idris2_vp_to_Bits32(p)                                                 \
+  ((idris2_vp_int_shift == 16)                                                 \
+       ? (((Value_Bits32 *)(p))->ui32)                                         \
+       : ((uint32_t)((uintptr_t)(p) >> idris2_vp_int_shift)))
```
buzden, line 58: This and similar checks seems to be runtime checks. Can't we turn these to compile-time ones, say with

```c
#if INTPTR_WIDTH >= 32
#define idris2_vp_to_Bits32(p) /* direct unboxed use */
#else
#define idris2_vp_to_Bits32(p) (((Value_Bits32 *)(p))->ui32)
#endif
```

Or, do you rely on the compiler's optimisation on the constant `if` conditions?

seagull-kamome: > Or, do you rely on the compiler's optimisation on the constant if conditions?

Yes, An optimizer of a modern C compiler is awesome. However INTPTR_WIDTH  or UINTPTR_WIDTH that I forgot one sounds better to me, I'll try it.  thanks!

seagull-kamome: After a little experimentation, I found that UINTPTR_WIDTH is an optional feature of C23. And some of my target compilers did not yet support it.
So I prepared a fallback just in case.

support/refc/_datatypes.h
```diff
@@ -50,15 +45,47 @@ typedef struct {
...
+#define idris2_vp_to_Int32(p) (((Value_Int32 *)(p))->i32)
+
+#else
+#error "unsupported uintptr_t width"
+#endif
+
+#define idris2_vp_to_Bits16(p)                                                 \
+  ((uint16_t)((uintptr_t)(p) >> idris2_vp_int_shift))
+#define idris2_vp_to_Bits8(p) ((uint8_t)((uintptr_t)(p) >> idris2_vp_int_shift))
+#define idris2_vp_to_Int64(p) (((Value_Int64 *)(p))->i64)
```
buzden, line 82: Could you explain why you are still boxing 64-bit values even in the case when `UINTPTR_WIDTH >= 64`?

support/refc/_datatypes.h
```diff
@@ -50,15 +45,47 @@ typedef struct {
...
+#define idris2_vp_to_Bits32(p)                                                 \
+  ((uint32_t)((uintptr_t)(p) >> idris2_vp_int_shift))
+#define idris2_vp_to_Int32(p): ((int32_t)((uintptr_t)(p) >> idris2_vp_int_shift))
+
+#elif UINTPTR_WIDTH >= 32
+#define idris2_vp_to_Bits32(p) (((Value_Bits32 *)(p))->ui32)
+#define idris2_vp_to_Int32(p) (((Value_Int32 *)(p))->i32)
+
+#else
+#error "unsupported uintptr_t width"
```
buzden, line 76: I think, the idea that the target architecture of the `refc` backend is assumed to have pointer type to have at least 32 bits should be documented, at least in the changelog, because this is potentially a breaking thing since Idris has some successful stories of running on 16-bit architectures in the past, so this information should be present somewhere.

support/refc/prim.c
```diff
@@ -82,7 +82,7 @@ Value *idris2_crash(Value *msg) {
 
 Value *idris2_Data_IOArray_Prims_prim__newArray(Value *erased, Value *_length,
                                                 Value *v, Value *_word) {
-  int length = extractInt(_length);
+  int length = idris2_vp_to_Int64(_length); // it sould Int32
```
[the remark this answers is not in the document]
seagull-kamome: I searched for Value_Int and replaced it to idris2_vp_to_.

That comment is just a doodling. I meant that Int32 is better in terms of performance if it is wide enough, because Int64 isn't unboxable in 64bit CPU. And in many cases 32 bit width is sufficient.

support/refc/_datatypes.h
seagull-kamome: Since bit 0 of the pointer is used as a flag, the maximum number of bits that can be unboxed is 63. (My code also took away bit 1, so 62 bits is the limit.) And since we are not thinking about 128-bit CPUs, int64 will not be able to unbox.

buzden: Ok. And could you comment why is this flag needed?

seagull-kamome: This flag is used to identify whether a pointer is actually a Value* or an integer.

RefC does not have complete static tracking of type information, so types are identified at runtime using Value_Header's tag field. However, Int that are pretending to be pointers cannot have that tag, so use that flag to identify them first.
Of course, this flag is not used if it is clear that Value* is actually an Int. But places like newReference/removeReference require this flag.

seagull-kamome: In fact. this unboxing is a classic idea, but It seems to be difficult to understand, so I added a comment to the source.

buzden: > But places like newReference/removeReference require this flag

Is it possible in principle to not to call to `removeReference` for unboxed values when still having the typing information?

seagull-kamome: Partially, yes. I guess.
If AVar had type information, it might reduce the number of calls to removeReference, and integers shouldn't even have to pretend to be pointers. Maybe until we bind it somewhere or use believe_me.

In the case of closures and constructors, static information is not very useful for values already bound to them, and eventually we have to rely on dynamic information. However, we can put the information in the header part of the closure instead of the pointer bit 0.

support/refc/_datatypes.h
```diff
@@ -50,15 +45,58 @@ typedef struct {
...
-  uint8_t ui8;
-} Value_Bits8;
-
-typedef struct {
-  Value_header header;
-  uint16_t ui16;
-} Value_Bits16;
+/*
+We expcted at least 4bytes for Value_header alignment, to use bit0 and bit1 of
+pointer.as flags.
```
buzden, line 50: You have couple of typos in this comment, I'd suggest something like

```diff
-We expcted at least 4bytes for Value_header alignment, to use bit0 and bit1 of
-pointer.as flags.
+We expect at least 4 bytes for `Value_header` alignment, to use bit0 and bit1 of
+pointer as flags.
```

The question is: whether is it documented that pointers are meant to be aligned and that we can rely on this? I know that in some architectures like ARM values that are used for addressing must be aligned, but it there some documented provisions of this for C pointers?

src/Compiler/RefC/RefC.idr
```diff
@@ -26,8 +26,6 @@ import System.File
 import Protocol.Hex
 import Libraries.Utils.Path
 
-%default covering
```
mattpolzin, line 29: This is a pretty big change. Are you sure we don't want any functions checked (that weren't already marked partial) in this module?

seagull-kamome: This is a wrong commit. I have revoked it.

support/refc/runtime.c
```diff
@@ -112,44 +113,41 @@ Value *tailcall_apply_closure(Value *_clos, Value *arg) {
   return (Value *)makeClosureFromArglist(f, newArgs);
 }
 
-int extractInt(Value *v) {
+int idris2_extractInt(Value *v) {
```
mattpolzin, line 116: I think ideally it would either be safe to use `idris2_extractInt` any time an integer value was in play or else documented otherwise very loudly at the top of the function declaration and/or definition.

What I think I see happening in this PR is that (a) some uses of `idris2_extractInt` are being swapped for explicit representation casts (mostly if not entirely `idris2_vp_to_Int64`) and (b) the default unboxed behavior of `idris2_extractInt` is to assume a 32-bit Int is used. I worry that at the very least this implementation only works when used against the FFI `Int` type because 64-bit integers cannot be unwrapped under the current implementation.

mattpolzin (commented):
This is looking really close to ready to merge I think. I've got to find time to do full passes over just a couple more files.

support/refc/runtime.c
seagull-kamome: Tests were added in the last commit.

8-bit, 16-bit and 32-bit Int are cast to 64-bit at first, And then transformed into pointers. So idris2_vp_to_Int32/Bits32 is almost equivalent to casting the original value 8bit -> 64bit -> 32bit in that order.
I was worried that I was cutting corners with the sign bit handling, but from the test results it seems to be working fine.
(Except for two problems with Int64 that I found accidentally.)

Note: extractInt is not safe at all because it implicitly casts large values and converts them to signed values. Also, now extractInt is only used in ConstCase, so the extract function is not necessary for us if we have a comparison function.

Note: I  agree that documentation should be provided. I will put it on my todo list.

mattpolzin (approved):
I'm sold. Once the CHANGELOG_NEXT conflict is resolved, let's merge this!

mattpolzin:
Thanks!

PR #3182, opened 2024-01-01
Improve `buildIdris` and expose the `idris2-api` via the Nix flake.

mattpolzin:
# Description

- Switch from deprecated `defaultPackage.${system}` to newer `package.${system}.default` flake structure.
- Rename `build` property of `buildIdris` result to `executable` (keep old property name around but deprecated).
- Rename `installLibrary` property of `buildIdris` result to `library` and turn it into a function that supports "install with source" (keep the old property name around but deprecated).
- Add the `idris2-api` package to the flake outputs.

      `CONTRIBUTORS.md`).

### DIFF
--- a/nix/buildIdris.nix
+++ b/nix/buildIdris.nix
@@ -6,7 +6,7 @@ let
   idrName = "idris2-${idris2-version}";
   libSuffix = "lib/${idrName}";
   lib-dirs =
-    lib.strings.concatMapStringsSep ":" (p: "${p}/${libSuffix}") idrisLibraries;
+    lib.strings.makeSearchPath libSuffix idrisLibraries;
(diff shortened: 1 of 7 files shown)


nix/buildIdris.nix
```diff
@@ -30,12 +30,18 @@ in rec {
...
+      installPhase = ''
+        runHook preInstall
+        mkdir -p $out/${libSuffix}
+        export IDRIS2_PREFIX=$out/lib
+        idris2 ${installCmd} ${ipkgName}
+        runHook postInstall
+      '';
+    });
+  # deprecated aliases:
+  build = builtins.trace "Warning: 'build' is a deprecated alias for 'executable'." executable;
```
mattpolzin, line 45: Note to self: swap trace out with lib.warn

Issue #3183, opened 2024-01-01
Delay running `%transform` rules


Z-snails:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

`%transform` rules should be run on `CExp` instead of `Term`, along with the other optimisations.

## Motivation

Currently `%transform` rules are run on the core, rather than on `CExp`. This can make it fragile, since the lack eta-expansion and inlining can prevent rules from matching

## The proposal

The `%transform` optimisation should be rewritten on `CExp` instead of `Term`. This might lead to more complexity, since `CExp` is more complex than `Term`, but it will make transform rules more reliable. 

### Examples

I have not managed to come up with a self-contained example yet, but while working on adding transform rules to prelude, I noticed the following fragility:

This definition allowed the rule to match:
```
Ord Nat where
    compare x y = compareNat x y
```
But this didnt:
```
Ord Nat where
    compare = compareNat
```

### Technical implementation

N/A

## Alternatives considered

Improve the current rewrite rule optimisation.
This may be complex, and will likely end up duplicating work, such as inlining and eta-expansion.

Matthew-Mosior:
Do we know if the fragility of `%transform` is still an issue?

PR #3184, opened 2024-01-01
Remove nat magic

Z-snails:
# Description

Remove compiler magic around `Nat`.

This uses `%transform` rules, which are currently fragile (see #3183) so it doesn't work currently.
Once that issue is fixed, then this should work well.

Yes, I will do this later

      `CONTRIBUTORS.md`).

### DIFF
--- a/libs/prelude/Prelude/Types.idr
+++ b/libs/prelude/Prelude/Types.idr
@@ -39,7 +39,7 @@ nonNegativeIntegerToNat :
 nonNegativeIntegerToNat 0 = Z
 nonNegativeIntegerToNat x =
     let -- x >= 0 and x != 0
-        -- so x >= 0 so x - 1 >= 0
+        -- so x > 0 so x - 1 >= 0
         prf = believe_me Refl
      in S $ nonNegativeIntegerToNat {prf}
           $ assert_smaller x $ x - 1
--- a/libs/prelude/Prelude/Types.idr
+++ b/libs/prelude/Prelude/Types.idr
@@ -30,17 +30,34 @@ prim__integerToNat i
       then believe_me i
       else Z
 
+||| Convert a non-negative integer to a `Nat`
+public export
+nonNegativeIntegerToNat :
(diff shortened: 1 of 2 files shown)


libs/prelude/Prelude/Types.idr
```diff
@@ -30,17 +30,34 @@ prim__integerToNat i
...
+||| Convert a non-negative integer to a `Nat`
+public export
+nonNegativeIntegerToNat :
+    (x : Integer) ->
+    {auto 0 prf : (x >= 0) === True} ->
+    Nat
+nonNegativeIntegerToNat 0 = Z
+nonNegativeIntegerToNat x =
+    let -- x >= 0 and x != 0
+        -- so x >= 0 so x - 1 >= 0
```
mattpolzin, line 42:
```suggestion
        -- so x > 0 so x - 1 >= 0
```

src/Compiler/CompileExpr.idr
```diff
@@ -142,77 +142,11 @@ mkDropSubst i es rest (x :: xs)
...
-magic__natUnsuc fc fc' [m]
+||| Unsucc of some positive natural number
+natUnsucc : FC -> FC -> forall vars. Vect 1 (CExp vars) -> CExp vars
+natUnsucc fc fc' [m]
   = COp fc (Sub IntegerType) [m, CPrimVal fc (BI 1)]
 
--- TODO: next release remove this and use %builtin pragma
-natHack : List Magic
-natHack =
-    [ MagicCRef (NS typesNS (UN $ Basic "natToInteger")) 1 (\ _, _, [k] => k)
```
mattpolzin, line 199: This makes Nat -> integer a no-op right? Is that optimization captured anymore once this code is removed?

Z-snails: Yes that optimisation is captured already by a different, much more general optimisation I added.

gallais:
Abandoned.

PR #3186, opened 2024-01-04
[ fix ] Disable top-level constants optimisation during incremental compilation

cypheon:
# Description

Since incremental compilation currently has no way to know about constants in other modules, this commit sets the list of constants to `empty` (as is done in the `ChezSep` backend) which allows successful compilation.

# Background

The optimisation for non-recursive top-level constants introduced in #2817 accidentally breaks incremental compilation when an optimised top-level constant is referenced in another module (noticeable in larger, non-trivial programs, e.g. when building Idris itself in incremental mode).

The resulting error message (during execution) can look like this:

```
Exception: variable PrimIO-unsafePerformIO is not bound
```

The above error is caused due to the order of the different `(load "xyz.so")` statements. But even when correcting the `(load)` order by hand, the following error occurrs:

```
Exception: attempt to apply non-procedure #(1)
```
This one is caused by the top-level constant being a value instead of a zero-argument function. The referencing module only knows about its own constants and therefore treats the constant in the other module as zero-argument function instead of a constant value.

Making top-level constants work together with incremental compilation would require to do both the following:
1. sorting the `(load)` statements, if this is even possible (might there be circular dependencies?)
2. somehow passing along the info which top-level names in other modules are values (instead of zero-arg functions) to the individual module during compilation

I believe this fix is fine, considering the incrementally compiled code is expected to be slower during runtime anyways.

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -658,7 +658,7 @@ incCompile c s sourceFile
                version <- coreLift $ chezVersion chez
                fgndefs <- traverse (getFgnCall version) ndefs
                (sortedDefs, constants) <- sortDefs ndefs
-               compdefs <- traverse (getScheme constants (chezExtPrim constants) chezString) sortedDefs
+               compdefs <- traverse (getScheme empty (chezExtPrim empty) chezString) sortedDefs
                let code = concat $ map snd fgndefs ++ compdefs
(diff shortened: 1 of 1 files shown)


mattpolzin:
This looks reasonable to me; the description is very well laid out. 

It would be really cool to have a test case that catches this or even similar regressions but that may not be essential in this case.

PR #3187, opened 2024-01-05
Fix static JS code: open process mode in support_system_file_popen

nukisman:
In JS switch cases with shared code should look like this:
```js
switch(smth) {
  case 'w':
  case 'a':
     // shared code
     break;
}
```
In case of original code, first case always ignored:
```js
switch(smth) {
  case 'w', 'a':   // Case 'w' always ignored
     // 'a'-only code
     break;
}
```

# Description

No, because currently NodeJS backend supports only 'r' mode, so this switch-case ignored anyway

### DIFF
--- a/support/js/support_system_file.js
+++ b/support/js/support_system_file.js
@@ -132,7 +132,8 @@ function support_system_file_popen (cmd, m) {
     case "r":
       io_setting = ['ignore', write_fd, 2]
       break
-    case "w", "a":
+    case "w": 
+    case "a":
       io_setting = [write_fd, 'ignore', 2]
       break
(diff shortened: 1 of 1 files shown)


andrevidela:
Any chance we can add a test case for it?

nukisman:
> Any chance we can add a test case for it?

I think no, because it this bad JS in dead code section right now. So it should not affect end user any way.
I could give a JS snippet only to reproduce switch-case behavior.

andrevidela:
Oh right, yes I see. Well this is definitely fixing something so might as well merge it now

PR #3190, opened 2024-01-07
Remove use of deprecated `getByte` function

mattpolzin:
# Description

In pursuit of being able to eventually remove the deprecated `getByte` function, remove its use in the `bufferData` function and all other locations aside from the one test case that specifically exists to test the `getByte` function.

Then, realizing that the `bufferData` function has the same problem as `getByte`, deprecate `bufferData` in favor of `bufferData'` with a similar change in the range as to `getByte` -> `getBits8`. Along with deprecating `bufferData` and introducing `bufferData'`, the compiler source gains a duplicate definition of `bufferData'` so that it both avoids using a deprecated function and also continues to build using the previous version of the `base` library; that duplicated function can be removed once we release the next version of the compiler.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -39,6 +39,10 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 
 * Added append `(++)` for `List` version of `All`.
 
+* Deprecate `bufferData` in favor of `bufferData'`. These functions are the same
+  with the exception of the latter dealing in `Bits8` which is more correct than
+  `Int`.
+
 #### Contrib
 
 * `Data.List.Lazy` was moved from `contrib` to `base`.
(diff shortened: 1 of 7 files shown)


buzden:
You are changing signature of one function from `Int` to `Bits8` and making a copy of another function with the same change in signature. Why not to just change signature in both places? Is it about bootstrapping the compiler? We could move only the old function to `src/Libraries` instead, couldn't we?

mattpolzin:
I’ll explain my change by eliminating alternatives. 

I couldn’t just leave the compiler using `bufferData` because I deprecated that function and the compiler as warnings as errors turned on so that would be a compile time error (specifically when self-hosting). 

I couldn’t just change the compiler to use `buffer data’` because that would be a compiler error when building from the previous version of the base library. 

So I could copy either the old `bufferData` or the new `bufferData’` into the compiler to avoid depending on either the old or new base library states. After the next release. I want the compiler to use `bufferData’`, so that’s the one I copied into the compiler.

mattpolzin:
Unless your suggestion was to delete base’s `bufferData` (moving it to the compiler for now). I want to deprecate before removing from base as was done with the `getByte` function leaving removal for some future release.

PR #3192, opened 2024-01-13
fix: singing dragon link

foxyseta:
# Description

YouTube removed the original singing dragon video that was linked in the FAQs.
Added the archive.org version.


andrevidela:
The contribution I didn't know we needed, good work!

PR #3195, opened 2024-01-18
Fix `buildIdris` on non-macos machines

mattpolzin:
## Description

We build the support library as its own derivation for Nix builds so we do not want to copy it as part of the `buildIdris` install phase and we certainly don't want to wrap it as if it were an executable.

### DIFF
--- a/nix/buildIdris.nix
+++ b/nix/buildIdris.nix
@@ -33,7 +33,7 @@ in rec {
         chmod +x $out/bin/*
       else
         cd build/exec/*_app
-        rm libidris2_support.so
+        rm -f ./libidris2_support.so
         for file in *.so; do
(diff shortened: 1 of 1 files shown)


nix/buildIdris.nix
```diff
@@ -33,6 +33,7 @@ in rec {
         chmod +x $out/bin/*
       else
         cd build/exec/*_app
+        rm libidris2_support.so
```
mattpolzin, line 36:
```suggestion
        rm -f ./libidris2_support.so
```

mattpolzin:
Tested on `x86_64-darwin`.

mattpolzin:
Tested on `x86_64-linux`

PR #3196, opened 2024-01-19
[RefC] [Test] Valgrind support & Fix invalid memory read of strSubstr

seagull-kamome:
# Description

* Bug fix
    - Invalid memory read of strSubStr. (String overrun)
    - Memory leaks of IORef.
* Perform memory leak analysis, if valgrind is installed.(Only for RefC)
* Keeping error messages at compile time of tests in the output to make debug easy.

      `CONTRIBUTORS.md`).

### DIFF
--- a/support/refc/stringOps.c
+++ b/support/refc/stringOps.c
@@ -74,11 +74,17 @@ Value *strAppend(Value *a, Value *b) {
 Value *strSubstr(Value *start, Value *len, Value *s) {
   char *input = ((Value_String *)s)->str;
   int offset = extractInt(start);
-  int l = extractInt(len);
+  if (offset < 0)
+    offset = 0;
 
-  int tailLen = strlen(input);
-  if (tailLen < offset + l) {
-    l = tailLen - offset;
+  int l = extractInt(len);
+  if (l < 0)
+    l = 0;
+
+  int tailLen = strlen(input) - offset;
+
+  if (tailLen < l) {
+    l = tailLen;
   }
 
   Value_String *retVal = makeEmptyString(l + 1);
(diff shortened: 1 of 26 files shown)


support/refc/stringOps.c
```diff
@@ -77,8 +77,8 @@ Value *strSubstr(Value *start, Value *len, Value *s) {
   int l = extractInt(len);
 
   int tailLen = strlen(input);
-  if (tailLen < l) {
-    l = tailLen;
+  if (tailLen < offset + l) {
+    l = tailLen - offset;
```
buzden, line 81: Could you add a test for this fix?

seagull-kamome: Added it.

It is not visually apparent because the NUL character is copied and the result is visually correct.
Therefore, it is difficult to detect in testing.

I never noticed this bug until I saw valgrind's report.

seagull-kamome:
I also fixed another bug. It's also hard to find with unit testing.

buzden:
I like the changes in general but after the last changes this PR looks like containing some non-trivial unrelated stuff. I mean, I think new tests and related fixes can go in a single PR, but addition of new functionality and/or support of something new should definitely go to a separate PR, or else such gigantic PR becomes unreviewable.

mattpolzin (commented):
I can almost approve this PR, but the one thing I am not familiar with was the original reasoning behind having the global IORef storage. I'll try and find some time to look back and get some insight on the history of that code in the coming days to see if I can get comfortable enough to approve/merge this.

mattpolzin (approved):
I can't find any historical problems with avoiding the global array of `IORefs`. I hope the global array wasn't somehow there for performance reasons, but I can't see how that would be the case.

This looks like a solid improvement in all the ways it purports to be. Thanks!

PR #3197, opened 2024-01-21
[RefC] [Cleanup] Erase the World (not a fantasy things)

seagull-kamome:
# Description

Removal of vestigial organs that are never used.
(organs? what? I don't think this translation is correct, but I will believe DeepL)

No. This was never used in the first place.


andrevidela (approved):
Looks good to me

PR #3199, opened 2024-01-21
Update starting.rst

computerdreamm:
# Description
On Arch Linux, you can also use the `yay` package manager instead of building from source.


computerdreamm:
Nothing too serious, just mentioned it in the docs :)

andrevidela:
Good idea!

PR #3202, opened 2024-01-22
[ cleanup ] Do not wrap the `mainExpression` into `lazy` in ES backends

buzden:
This seems to be unnecessary, and, as was discussed [in the Discord](https://discord.com/channels/827106007712661524/827109161857712128/1199083661769838685), is as safe as the original code. This allows the tree-shaked code to not to contain unneeded definition of `__lazy` when it is indeed is not needed.

I added a test that checks that `mainExpression` is indeed compiled without `__lazy`, and that `__lazy` indeed was inserted when it was inserted before this PR.


buzden:
I updated the `run` script, and decided that description should be in the `expected` file just in case of change to be clear what is for what. @mattpolzin is description clear enough?

mattpolzin (approved):
Thanks!

Issue #3161, opened 2023-12-09
Backends that use C FFI are inconsistent with the others on strings with zeros


buzden:
This issue looks really similar to #3158, but this issue, I think, is different. The #3158 is about particular backend, but this feature is about C FFI, available to (and acually used by) many backends.

# Steps to Reproduce

Have a program

```idris
main : IO ()
main = putStrLn "abc\00d"
```

Run it with several standard backends:

```console
$ IDRIS2_CG=chez idris2 --exec main X.idr
$ IDRIS2_CG=racket idris2 --exec main X.idr
$ IDRIS2_CG=node idris2 --exec main X.idr
$ IDRIS2_CG=refc idris2 --exec main X.idr
```

For the clear repoducibility and independence on the particular terminal, you can pipe those commands with `... | wc -c` to count number of printed bytes.

# Expected Behavior

All runs print the same thing, namely `abcd`, or `6` if you run using `wc`

# Observed Behavior

All backends that use C FFI for implementation of `putStrLn` (and similar functions like `fPutStr`) print only `abc`, or `3` if you use `wc`. The `node` backend prints `abcd` (or `6`), as expected.

Issue #3158, opened 2023-12-05
`String`s with zeros inside are inconsistent between refc and other backends


buzden:
# Steps to Reproduce

Consider the following program

```idris
main : IO ()
main = do
  printLn $ String.length "ab"
  printLn $ String.length "a\0b"
```

and run is with different backends:

```console
$ IDRIS2_CG=chez idris2 --exec main Main.idr
$ IDRIS2_CG=racket idris2 --exec main Main.idr
$ IDRIS2_CG=refc idris2 --exec main Main.idr
$ IDRIS2_CG=node idris2 --exec main Main.idr
```

# Expected Behavior

All backends would give a consistent output, presumably

```
2
3
```

# Observed Behavior

chez, racket and node give

```
2
3
```

and refc gives

```
2
1
```

andrevidela:
oh boy I was waiting for this to happen, exciting

seagull-kamome:
I have started experimenting with having the length information in the Value_String. The code can be found at https://github.com/seagull-kamome/Idris2/tree/issue3161.

I couldn't figure out a good way to preserve information such as length and memory ownership in FFI, so I used ugly method to handle fastPack, fastUnpack and fastConcat. At least, it seems to work as  well, but I have no clue this method is always safe at all backends.
If we want to hide Value_String from the eyes of the C library and keep compatibility, we will need some way to switch the calling convention.

PR #3203, opened 2024-01-23
[RefC] [Cleanup] Pattern matching generates simpler code.

seagull-kamome:
# Description

The jump tables generated by pattern matching seem completely unnecessary. A simpler conditional branch code would reduce malloc/free costs. It also seems to me that debugging would be easier.

      `CONTRIBUTORS.md`).


seagull-kamome:
I am thinking of eliminating special constructors as my next plan, and this PR is in preparation for that.

andrevidela:
Anything that deletes looks good to me, any objectsion?

mattpolzin:
> any objections

only that I know dependent types make case statements deceivingly tricky so (without knowing whether this is a valid concern) I wonder if there are certain dependent case patterns that aren’t fully covered by test cases that would stop working. 

On the plus side, there are already certain limitations with the RefC backend so I’m not totally opposed to merging and keeping this in mind if anyone reports new problems that sound like they could be related to this change.

seagull-kamome:
I added a test in the previous commit.
This test should be placed in allbackends, but for now it is in tests/refc because of the issues in Issue #3161 and #3158.

andrevidela:
I think we're good to go here then

PR #3206, opened 2024-01-27
[RefC] [Cleanup] Erase trivial constructors; Stop copying constructor name.

seagull-kamome:
# Description

(This PR depends on PR #3203)

If I understand correctly, special constructors such as Nil and Nothing can be eliminated by mapping them to null.

 In newConstructor, the string is copied, but considering that the original string is always literal, this is a complete waste and can simply be replaced with a pointer assignment, and in the case of constructors with tags, even pointers should be unnecessary (except for debugging purposes).

As for the argument part, simply inlining it would reduce the cost of malloc/free.

      `CONTRIBUTORS.md`).


andrevidela:
#3203 is merged, we should look at this next

tests/refc/callingConvention/expected
```diff
@@ -10,38 +10,24 @@ Value *Main_last
...
-        break;
-      }
-      case 1:
-      {
-        Value * var_2 = ((Value_Constructor*)var_0)->args[0];
-        Value * var_3 = ((Value_Constructor*)var_0)->args[1];
+    Value * tmp_128 = NULL;                                  // Main:5:8--5:14
+    if (NULL == var_0 /* Prelude.Basics.Nil [nil] */) {
+        tmp_128 = newReference(var_1);
+    } else if (NULL != var_0 /* Prelude.Basics.(::) [cons] */) {
```
gallais, line 16: IIUC we are generating these redundant checks because we are emitting code in
such a way that each `MkAConAlt` is dealt with in isolation.

Is it not the case however that in the `NULL`-discrimination case the same mechanism
deciding whether to print an `else` could be used to decide to bypass the check?

(Side-remark: we should *really* move away from string-printing as the main mechanism
to generate C code. I have once again offered designing a C-- IR as an MSc project for
this year. Not sure whether there'll be any takers).

gallais:
I haven't checked the implementation (the string emission business does my head in)
but the core idea is very sensible and it's nice to see the extra information about list-like
and maybe-like data being used outside of the scheme backends.

tests/refc/callingConvention/expected
seagull-kamome: I also thought this redundancy was a bit ugly. But modern C compilers can easily remove it. So I left the task to them. In other words, I simply ignored it.

In technically, it seems that it could be easily removed by simply sorting the case clause. For example, transform "[] => ... ;x::xs => ...." into "x::xs => ... ;_ => ..." .
On the other hand, such term simplification seems to be useful for all backends. In other words, I guess, this is not a backend's job, but something that should be solved a bit more upstream.

gallais: Fair enough.

seagull-kamome:
I experimented with sorting ConAlt, but there seems to be a problem with the Identity checker not being able to detect it properly. It seems that we need to think carefully about where to implement this.

````
sortCase : CExp vars -> CExp vars
sortCase (CConCase fc sc [] (Just c1)) = c1
sortCase exp@(CConCase fc sc alts Nothing) =
  case alts of
    [MkConAlt _ UNIT _ [] c1] => c1
    [MkConAlt _ NIL _ [] c1, a2@(MkConAlt _ CONS _ _ _)] => CConCase fc sc [a2] (Just c1)
    [a1@(MkConAlt _ CONS _ _ _), MkConAlt _ NIL _ [] c2] => CConCase fc sc [a1] (Just c2)
    [MkConAlt _ NOTHING _ [] c1, a2@(MkConAlt _ JUST _ _ _)] => CConCase fc sc [a2] (Just c1)
    [a1@(MkConAlt _ JUST _ _ _), MkConAlt _ NOTHING _ [] c2] => CConCase fc sc [a1] (Just c2)
    [MkConAlt _ ZERO _ [] c1, a2@(MkConAlt _ SUCC _ _ _)] => CConCase fc sc [a2] (Just c1)
    [a1@(MkConAlt _ SUCC _ _ _), MkConAlt _ ZERO _ [] c2] => CConCase fc sc [a1] (Just c2)
    _ => exp
sortCase exp = exp
````

mattpolzin:
I realize I'm working my way backwards through the RefC PRs, but I'll take a look at this one next if the conflicts get fixed.

seagull-kamome:
My old branch is already broken, so it is correct to look at them in reverse order.

mattpolzin (approved):
Nice. This all looks sensible to me.

PR #3208, opened 2024-01-31
[ ci ] Bump deprecated versions of used CI actions to avoid warnings

buzden:
Currently CI shows warnings that (transivetely) deprecated CI actions are used. Let's bump


andrevidela (approved):
let's do it

PR #3217, opened 2024-02-17
[ doc ] document %cg pragma

dunhamsteve:
# Description

Add documentation for the `%cg` pragma. I missed this one when documenting the pragmas because it is hiding in the lexer.

### DIFF
--- a/docs/source/reference/pragmas.rst
+++ b/docs/source/reference/pragmas.rst
@@ -140,7 +140,16 @@ using ``--directive extraRuntime=mycode.ss`` on the command line for the chez ba
 
     %cg chez extraRuntime=mycode.ss
 
-See the code generator documentation for available directives.
+The ``%cg`` pragma is followed by the name of a codegen and a directive for that codegen, terminated by
+newline.  Directives from imported modules, including transitive imports, will aggregate. All of the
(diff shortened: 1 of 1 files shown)


docs/source/reference/pragmas.rst
```diff
@@ -130,6 +130,18 @@ Set the expression search timeout in milliseconds.  The default is 1000.
...
+
+Codegen directives can be included in source code with the ``%cg`` pragma. For example, instead of
+using ``--directive extraRuntime=mycode.ss`` on the command line for the chez backend, you can write:
+
+.. code-block:: idris
+
+    %cg chez extraRuntime=mycode.ss
+
+See the code generator documentation for available directives.
+
```
buzden, line 144: I think it should be added whether these directives sum up across the modules of the project, and clarify how successive pragmas behave in one module being applied to the same or different backends

dunhamsteve:
I meant to comment on that review, but accidentally hit resolve (I prefer to let the original reviewer resolve).  The directives for all supplied codegen are stored in the ttc, and directives for the current codegen are used when compiling.  How they are aggregated depends on the directive, but `extraRuntime` in chez is collected and deduplicated.

andrevidela:
That looks good to me! Any reason we should delay this?

PR #3225, opened 2024-03-05
Generalize `succNotLTEpred` 

rvs314:
# Description

This re-implements `succNotLTEpred` such that the implicit argument of `succNotLTEpred` no longer needs to be available at runtime. It also adds `succNotLTEpred` as a `Uninhabited` instance, which parallels how `succNotLTEzero` is available as an `Uninhabited` instance.

It is technically user-facing so I've added it to the `CHANGELOG_NEXT` file. However, I'm happy to remove the addition to the change-log if it's superfluous.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -83,6 +83,9 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
   implemented through the newly added one. The similar alternative for `mapMTTImp`
   is added too.
 
+* Removed need for the runtime value of the implicit argument in `succNotLTEpred`,
+  and added its result as an `Uninhabited` instance.
+
 #### Contrib
 
 * `Data.List.Lazy` was moved from `contrib` to `base`.
--- a/libs/base/Data/Nat.idr
+++ b/libs/base/Data/Nat.idr
@@ -179,9 +179,8 @@ fromLteSucc : LTE (S m) (S n) -> LTE m n
 fromLteSucc (LTESucc x) = x
 
 export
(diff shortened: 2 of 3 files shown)


rvs314:
It looks the test `tests/idris2/basic/basic044` is failing the differential testing, though the change I added doesn't change the results, only the logging information. The diffs either look like this:

```
-LOG unify.meta:5: Adding new meta ({P:cut:2}, (Term:5, Rig0))
-LOG unify.meta:5: Adding new meta ({P:vars:2}, (Term:6, Rig0))
+LOG unify.meta:5: Adding new meta ({P:cut:3}, (Term:5, Rig0))
+LOG unify.meta:5: Adding new meta ({P:vars:3}, (Term:6, Rig0))
```

or this:

```
-$resolved6
+$resolved7
```

I don't know much about the testing infrastructure, so I may be wrong, but it doesn't look like a real failure, only that the differential testing is catching some auxiliary information. If it is actually an issue, I'm not sure how to fix it.

buzden:
Please add call to `clean_names` to the `run` of this failing test like it's done here: https://github.com/idris-lang/Idris2/pull/3156/files#diff-c8e4bb1ca88a1c8330674ecb0abdbf2672760629df376f2e6a05608fe3f8edd8

gallais (approved):

mattpolzin:
> It is technically user-facing so I've added it to the `CHANGELOG_NEXT` file.

It's a nice addition to the changelog. Nothing wrong with letting folks know what nice non-breaking things have happened!

PR #3228, opened 2024-03-08
[fix] ambiguity error with Uninhabited interface implementations.

mattpolzin:
closes https://github.com/idris-lang/Idris2/issues/3230

# Description
This PR reverts the addition of a new Uninhabited implementation found in https://github.com/idris-lang/Idris2/pull/3225.

That addition looked good at face value, but with small Nats it causes ambiguity with another pre-existing implementation of the same interface. I think the most reasonable solution is to remove the new interface implementation because it is in essence already a named implementation of Uninhabited (`absurd x` can be spelled `succNotLTEpred x`).

I've added a test that failed with the following ambiguity prior to reverting the interface imoplementation:
```
Error: While processing right hand side of lteZeroSuccAbsurd. Multiple solutions found in search of:
    Uninhabited (LTE 1 0)

Properties:4:23--4:31
 1 | import Data.Nat
 2 | 
 3 | lteZeroSuccAbsurd : LTE (S Z) Z -> Void
 4 | lteZeroSuccAbsurd y = absurd y
                           ^^^^^^^^

Possible correct results:
    Uninhabited implementation at Data.Nat:96:1--98:33
    Uninhabited implementation at Data.Nat:185:1--187:31
```

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -83,8 +83,7 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
   implemented through the newly added one. The similar alternative for `mapMTTImp`
   is added too.
 
-* Removed need for the runtime value of the implicit argument in `succNotLTEpred`,
-  and added its result as an `Uninhabited` instance.
+* Removed need for the runtime value of the implicit argument in `succNotLTEpred`.
 
 #### Contrib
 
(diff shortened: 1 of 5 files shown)


mattpolzin:
The failing test here should no longer fail after the fix at https://github.com/idris-lang/Idris2/pull/3227

PR #3233, opened 2024-03-09
prefer chez scheme 10+ over racket fork

mattpolzin:
# Description

Chez 10.0.0 merged a lot of work from the Racket fork of Chez including support for new systems. Importantly, the Racket team has decided _not_ to maintain a fork going forward now that a sufficient amount of work has been upstreamed, so we should prefer Chez Scheme and not the Racket fork when possible going forward.

      `CONTRIBUTORS.md`).


dunhamsteve:
For simplicity, we might drop the mention of the Racket Chez (and its build instructions).  

But since I added that, the instructions have simplified to:
```sh
git clone git@github.com:racket/ChezScheme.git
cd ChezScheme
./configure --threads
make
sudo make install
```
The other steps seem to be handled by the configure script now.

dunhamsteve:
Also, homebrew is updated to include chez 10 for aarch64, so that should be the easiest path for mac users. (I don't use ports, so I don't know the situation there.)

mattpolzin:
Thanks for the suggestions. I'll incorporate them soon hopefully. I'm also digging into what appears to be evidence that Chez 10 breaks a couple of edge-case things with Idris's scheme support code so that would need to be fixed before recommending Chez 10 as well.

dunhamsteve:
I'd love to hear details when you get it sorted out.  Homebrew is on chez 10 now (and ships idris2 against it).

If it involves signal handling, chez bug 809 might be related.

mattpolzin:
It looks like the last error I saw was actually a skill issue; it resolved when I did a cycle of clean, build, and install; somewhere I must have had stale (e.g. built with Scheme 9.x) support files.

PR #3235, opened 2024-03-15
[ elab ] Let elab scripts access visibility modifiers

jacobjwalters:
# Description
This is a small extension to the elaborator that allows an elab script to query the visibility of a declaration.

As a quick example, given the following file:
```idris
private foo : Bar
export baz : Quux
namespace A
  public export foo : Bar
```

We can get do the following in the REPL:
```idris
> :let %language ElabReflection
> %runElab getVis `{foo}
[(NS (MkNS ["A", "Main"]) (UN (Basic "foo")), Public),
 (NS (MkNS ["Main"]) (UN (Basic "foo")), Private)]
> %runElab getVis `{baz}
[(NS (MkNS ["Main"]) (UN (Basic "baz")), Export)]
```

The visibilities of all declarations matching the given `Name` are returned.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -30,6 +30,9 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
   customise the syntax of operator to look more like a binder.
   See [#3113](https://github.com/idris-lang/Idris2/issues/3113).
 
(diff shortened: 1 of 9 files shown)


andrevidela:
I don't know if I can do this for you, but if you rebase from `main` you should be able to get all the tests pass

andrevidela:
LGTM

PR #3237, opened 2024-03-16
Use overwrite install to fix glib installation

andrevidela:
# Description

fix brew (for real this time)

      `CONTRIBUTORS.md`).

### DIFF
--- a/.github/workflows/ci-idris2-and-libs.yml
+++ b/.github/workflows/ci-idris2-and-libs.yml
@@ -176,6 +176,7 @@ jobs:
         run: |
           brew update
           brew install --overwrite python@3.12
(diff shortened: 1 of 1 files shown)


andrevidela:
this fixes the `brew link` issue but now tests don't pass :|

andrevidela:
Is it failing because there is no racket instance but only chez scheme? If that's the case I'm confused as to how it work in the first place

mattpolzin:
It's failing the nodejs all-backend tests. I see it seems to not be printing very much of the expected output for some reason.

mattpolzin (approved):
I've got a good feeling about this passing CI.

PR #3242, opened 2024-03-31
[RefC] Object Immortalization and Pre-Generation of Constants

seagull-kamome:
# Description

This PR immortalizes objects whose reference counter has reached its maximum value and makes them excluded from GC. This immortalization is used to make the following improvements.

* Since there is no longer any need to worry about the refCounter overflowing to 0, change the refCounter from int to uint16_t to reduce the size of the Value_Header and improve memory usage efficiency.
* Statically generate constants to reduce runtime cost and share data. Constants are naturally immortal.
Statically generate values that often appear at runtime, such as empty strings and integers less than 100, and share them at runtime to reduce allocation costs. This heuristic may attract many opinions, but it is a good first-try choice.

On the other hand, there is no way to release immortal objects, so they leak. Dynamic immortalization rarely occurs, but it is a problem for programs that run for long periods of time, such as servers. A secondary GC is needed to solve this, but is not implemented in this PR.

      `CONTRIBUTORS.md`).


src/Compiler/RefC/RefC.idr
```diff
@@ -635,7 +621,53 @@ mutual
...
+           Nothing => dyngen
+     where
+        constantName : Constant -> String -> Core String
+        constantName c n = case c of
+           I x   => pure "((Value*)&idris2_constant_Int64_\{cCleanString $ show x})"
+           I64 x => pure "((Value*)&idris2_constant_Int64_\{cCleanString $ show x})"
+           B64 x => pure "((Value*)&idris2_constant_Bits64_\{show x})"
+           Db x  => pure "((Value*)&idris2_constant_Double_\{cCleanString $ show x})"
+           Str x => pure "((Value*)&idris2_constant_String_\{n})"
+           _ => throw $ InternalError "[refc] Unsupported type of constant."
```
mattpolzin, line 638: this runtime error feels like it could be fairly easily remediated by a new smaller data-type than `Constant` so that we aren't relying on the invariant that no new type of constant is stored in `ConstDef` without a new case being added here.

support/refc/memoryManagement.c
```diff
@@ -1,6 +1,47 @@
...
+      idris2_memory_stat.n_newValue, idris2_memory_stat.n_newReference,
+      idris2_memory_stat.n_actualNewReference,
+      idris2_memory_stat.n_immortalized, idris2_memory_stat.n_removeReference,
+      idris2_memory_stat.n_tried_to_kill_immortals, idris2_memory_stat.n_freed);
+}
+
+#else
+#define IDRIS2_INC_MEMSTAT(x)
+// don't inline this, Because IDRIS2_MEMSTAT works only at compiling support
+// libralies to suppressing overhead.
```
mattpolzin, line 41:
```suggestion
// libraries to suppressing overhead.
```

support/refc/memoryManagement.c
```diff
@@ -134,39 +188,45 @@ Value_Array *idris2_makeArray(int length) {
...
-  // recursively remove all references to all children
-  {
+  else if (elem->header.refCounter == IDRIS2_VP_REFCOUNTER_MAX) {
+    IDRIS2_INC_MEMSTAT(n_tried_to_kill_immortals);
+    return;
+  } else if (elem->header.refCounter != 1) {
+    --elem->header.refCounter;
+    return;
+  } else {
+    IDRIS2_INC_MEMSTAT(n_freed);
```
mattpolzin, line 214: I notice that whereas `elem->header.refCounter` used to be decremented prior to freeing, now it is not; that seems theoretically ok to me, but I wanted to make sure it was intentional because I have not verified that nothing downstream of this code would double check the ref count.

andrevidela:
@seagull-kamome @mattpolzin any chance we can get this closer to a resolution?

mattpolzin:
I had some requests related to the Idris2 code that I’d like to see discussed if not acted upon but overall the LR felt close the last time I looked it over.

support/refc/memoryManagement.c
seagull-kamome: I decided not to write back the 0 value of the counter because I thought the overhead of updating the contents of a deleted object would be wasted.

The improvements are so small that they have little real effect.

src/Compiler/RefC/RefC.idr
seagull-kamome: I pushed the changes.

I hope I understand the problem correctly. Anyway, this eliminated the runtime error

src/Compiler/RefC/RefC.idr
```diff
@@ -635,7 +621,53 @@ mutual
...
+            B16 x => pure "idris2_mkBits16(UINT16_C(\{show x}))"
+            B32 x => pure "idris2_mkBits32(UINT32_C(\{show x}))"
+            B64 x => if x >= 0 && x < 100
+               then pure "(Value*)(&idris2_predefined_Bits64[\{show x}])"
+               else orStagen
+            Db _ => orStagen
+            Ch x  => pure "idris2_mkChar(\{escapeChar x})"
+            Str _ => orStagen
+            PrT t => pure $ cPrimType t
+            _ => pure "NULL"
```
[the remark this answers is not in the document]
seagull-kamome: In the previous commit, I fixed this to improve coverage.

seagull-kamome:
I apologize for any inconvenience caused to you due to my poor time control.

I just committed the latest code. I hope this clears up all doubts.

mattpolzin:
No apology needed. None of us are making any implicit promises by volunteering our time including you. Thanks for following up. It’s an exceedingly busy time for me with a 3 day old baby but I’ll re-review this when I can ❤️.

mattpolzin (approved):
Looks good. Thanks for another nice bit of machinery in the RefC backend.

Issue #3251, opened 2024-04-05
Postfix functions are parsed as if inside unquote when are right after unquote


buzden:
# Steps to Reproduce

Consider the following code:

```idris
import Language.Reflection

(.fun) : Nat -> Nat

x : TTImp

f : Nat -> TTImp

useX : TTImp
useX = `(g (~x).fun)

useX' : TTImp
useX' = `(g ~x.fun)

useFX : TTImp
useFX = `(g (~(f 5)).fun)

useFX' : TTImp
useFX' = `(g ~(f 5).fun)
```

# Expected Behavior

All `use*` function typecheck successfully, primed and un-primed variants having the same meaning.

# Observed Behavior

Primed versions do not typecheck:

```
Error: While processing right hand side of useX'. When unifying:
    TTImp
and:
    Nat
Mismatch between: TTImp and Nat.

X:13:14--13:15
 09 | useX : TTImp
 10 | useX = `(g (~x).fun)
 11 | 
 12 | useX' : TTImp
 13 | useX' = `(g ~x.fun)
                   ^
```

```
Error: While processing right hand side of useFX'. When unifying:
    TTImp
and:
    Nat
Mismatch between: TTImp and Nat.

X:19:16--19:19
 15 | useFX : TTImp
 16 | useFX = `(g (~(f 5)).fun)
 17 | 
 18 | useFX' : TTImp
 19 | useFX' = `(g ~(f 5).fun)
                     ^^^
```

Seems like `(.fun)` function was tried to be applied to the `TTImp` being unquoted, despite being outside the unquote brackets.

dunhamsteve:
I believe this is fixed by #3253.

PR #3253, opened 2024-04-09
[ parser ] Fix issue parsing unquote

dunhamsteve:
# Description

This PR addresses #3251 - where unquote was picking up postfix expressions where it wasn't expected. 

Previously, `(g ~(f 5).fun)` would unquote all of `(f 5).fun` and now it only unquotes `(f 5)`.  Similarly `~x.fun` would unquote `x.fun` and now it only unquotes `x`.

The code from #3251 is used as a test case.

I've added it to CHANGELOG_NEXT since it may break existing code.


andrevidela:
while I'm here, this looks good

PR #3254, opened 2024-04-11
[ new ] added util functions for SortedMap

claudio-etterli:
# Description

Added `insertWith`, `insertFromWith` and `fromListWith` to the SortedMap module.
The idea is to insert key-value pairs and decide how duplicated values should be merged.
Unique key-values pairs will be appended to the SortedMap.

      `CONTRIBUTORS.md`).


libs/base/Data/SortedMap.idr
```diff
@@ -26,6 +26,15 @@ export
...
 insert k v = M . insert k v . unM
 
+||| Inserts a key value pair into a map and merges duplicated values
+||| with the given function.
+export
+insertWith : (v -> v -> v) -> k -> v -> SortedMap k v -> SortedMap k v
+insertWith f k v xs =
+  case lookup k xs of
+    Just x  => insert k (f v x) xs
+    Nothing => insert k v xs
```
gallais, line 36: that's a direct corollary of `update`  so not sure it's needed?

stefan-hoeck:
I had a look at Haskell's [Data.Map](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html) today, and I am in strong favour of adding the utilities from this PR (including `insertWith`). The Haskell module has a function like `Data.SortedMap.update` (it's called `alter` there) as well as `insertWith`, `fromListWith`, and several other related utilities.

In my own code, the need for `insertWith` comes up quite frequently, while I hardly ever used `update` so far, probably because it is much more comon to add something to a `SortedMap` than to delete stuff.

PR #3256, opened 2024-04-14
Doc: module naming constraints

foxyseta:
# Description

The Crash Course material on naming modules is apparently out of date. It does not mention that identifiers in module names must be capitalized (I used UK spelling in my revision), or that they must match the filename path.

      `CONTRIBUTORS.md`).

### DIFF
--- a/docs/source/tutorial/modules.rst
+++ b/docs/source/tutorial/modules.rst
@@ -76,15 +76,13 @@ error messages: ``with MyModule.(>>=) do ...``
 
 If a file contains a module declaration ``module Foo.Bar.MyModule``, its
 path relative to the ``sourcedir`` specified in the ``.ipkg`` project file
-(defaults to `.`) must be ``./Foo/Bar/MyModule.idr``. If you are not using an
+(defaults to ``.``) must be ``./Foo/Bar/MyModule.idr``. If you are not using an
 ``.ipkg`` project file, the path must be relative to the directory you are
 running Idris from. Similarly, an ``import`` statement also refers to such a
 relative filepath stripped of its file extension, using dots to separate
 directories. As in the example above, all modules names and directories must be
-capitalised identifiers. The main module, with the ``main`` function, must be
-called ``Main`` — although its filename need not be ``Main.idr``. This is the
-only case where a filename does not have to match with the respective module
-name.
+capitalised identifiers. If a file does not contain a module declaration, it
+is considered to be a module whose identifier is ``Main``.
 
 Export Modifiers
 ================
--- a/docs/source/tutorial/modules.rst
+++ b/docs/source/tutorial/modules.rst
@@ -74,14 +74,14 @@ The ``with`` keyword in expressions comes in two variants:
 This is particularly useful with ``do`` notation, where it can often improve
 error messages: ``with MyModule.(>>=) do ...``
 
-There is no formal link between the module name and its filename,
-although it is generally advisable to use the same name for each. An
-``import`` statement refers to a filename, using dots to separate
-directories. For example, ``import foo.bar`` would import the file
-``foo/bar.idr``, which would conventionally have the module declaration
-``module foo.bar``. The only requirement for module names is that the
-main module, with the ``main`` function, must be called
-``Main`` — although its filename need not be ``Main.idr``.
+If a file contains a module declaration ``module Foo.Bar.MyModule``, its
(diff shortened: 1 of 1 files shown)


docs/source/tutorial/modules.rst
```diff
@@ -74,14 +74,14 @@ The ``with`` keyword in expressions comes in two variants:
...
-There is no formal link between the module name and its filename,
-although it is generally advisable to use the same name for each. An
-``import`` statement refers to a filename, using dots to separate
-directories. For example, ``import foo.bar`` would import the file
-``foo/bar.idr``, which would conventionally have the module declaration
-``module foo.bar``. The only requirement for module names is that the
-main module, with the ``main`` function, must be called
-``Main`` — although its filename need not be ``Main.idr``.
+If a file contains a module declaration ``module Foo.Bar.MyModule``, its
+full path must end in ``Foo/Bar/MyModule.idr``. Similarly, an ``import``
```
buzden, line 78: Not only *end*, it must *be* this path relative to the source directory from the `ipkg` file (if present)

foxyseta: Should be okay now.

foxyseta:
@buzden fixed it!

docs/source/tutorial/modules.rst
```diff
@@ -74,14 +74,15 @@ The ``with`` keyword in expressions comes in two variants:
...
-There is no formal link between the module name and its filename,
-although it is generally advisable to use the same name for each. An
-``import`` statement refers to a filename, using dots to separate
-directories. For example, ``import foo.bar`` would import the file
-``foo/bar.idr``, which would conventionally have the module declaration
-``module foo.bar``. The only requirement for module names is that the
-main module, with the ``main`` function, must be called
-``Main`` — although its filename need not be ``Main.idr``.
+If a file contains a module declaration ``module Foo.Bar.MyModule``, its
+path relative to the ``.ipkg`` project file (or the directory you are running
```
buzden, line 78: This path is relative not to the `ipkg` project file, but source directory mentioned in this file. Say, it's written in the `ipkg` that `sourcedir = "src"` -- this means that path must be `src/Foo/Bar/MyModule.idr`. But by default `sourcedir` is indeed `.`, i.e. the directory where the `ipkg` file resides.

foxyseta: You are right! Misread your comment at first, sorry. is this better?

buzden: > is this better?

It is

docs/source/tutorial/modules.rst
```diff
@@ -74,14 +74,17 @@ The ``with`` keyword in expressions comes in two variants:
...
-``Main`` — although its filename need not be ``Main.idr``.
+If a file contains a module declaration ``module Foo.Bar.MyModule``, its
+path relative to the ``sourcedir`` specified in the ``.ipkg`` project file
+(defaults to `.`) must be ``./Foo/Bar/MyModule.idr``. If you are not using an
+``.ipkg`` project file, the path must be relative to the directory you are
+running Idris from. Similarly, an ``import`` statement also refers to such a
+relative filepath stripped of its file extension, using dots to separate
+directories. As in the example above, all modules names and directories must be
+capitalised identifiers. The main module, with the ``main`` function, must be
+called ``Main`` — although its filename need not be ``Main.idr``. This is the
```
buzden, line 85: Sorry, I found this only now. There is not requirement that the main module is called `Main`. Again, in `ipkg` file you can set any module to be the main one with the `main` property.

buzden: But if module does not have any `module <Name>` declaration in the beginning it indeed automatically gets its name to be `Main`. But it does not have to in order to be the starting point of an application

foxyseta: No problem! I mentioned what happens to modules with no `module <Name>` declaration. I did not add the part about the `main` property for `.ipkg` files yet. Should I?

buzden: > I did not add the part about the `main` property for `.ipkg` files yet. Should I?

Maybe. What I mean is that phrase

> The main module, with the ``main`` function, must be called ``Main``

is not correct

foxyseta: Yep, I feel the same way about it, too.

gallais (approved):

buzden (approved):

PR #3260, opened 2024-04-17
[ base ] Data.SortedSet.leftMost and .rightMost

MithicSpirit:
# Description

Implement `leftMost` and `rightMost` for `SortedSet` in terms of the
functions with the same name in `Data.SortedMap`.

      `CONTRIBUTORS.md`).


MithicSpirit:
Now that I think about this again, it might be better to call these `minimum`
and `maximum` since they correspond to the actual value that is returned. This
makes it more in-line with `Data.List` rather than `Data.SortedMap`, which I
think is preferable since the use case of sets is closer to that of lists than
that of maps. Please let me know if you would like this change.

PR #3269, opened 2024-04-27
[ fix ] fix macos ci

dunhamsteve:
# Description

On ARM macs, homebrew stashes libgmp in `/opt/homebrew`.  The compiler build fails unless you set CPATH or CPPFLAGS.   It looks like github is now using ARM macs and the CI is failing.  Pack updates also fail for this reason (stefan-hoeck/idris2-pack#283).

The PR changes `config.mk` to look for `/opt/homebrew/include/gmp.h` and update `CPPFLAGS` and `LDFLAGS` if it is there.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,7 @@ test: testenv
 	@echo "NOTE: \`${MAKE} test\` does not rebuild Idris or the libraries packaged with it; to do that run \`${MAKE}\`"
 	@if [ ! -x "${TARGET}" ]; then echo "ERROR: Missing IDRIS2 executable. Cannot run tests!\n"; exit 1; fi
 	@echo
-	@${MAKE} -C tests only=$(only) except=$(except) IDRIS2=${TARGET} IDRIS2_PREFIX=${TEST_PREFIX}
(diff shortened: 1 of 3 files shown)


andrevidela:
I'm confused why this CI suite doesn't pass now

andrevidela:
should we maybe use `$HOMEBREW_PREFIX` instead?

dunhamsteve:
The change in this PR got `macos-bootstrap-chez` working, but a refC test is failing for a similar reason. It looks like CPPFLAGS and LDFLAGS are not being passed in to the make process for tests.

I believe the build and the test are failing because the macos CI is now running on ARM architecture macs instead of Intel.  This is how Idris behaves on ARM macs.  Prior to this change, `make` never worked without setting certain shell variables (either `CPATH` or `CPPFLAGS`) and neither did the refc backend tests.

The root cause is that homebrew puts its files into `/opt/homebrew` on ARM macs, and the compilers don't see it there without help. `pkg-config` does find them, but we're not using that. Locally, I've been using `export CPATH=/opt/homebrew` in the environment to correct for this, but here I set `CPPFLAGS` and `LDFLAGS` in the makefile, because I think they're the more appropriate variables to set.

dunhamsteve:
I can switch to `${HOMEBREW_PREFIX}`.  I'm thinking we still check for `${HOMEBREW_PREFIX}/include/gmp.h` rather than just `${HOMEBREW_PREFIX}` in case the user wants to get gmp in another way.

andrevidela:
let's gooooo

Issue #964, opened 2021-01-20
Idris hangs eating all the memory when one function returns `Refl`


buzden:
# Steps to Reproduce

Consider two functions with the same signature:

```idris
r1 : a -> Stream a
r1 x = xs where
  xs : Stream a
  xs = x :: xs

r2 : a -> Stream a
r2 x = x :: r2 x
```

Consider the following function for these functions' property:

```idris
repeatsTheSame : (x : a) -> r1 x = r2 x
repeatsTheSame x = Refl
```

# Expected Behavior

Some result (unification error?)

# Observed Behavior

Idris hangs eating all memory.

When there is a hole instead of `Refl`, compiler finishes typecheck.

Russoul:
@gallais Any chance you have an insight into what's going on here?
I'm quite willing to fix that now.

gallais:
Probably Section 1 of: http://strictlypositive.org/Ripley.pdf
We should be extremely careful about unfolding definitions when unifying under an `Inf`
(as in: it's probably never OK to do so).

Russoul:
Thanks for the pointer!

yellowsquid:
The non-termination stops Idris reporting obvious errors too. For example, Idris never complains that the following is nonsense.

```idris
notTheSame : r1 0 = r1 1
notTheSame = Refl
```

mjustus:
Idris' evaluator is re-used for purposes other than just ordinary weak-head evaluation, chiefly(?) inlining meta-variable solutions and inlining functions marked as `%tcInline` to help out the termination checker. This means that sometimes we do have to "evaluate" under delays. In those cases, we should not unfold recursive definitions.

Resolving this issue requires:

1. Making the evaluator aware that it is operating under a delay. In my [experiments](https://github.com/idris-lang/Idris2/compare/main...mjustus:Idris2:fix-tcInline-delayed), I added a flag to `EvalOpts` but arguably it could also be recorded as part of the evaluation environment. Both are stored inside closures.
2. Marking whether a function definition is recursive, maybe as part of `PMDefInfo`. The `getRefs` should be useful to implement this check, see [References.idr](https://github.com/idris-lang/Idris2/blob/f840d1b771be59000b478780dcfc1dfa2bfde624/src/Core/Termination/References.idr) for an example of its usage.
[...]

PR #3272, opened 2024-05-02
fixed bugs that caused compiler to hang forever when there is %tcinline pragma

### LINKED ISSUE #2995: Addition of `%tcinline` may hang compilation
# Steps to Reproduce

Consider the following seemingly equivalent definitions:

```idris
%default total

incAll : Stream Nat -> Stream Nat
incAll (x::xs) = S x :: incAll xs

incAll' : Stream Nat -> Stream Nat
incAll' = \(x::xs) => S x :: incAll' xs
```

Now, add `%tcinline` to each of the definition.

# Expected Behavior

The same behaviour on both definitions, either typecheck success (preferrable), or some meaningful error.

# Observed Behavior

Addition of `%tcinline` to the first function typechecks, addition to the second function hangs the compiler taking up all available memory.

Simplified examples also show the same hanging behaviour:

```idris
%tcinline
zs : Stream Nat -> Stream Nat
zs xs = Z :: zs xs

%tcinline
zs' : Stream Nat
zs' = Z :: zs'
```

I suspect that the first function typechecks due to #2737, but this is no more than a guess.

AntonPing:
# Description
(This pull request is a continuation of pull request #3257, since github doesn't allow me to re-open it)
close #2995 

This pull request consists of two part:
- Add new pragma `%tcinline_fuel` to avoid totality checker from infinitely inlining recursive function with `%tcinline` and hang forever. The default value of `%tcinline_fuel` is `1000`.
- Fix a bug in `CallGraph.idr` that caused compiler to hang forever even when `%tcinline_fuel` is set.

The second part is somehow rely on the first part, so it can't be divided into two pull requests .

## Description of the first part
As issue #2995 mentioned, this code will cause totality checker to hang.
```
-- See test `total025`
%tcinline
zs : Stream Nat
zs = Z :: zs
```
By adding new pragma `%tcinline_fuel`, this problem is solved. A new test `total027` is added to test this new pragma.

## Description of the second part
And also as issue #2995 mentioned, this code also hangs compilation.
```
-- See test `total026`
%tcinline
incAll : Stream Nat -> Stream Nat
incAll = \(x::xs) => S x :: incAll xs
```
But surprisingly, after adding new pragma `%tcinline_fuel`, it still hangs compilation. It was caused by a bug in `CallGraph.idr`.
The lambda function with pattern matching `\(x::xs) => ...` is actually encoded in such way(`f` is a fresh variable):
```
incAll : Stream Nat -> Stream Nat
f : Stream Nat -> Stream Nat
incAll = f
f (x::xs) => S x :: incAll xs
```
In CallGraph.idr, such `f` is called a case function, and they are treated specially: each time `findSCall` meets a case function, it will jump to the definition of the case function, and it won't be treated as regular function call.
After tcinlining, `f` becomes a recursive definition:
```
incAll = f
f (x::xs) => S x :: f xs
```
In such way, `findSCall` in `CallGraph.idr` will loop forever here, since it corresponds to a program of infinite length:
```
incAll =  \(x::xs) => S x :: (\(x::xs) => S x :: (\(x::xs) => S x :: (\(x::xs) =>  ... ))) ...)
```
The fix in CallGraph.idr introduce a new parameter `cbs`. Each time `findSCall` enters a case function, it appends the function name to `cbs`. If it's already there, `findSCall` then will treat it as a regular function call.

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -52,6 +52,13 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 * The compiler now parses `~x.fun` as unquoting `x` rather than `x.fun`
   and `~(f 5).fun` as unquoting `(f 5)` rather than `(f 5).fun`.
 
+* Add new pragma `%tcinline_fuel` to avoid totality checker from infinitely
+  inlining recursive function with `%tcinline` and hang forever. The default
+  value of `%tcinline_fuel` is `1000`.
+
+* Fix a bug in `CallGraph.idr` that caused compiler to hang forever even when
+  `%tcinline_fuel` is set.
+
 ### Backend changes
 
 #### RefC Backend
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -14,6 +14,7 @@ Andre Kuhlenschmidt
 André Videla
 Andy Lok
 Anthony Lodi
+Anton Ping
 Arnaud Bailly
 Brian Wignall
 Bryn Keller
--- a/docs/source/reference/pragmas.rst
+++ b/docs/source/reference/pragmas.rst
@@ -151,6 +151,12 @@ the ``minimal`` directive priority over the ``compact`` directive if both are pr
 
 See the section for each codegen under :ref:`sect-execs` for available directives.
 
+``%tcinline_fuel``
+--------------------
+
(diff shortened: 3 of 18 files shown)


mjustus:
I believe the examples involving `Stream` are instances of #964, where I just posted an outline of how to fix evaluation under delays.

Outside of those, do you have a test case that requires inlining a recursive function to show termination? I am not in principle opposed to introducing fuel for that purpose but it would be nice to have a real-world derived example.

PR #3274, opened 2024-05-05
dont execute effect in toMaybeT if condition is not met

joelberkeley:
# Description

I'm not sure about this, but is it a bug that
```
toMaybeT : Functor m => Bool -> m a -> MaybeT m a
toMaybeT b m = MkMaybeT $ map (\a => toMaybe b a) m
```
always executes its `m` effect.

If so, this fixes that.

See also
```
toMaybe : Bool -> Lazy a -> Maybe a
toMaybe True  j = Just j
toMaybe False _ = Nothing
```

      `CONTRIBUTORS.md`).

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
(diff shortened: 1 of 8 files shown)


joelberkeley:
I think it was right before, but I'll change the docstring to explain why

Issue #3271, opened 2024-04-29
Type information is lost/changed when inlining a function (and it's not clear why)


JavierGelatti:
Note: I'm not sure whether this is a bug or I don't fully understand how some language feature is supposed to work (I'm currently learning Idris). If you find it's the latter, I'd greatly appreciate it if you could direct me to resources that can fill the gaps in my understanding :pray: 

# Steps to Reproduce
In this example, which —correctly— typechecks:
```idris
data IsEmpty : List Char -> Type where
    Empty : IsEmpty []

Uninhabited (IsEmpty (x::xs)) where
    uninhabited Empty impossible

decideIsEmpty : (cs : List Char) -> Dec (IsEmpty cs)
decideIsEmpty cs = case cs of
    [] => Yes Empty
    (x :: xs) => No uninhabited

decideIsEmptyString : (s : String) -> Dec (IsEmpty (unpack s))
decideIsEmptyString s = decideIsEmpty (unpack s)
```

replace the application of `decideIsEmpty` by its definition (i.e. _inline_ the `decideIsEmpty` function in `decideIsEmptyString`).
You're left with:
```idris
-- ...

decideIsEmptyString : (s : String) -> Dec (IsEmpty (unpack s))
decideIsEmptyString s = case unpack s of
    [] => Yes Empty
    (x :: xs) => No uninhabited
```

Note that the type of `unpack s` is `List Char`, and the expected type for the expression is still `Dec (IsEmpty (unpack s))`.

# Expected Behavior
The resulting code is equivalent to the original program, so it should typecheck.

# Observed Behavior
Both `Yes Empty` and `No uninhabited` stop being valid, and the program no longer typechecks. The errors are:
- In `Empty`: 
  > While processing right hand side of decideIsEmptyString. Can't solve constraint between: [] and **_`<<(unpack s)-implementation>>`_**.
- In `uninhabited`:
  > While processing right hand side of decideIsEmptyString. Can't find an implementation for Uninhabited (IsEmpty (**_`<<(unpack s)-implementation>>`_**)).

Where **_`<<(unpack s)-implementation>>`_** is:
[...]

PR #3277, opened 2024-05-11
[ doc ] Update documentation for `with`

dunhamsteve:
# Description

The documentation in `:doc with` mentions that `with` rewrites the target type. This doesn't seem to be mentioned in the html documentation, so I'm copying it here.  This came up in the discussion in #3271. 

I've also added documentation for multi-`with` and removed the warning about the documentation not being updated for Idris 2.  The documentation appears to be accurate and all of the code snippets run in Idris 2.

#3271 also asks when to use `with` vs `case`.  I'm not sure what to say or where to say it, so I haven't addressed that.

### DIFF
--- a/docs/source/tutorial/views.rst
+++ b/docs/source/tutorial/views.rst
@@ -4,10 +4,6 @@
 Views and the “``with``” rule
 *****************************
 
-.. warning::
-
-   NOT UPDATED FOR IDRIS 2 YET
-
 Dependent pattern matching
 ==========================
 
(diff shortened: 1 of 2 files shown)


buzden:
> #3271 also asks when to use `with` vs `case`. I'm not sure what to say or where to say it, so I haven't addressed that.

The great difference is that `with` rewrites types of all arguments according to the information given by dependent pattern matching. Say, you have `xs : Vect n a` and `ys : Vect m a`. Say, you pattern match on a value `Dec (n = m)` in your `with`. This means that in the clause `Yes Refl` you will get both `xs` and `ys` to be of type `Vect m a`. And wherever `n` appeared in the types of arguments, it will be replaced by `m` in this clause.

PR #3283, opened 2024-05-17
[ fix ] fix windows CI, aligned_alloc not supported on win32

dunhamsteve:
# Description

The support code calls `aligned_alloc` which does not appear to be supported on windows, and is not defined in the `stdlib.h` header file. The latest GitHub windows image has turned the warning about this into an error, causing ci to fail.

I’ve updated the condition in `memoryManagement.c` to use the fallback branch which calls `malloc` when building on windows.

### DIFF
--- a/support/refc/memoryManagement.c
+++ b/support/refc/memoryManagement.c
@@ -2,7 +2,8 @@
 #include "runtime.h"
 
 Value *idris2_newValue(size_t size) {
-#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* C11 */
+#if !defined(_WIN32) && defined(__STDC_VERSION__) &&                           \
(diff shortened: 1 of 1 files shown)


dunhamsteve:
Additional details:

It looks like we've been getting warnings about `aligned_alloc` being implicitly defined for a while, but the latest github windows image contains a newer version of GCC that [escalates this to an error](https://gcc.gnu.org/gcc-14/porting_to.html). I also found [cppreference.com](https://en.cppreference.com/w/c/memory/aligned_alloc) which says that "This function is not supported in Microsoft C Runtime library".

It looks like, despite not being defined in the header files, the linker was finding this function.  But I presume `malloc` is fine here, since it is declared and is already being used on some systems.

andrevidela:
Astounding work, thank you

PR #3284, opened 2024-05-19
Add pipeline operators

observant2:
# Description
Adds the pipeline operators `|>` and `<|` known from other functional languages like F# and Elixir.

      `CONTRIBUTORS.md`).

### DIFF
--- a/tests/idris2/operators/operators001/Test.idr
+++ b/tests/idris2/operators/operators001/Test.idr
@@ -6,7 +6,7 @@ import Data.Vect
 private typebind infixr 0 =@
 private infixr 0 -@
 
-%hide Prelude.Ops.infixr.(|>)
+%hide Prelude.Ops.infixl.(|>)
 
 -- typebind infixr 1 =@@
 
@@ -79,4 +79,3 @@ compose (a |> a') (b |> b') =
   (x : (y : a) @@ (a' y -> b)) |>
        (y : a' x.fst) @@
             b' (x.snd y)
-
--- a/tests/idris2/operators/operators001/Test.idr
+++ b/tests/idris2/operators/operators001/Test.idr
@@ -6,6 +6,8 @@ import Data.Vect
 private typebind infixr 0 =@
 private infixr 0 -@
 
+%hide Prelude.Ops.infixr.(|>)
+
 -- typebind infixr 1 =@@
 
 0 (=@) : (a : Type) -> (a -> Type) -> Type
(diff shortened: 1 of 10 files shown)


andrevidela:
Looks like there are some test to fix because now there are two conflicting fixity definitions for `(<|)`

observant2:
Unfortunately I'm stuck here. When I execute "tests/idris2/operators/operators001/Test.idr"
with `%hide Prelude.Ops.infixr.(|>)` I still get the errors:
```
1/1: Building Test (Test.idr)
[-Warning: operator fixity is ambiguous, we are picking Prelude.Ops.infixr.(|>) out of :-]
[- - Prelude.Ops.infixr.(|>), precedence level 9-]
[- - Test.infixr.(|>), precedence level 0-]

[-To remove this warning, use `%hide` with the fixity to remove-]
[-For example: %hide Prelude.Ops.infixr.(|>)-]
```
It suggests me to do what I already did.

andrevidela:
Do the tests run locally? it seems there are still conflicting fixity definitions in the file `Import.idr` for the test `idris2/misc/import009`

observant2:
Now they should pass.

observant2:
Oh, I forgot, someone commented in Discord:
> shouldn't `(|>)` be `infixl` in the prelude?
> also, I think it would make more sense if both pipeline operators had the same precedence

So should I make |> infixl? If they should get the same precedence, which should it be?

buzden:
> So should I make |> infixl?

You definitely should. Even example from your docstring shouldn't typecheck if this operator leaves `infixr`.

> If they should get the same precedence, which should it be?

Precedence should be low enough to be able to form the value you are operating on. Again, in your docstring, you are, actually, applying `|>` after a bunch of `::` (which `[a, b c]` desugars to), so `|>` should have a precedence *lower* than that (and now it's very high). I think that 0 for both is rather a good choice.

Maybe you should add the example from your docstring as a test in order to check that what you intended still works.

observant2:
Okay, that's a good idea, where should this test go? Is there an existing file where it would fit or should I create a new `operators011` folder?

buzden:
> Okay, that's a good idea, where should this test go? Is there an existing file where it would fit or should I create a new `operators011` folder?

It should go here, I think:
https://github.com/idris-lang/Idris2/tree/main/tests/prelude

tests/idris2/operators/operators001/Test.idr
```diff
@@ -6,6 +6,9 @@ import Data.Vect
 private typebind infixr 0 =@
 private infixr 0 -@
 
+%hide Prelude.Ops.infixl.(|>)
+%hide Prelude.Ops.infixr.(|>)
```
observant2, line 10: Without this line I would still get warnings, although that operator doesn't exist.

gallais (changes requested):
Please fix the CI errors

gallais (approved):

PR #3293, opened 2024-06-02
Add an option that dumps package details to JSON

mattpolzin:
# Description

Adds a `dump-ipkg-json` option. It would be helpful to be able to "export" Idris2's package information in a format that is fairly universally understood. I have one definite use-case in mind, but I suspect that having authoritative access to this information would be useful in any number of build environments where build tooling is not written as an Idris app directly based on the compiler's API.

      `CONTRIBUTORS.md`).


src/Idris/CommandLine.idr
```diff
@@ -279,6 +281,9 @@ options = [MkOpt ["--check", "-c"] [] [CheckOnly]
            MkOpt ["--init"] [Optional "package file"]
               (\ f => [Package Init f])
               (Just "Interactively initialise a new project"),
+           MkOpt ["--dump-ipkg-json"] [Optional "package file"]
+              (\ f => [Package DumpJson f])
+              (Just "Interactively initialise a new project"),
```
gallais, line 286: copy/pasting mistake

src/Idris/Package/ToJson.idr
```diff
@@ -0,0 +1,116 @@
...
+import Libraries.Data.String.Extra
+import Data.List
+import Core.FC
+import Core.Name.Namespace
+import Data.Maybe
+
+%default total
+
+-- We don't bother with a robust JSON implementation
+-- for this one-off JSON serialization.
```
gallais, line 13: famous last words 😅

mattpolzin: hah. I guess the irony is that I’m claiming JSON is so useful that this PR is worthwhile at the same time as I am commenting that JSON is going to be used so narrowly that it can be implemented ad-hoc.

PR #3306, opened 2024-06-12
[ fix ] implicits are not in scope under an implicit parameter block

### LINKED ISSUE #2444: function in `parameters` block unable to access implicits without bounding on LHS
# Steps to Reproduce
```idris
parameters {0 res : Type}
    fun : {x : Nat} -> Nat
    fun = x
```
# Expected Behavior
It type-checks.

# Observed Behavior
```
Error: While processing right hand side of fun. Undefined name x. 

error:4:11--4:12
 1 | 
 2 | parameters {0 res : Type}
 3 |     fun : {x : Nat} -> Nat
 4 |     fun = x
               ^

Error(s) building file
```

`x` is not available in a parameters block when it would be without the `parameters` block
ie this works
```idris
fun : {x : Nat} -> Nat
fun = x
```

dunhamsteve:
# Description

This PR fixes #2444, an issue with implicits not showing up under a parameter block containing an implicit parameter.

If a function is in the scope of a an implicit `parameter` block, its implicits are not visible in the body of the function.  For example in this code:
```idris
parameters {n : Nat}
    foo : Vect m Nat -> Nat
    foo xs = ?hole
```
Idris reports the scope of `?hole` as:
```
Main> :ti hole
   {n : Nat}
   xs : Vect m Nat
hole : Nat
```
Note that `m` is missing.  It behaves the same way after explicitly adding `{m : Nat} ->` to the signature of `foo`.  But if we change the parameter `n` to be explicit, `m` shows up in the hole.

The root cause is in `findImps`.  It attempts to skip over locals, but assumes that they are `Explicit`.  They are explicit in the case when the locals come from a surrounding function, but not necessarily the case when they come from a `parameter` block. In that case, the default case is hit and the function returns `[]`.


src/TTImp/TTImp.idr
```diff
@@ -728,7 +728,7 @@ implicitsAs n defs ns tm
         -- corresponding to the variables bound in the parent function.
         -- So we first peel off all of the explicit quantifiers corresponding
         -- to these variables.
```
andrevidela, line 730: Shouldn't this comment be changed?

dunhamsteve: Thanks. Yes, it should.  I forgot to go back and take care of this.  I've updated it to mention that parameter blocks also contribute variables that may not be explicit.

andrevidela:
looks good to me

PR #3312, opened 2024-06-15
Update typedd.rst

m-rinaldi:
# Description

Entering the expression `6.0 + 3 * 12` in the REPL produces the following error on Idris 2:

> Error: Can't find an implementation for `FromDouble Integer`.

Instead, entering `the Double (6.0 + 3 * 12)` behaves as stated in the book.

No


gallais:
Btw thanks for all of these!

PR #3314, opened 2024-06-15
[ fix ] auto search returns no solution instead of ambiguous solution #3313

### LINKED ISSUE #3313: Inconsistent error message when finding more than one implementation of an interface
# Context
When more than one implementation of an interface is found, Idris (correctly) generates an error notifying the problem. The error message includes an explanation of the issue.

For example:
```idris
data Expr : Type -> Type where
    Add : (Num n) => n -> n -> Expr n

eval : (Num n) => Expr n -> n
eval (Add x y) = x + y
--               ^^^^^
--               Error marked here
```
The generated error is:
```
While processing right hand side of eval. Multiple solutions found in search of:
    Num n

Possible correct results:
    conArg (implicitly bound at <location>)
    conArg (implicitly bound at <location>)
```

# Problem
However, when more than one implementation is _indirectly_ found (i.e. by interface extension), the error message is inconsistent and can be misleading.

# Steps to Reproduce
The following is the same code as before, but replacing `Num` with `Integral`.
Note that the applied function (`+`) is declared in the `Num` interface, and `Integral` extends `Num`.
```idris
data Expr : Type -> Type where
    Add : (Integral n) => n -> n -> Expr n

eval : (Integral n) => Expr n -> n
eval (Add x y) = x + y
--               ^^^^^
--               Error marked here
```

# Expected Behavior
The generated error should explain that more than one instance of `Num` was found:
```
While processing right hand side of eval. Multiple solutions found in search of:
    Num n

Possible correct results:
    ...
```
Knowing the search paths that were used to find the implementations could also be useful.

# Observed Behavior
The error message says that it couldn't find an implementation for `Num`: 
```
While processing right hand side of eval. Can't find an implementation for Num n.
```
This is not only inconsistent with the first error but can also be misleading because it's equivalent to the error that's generated when _no_ implementation of `Num` could be found, e.g.:
```idris
data Expr : Type -> Type where
    Add : n -> n -> Expr n

eval : Expr n -> n
eval (Add x y) = x + y
--               ^^^^^
-- While processing right hand side of eval.
-- Can't find an implementation for Num n.
```

dunhamsteve:
# Description

This fixes #3313, where auto search reports no solution instead of an ambiguous solution. 

The root cause it that the `AmbiguousSearch` error is caught and then re-emitted as `CantSolveGoal`.

Prior to this change the code:
```idris
data Expr : Type -> Type where
    Add : (Integral n) => n -> n -> Expr n

eval : (Integral n) => Expr n -> n
eval (Add x y) = x + y
```
reports
```
1/1: Building src.Issue3313 (src/Issue3313.idr)
Error: While processing right hand side of eval. Can't find an implementation for Num n.

src.Issue3313:5:18--5:23
 1 | data Expr : Type -> Type where
 2 |     Add : (Integral n) => n -> n -> Expr n
 3 | 
 4 | eval : (Integral n) => Expr n -> n
 5 | eval (Add x y) = x + y
                      ^^^^^
```
and after the change it reports
```
1/1: Building Issue3313 (src/Issue3313.idr)
Error: While processing right hand side of eval. Multiple solutions found in search of:
    Integral n

Issue3313:5:18--5:23
 1 | data Expr : Type -> Type where
 2 |     Add : (Integral n) => n -> n -> Expr n
 3 | 
 4 | eval : (Integral n) => Expr n -> n
 5 | eval (Add x y) = x + y
                      ^^^^^

Possible correct results:
    conArg (implicitly bound at Issue3313:5:1--5:23)
    conArg (implicitly bound at Issue3313:5:1--5:23)
```

### DIFF
--- a/src/Core/AutoSearch.idr
+++ b/src/Core/AutoSearch.idr
@@ -142,6 +142,7 @@ anyOne fc env top [elab]
     = catch elab $
          \case
            err@(CantSolveGoal _ _ _ _ _) => throw err
+           err@(AmbiguousSearch _ _ _ _) => throw err
            _ => throw $ CantSolveGoal fc (gamma !(get Ctxt)) [] top Nothing
 anyOne fc env top (elab :: elabs)
     = tryUnify elab (anyOne fc env top elabs)
--- a/tests/idris2/error/error028/Issue3313.idr
+++ b/tests/idris2/error/error028/Issue3313.idr
@@ -1,0 +1,5 @@
+data Expr : Type -> Type where
+    Add : (Integral n) => n -> n -> Expr n
+
+eval : (Integral n) => Expr n -> n
+eval (Add x y) = x + y
--- a/tests/idris2/error/error028/expected
+++ b/tests/idris2/error/error028/expected
@@ -1,0 +1,15 @@
+1/1: Building Issue3313 (Issue3313.idr)
+Error: While processing right hand side of eval. Multiple solutions found in search of:
+    Integral n
+
+Issue3313:5:18--5:23
+ 1 | data Expr : Type -> Type where
+ 2 |     Add : (Integral n) => n -> n -> Expr n
+ 3 | 
+ 4 | eval : (Integral n) => Expr n -> n
+ 5 | eval (Add x y) = x + y
+                      ^^^^^
+
+Possible correct results:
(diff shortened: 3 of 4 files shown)


gallais:
Ideally we'd want the bounds on these implicitly introduced names to be tighter so that one points to `eval` and one points to `Add`  instead of both pointing to the whole LHS.

PR #3323, opened 2024-06-19
[ cleanup ] remove unused code left in #3108

dunhamsteve:
The changes in #3108 remove the `arity` argument from `findSCcall`, but left some code that was doing a type lookup to compute the arity. This PR removes that vestigial code.


mjustus:
Thanks!

PR #3324, opened 2024-06-21
idris_support: fix environ for macOS

barracuda156:
# Description

Fix environ for Apple: this version works on every macOS, unlike the generic one.

### DIFF
--- a/support/c/idris_support.c
+++ b/support/c/idris_support.c
@@ -23,6 +23,9 @@ char **_argv;
 extern char **_environ;
 #include "windows/win_utils.h"
 #define environ _environ
+#elif defined(__APPLE__)
+#include <crt_externs.h>
+#define environ (*_NSGetEnviron())
 #else
 extern char **environ;
 #endif


mattpolzin (commented):
This looks good to me. Do you mind adding a really simple test to `tests/base` that exercises `getEnvironment` in Idris which in turn relies on the `environ` variable? If you'd like some pointers on adding new tests, just let me know.

barracuda156:
@mattpolzin This is a canonical definition: 
https://github.com/gcc-mirror/gcc/blob/573f11ec34eeb6a6c3bd3d7619738f927236727b/include/environ.h#L25-L35

https://github.com/libuv/libuv/blob/ba24986f8deda3a0228e1e06e4a5ee87451ede92/src/unix/core.c#L59-L65

https://opensource.apple.com/source/Libc/Libc-1244.50.9/stdlib/FreeBSD/system.c.auto.html

But I do not mind a test, of course, if you think it is needed. But yes, I will need help with it then.

mattpolzin:
Idris uses golden tests that are automatically discovered (for the test directory we are concerned about) so the easiest thing to do is copy an existing test and modify.

I'd recommend copying `tests/base/system_env` to `tests/base/system_get_environment` as a starting place.

Next, delete the `expected` file (you'll generate a new one when running the test after modifying the copied folder).

In the `run` file, you can delete the comment and also the second invocation of `idris2` (`getEnvironment` does not support the Node backend at the moment).

In `Test.idr`, you can replace just about everything. Since we can't rely on the full list of environment variables being the same in every environment, I suggest simply checking that the list of variables is not empty in the test. You can just print out "good" or "bad" depending on success (or any number of other strategies for indicating success) because the golden test will pass if the output is the same and fail if it is different.

Then, to generate a value for your test, run the following from the root of the repo:
```
make test only=system_get_environment`
```

You'll see "Golden value missing..." and it should suggest the expected success result to you. Type `y` and hit enter to accept the value. It'll indicate tests failed, but if you immediately re-run the test command, you'll see it succeeds.

mattpolzin:
Depending on your familiarity with the Idris2 language itself, I may or may not have left you with some things to puzzle through. Feel free to ask if you're just getting started with programming in Idris and haven't yet worked out how to write IO do blocks.

mattpolzin:
Your change here definitely looks good to go, so I hope you don't mind that I pushed up that new test I was suggesting myself. We can get this merged once CI finishes.

mattpolzin (approved):

barracuda156:
@mattpolzin Thank you very much! And sorry that I did not return to test issue, too much stuff to handle, and I am not familiar with the language syntax here, so need to read documentation first to do anything.

PR #3326, opened 2024-06-21
idris2.ss: support powerpc

barracuda156:
# Description

Fix building on macOS PowerPC

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -38,6 +38,8 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 * A new `idris2 --dump-ipkg-json` option (requires either `--find-ipkg` or
   specifying the `.ipkg` file) dumps JSON information about an Idris package.
 
+* Support for macOS PowerPC added.
+
 ### Language changes
 
 * Autobind and Typebind modifier on operators allow the user to
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -95,7 +95,7 @@ schHeader chez libs whole
     (case (machine-type)
       [(i3fb ti3fb a6fb ta6fb) #f]
       [(i3le ti3le a6le ta6le tarm64le) (load-shared-object "libc.so.6")]
-      [(i3osx ti3osx a6osx ta6osx tarm64osx) (load-shared-object "libc.dylib")]
+      [(i3osx ti3osx a6osx ta6osx tarm64osx tppc32osx tppc64osx) (load-shared-object "libc.dylib")]
       [(i3nt ti3nt a6nt ta6nt) (load-shared-object "msvcrt.dll")]
       [else (load-shared-object "libc.so")])
 
--- a/src/Compiler/Scheme/ChezSep.idr
+++ b/src/Compiler/Scheme/ChezSep.idr
@@ -40,7 +40,7 @@ schHeader libs compilationUnits = fromString """
       \{ unwords ["(" ++ cu ++ ")" | cu <- compilationUnits] })
   (case (machine-type)
     [(i3le ti3le a6le ta6le tarm64le) (load-shared-object "libc.so.6")]
-    [(i3osx ti3osx a6osx ta6osx tarm64osx) (load-shared-object "libc.dylib")]
+    [(i3osx ti3osx a6osx ta6osx tarm64osx tppc32osx tppc64osx) (load-shared-object "libc.dylib")]
     [(i3nt ti3nt a6nt ta6nt) (load-shared-object "msvcrt.dll")]
     [else (load-shared-object "libc.so")]
   \{ unlines ["  (load-shared-object \"" ++ escapeStringChez lib ++ "\")" | lib <- libs] })
--- a/support/chez/support.ss
+++ b/support/chez/support.ss
@@ -4,7 +4,7 @@
     [(i3ob ti3ob a6ob ta6ob tarm64ob) "unix"]  ; OpenBSD
     [(i3fb ti3fb a6fb ta6fb tarm64fb) "unix"]  ; FreeBSD
     [(i3nb ti3nb a6nb ta6nb tarm64nb) "unix"]  ; NetBSD
-    [(i3osx ti3osx a6osx ta6osx tarm64osx) "darwin"]
+    [(i3osx ti3osx a6osx ta6osx tarm64osx tppc32osx tppc64osx) "darwin"]
     [(i3nt ti3nt a6nt ta6nt tarm64nt) "windows"]
     [else "unknown"]))
 


buzden:
Every change in the bootstrap file needs a change in the release checklist here https://github.com/idris-lang/Idris2/blob/main/Release/CHECKLIST or else it will be just overwritten on the next release

barracuda156:
@buzden Thank you. Could you please take a look and say if that will do? I am not really sure of conventions in that regard.

CodingCellist:
Is this something that needs to be added manually, or can we configure Chez Scheme to consider PowerPC support valid from the get-go? As far as I can tell, there is nothing functionally changing, other than adding the right magic strings to let the pattern match load the library on those architectures as well?

Release/CHECKLIST
```diff
@@ -19,5 +19,6 @@
     string "__PREFIX__" rather than your local prefix
 [x] Remove __collect_safe from generated chez (to avoid need for chez >9.5)
 [x] Tag on github with version number (in the form vX.Y.Z)
+[ ] Add "tppc32osx tppc64osx" into bootstrap/idris2_app/idris2.ss
```
CodingCellist, line 22: * Could you please add two sub-checkboxes to this? One specifying we need to append the architectures to the list before `load-shared-object "libc.dylib")`, and one specifying that this also needs to be done in the `[...]Chez-schHeader` and `[...]ChezSep-schHeader` functions.
* Nit: I would reword to "Patch support for 'tppc[...]' into [...]"

barracuda156:
> @barracuda156 Is this something that needs to be added manually, or can we configure Chez Scheme to consider PowerPC support valid from the get-go? As far as I can tell, there is nothing functionally changing, other than adding the right magic strings to let the pattern match load the library on those architectures as well?

@CodingCellist Perhaps you can, but I have no idea where that should be done. Locally I had to patch this file to make it work.

Release/CHECKLIST
barracuda156: Done

dunhamsteve:
This is only changing the bootstrap file, which is an output of the Idris compiler. That's fine for a one-time fix of bootstrap.
But it's actually changing the source to the compiler itself.  If you recompile Idris with this bootstrap compiler, it will no longer produce compatible programs.

The PR changes `CompilerC-45SchemeC-45Chez-schHeader` which is the output of compiling `schHeader` in `src/Compiler/Scheme/Chez.idr` (around line 98).  That is the file needs to be changed for a permanent fix.

After that, I don't think we need the additional checklist item, when the next bootstrap file is generated, it should be built with the updated compiler.

barracuda156:
@dunhamsteve Thank you! For some reason BBEdit did not find that instance for me. I will fix it now and drop checklist.

barracuda156:
@dunhamsteve Is it correct now?

dunhamsteve:
I think you were doing `ChezSep.idr` too (I missed that until I looked closer at the .ss patch). I don't know if that's used anymore, but might as well.

barracuda156:
> I think you were doing `ChezSep.idr` too (I missed that until I looked closer at the .ss patch). I don't know if that's used anymore, but might as well.

@dunhamsteve Thank you, added that.

dunhamsteve:
Looks good to me.

mattpolzin (commented):
There's one other spot where chez handles the machine type: `support/chez/support.ss`.

I think it's fine if support is selectively added for Chez, but it might be just as well (and a bit more robust) to also add the same additional machine types for the Racket backed if they are supported by Racket as well.

CodingCellist:
Seems like Racket does indeed support Mac PowerPC: https://github.com/racket/racket/blob/cb410f3e9c349588f8c55aa687aa16585f16e8c1/racket/src/bc/sconfig.h#L647-L660

barracuda156:
> Seems like Racket does indeed support Mac PowerPC: https://github.com/racket/racket/blob/cb410f3e9c349588f8c55aa687aa16585f16e8c1/racket/src/bc/sconfig.h#L647-L660

On a side note, this is slightly incorrect, it seems: it will hardcode the arch to `ppc`, even if the build is for `ppc64` (`__POWERPC__` evaluates as true for both).

barracuda156:
@mattpolzin This makes sense, but I need to verify if it actually works. Turns out, there is no port for `racket`, so I do not have it installed. Let me try writing a port and building it.

barracuda156:
@mattpolzin @dunhamsteve Looks like Racket does not specify arch anywhere besides the bootstrap file? How is it generated?

For chez, I have added ppc to `support.ss` now.

barracuda156:
Re `Racket` itself on `ppc`: it does not look too bad, I almost got it built, but destroot failed with a strip error: https://github.com/racket/racket/issues/5021
Perhaps should be disabled or done with different options.

mattpolzin (approved):
Let's not worry too much about Racket in this PR given the difficulty with even checking that support for Racket on PowerPC is working.

The looks good to me. Thanks!

As a side note, I think there's something fishy with our Racket bootstrap code. It does not appear to be the result of the Racket.idr compiler sourcecode... Nothing to be addressed in this PR. I could also be making the wrong assumption about whether the Racket backend uses any of the Chez backend's code (other than the common scheme files which of course are shared).

CodingCellist:
> Looks like Racket does not specify arch anywhere besides the bootstrap file? How is it generated?
> 
> For chez, I have added ppc to `support.ss` now.

(and @mattpolzin)

I am also scratching my head here... As far as I can tell, based on the bootstrap files going all the way back to `v0.2.0`, it seems it just uses the Chez Scheme arch library code-string. Not entirely sure from where, as `Compiler.Chez` doesn't `export` its `schHeader`? Also, how on Earth has the Racket CI ever been passing then?... Currently doing a complete bootstrap of Idris2 backed by Racket only, to see if I can repro the Chez code being present.

But indeed, this is a problem for a different PR. I am also happy with this as-is  : )

CodingCellist:
I think I've narrowed it down: we include the Chez Scheme header in the Racket bootstrap file because it's derived from building `idris2.ipkg` and then extracting the resulting `build/exec/idris2_app/idris2.rkt`, which includes the Chez Scheme compiler because, well, it's part of Idris2 (afaiu this means the self-hosting version technically supports any backend code, just not the ability to run it).

I'm basing this off that afaict, `CompilerC-45SchemeC-45Chez-schHeader` is only invoked in the compiled Racket file as part of the `CompilerC-45SchemeC-45Chez-compileToSS` and `CompilerC-45SchemeC-45Chez-compileToSSInc` functions.

So maybe Racket handles architecture-specific things at the time of building Racket itself and then it is forever part of that install?

CodingCellist:
@mattpolzin @barracuda156 do you think it would make sense to revert the changes to the bootstrap files? These will be overwritten with fresh builds on the next release (whenever that may be), and with the other changes from this PR landing, should automatically include support for ppc (at least for the Chez compiler).

mattpolzin:
I think it’s worth making the changes to the bootstrap files here or else we would need to wait for a release to actually support bootstrapping on a PowerPC as is the original goal here.

barracuda156:
Anything different from these?
https://github.com/idris-lang/Idris2/pull/3326/commits/76b02f5459043af34ea59e9101acf31051cf41a2
https://github.com/idris-lang/Idris2/pull/3326/commits/e4f6dc12ccaac35326d55ebbc7e5a0b3380f5f1e
It works locally on `ppc` for me.

Issue #3327, opened 2024-06-21
Support type annotations on LHS of a monadic bind `<-`


joelberkeley:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

I want to be able to annotate the LHS of a monadic bind with types and quantities, just like with `let`. I usually want this when debugging type errors.

## Motivation

If my function is not type checking, I scatter the code with annotations until I find the problem. This is easy with `let`, but with monadic bind, I can only pseudo-annotate the RHS with `the`, like
```idris
foo : IO Nat

main : IO ()
main = do
  x <- the (IO Nat) foo
  printLn x
```
However, this is verbose, in part because I have to name the monad in the "annotation". I'd much rather be able to write
```idris
main : IO ()
main = do
  x : Nat <- foo
  printLn x
```

## The proposal

On the LHS of a monadic bind, rather than just a pattern, you can also add quantity and/or type annotations, like
```idris
f = do
  0 x <- foo
  ?res

g = do
  x : Nat <- foo
  ?res

h = do
  0 x : Nat <- foo
  ?res
```
I don't believe there would be any breaking changes, as it's optional, and I can't think of any valid syntax like it. The syntax would identical to the syntax between `let` and `=`.

### Examples

see above

### Technical implementation

This is beyond me atm.

## Alternatives considered

No other options come to mind

## Conclusion

This feature has received interest from core contributors. I can also imagine it using a similar grammar to the monadic `let` binding, even if it's interpreted differently, so I don't expect it to diverge from the current grammar, just add to it.

buzden:
It also needs to be thought how this feature would interplay with pattern matching and shortcut alternatives:

```idris
do (a, b) : (X, Y) <- foo
   1 (c, d, e) : (A, B, C) <- bar
   5 : Nat <- baz
     | 6 => unbaz
     | _ => unbaf
   unfoo
```

joelberkeley:
Those appear to me to all agree with how `let` bindings work (which is a very good thing, as having subtly different syntax for each would be a recipe for confusion)

dunhamsteve:
I don't think the `5` example is consistent with the current behavior of `let`.  Because of parsing for multiplicities, `let` will give a parse error for:
```idris
let 5 := ...
```
and 
```idris
let 5 : Nat := ...
```
but accepts:
```idris
[...]

PR #3329, opened 2024-06-23
Add type annotations to monadic bind #3327

dunhamsteve:
# Description

This PR implements most of the proposal in #3327. With this change we can optionally provide a type annotation in a monadic bind, following the syntax of the same in a `let` statement:
```idris
test1 = do
  x : Nat <- foo
  printLn x
```
and with pattern matching
```idris
test6 = do
  Just x : Maybe Nat <- bar
    | Nothing => pure ()
  printLn x
```

The discussion in #3327 (and on discord) mentions support for declaring multiplicities on the binds like in `let` statements.  This is implemented when binding a variable name, but not for a pattern matching bind.  The syntax for pattern quantities would have broken existing code and would not function correctly because of issues related to #2513 (`ICase` does not have a quantity).

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Idris/Desugar.idr
+++ b/src/Idris/Desugar.idr
@@ -743,7 +743,7 @@ mutual
                         (\ty => desugarDo side ps ns ty) ty
            pure $ bindFun fc ns tm'
                 $ ILam nameFC rig Explicit (Just n) ty' rest'
-  expandDo side ps topfc ns (DoBindPat fc pat rig ty exp alts :: rest)
+  expandDo side ps topfc ns (DoBindPat fc pat ty exp alts :: rest)
       = do pat' <- desugarDo LHS ps ns pat
            (newps, bpat) <- bindNames False pat'
            exp' <- desugarDo side ps ns exp
@@ -756,7 +756,7 @@ mutual
            ty' <- maybe (pure $ Implicit fc False)
                         (\ty => desugarDo side ps ns ty) ty
            pure $ bindFun fc ns exp'
-                $ ILam EmptyFC rig Explicit (Just (MN "_" 0))
+                $ ILam EmptyFC top Explicit (Just (MN "_" 0))
                           ty'
                           (ICase fc [] (IVar patFC (MN "_" 0))
                                (Implicit fc False)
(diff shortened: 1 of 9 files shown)


src/Idris/Desugar.idr
```diff
@@ -735,14 +735,15 @@ mutual
       = do tm' <- desugarDo side ps ns tm
            rest' <- expandDo side ps topfc ns rest
            pure $ seqFun fc ns tm' rest'
-  expandDo side ps topfc ns (DoBind fc nameFC n tm :: rest)
+  expandDo side ps topfc ns (DoBind fc nameFC n rig ty tm :: rest)
       = do tm' <- desugarDo side ps ns tm
            rest' <- expandDo side ps topfc ns rest
            whenJust (isConcreteFC nameFC) $ \nfc => addSemanticDecorations [(nfc, Bound, Just n)]
+           ty' <- maybe (pure $ Implicit (virtualiseFC fc) False)
+                        (\ty => desugarDo side ps ns ty) ty
```
gallais, line 743: Isn't it bizarre that we are using `desugarDo` to elaborate the type?
Does it even make sense to have an effectful function returning a type here?

dunhamsteve: Yeah, that struck me as odd, but I followed the pattern of `DoLet` and `DoLetPat`. It looks like they were changed from a full `desugar` to `desugarDo` in  ee063a5 to propagate the `do` namespace.

gallais: Right, I think that should be reverted.
Unless @Russoul has a motivation for it.

dunhamsteve: I'll let you guys sort out what is wanted here. It looks like `desugar` simply calls `desugarDo` with a namespace of `Nothing`.

dunhamsteve: Yeah I think the function name threw me off, but either looks fine. The difference seems to be what namespace ends up on comprehensions, which I wouldn't expect to appear in types.

What did you want to revert? Is it just changing the other`ty` elaborations to `desugar` or more of #1700?

PR #3336, opened 2024-06-29
Support for non-glibc versions of linux (musl, android)

spocino:
check if a linux implementation uses glibc to decide whether to load "glibc.so.6" or "glibc.so"

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -94,13 +94,9 @@ schHeader chez libs whole
     (import (chezscheme))
     (case (machine-type)
       [(i3fb ti3fb a6fb ta6fb) #f]
-      [(i3le ti3le a6le ta6le tarm64le) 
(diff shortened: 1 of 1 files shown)


Z-snails:
Would it be easier/faster to try loading `libc.so.6` first, then if that fails try `libc.so` (using `with-exception-handler`) - that way we don't need to spawn a child process.

mattpolzin:
I think that suggestion makes sense to me too; I think we could go even further and avoid the runtime check or fallback entirely too since our goal in loading the dynamic library is not actually to build once and then run in multiple different environments so we ought to be able to determine which library to use at compile time.

mattpolzin:
I’ll try and find time to play with this branch in an alpine Linux docker container today.

mattpolzin:
Seems to be working!

No `libc.so.6` in sight but `getEnv` is working.

mattpolzin (approved):

PR #3340, opened 2024-07-01
[ fix #3339 ] Set the global test locale

### LINKED ISSUE #3339: Test idris2/reflection/reflection024 fails due to not accounting for locales
When attempting to install idris2 on Arch Linux (using the [idris2](https://aur.archlinux.org/packages/idris2) package in AUR) test `idris2/reflection/reflection024` fails. As can be seen in the log (lines 3741 and 3804), this happens because `cat: src/nonExistentToRead: No such file or directory` is expected to be in the output, but with a different locale `cat` can print the error in a different language, and the test fails. Rerunning the build with `LC_ALL=C` fixes the issue.

[idris2_install.log](https://github.com/user-attachments/files/16054851/idris2_install.log)

image

buzden:
# Description

Sets the locale in the `testutils.sh`. I think, this is right place do to this, since no other test wants to observe the same problem in the future.

### DIFF
--- a/tests/testutils.sh
+++ b/tests/testutils.sh
@@ -109,6 +109,9 @@ if [ -z "$PREFIX_CHANGED" ] && [ -n "$IDRIS2_PREFIX" ]; then
     export PREFIX_CHANGED=1
 fi
 
+# Set the most neutral locale for reproducibility
+export LC_ALL=C.UTF-8
+
 # Remove test directory from output
 # Useful for consistency of output between machines
 # Usage: run SomeTest.idr | filter_test_dir


tests/testutils.sh
```diff
@@ -109,6 +109,9 @@ if [ -z "$PREFIX_CHANGED" ] && [ -n "$IDRIS2_PREFIX" ]; then
     export PREFIX_CHANGED=1
 fi
 
+# Set the most neutral locale for reproducibility
+LC_ALL=C.UTF-8
```
mattpolzin, line 113:
```suggestion
export LC_ALL=C.UTF-8
```

buzden: Well, I didn't do this because I was afraid of polluting the external context. But I tested that this does not which confuses me -- it means that I don't understand the difference. Fix without `export` works too.

PR #3348, opened 2024-07-13
Add `fromRight` and `fromLeft` for extracting values out of `Either`

joelberkeley:
I'm using this to provide more informative proofs on function arguments than `IsJust`.

### DIFF
--- a/libs/base/Data/Either.idr
+++ b/libs/base/Data/Either.idr
@@ -41,6 +41,11 @@ export
 Uninhabited (IsRight (Left x)) where
   uninhabited ItIsRight impossible
 
+||| Returns the `r` value of an `Either l r` which is proved `Right`.
+fromRight : (e : Either l r) -> {auto 0 isRight : IsRight e} -> r
+fromRight (Right r) = r
+fromRight (Left _) impossible
+
 ||| Proof that an `Either` is actually a Left value
 public export
 data IsLeft : Either a b -> Type where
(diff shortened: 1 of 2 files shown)


joelberkeley:
It might be the case that `f x` in `IsRight (f x)` doesn't always reduce, in which case using `IsRight` might not serve much benefit over `IsJust`

libs/base/Data/Either.idr
```diff
@@ -41,6 +41,11 @@ export
 Uninhabited (IsRight (Left x)) where
   uninhabited ItIsRight impossible
 
+||| Returns the `r` value of an `Either l r` which is proved `Right`.
+fromRight : (e : Either l r) -> {auto 0 isRight : IsRight e} -> r
```
andrevidela, line 45: `auto` should be removed to avoid unpredictable search results and to match the existing API of `fromJust` https://github.com/idris-lang/Idris2/blob/be2ec7d1fd8897ea0b786de08dd366089e483d29/libs/base/Data/Maybe.idr#L43

libs/base/Data/Either.idr
```diff
@@ -50,6 +55,11 @@ export
 Uninhabited (IsLeft (Right x)) where
   uninhabited ItIsLeft impossible
 
+||| Returns the `l` value of an `Either l r` which is proved `Left`.
+fromLeft : (e : Either l r) -> {auto 0 isLeft : IsLeft e} -> l
```
andrevidela, line 59: ditto here

andrevidela:
Looks good

PR #3349, opened 2024-07-14
Support more flexible requirements for Golden tests

mattpolzin:
# Description

This change to how Golden tests handle requirements for a pool of tests makes the test framework much more versatile. It is a breaking change but most code using the test lib (including the compiler test suite) won't need to change in response to it.

## Details

I've introduced a new `Requirement` record type to allow projects using this library to define their own types of requirements. Existing constructors of the old `Requirement` type have been rewritten as same-named functions returning values of the new record type.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -242,28 +242,8 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 
 #### Test
 
-* Renamed `Requirement` to `BackendRequirement`. A new interface has been
-  introduced with the name `Requirement`, so you should proactively adopt the new
-  `BackendRequirement` name anywhere you've used `Test.Golden.Requirement`
-  previously to avoid potentially difficult to understand compiler errors.
-
-* Both the `TestPool` type and the `testsInDir` helper have been reworked to
-  support a more flexible definition of a "requirement." Any `TestPool` or use
-  of `testsInDir` that does _not_ specify any constraints/requirements need not
-  change at all. A `TestPool` with existing requirements shouldn't need to
-  change, either, unless you'd like to use your own `Requirement` type going
-  forward instead of the default `BackendRequirement`. If you have a
-  `testsInDir` with requirements, it will need to change to a `testsInDir'`;
-  note the tick (apostrophe) in the name. In full, make the following change.
-
-      testsInDir "testdir" "My Tests" {requirements = [C, Node]
-
-  Should be refactored to:
-
-      testsInDir' "testdir" "My Tests" [C, Node]
-
-* The new `Requirement` interface in `Test.Golden` can be implemented for any
-  type you would like to use to specify requirements for a test pool. This makes
-  it quite easy to write test pools that are skipped based on any number of
-  things.
+* Replaced `Requirement` data type with a new record that can be used to create
+  any requirement needed. The constructors for the old `Requirement` type are
+  now functions of the same names that return values of the new record type so
+  in most situations there should be no compatibility issues.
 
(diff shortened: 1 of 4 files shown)


CHANGELOG_NEXT.md
```diff
@@ -238,3 +238,32 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
...
+  change, either, unless you'd like to use your own `Requirement` type going
+  forward instead of the default `BackendRequirement`. If you have a
+  `testsInDir` with requirements, it will need to change to a `testsInDir'`;
+  note the tick (apostrophe) in the name. In full, make the following change.
+
+      testsInDir "testdir" "My Tests" {requirements = [C, Node]
+
+  Should be refactored to:
+
+      testsInDir' "testdir" "My Tests" [C, Node]
```
gallais, line 263: I'm guessing the change comes from the fact it's impossible to infer the `req` type from an empty list.
Why not directly define
```
record Requirement where
  constructor MkReq
  reqName : String
  reqTest : IO (Maybe String)
```
instead and then introduce definition for `C`, `Chez`, etc. as top-level constants of type `Requirement`
rather than constructors? It would then allow you to keep the same `testInDir` interface with the
requirements still defaulting to the empty list.

mattpolzin: @gallais I've pushed your suggestion as the newest commit (without updating the CHANGELOG or README changes just yet).

Overall, it's definitely less disruptive I'd say. The key drawbacks to the new approach  (not that they outweigh the benefits) are:
1. You can't implement equality on the `C`, `Node`, etc. requirements anymore (you need to use their name as a proxy).
2. You can't readily distinguish between a use of the `C`, `Node`, etc. requirements as a name vs. as the name of the codegen that requires them -- I ended up just changing the names to those of the codegens since that is effectively how the builtin requirements are used.

gallais: You can define a `Eq` implementation on `Requirement` to be equality of names.
This way you avoid the danger of inadvertently using the wrong string literal in the guards.

mattpolzin: That did cross my mind. I originally didn't do that because it kind of hides the fact that there's not anything stopping you from adding two requirements with the same name and different `satisfy` implementations. Still, I think you're right that avoiding the fragility of typing the name out when comparing is probably preferable. I know we could protect against duplicate names, just wasn't sure I wanted to go there with this PR.

libs/test/Test/Golden.idr
```diff
@@ -328,66 +328,78 @@ pathLookup names = do
                                          y <- extensions]
   firstExists candidates
 
-
-||| Some test may involve Idris' backends and have requirements.
-||| We define here the ones supported by Idris
+||| A test requirement. The String value returned from `satsify` witnesses requirement.
```
gallais, line 331:
```suggestion
||| A test requirement. The String value returned from `satisfy` witnesses requirement.
```

libs/test/Test/Golden.idr
```diff
@@ -614,6 +630,14 @@ poolRunner opts pool
 
   where
 
+    checkReq : {0 req : Type} -> Requirement -> IO (Maybe String, String)
+    checkReq req = do
+      mfp <- satisfy req
+      let msg = case mfp of
+                  Nothing => "✗ " ++ (name req) ++ " not found"
+                  Just fp => "✓ Found " ++ (name req) ++ " at " ++ fp
```
gallais, line 638:
```suggestion
                  Nothing => "✗ \{name req}  not found"
                  Just fp => "✓ Found \{name req} at \{fp}"
```

gallais (approved):

Issue #1918, opened 2021-09-12
Known issues of the `--mkdoc` command


gallais:
Copy/pasted from the original PR (#903):

* [ ] types are not shown where they should be
      (although most of the time, the type is in the first line of the documentation)
      [NDLA: Not sure what this means?]
* [ ] "documentation" is currently just a <pre> tag with the output equivalent to :doc in the REPL
* [ ] namespace members are sorted alphabetically, this may not be the best choice
      [NDLA e.g. we could use file position instead to have a more scope-friendly presentation.
      This would also allow us to include things like headers in between sections]
* [ ] `--clean` does not remove the generated docs
* [X] code needs some cleanup (and probably moved into a separate file, as it pollutes
      `Package.idr` with lots of HTML output-specific code, which will get worse as more
      features are implemented)

Additional issues:

* [ ] Each doc file corresponds to a physical file and only contains the definitions in that
      exact namespace. We should either fully embrace the namespace-based approach and have
      one file per namespace or make sure the doc file also contains the definitions introduced
      in a nested namespace

gallais:
#2411 addresses
1. alphabetical order of definition (now changed to `FC`'s `startPos` order)
2. content of namespace rather than source file

Matthew-Mosior:
#3353 addresses `--clean does not remove the generated docs`.

PR #3353, opened 2024-07-25
Clean removes generated docs

Matthew-Mosior:
This PR enables `--clean` to remove documentation generated via `--mkdoc`.

This addresses one of the tasks in #1918.

### DIFF
--- a/src/Idris/Package.idr
+++ b/src/Idris/Package.idr
@@ -847,16 +847,16 @@ clean pkg opts -- `opts` is not used but might be in the future
          let build = build_dir (dirs (options defs))
          let docBase = build </> "docs"
          let docDir = docBase </> "docs"
-         Right docfiles' <- coreLift $ listDir docDir
-           | Left err => pure ()
-         Right docfiles'' <- coreLift $ listDir docBase
-           | Left err => pure ()
-         traverse_ (\x => delete $ docDir </> x)
-                   docfiles'
-         traverse_ (\x => delete $ docBase </> x)
-                   docfiles''
-         deleteFolder docDir []
-         deleteFolder docBase []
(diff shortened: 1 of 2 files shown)


src/Idris/Package.idr
```diff
@@ -843,6 +843,20 @@ clean pkg opts -- `opts` is not used but might be in the future
          deleteFolder builddir []
          maybe (pure ()) (\e => delete (outputdir </> e))
                (executable pkg)
+         -- clean out the generated docs
+         let build = build_dir (dirs (options defs))
+         let docBase = build </> "docs"
+         let docDir = docBase </> "docs"
+         () <- do Right docfiles' <- coreLift $ listDir docDir
```
andrevidela, line 850: Looks great! Can we refactor the two do-blocks in their own auxiliary function?

Matthew-Mosior: Thank you!  That sounds good, should be resolved by commit ccff608e1970f2b205b3e8ea7afdb967eb85c9dc.

andrevidela: Sorry that's not quite what I meant, the two functions you put in auxiliary definitions are essentially the same, except for the file path, it would be nice to remove this redundancy.

Matthew-Mosior: Ah I see!  I think I've addressed this with d788e07c327b9bf57d154d7b75304e9481c1153f.

Matthew-Mosior: Not sure why [this test](https://github.com/idris-lang/Idris2/actions/runs/10199548283/job/28217862470?pr=3353) is failing ...

Matthew-Mosior:
@andrevidela I think this should be ready for you now :)

andrevidela:
Looks good!

PR #3360, opened 2024-07-26
`idris2 --init` doesnt check the name of a package

### LINKED ISSUE #3232: `idris2 --init` doesn't check a name of a package
# Steps to Reproduce
1. Call `idris2 --init`
2. Give a name for a package that is not a correct Idris identificator, i.e. `123_Issue`
3. Finish the package in any reasonable way (at least set everything else empty)

# Expected Behavior
The program doesn't allow such a name for a package since it doesn't compile

# Observed Behavior
The program allows to create a package with such a name

Matthew-Mosior:
This PR closes #3232.

This issue stemmed from the lack of parsing in `interactive` found in `src/Idris/Package/Init.idr`.

### DIFF
--- a/src/Idris/Package/Init.idr
+++ b/src/Idris/Package/Init.idr
@@ -68,7 +68,6 @@ prompt p = putStr p >> fflush stdout >> getLine
 
 export
 covering
---interactive : IO PkgDesc
 interactive : IO (Either () PkgDesc)
 interactive = do
   pname    <- prompt "Package name: "
--- a/src/Idris/Package/Init.idr
+++ b/src/Idris/Package/Init.idr
@@ -70,7 +70,7 @@ export
 covering
 interactive : IO (Either () PkgDesc)
 interactive = do
-  pname    <- prompt "Package name: "
+  pname <- prompt "Package name: "
   -- check to ensure that pname is a valid Idris2 identifier.
   case checkPackageName $ fastUnpack pname of 
     False => do () <- putStrLn "Package name is not a valid Idris Identifier."
--- a/src/Idris/Package/Init.idr
+++ b/src/Idris/Package/Init.idr
@@ -112,6 +112,6 @@ interactive = do
                                 True  => isIdentTrailing xs
 
     checkPackageName : List Char -> Bool
-    checkPackageName []      = False
+    checkPackageName []      = True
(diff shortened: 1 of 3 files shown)


src/Idris/Package/Init.idr
```diff
@@ -67,25 +68,50 @@ prompt p = putStr p >> fflush stdout >> getLine
...
-  let pkg : PkgDesc =
-        { authors   := mstring pauthors
-        , options   := (emptyFC,) <$> mstring poptions
-        , modules   := modules
-        , sourcedir := sourcedir
-        } (initPkgDesc (fromMaybe "project" (mstring pname)))
-  pure pkg
+  pname <- prompt "Package name: "
+  -- check to ensure that pname is a valid Idris2 identifier.
+  case checkPackageName $ fastUnpack pname of
```
andrevidela, line 75: Might be worth using inline pattern matching 
```suggestion
  let True = checkPackageName $ fastUnpack pname 
    | False => printLn >> Left () 
```

src/Idris/Package/Init.idr
```diff
@@ -67,25 +68,50 @@ prompt p = putStr p >> fflush stdout >> getLine
 
 export
 covering
-interactive : IO PkgDesc
+interactive : IO (Either () PkgDesc)
```
andrevidela, line 71: Using Either () instead of Maybe is a bit unconventional, is there a specific reason for that?

src/Idris/Package/Init.idr
```diff
@@ -67,25 +68,50 @@ prompt p = putStr p >> fflush stdout >> getLine
...
+                pure $ Right pkg
 
   where
 
     mstring : String -> Maybe String
     mstring str = case trim str of
       "" => Nothing
       str => Just str
+
+    isIdentStart : Char -> Bool
```
andrevidela, line 98: I suspect that code (and below) came from the lexer but there is no way to
import and reuse the bits for identifiers, is that right?

Matthew-Mosior: @andrevidela  I don't think there's a straightforward to use that code in this context (could be wrong), but totally open to changing the way I'm doing this if there is a better way (or a way that can utilize the lexer code).

PR #3363, opened 2024-08-02
[ fix ] Report chez failures

dunhamsteve:
# Description

If chez exits with a status code other than 0 during compilation, Idris ignores the error and completes successfully, leaving a non-working executable. For example, this occurs when an out of memory error occurs during compilation.  I've changed `Chez.idr` to check the return code and throw an `InternalError` reporting the code if it is not zero.

Before this change, `CHEZ=false idris2 -c Hello.idr -o hello` would complete successfully, but `./build/exec/hello` would report:
```
./build/exec/hello: line 15: /Users/dunham/code/i2/Idris2/build/exec/hello_app/hello.so: No such file or directory
```
after the change, the compilation step fails with:
```
Error: INTERNAL ERROR: Chez exited with return code 1
```

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -535,7 +535,8 @@ compileToSO prof chez appDirRel outSsAbs
          Right () <- coreLift $ writeFile tmpFileAbs build
             | Left err => throw (FileErr tmpFileAbs err)
          coreLift_ $ chmodRaw tmpFileAbs 0o755
-         coreLift_ $ system [chez, "--script", tmpFileAbs]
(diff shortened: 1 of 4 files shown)


mattpolzin (approved):
Well that's a pretty clear win.

gallais:
Can we have the same for the other backends too?

dunhamsteve:
It looks like the other backends check the return value for compilation.  None of them check the return value when running the result of compilation for REPL `:eval`.

The windows CI failure is for `chez/forkjoin001`, which produced no output. I believe it's unrelated to this change.  It did not fail for me when github ran CI for this commit in my repository.  Possibly something timing sensitive?

PR #3366, opened 2024-08-06
[ fix ] Make searching in pairs to not cause ambiguity when there isn't

### LINKED ISSUE #3365: Searching of `auto` inside pairs introduces ambiguity error between actually same things
# Steps to Reproduce

Consider a data type, a function that takes a value of this type as an `auto`-implicit argument, and a function that produces a value of this type paired with some other value:

```idris
data X = MkX

useX : X => Nat

v : (X, String)
```

Now consider that you take this pair, pattern match on it and call a function that takes an `auto`-implicit:

```idris
f : Nat
f = do
  let (x, _) = v
  useX
```

The function typechecks, as expected.

You can also do not pattern match at all, and the `auto`-implicit value would still be correctly found:

```idris
f' : Nat
f' = do
  let xx = v
  useX
```

This is done because Idris treats pairs specially during the `auto`-implcit search in order to be able Haskell-like constraints work as well as native Idris-like ones:
```idris
cfH : (Eq a, Monoid a) => ... -- Haskell-like style
cfI : Eq a => Monoid a => ... -- Idris-like style
```

But now consider that you both named the whole pair while still pattern matching it:

```idris
g : Nat
g = do
  let xx@(x, _) = v
  useX
```

Or consider that you did the original pattern-matching in a, say, monadic code:

```idris
g' : Monad m => m Nat
g' = do
  (x, _) <- pure v
  pure useX
```

Or consider you did both:

```idris
g'' : Monad m => m Nat
g'' = do
  xx@(x, _) <- pure v
  pure useX
```

# Expected Behavior

All `g*` functions typecheck well

# Observed Behavior

All `g*` functions fail with a search ambiguity error.

For the `g` it is

```
Error: While processing right hand side of g. Multiple solutions found in search of:
    X

PairErroneouslyFound:20:3--20:7
 16 | 
 17 | g : Nat
 18 | g = do
 19 |   let xx@(x, _) = v
 20 |   useX
        ^^^^

Possible correct results:
    fst xx
    x
```

For `g'` it is

```
Error: While processing right hand side of g'. Multiple solutions found in search of:
    X

PairErroneouslyFound:25:8--25:12
 21 | 
 22 | g' : Monad m => m Nat
 23 | g' = do
 24 |   (x, _) <- pure v
 25 |   pure useX
             ^^^^

Possible correct results:
    fst _
    x
```

And for `g''` it is

```
Error: While processing right hand side of g''. Multiple solutions found in search of:
    X

PairErroneouslyFound:30:8--30:12
 26 | 
 27 | g'' : Monad m => m Nat
 28 | g'' = do
 29 |   xx@(x, _) <- pure v
 30 |   pure useX
             ^^^^

Possible correct results:
    fst xx
    fst _
    x
```

# Analysis

In all these cases "the whole pair" value (being applied `fst` to it) is used in the search along with the pattern matched value. In monadic case it is unnamed but still present in the context because of the `pure v >>= \case (x, _) => useX` translation, i.e. this "whole value" is present no matter what user is doing.

I think, that this search should not lead to an ambiguity error because of two reasons:
1) `fst _`, `fst xx` and `x` are literally the same thing in these examples;
2) searching inside pairs is an additional functionality, which should help to find values from inside pairs, but shouldn't break the search if there are pairs from another reason.

All this happends inside `Core.AutoSearch` inside `searchLocalWith` function, in particular, inside [`findPos`](https://github.com/idris-lang/Idris2/blob/main/src/Core/AutoSearch.idr#L320) one: it tries to find the name directly and falls back to trying to apply `fst` and/or `snd` and repeat the search recursively. I think, it does what it does a bit hastily, so that looking inside happens even if nice value still can be found without it. However, I don't know if they could be prioritised, because we currently typecheck these two functions with the same error and it looks reasonable:
```idris
failing "Multiple solutions found"
  bad1 : X => Y => X => Nat
  bad1 = useX

failing "Multiple solutions found"
  bad2 : (X, Y) => X => Nat
  bad2 = useX
```

buzden:
Fixes #3365. Also contains two commits with mostly unrelated cleaning up, which is small enough to not to bother with separate PRs for it.

In this PR it was decided that we would normalise only `fst`s and `snd`s introduced by special treatment of searching inside the pairs, so that we do not overnormalise and do not spend to much time on it. It was also decided to be okay to call `nub` on the intermediate results in `exactlyOne` because normally those lists and terms can't be big.

Preparational refactoring and actual change are in separate commits to make review easier.


buzden:
Just for a record: this fix still does not help in the case of `case`, i.e. when `auto`-search is behind `case` comparing to the place of the pattern matching, e.g.

```idris
gC : Nat
gC = do
  let xx@(x, _) = v
  case MkUnit of
    _ => useX
```

This is likely because `case` expressions (now) are elaborated as special case functions, and both arguments `xx` and `x` are passed independently, thus, all interconnection between them is lost, thus leading to the same ambiguity error as before this PR. This particular problem, probably, may be fixed by re-pattern matching all `as`-variables in the case function, but the case thing should be done for the anonymous argument of a lambda in `>>=`, which is a much harder thing.

Issue #3157, opened 2023-12-04
IDE command to print docs for the type at point


ohad:
- [X ] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [X ] I have checked that there is no existing PR/issue about my proposal.

## Summary

The IDE protocol currently supports a `DocsFor` command for presenting the documentation for some name.
When that name is a hole, what the user wants is the documentation for the type of hole, rather than the non-existing documentation for the hole.

## Motivation

```idris
data MyFancyType : Type where
  C1 : List String -> MyFancyType
  C2 : (Bool,Type) -> MyFancyType
  C3 : Either MyFancyType Nat -> MyFancyType

Foo : MyFancyType
--    ^1
Foo = ?whatToChoose
--    ^2
```
If the user places their cursor at the character pointed at by caret 1 (on `MyFancyType`) and issues a `DocsFor` command (e.g., `idris-docs-at-point (C-c C-d d)` in emacs's idris-mode), we get a lot of useful information:
```idris
data Main.MyFancyType : Type
  Totality: total
  Visibility: private
  Constructors:
    C1 : List String -> MyFancyType
    C2 : (Bool, Type) -> MyFancyType
    C3 : Either MyFancyType Nat -> MyFancyType

[back]
```
But if we're pointing at the hole `whatToChoose` (caret 2), then asking for the docs gives us nothing:
```idris


[back]
```
What we really want at this point is the docs for the type of `whatToChoose`, which will give us the earlier output, which will help us choose the right constructor.

In this example, we can work around by navigating to `Foo`'s type declaration and asking for the docs, but if we're working on a hole nested inside another term, we can't do it so easily.  What we need to do in that case is ask idris for the type of the name under the cursor (`idris-type-at-point (C-c C-t)`, switch to the info buffer so that the cursor now points at the type, and then ask the IDE protocol for the docs.

## The proposal

Add a `DocsForTypeOf` command to the IDE protocol that returns the docs for the type of the name given.

### Examples
In the example above, placing the cursor at caret 2 and issuing a `DocsForTypeOf` (say, `idris-docs-for-type-at-point (C-c C-d t)`) would display:
```idris
Documentation for type of whatToChoose : MyFancyType:
data Main.MyFancyType : Type
  Totality: total
  Visibility: private
  Constructors:
    C1 : List String -> MyFancyType
    C2 : (Bool, Type) -> MyFancyType
[...]

PR #3371, opened 2024-08-16
[ new ] add docs-for-type-of IDE command

DanMax03:
# Description

I've decided to try to implement the issue #3157 in this PR. Right now it's a draft as long as I haven't resolved several issues:
1) I haven't tested it yet, because I don't see any clear way to install Idris without touching the main one (which is connected to $HOME/.idris2)
2) The problem is, after getting the type of the hole, I have only `IPTerm` representation of it. Thus, for calling `process (DocsFor ? modeOpt)` I need to use either already done function, or to extract the type name somehow

Definitely yes as a feature implementation

- [ ]  If this is a fix, user-facing change, a compiler change, or a new paper
      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Idris/IDEMode/REPL.idr
+++ b/src/Idris/IDEMode/REPL.idr
@@ -199,6 +199,10 @@ process (MakeWith l n)
     = replWrap $ Idris.REPL.process (Editing (MakeWith False (fromInteger l) (UN $ mkUserName n)))
 process (DocsFor n modeOpt)
     = replWrap $ Idris.REPL.process (Doc $ APTerm (PRef EmptyFC (UN $ mkUserName n)))
+process (DocsForTypeOf n modeOpt)
+    = do (REPL (TermChecked itm ity)) <- Idris.REPL.process $ (Check $ PRef replFC (UN $ mkUserName n))
+           | err => pure $ REPL $ REPLError (pretty0 $ show err)
+         process (DocsFor (prettyBy Syntax ity) modeOpt)
 process (Apropos n)
     = do todoCmd "apropros"
          pure $ REPL $ Printed emptyDoc
@@ -513,5 +517,5 @@ replIDE
          case res of
               REPL _ => printError $ reflow "Running idemode but output isn't"
               IDEMode _ inf outf => do
-                send outf (ProtocolVersion 2 1) -- TODO: Move this info somewhere more central
+                send outf (ProtocolVersion 2 2) -- TODO: Move this info somewhere more central
                 loop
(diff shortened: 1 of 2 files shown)


DanMax03:
Requesting @ohad @gallais to comment

src/Idris/IDEMode/REPL.idr
```diff
@@ -199,6 +199,10 @@ process (MakeWith l n)
     = replWrap $ Idris.REPL.process (Editing (MakeWith False (fromInteger l) (UN $ mkUserName n)))
 process (DocsFor n modeOpt)
     = replWrap $ Idris.REPL.process (Doc $ APTerm (PRef EmptyFC (UN $ mkUserName n)))
+process (DocsForTypeOf n modeOpt)
+    = do (REPL (TermChecked itm ity)) <- Idris.REPL.process $ (Check $ PRef replFC (UN $ mkUserName n))
```
DanMax03, line 203: REPL constructor is an error here

src/Idris/IDEMode/REPL.idr
```diff
@@ -199,6 +199,10 @@ process (MakeWith l n)
...
+           | err => pure $ REPL $ REPLError (pretty0 $ show err)
+         process (DocsFor (prettyBy Syntax ity) modeOpt)
```
DanMax03, line 205: I have no idea how to work with ity properly in this place. Convert `ity : IPTerm` to `PTerm` somehow?

gallais:
If you chase various definitions you can see:

```idris
-- In Idris.IDEMode.REPL
process (DocsFor n modeOpt)
    = replWrap $ Idris.REPL.process (Doc $ APTerm (PRef EmptyFC (UN $ mkUserName n)))
```

```idris
-- Idris.REPL
process (Doc dir)
    = do doc <- getDocs dir
         pure $ PrintedDoc doc
```

```idris
-- In Idris.Doc.String
getDocsForPTerm (PRef fc name) = getDocsForName fc name MkConfig
```

So if you can get the name of the type's head constructor you should be golden with `getsDocsForName`.
Luckily there's a bunch of functions for you:

```idris
-- If you want to go under Pis in case the term is a function that's not fully applied
-- In Core.TT.Views
underPis : (n : Int) -> Env Term vars -> Term vars ->
           (bnds : SnocList Name ** (Env Term (bnds <>> vars), Term (bnds <>> vars)))

-- To grab the head of the application (after stripping off types if necessary)
-- In Core.TT.Term
getFn : Term vars -> Term vars
```

And you can then check that whatever you get back from `getFn` is indeed a `Ref`
(you probably also want to handle types that are `PrimVal` or `TType`).

gallais:
Marking as abandoned for now.

PR #3376, opened 2024-08-23
[ codegen ] get rid of artifacts introduced when optimizing `IO`

### LINKED ISSUE #3375: [ codegen ] Idris generates non-productive artifacts when optimizing `IO`
Idris is already quite good at optimizing / inlining `do` blocks for `IO` (and possibly - with proper inlining - for other monads), but this process can introduce quite a few non-productive artifacts, which makes the generated code less readable and potentially less efficient.

# Steps to Reproduce

Compute and run the following program and inspect the generated scheme and JavaScript code:

```idris
module Ref

import Data.IORef

%default total

%inline
release : IORef Nat -> IO ()
release ref = pure ()

%inline
readAndRelease : IORef Nat -> IO Nat
readAndRelease ref = do
  v <- readIORef ref
  release ref
  pure v

setget : IORef Nat -> IORef Nat -> IO (Nat,Nat)
setget r1 r2 = do
  writeIORef r1 100
  x <- readAndRelease r1
  y <- readAndRelease r2
  pure (x,y)

main : IO ()
main = do
  r1 <- newIORef Z
  r2 <- newIORef Z
  p  <- setget r1 r2
  printLn p
```

# Observed Behavior

The generated JavaScript for function `setget` looks as follows:

```javascript
/* Ref.setget : IORef Nat -> IORef Nat -> IO (Nat, Nat) */
function Ref_setget($0, $1, $2) {
 const $3 = ($0.value=100n);
 const $9 = ($0.value);
 const $d = undefined;
 const $8 = $9;
 const $f = ($1.value);
 const $13 = undefined;
 const $e = $f;
 return {a1: $8, a2: $e};
}
```

As can be seen, several lines in the code above are completely redundant. While the introduction of `const $d = undefined` should be easy to get rid of, the non-productive `let` bindings (`const $8 = $9`) are interesting as these should actually be caught during constant folding. Since the JS codegen flattnes and rearranges nested `let` bindings to a certain degree, it is easier to spot what's going on when inspecting the generated scheme code:

```scheme
(define Ref-setget
  (lambda (arg-0 arg-1 ext-0)
    (let ((act-1 (set-box! arg-0 100)))
      (let ((act-2
              (let ((act-2 (unbox arg-0)))
                (let ((act-3 (vector 0 ))) act-2))))
        (let ((act-3
                (let ((act-3 (unbox arg-1)))
                  (let ((act-4 (vector 0 ))) act-3))))
          (cons act-2 act-3))))))
```

In the code above, we see nested `let` bindings of the type `let x := y in x`. These can always be replaced with `y`, because they are only required for the side-effects being run in `y`.

Getting rid of these two kinds of artifacts should be straight forward with the already existing constant-folding machinery. I'll open a PR in a moment.

stefan-hoeck:
# Description

This fixes #3375 by adding two additional rules to constant folding.

### DIFF
--- a/src/Compiler/Opts/ConstantFold.idr
+++ b/src/Compiler/Opts/ConstantFold.idr
@@ -92,7 +92,6 @@ constFold rho (CLam fc x y)
 -- Expressions of the type `let x := y in x` can be introduced
 -- by the compiler when inlining monadic code (for instance, `io_bind`).
 -- They can be replaced by `y`.
-constFold rho (CLet fc x inl y $ CLocal {idx = 0} _ _) = constFold rho y
 constFold rho (CLet fc x inl y z) =
     let val := constFold rho y
      in case replace val of
--- a/src/Compiler/Opts/ConstantFold.idr
+++ b/src/Compiler/Opts/ConstantFold.idr
@@ -101,6 +101,8 @@ constFold rho (CApp fc (CRef fc2 n) [x]) =
             v                   => CApp fc (CRef fc2 n) [v]
      else CApp fc (CRef fc2 n) [constFold rho x]
 constFold rho (CApp fc x xs) = CApp fc (constFold rho x) (constFold rho <$> xs)
+constFold rho (CCon fc x UNIT tag []) = CErased fc
+constFold rho (CCon fc x RECORD tag [y]) = constFold rho y
 constFold rho (CCon fc x y tag xs) = CCon fc x y tag $ constFold rho <$> xs
 constFold rho (COp fc BelieveMe [CErased _, CErased _ , x]) = constFold rho x
 constFold rho (COp {arity} fc fn xs) =
--- a/src/Compiler/Opts/ConstantFold.idr
+++ b/src/Compiler/Opts/ConstantFold.idr
@@ -88,6 +88,11 @@ constFold rho (CLocal fc p) = lookup fc (MkVar p) rho
 constFold rho e@(CRef fc x) = CRef fc x
 constFold rho (CLam fc x y)
   = CLam fc x $ constFold (wk (mkSizeOf [x]) rho) y
+
+-- Expressions of the type `let x := y in x` can be introduced
+-- by the compiler when inlining monadic code (for instance, `io_bind`).
+-- They can be replaced by `y`.
+constFold rho (CLet fc x inl y $ CLocal {idx = 0} _ _) = constFold rho y
 constFold rho (CLet fc x inl y z) =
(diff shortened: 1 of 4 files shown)


src/Compiler/Opts/ConstantFold.idr
```diff
@@ -88,6 +88,11 @@ constFold rho (CLocal fc p) = lookup fc (MkVar p) rho
 constFold rho e@(CRef fc x) = CRef fc x
 constFold rho (CLam fc x y)
   = CLam fc x $ constFold (wk (mkSizeOf [x]) rho) y
+
+-- Expressions of the type `let x := y in x` can be introduced
+-- by the compiler when inlining monadic code (for instance, `io_bind`).
+-- They can be replaced by `y`.
+constFold rho (CLet fc x inl y $ CLocal {idx = 0} _ _) = constFold rho y
```
gallais, line 95: You should start by `constFold` the body of the let before checking
whether it is `CLocal` because that may very well reveal that it is!

gallais:
Can we add a test case demonstrating it indeed fixes #3375?
You should be able to use something like
```
awk -v RS= '/\/\* Main.main/' build/exec/main
```
to extract just the code of the main function in the test runner.

stefan-hoeck:
I added a test case using `grep` instead of `awk`. It checks the whole `.js` file for the patterns described in #3375 . None must be found. Funny enough, the new version no longer got rid of all the `undefined` cases. I therefore simplified the case where the variable term can be replaced.

src/Compiler/Opts/ConstantFold.idr
gallais: This reverses the change allowing a  `CLocal 0` being discovered for code
of the form `let z := x in let y := z in y`.

stefan-hoeck: > let z := x in let y := z in y

I added two additional functions to the test prog dealing with this kind of nested `let` statement:

```idris
%noinline
nestedLet : Nat -> Nat
nestedLet x =
  let z := x in let y := z in y

%noinline
nestedLetNoArg : Nat
nestedLetNoArg =
  let z := 10 * 10 in let y := z in y
```

Both get properly reduced:

```javascript
/* Prog.nestedLetNoArg : Nat */
const Prog_nestedLetNoArg = __lazy(function () {
 return 100n;
});

/* Prog.nestedLet : Nat -> Nat */
function Prog_nestedLet($0) {
 return $0;
}
```

PR #3381, opened 2024-09-11
[ new ] Support for dumping a package's install location

mattpolzin:
# Description

Historically I have hard coded assumptions about where Idris will install a package when `idris2 --install` is called. This works, but it's fragile, inconvenient, and far from obvious.

See, for example, this snippet from one of my C support library [Makefiles](https://github.com/mattpolzin/ncurses-idris/blob/main/support/Makefile):
```make
PACKAGE_INSTALLDIR = `${IDRIS} --libdir`
INSTALLDIR = ${PACKAGE_INSTALLDIR}/ncurses-idris-${TARGET_VERSION}/lib
```

This PR introduces `idris2 --dump-installdir` which could make the above Makefile snippet much more reasonable:
```make
DESTDIR = `${IDRIS} --dump-installdir --find-ipkg`
INSTALLDIR = ${DESTDIR}/lib
```

While I was adding `--dump-installdir` anyway, I also used it in the existing Nix `buildIdris` function to better support a "Pack-ism" where C support libraries are installed to the `./lib` folder and Pack takes care of the rest. Now `buildIdris` also takes care of the rest.

      `CONTRIBUTORS.md`).


gallais:
Can we get a test of this command-line option so that we are forced to think about any changes affecting the printed string?

mattpolzin:
Yeah, that’s a good call. I’ll add a test later today.

PR #3384, opened 2024-09-14
Correct ipkg comment docs

stephen-smith:
# Description

stefan-hoeck/idris2-pack#307  -- Neither Idris2 nor Pack actually support multiline comments in ipkg files.

Not IMO.


mattpolzin (approved):
I can confirm the multiline syntax does not work right now and this seems like a good documentation change until/unless we get multiline comments working in ipkg files.

Thanks!

stephen-smith:
I think I have a better resolution in #3386.  That one should be refined as needed instead of merging this one.

PR #3386, opened 2024-09-15
Handle multiline comments in Package (ipkg)

stephen-smith:
# Description

Alternative to #3384.

Implements multiline comments in ipkg to fulfill the promises of the existing documentation.

Added to _NEXT.

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Parser/Lexer/Package.idr
+++ b/src/Parser/Lexer/Package.idr
@@ -99,6 +99,7 @@ andop = is '&' <+> is '&'
 rawTokens : TokenMap Token
 rawTokens =
   [ (comment, Comment . drop 2)
+  , (blockComment, Comment . shrink 2)
   , (namespacedIdent, uncurry DotSepIdent . mkNamespacedIdent)
   , (identAllowDashes, DotSepIdent Nothing)
   , (separator, const Separator)
--- a/tests/idris2/pkg/multiline-comments/test.ipkg
+++ b/tests/idris2/pkg/multiline-comments/test.ipkg
@@ -1,0 +1,12 @@
+package test
+{-
+A test of
+multiline
+comments.
+-}
+version = 0.1.0
+authors = "Boyd Stephen Smith Jr."
+main = Main
+executable = "test"
+{- "." is the top-level package directory -}
+sourcedir = "."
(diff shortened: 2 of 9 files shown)


stephen-smith:
Is CONTRIBUTORS supposed to be maintained the alphabetical (or some other) order?

CodingCellist:
> Is CONTRIBUTORS supposed to be maintained in alphabetical (or some other) order?

It is meant to be in alphabetical order, yes (or at least, we try)

tests/idris2/pkg/multiline-comments/test.ipkg
```diff
@@ -0,0 +1,7 @@
+package test
+version = 0.1.0
+authors = "Boyd Stephen Smith Jr."
+main = Main
+executable = "test"
+{- "." is the top-level package directory -}
```
mattpolzin, line 6: May as well throw in a multi-line comment that is more than one line long (in addition to this comment).

stephen-smith: Done.

src/Parser/Lexer/Package.idr
```diff
@@ -99,6 +99,7 @@ andop = is '&' <+> is '&'
 rawTokens : TokenMap Token
 rawTokens =
   [ (comment, Comment . drop 2)
+  , (blockComment, Comment . shrink 2)
```
mattpolzin, line 102: I suppose this lacks a small amount of fidelity since the only useful meaning of the string within the comment would probably account for the fact that a block comment can start with a bracket followed by any number of dashes, but seeing as how 2 characters is the minimum at the start and end that doesn't feel too critical.

mattpolzin (approved):

src/Parser/Lexer/Package.idr
stephen-smith: Since we neither require a matching dash count on the end-of-comment token, nor strip excess leading dashes from eol comments, I don't feel this needs to change.

Parser/Lexer/Source.idr doesn't drop OR shrink for comments (of either type), but it also doesn't attempt to preserve their content in the tokens.

Issue #3276, opened 2024-05-11
A proof of Void


petithug:
The content of the namespace is inspired by Data.Int.Order in contrib:

```
%default total

namespace Bits8
  export
  data LessThan : Bits8 -> Bits8 -> Type where
    MkLessThan : (a < b) === True -> LessThan a b

  export
  %hint
  mkLessThan : (a < b) === True -> LessThan a b
  mkLessThan = MkLessThan

data Digit : Bits8 -> Type where
  MkDigit : (x : Bits8) -> {auto prf1 : LessThan 47 x} -> {auto prf2 : LessThan x 58} -> Digit x

prf1 : Digit 51
prf1 = MkDigit 51

dis0 : Digit 51 -> Void
dis0 (MkDigit _ {prf1=Refl}) impossible
```

# Steps to Reproduce

Load in Idris2 or Yaffle

# Expected Behavior

The line containing `impossible` should be rejected.

# Observed Behavior

Proving Void.

gallais:
Pretty sure that's #2250: you're using `Refl` instead of `MkLessThan` for the pattern.

Issue #2250, opened 2022-01-05
Definition with no RHS is incorrectly accepted as total


gallais:
Found while looking at the code interactively generated by case splitting in #2242.

```idris
%default total

g : Not Bool
g () impossible

f : Void
f = g True
```

Edit: Happy New Year!

buzden:
Looks like another example for the same problem, but with a bit deeper type problem:

```idris
data Even : Nat -> Type where
  EZ  : Even Z
  ESS : Even n -> Even (2 + n)

notE : (n : Nat) -> Not (Even n)
notE (S $ S n) (ESS n) impossible
notE 0 EZ = notE 2 (ESS EZ)
```

gallais:
This is rejected with `%default total` though:

> notE is not total, possibly not terminating due to recursive path Main.notE

buzden:
> This is rejected with `%default total` though:
> 
> > notE is not total, possibly not terminating due to recursive path Main.notE

Right, this is because I incorrectly shrunk the original example:

```idris
%default total

data Even : Nat -> Type where
  EZ  : Even Z
  ESS : Even n -> Even (2 + n)

data Odd : Nat -> Type where
  OSZ : Odd 1
  OSS : Odd n -> Odd (2 + n)

notE : (n : Nat) -> Not (Even n, Odd n)
notE 0         (_, y) = case y of _ impossible
notE 1         (x, _) = case x of _ impossible
notE (S $ S n) (ESS n, OSS n) impossible
```

(BTW, I don't know why `0` and `1` cases do not work with simple `impossible`, why I need to `case` of a part manually.

buzden:
> (BTW, I don't know why `0` and `1` cases do not work with simple `impossible`, why I need to `case` of a part manually.

One can do the following to avoid

```idris
notE : (n : Nat) -> Not (Even n, Odd n)
notE Z         (_, _) impossible
notE (S Z)     (_, _) impossible
notE (S $ S n) (ESS n, OSS n) impossible
```

Looks, like a separate bug...

ohad:
Another one (or related):
```idris
View : Type -> Type
View a = a -> Type

-- Pre-image view
data (.inv) : (f : a -> b) -> View b where
  Choose : (x : a) -> f.inv (f x)

LessThanOrEqualTo : Nat -> Nat -> Type
LessThanOrEqualTo n k = (+n).inv k

Ex1 : 3 `LessThanOrEqualTo` 5
Ex1 = Choose 2

[...]

PR #3396, opened 2024-10-05
[ fix ] Address some proofs of void via impossible from issues #2250 and #3276

dunhamsteve:
# Description

This PR addresses a few issues in #2250 and issue #3276.  They are caused by an issue with `IAlternative` in impossible clauses, a type confusion issue in impossible clauses, and treatment of out of hidden constructors as impossible. An additional issue was brought up in #2250 that is not addressed here.  I'll first discuss the issues and how I resolved them and then discuss a frex issue that should be patched before merging.

## Issues fixed

The `IAlternative` issue is illustrated at the top of #2550:
```idris
%default total

g : Not Bool
g () impossible

f : Void
f = g True
```
Here the `impossible` clause is accepted because the type doesn't match at all, but when it is converted by `mkTerm` to a term for coverage checking, the ambiguous `()` is converted to a pattern variable which is then considered to cover everything. I addressed this by choosing the default alternative.  I didn't see a way to determine the appropriate alternative at that point in the code.  A user will have to explicitly say `Unit` or `Pair` to get the other option in an impossible clause.

The type confusion issue is illustrated in #2250 by:
```idris
import Data.Nat
%default total

f : n `LTE` m -> Void
f Z impossible
f (LTESucc x) = f x

x : Void
x = f $ LTEZero {right=Z}
```
Here `Z` is impossible because it has the wrong type, but during coverage checking only the constructor tag is looked at and `Z` is confused with `LTEZero`.  I addressed this by also checking that the name matches.  I coerce the names to resolved names in `getCons` because some constructor names were resolved in the context and others were not.

Finally, in #3276 there was a false impossible because a constructor was not visible. I adjusted `recoverableError` to not consider `InvisibleName` to be evidence of impossible.

The issues fixed in #2250 and #3276 are included as test cases.

## Frex

Frex is relying on the ambiguity bug with Pair/MkPair in one location.  The function is indeed covering, but Idris cannot capable of determining that if `MkPair` is used instead of `(,)`. The current code is only working because of the above bug.  This can be addressed with the following patch to frex, which pushes the impossible match down to a separate case tree:

```diff
diff --git a/src/Frexlet/Monoid/Frex/Structure.idr b/src/Frexlet/Monoid/Frex/Structure.idr
index 8ccc22e..0164d67 100644
--- a/src/Frexlet/Monoid/Frex/Structure.idr
+++ b/src/Frexlet/Monoid/Frex/Structure.idr
@@ -159,8 +159,10 @@ MultHomomorphism a s (i, ConsUlt i1 x is) (j,ConsUlt j1 y js)
   = ( a.cong 2 (Dyn 0 .*. Dyn 1) [_,_] [_,_] [i_eq_j, i1_eq_i2]
     , x_eq_y
     ) :: is_eq_js
-MultHomomorphism _ _ (_, Ultimate _) (_, ConsUlt _ _ _) (MkAnd _ _) impossible
-MultHomomorphism _ _ (_, ConsUlt _ _ _) (_, Ultimate _) (MkAnd _ _) impossible
+MultHomomorphism _ _ (_, Ultimate _) (_, ConsUlt _ _ _) x with (x)
+  _ | MkAnd _ _ impossible
+MultHomomorphism _ _ (_, ConsUlt _ _ _) (_, Ultimate _) x with (x)
+  _ | MkAnd _ _ impossible
 
 public export
 AppendHomomorphismProperty : (a : Monoid) -> (x : Setoid) ->
```

### DIFF
--- a/src/Core/Coverage.idr
+++ b/src/Core/Coverage.idr
@@ -148,7 +148,7 @@ isEmpty defs env _ = pure False
 altMatch : CaseAlt vars -> CaseAlt vars -> Bool
 altMatch _ (DefaultCase _) = True
 altMatch (DelayCase _ _ t) (DelayCase _ _ t') = True
-altMatch (ConCase n t _ _) (ConCase n' t' _ _) = t == t' && n == n'
+altMatch (ConCase n t _ _) (ConCase n' t' _ _) = t == t'
 altMatch (ConstCase c _) (ConstCase c' _) = c == c'
 altMatch _ _ = False
 
--- a/src/Core/Coverage.idr
+++ b/src/Core/Coverage.idr
@@ -148,7 +148,7 @@ isEmpty defs env _ = pure False
 altMatch : CaseAlt vars -> CaseAlt vars -> Bool
 altMatch _ (DefaultCase _) = True
 altMatch (DelayCase _ _ t) (DelayCase _ _ t') = True
-altMatch (ConCase _ t _ _) (ConCase _ t' _ _) = t == t'
+altMatch (ConCase n t _ _) (ConCase n' t' _ _) = t == t' && n == n'
 altMatch (ConstCase c _) (ConstCase c' _) = c == c'
 altMatch _ _ = False
 
--- a/src/Core/Case/Util.idr
+++ b/src/Core/Case/Util.idr
@@ -26,11 +26,11 @@ getCons defs (NTCon _ tn _ _ _)
   where
     addTy : Name -> Core (Maybe DataCon)
     addTy cn
-        = do Just (idx, gdef) <- lookupCtxtExactI cn (gamma defs)
+        = do Just gdef <- lookupCtxtExact cn (gamma defs)
                   | _ => pure Nothing
              case (gdef.definition, gdef.type) of
                   (DCon t arity _, ty) =>
-                        pure . Just $ MkDataCon (Resolved idx) t arity
+                        pure . Just $ MkDataCon cn t arity
                   _ => pure Nothing
 getCons defs _ = pure []
 
--- a/src/Core/Case/Util.idr
+++ b/src/Core/Case/Util.idr
@@ -26,11 +26,11 @@ getCons defs (NTCon _ tn _ _ _)
   where
     addTy : Name -> Core (Maybe DataCon)
     addTy cn
-        = do Just gdef <- lookupCtxtExact cn (gamma defs)
+        = do Just (idx, gdef) <- lookupCtxtExactI cn (gamma defs)
                   | _ => pure Nothing
              case (gdef.definition, gdef.type) of
                   (DCon t arity _, ty) =>
-                        pure . Just $ MkDataCon cn t arity
+                        pure . Just $ MkDataCon (Resolved idx) t arity
                   _ => pure Nothing
 getCons defs _ = pure []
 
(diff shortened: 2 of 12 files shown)


src/Core/Coverage.idr
```diff
@@ -148,7 +148,7 @@ isEmpty defs env _ = pure False
 altMatch : CaseAlt vars -> CaseAlt vars -> Bool
 altMatch _ (DefaultCase _) = True
 altMatch (DelayCase _ _ t) (DelayCase _ _ t') = True
-altMatch (ConCase _ t _ _) (ConCase _ t' _ _) = t == t'
+altMatch (ConCase n t _ _) (ConCase n' t' _ _) = t == t' && n == n'
```
mjustus, line 151: I'd be happy for it to be merged but I am still in two minds about this change in the coverage checker.

It is clearly correct and fixes an existing bug but I was under the impression that coverage checking implicitly assumes that all patterns passed to it are already well-typed. So I'd prefer if ill-typed patterns like

```
Bug : Not (3 `LessThanOrEqualTo` 5)
Bug Refl impossible
```

are rejected before the coverage checker is called. On the other hand, that would require a much more involved change and it may be a while before somebody finds the time to work on it.

dunhamsteve: Sorry, I chased this stuff back and saw the `mkTerm` issue, but I didn't see what else was happening in `Impossible.idr` aside from `mkTerm` itself.  It seems that most of that checking is already in place there, but line 48 is short circuiting it:

https://github.com/idris-lang/Idris2/blob/f840d1b771be59000b478780dcfc1dfa2bfde624/src/TTImp/Impossible.idr#L45-L52

I'm not sure why the type check is being short circuited, but if I remove that line and remove the change you quoted above, all of the tests pass. 

Further, for the `IAlternative` case, I didn't realize that I may have the type at hand (as a `Maybe (Closure [])`). I think I can take the same approach instead of picking the default alternative. (The type is indeed there in my test cases.)

A third thing - I tracked down the weird covering error that I was hitting when trying to fix frex (it was covering if I put ` = ?rhs` but not covering if I put `impossible`).  It appears that line 1131 of `ProcessDef.idr` silently drops an impossible clause when an error is raised in `getImpossibleTerm`.  In the case of frex, it raises an `AmbiguousName` because of `Ultimate`:

https://github.com/idris-lang/Idris2/blob/f840d1b771be59000b478780dcfc1dfa2bfde624/src/TTImp/ProcessDef.idr#L1126-L1133

I think we would want to give an error or warning if an impossible clause is discarded due to an error.

dunhamsteve: I've reverted the changes to constructor checking in `Coverage.idr` and instead changed `dropNoMatch` to also check cases where there is only one definition matching the name.  It uses `TTImp.Impossible.match` to see if constructors are valid. Previously this was only used in the case of an ambiguous name to guess the correct name.

Note that because we're dealing with raw terms, it is not precise type checking. It looks like it chases down the type constructor in the return type and checks if it has the same name as the expected type.

As I noted above, the impossible clause is silently dropped if it doesn't match. That is the existing behavior for cases where the name many or zero options with an acceptable type. For example, in the current Idris the following line in frex is silently dropped because there are two `Refl` in scope:
```idris
  sumDegenerate a (x :: xs) (FS i) prf | Left  Refl impossible
```
It would be nice if this were a warning or error, but perhaps that should be a separate PR.

andrevidela:
@mjustus I think this is waiting on you to approve, right?

src/Core/Coverage.idr
dunhamsteve: I think I've resolved mjustus' concerns, but I don't know who is supposed to mark resolve conversation.

For IAlternative, I have code that checks each option, but it risks being exponential (unlikely), so I've left that at choosing the default.

andrevidela:
I want this badly enough that I'll merge it on a Friday evening

PR #3403, opened 2024-10-26
[ literate ] Support `typst` in literate Idris

buzden:
# Description

[Typst](https://github.com/typst/typst) is relatively young typesetting system which aims to be as powerful as TeX while being lightweight more like Markdown and extensible by programmers with custom libraries (and tens of them are already around).

I personally use Pandoc Markdown + raw TeX insertions with Literate Idris for my papers and presentations, and I find this facinating, especially an ability to compile and run my slides and thus to check all the code on them during build of a PDF. I'd like to use typst too for that, but now compiler does not support it, which is not much a change.

So, I suggest to add the support for typst files. Also I made a small refactoring in a separate commit, and I tried to correct/extend existing literate Idris documentation.

      `CONTRIBUTORS.md`).


docs/source/reference/literate.rst
```diff
@@ -156,3 +179,44 @@ With one such example using ``fancyverbatim`` and ``comment`` packages as::
     \usepackage{comment}
 
     \excludecomment{hidden}
+
+Typest
```
dunhamsteve, line 183: Should this be Typst?

buzden: It should, it's my typost ;-) Fixed.

gallais (approved):

andrevidela:
@gallais are we waiting on something to merge?

gallais:
Would be nice to have a test

buzden:
I've added a test to the last commit

gallais (approved):

PR #3405, opened 2024-10-29
[ cleanup ] Make `Nat`'s `NonZero` to be an alias for `IsSucc`

buzden:
# Description

Historically for some strange reason in `Data.Nat` we have two equivalent definitions: `IsSucc` and `NonZero`. Other standard types have standard pattern of `Is<Constructor>` type, `ItIs<Constructor>` constructor of proof and `isIt<Constructor>` for `Dec Is<Consturctor>`. `IsSucc` follows this pattern, when `NonZero` does not. However, `NonZero` is a handy name for this.

But the fact that there are two different types for this and the fact that some standard and non-standard functions use one of them, and others use the other, triggers a lot of people (I personally know around five of them).

What I suggest is to make `NonZero` to be an alias for `IsSucc` and to declare a deprecated alias for original constructor of `NonZero` to be backward compatible at least on usages at RHS.

      `CONTRIBUTORS.md`).

### DIFF
--- a/libs/base/Data/Nat.idr
+++ b/libs/base/Data/Nat.idr
@@ -32,6 +32,9 @@ isSucc : Nat -> Bool
 isSucc Z     = False
 isSucc (S n) = True
 
+||| A definition of non-zero with a better behaviour than `Not (x = Z)`
+||| This is amenable to proof search and `NonZero Z` is more readily
+||| detected as impossible by Idris
 public export
 data IsSucc : (n : Nat) -> Type where
   ItIsSucc : IsSucc (S n)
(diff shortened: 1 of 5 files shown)


andrevidela:
If we can fix frex, this looks good to me

buzden:
> If we can fix frex, this looks good to me

Right, this seems to be not much a strech for `frex`, I made a PR there

libs/base/Data/Nat.idr
```diff
@@ -45,6 +48,17 @@ isItSucc : (n : Nat) -> Dec (IsSucc n)
...
 isItSucc (S n) = Yes ItIsSucc
 
+||| A hystorical synonym for `IsSucc`
+public export
+0 NonZero : Nat -> Type
+NonZero = IsSucc
+
+-- Remove as soon as 0.8.0 (or greater) is released
+public export %inline
+%deprecate -- use ItIsSucc instead
```
mattpolzin, line 58: If you put the suggestion to use ItIsSucc instead into a code doc then it should show up in the warning for deprecation which is a nice touch.

mattpolzin (commented):
Looks good to me too. Seems like we can get this merged whenever.

libs/base/Data/Nat.idr
buzden: Sounds reasonable, done

buzden:
> Seems like we can get this merged whenever.

Now `frex` is ready too, so CI is green

Issue #3402, opened 2024-10-25
let binders in record 


andrevidela:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

Allow `let` bindings in the body of record definitions.

## Motivation

Records abstract over a lot of boilerplate over using `data`. However, they are limited in ways that `data` isn't. One thing that `data` declarations allow but `record` do not is the ability to write `let`-binders in the type of a constructor. For example:

```
record Monoid where
  constructor MkMon
  carrier : Type
  neutral : carrier
  op : carrier -> carrier -> carrier
  
data MonoidLaw : Type where
  MkMonLaw : (mon : Monoid) -> let
    ty : Type
    ty = mon.carrier
    (+) : ty -> ty -> ty
    (+) = mon.op
    i : ty
    i = mon.neutral
    in (idR : forall a . a + i = a) ->
       (idL : forall a . i + a = a) ->
       (assoc : forall a, b, c : a + (b + c) = (a + b) + c) ->
       MonoidLaw
```

This is currently impossible with a record. And in itself it's not too problematic, but `data` does not synthesize projections functions, unlike records. This proposal fill the hole in the design space by providing the ability to write `let` in record defintions. 

## The proposal

Allow `let` blocs in the body of records that allow definitions to be reused multiple times in subsequent fields.

For typechecking reasons, the `let` blocks in records only allow clauses with type information. This is because each `let` binder must be lifted to the top-level where type inference is brittle and because the types of each field must be easily computable, we require each field to be given a type 

### Examples

```
record Monoid where
  constructor MkMon
  carrier : Type
  neutral : carrier
  op : carrier -> carrier -> Type

record MonLaws (mon : Monoid) where
  constructor MkMonLaws
  let (+) : mon.carrier -> mon.carrier -> Type
      (+) = mon.op
      i : mon.carrier
      i = mon.neutral
  neutralIdL : forall a . a + i = a
  neutralIdR : forall a . i + a = a
  opAssoc : forall a, b, c . a + (b + c) = (a + b) + c
```

The block of definitions can be written after the `let`:

```
record MonLaws (mon : Monoid) where
  constructor MkMonLaws
  let 
    (+) : mon.carrier -> mon.carrier -> Type
    (+) = mon.op
    i : mon.carrier
    i = mon.neutral
  …
```

[...]

PR #3406, opened 2024-10-30
WithFC, a datastructure to keep track of locations

andrevidela:
# Description

This was meant to be a PR about #3402 but the management of _File Context_ got out of hand and so I'm making a preliminary PR to get it sorted. 

This should not change, modify, or add any functionality, but should make future changes to file-location easier to do and read. It should also make it easier to report appropriate locations for error messages.

This takes steps toward #3408 but does not fix it yet.

No, this is purely refactoring


andrevidela:
I'm pretty sure the FC code is "right" but the highlight code relies on the previously incorrect FC code so I need to figure out what's going on.

andrevidela:
LSP fix available here https://github.com/idris-community/idris2-lsp/pull/230

PR #3412, opened 2024-11-14
refactor Uninhabited implementation for Elem types

ihor-rud:
# Description

Refactor `Uninhabited` implementation for `Data.List.Elem`, `Data.List1.Elem`, `Data.SnocList.Elem` and `Data.Vect.Elem` so it can be used for homogeneous and heterogeneous equality.


libs/base/Data/List/Elem.idr
```diff
@@ -19,11 +19,11 @@ data Elem : a -> List a -> Type where
      There : Elem x xs -> Elem x (y :: xs)
 
 export
-Uninhabited (Here = There e) where
+Uninhabited (Here {x} {xs} = There {x = x'} {y} {xs} e) where
```
gallais, line 22: If you're going fully heterogeneous, why not have two different `xs` on each sides?

ihor-rud: good point. `xs` are now different

ihor-rud: Since `xs` are now different maybe we should add something like that? 
`Either (Not (x = x')) (Not xs = xs') => Uninhabited (Here {x} {xs} = Here {x = x'} {xs = xs'}) where`
Basically, if `x` or `xs` are different, `Here` and `There` will be different as well.

buzden: > Either (Not (x = x')) (Not xs = xs') => Uninhabited (Here {x} {xs} = Here {x = x'} {xs = xs'})

I think this can be proven for any `Biinjective`, not only for this particular one

CHANGELOG_NEXT.md
```diff
@@ -227,6 +227,8 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 
 * Added `Data.IORef.atomically` for the chez backend.
 
+* Refactor `Uninhabited` implementation for `Data.List.Elem`, `Data.List1.Elem`, `Data.SnocList.Elem` and `Data.Vect.Elem`.
```
CodingCellist, line 230: Would you mind adding a brief sentence explaining why / explaining what the refactor achieves?

CodingCellist (changes requested):
This might be more of a stylistic choice, I'm not sure, but I would personally prefer if at least one of the test code files used the explicit syntax for which equality was being used (i.e. `===` or `~=~`), just so we have something clearly showing the different ones - since that was what caused the original confusion on the discord - as well as testing the magic elaboration of `=` to the hopefully correct equality  : )

CHANGELOG_NEXT.md
ihor-rud: done

ihor-rud:
> This might be more of a stylistic choice, I'm not sure, but I would personally prefer if at least one of the test code files used the explicit syntax for which equality was being used (i.e. `===` or `~=~`), just so we have something clearly showing the different ones - since that was what caused the original confusion on the discord - as well as testing the magic elaboration of `=` to the hopefully correct equality : )

I added tests for all combinations

CodingCellist (approved):
Thank you for making the minor changes, I'm happy with it as-is now. If nobody has any other comments before Friday at noon GMT (UTC+0), I'll merge things around then  : )

mattpolzin:
I'll go ahead and merge this since that appears to have been the goal any time after this past week.

PR #3414, opened 2024-11-17
Add between parser combinator to Data.String.Parser

justjoheinz:
# Description

between - an additional parser combinator in Data.String.Parser (contrib).

### DIFF
--- a/libs/contrib/Data/String/Parser.idr
+++ b/libs/contrib/Data/String/Parser.idr
@@ -285,10 +285,17 @@ covering
 spaces1 : Monad m => ParseT m ()
 spaces1 = skip (some space) <?> "whitespaces"
 
+
+||| Run the parser p between the opening parser `o` and the closing parser `c`,
+||| returning the result of p.
+export
+between : Monad m => ParseT m o -> ParseT m c -> ParseT m p -> ParseT m p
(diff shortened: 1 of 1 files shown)


libs/contrib/Data/String/Parser.idr
```diff
@@ -406,3 +406,14 @@ export
...
+
+||| Run the parser p between the opening parser `o` and the closing parser `c`,
+||| returning the result of p.
+export
+between : Monad m => ParseT m o -> ParseT m c -> ParseT m p -> ParseT m p
+between o c p = do
+  _ <- o
+  result <- p
+  _ <- c
+  pure result
```
gallais, line 419:
```suggestion
between o c p = id <$ o <*> p <* c
```

justjoheinz: I have a bit of a hard time understanding the suggestion - is it equivalent to 
```
between o c p = o *> p *< c
```
?

gallais: The core idea is to never mix left & right pointed combinators. Only have `f arg1 arg2 arg3` with `<*` before unused arguments and `<*>` before used ones.

andrevidela:
with AOC upon us this is more useful than ever

PR #3415, opened 2024-11-18
[ new ] Quantity for proof in with-clauses

spcfox:
# Description

Currently the quantity of proof in with-clause is the same as the quantity of expression. But often proofs are needed only at compile time. Example:
```idris
filterSquared : (p : a -> Bool) -> (xs : List a) ->
                filter p (filter p xs) === filter p xs
filterSquared p [] = Refl
filterSquared p (x :: xs) with (p x) proof 0 eq
  filterSquared p (x :: xs) | False = filterSquared p xs
  filterSquared p (x :: xs) | True
    = rewrite eq in cong (x ::) (filterSquared p xs)
```
So I suggest adding the ability to separately specify quantities for proofs.

Note that `prf` is currently erased in the following code, but in this version it will now be unrestricted, as will all variables for which multiplicity is not explicitly specified. This may affect existing code and probably requires discussion.
```idris
foo x with 0 (x) proof eq
```

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Idris/REPL.idr
+++ b/src/Idris/REPL.idr
@@ -225,8 +225,8 @@ printClause l i (WithClause _ lhsraw rig wvraw prf flags csraw)
          cs <- traverse (printClause l (i + 2)) csraw
          pure (relit l ((pack (replicate i ' ')
                 ++ show lhs
-                ++ " with " ++ elimSemi "0 " "1 " (const "") rig ++ "(" ++ show wval ++ ")"
-                ++ maybe "" (\ nm => " proof " ++ show nm) prf
+                ++ " with " ++ showCount rig ++ "(" ++ show wval ++ ")"
+                ++ maybe "" (\ nm => " proof " ++ showCount (fst nm) ++ show (snd nm)) prf
                 ++ "\n"))
                ++ showSep "\n" cs)
 printClause l i (ImpossibleClause _ lhsraw)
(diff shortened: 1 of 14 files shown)


src/Idris/REPL.idr
```diff
@@ -225,8 +225,8 @@ printClause l i (WithClause _ lhsraw rig wvraw prf flags csraw)
          cs <- traverse (printClause l (i + 2)) csraw
          pure (relit l ((pack (replicate i ' ')
                 ++ show lhs
-                ++ " with " ++ elimSemi "0 " "1 " (const "") rig ++ "(" ++ show wval ++ ")"
-                ++ maybe "" (\ nm => " proof " ++ show nm) prf
+                ++ " with " ++ showCount rig ++ "(" ++ show wval ++ ")"
+                ++ maybe "" (\ nm => " proof " ++ showCount (fst nm) ++ show (snd nm)) prf
```
spcfox, line 229: Cannot match on pair. See https://github.com/idris-lang/Idris2/issues/3418

spcfox:
This PR is ready to review.
I made a PR https://github.com/stefan-hoeck/idris2-elab-util/pull/79 in elab-util to fix a build issue. pack and lsp build without problems, but I also made a PR https://github.com/idris-community/idris2-lsp/pull/234 in lsp that fixes pretty printing.

gallais (approved):

gallais:
Happy to merge once the build is successful.

buzden:
We can merge both downstream PRs, but this `pack` job is harder, since we need newer `elab-util` AND this PR to be merged in order to have new successfully building pack collection for that. I propose this PR #3446 in order to not to require this. After that PR we can merge downstreams and then all three failing CI jobs from this PR should pass.

PR #3425, opened 2024-11-25
[ base ] Better set's `toList` implementation + preparation for cleanup

buzden:
# Description

This is mainly the less controversial part of #3392, reimplementing and deprecating completely useless specialised `toList` function from `SortedSet`.

Additionally, this PR contains some preparation to future removing the redundant `Libraries.Data.Sorted{Set|Map}` from the compiler's internal lib, the fact coming from times when `Sorted{Set|Map}` were in `contrib`. But currently, they cannot be replaced, because implementations in the compiler's lib have several functions that are missing in the implementation at `base`, so I suggest adding them there, while also doing the same thing with deprecating specialised `toList`, so that we just could remove those redundant modules once we've released new version of the compiler.

I made all changed in different commits in order to make review easier.

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -244,6 +244,8 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
   `SIsNonZero` was made to be an alias for `ItIsSucc`, was marked as deprecated,
   and won't work on LHS anymore.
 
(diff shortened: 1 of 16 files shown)


mattpolzin (approved):

mattpolzin:
@buzden trivial merge conflict in changelog

PR #3426, opened 2024-11-27
Export Idris.Package.build and Idris.Package.check; enhance path normalisation in Core.Directory.cleanPath

Russoul:


src/Core/Directory.idr
```diff
@@ -224,9 +238,9 @@ corePathToNS : String -> Maybe String -> String -> Core ModuleIdent
 corePathToNS wdir sdir fname = do
   let err = UserError $
           "Source file "
-       ++ show fname
+       ++ show (cleanPath fname)
        ++ " is not in the source directory "
```
Z-snails, line 242: I don't think this error is being tested for yet - could you add one please?

Z-snails: Ignore this, I didn't read the code thoroughly enough

PR #3429, opened 2024-11-30
Error Message Potpourri

mattpolzin:
# Description

I tried to keep the following isolated in their own commits so we can just ditch anything folks don't like.

----
Whereas the CLI for the REPL currently only pays attention to one argument as an input file but quietly ignores the rest, I've made it an error to specify multiple input files:
```
Uncaught error: Error: Expected at most one input file but was given: file1.idr, file2.idr
```
This is further improved (described in the last section below) by using `die` instead of an uncaught `throw` so the error doesn't look so unexpected.

----
Whereas the REPL previously printed an error about a file loaded with `:l` in monochrome without fanfare, I've indented the error message, given it room to stand out, and written the context of the error in the standard `IdrisAnn` color for error messages.
```
Main> :l "file1.idr"

  Error loading file "file1.idr": File Not Found

Main> [...]
```

----
Whereas on launch the REPL previously printed a subtly different error when a requested file could not be loaded than it did when using the `:l` command inside the REPL, I've aligned the two errors (and indeed refactored them to use the same code). I've also added a suggestion specifically for this startup use-case because it is common to see people intend to run e.g. `idris2 --repl ...` but instead type `idris2 repl ...` which causes the REPL to attempt to load a file named "repl" -- the suggestion is that the user may have meant to write `idris2 --repl ...` so this should ease the pain of a common mistake.
```
 $ ./build/exec/idris2 repl
     ____    __     _         ___
    / // __  / ___/ / ___/   __/ /     Version 0.7.0-c1fc0b346

  Error loading file "repl": File Not Found
  Did you mean to type "--repl"?

Main> [...]
```

----
As mentioned above, this PR finally switches from uncaught `throw` for rendering errors from the CLI over to using `die` so we get a less buggy looking printout. This commit also adds in a suggestion when relevant similar to the one seen within the REPL session for the above case.
```
 $ ./build/exec/idris2 build idris2.ipkg
Error: Expected at most one input file but was given: build, idris2.ipkg
Did you mean to type "--build"?
```

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -27,6 +27,8 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 
 * Remove reference to column number parameter in help menu for `refine` command.
 
+* CLI errors will now be printed to `stderr` instead of `stdout`.
+
 ### Building/Packaging changes
 
 * The Nix flake's `buildIdris` function now returns a set with `executable` and
--- a/src/Idris/CommandLine.idr
+++ b/src/Idris/CommandLine.idr
@@ -2,6 +2,9 @@ module Idris.CommandLine
 
 import Idris.Env
 import Idris.Version
+import Idris.Doc.Display
+import Idris.Doc.String
+import Idris.Pretty
 
 import Core.Options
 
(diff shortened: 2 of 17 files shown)


mattpolzin:
I'll merge this soon on the grounds that it is unlikely to offend anyone and unlikely to trip anyone up too badly. It can always be reverted.

PR #3430, opened 2024-11-30
[ fix ] Fix option loss when an optional argument is missed

spcfox:
# Description

If an optional argument is expected and the following option comes instead, it is skipped. For example, in the command
```shell
idris2 --build --quiet
```
`--quiet` will be missed. And the command
```shell
idris2 --clean --log 0
```
will be processed as
```shell
idris2 0 --clean
```

      `CONTRIBUTORS.md`).


mattpolzin (approved):
The change looks great. I've got CI kicked off now.

PR #3433, opened 2024-12-03
TDDwI: Minor changes to chapter10

sagehane:
I hope this is the right place to suggest changes to https://idris2.readthedocs.io/en/latest/typedd/typedd.html.

While following the TypeDD book with idris2, I noticed that the code can be made total whilst resembling [the original code](https://github.com/edwinb/TypeDD-Samples/blob/5e648f36a461038e09f0fa269fb0a4acf3d64e14/Chapter10/IsSuffix.idr) more.

I checked the code with `idris2 --check tests/typedd-book/chapter10/IsSuffix.idr` but didn't run any tests because I didn't really understand how it works.

---

Unrelated, but I feel like it would be nicer if the `Snoc` constructor was changed from:
```idris2
Snoc : (x : a) -> (xs : List a) -> SnocList xs -> SnocList (xs ++ [x])
```
to something better resembling idris1:
```idris2
Snoc : {x : _} -> {xs : _} -> SnocList xs -> SnocList (xs ++ [x])
```
I assume there is a valid motivation for this change that I didn't catch, as I've never used Idris outside of the purposes of this book.

### DIFF
--- a/docs/source/typedd/typedd.rst
+++ b/docs/source/typedd/typedd.rst
@@ -299,6 +299,18 @@ It's no longer necessary to give ``{input}`` explicitly in the patterns for
       isSuffix (xs ++ [x]) (ys ++ [y]) | (Snoc x xs xsrec) | (Snoc y ys ysrec)
         = if x == y then isSuffix xs ys | xsrec else False
 
+For the ``SplitRec`` view in the exercise 2 after Chapter 10-2 ``import Data.Vect.Views.Extra``
+from ``contrib`` library. The ``n`` argument needs to be made explicit, and so
+does the ``xs`` and ``ys`` for the ``SplitRecPair`` constructor.
+
+.. code-block:: idris
+
+   mergeSort : {n : _} -> Ord a => Vect n a -> Vect n a
+   mergeSort xs with (splitRec xs)
+     mergeSort [] | SplitRecNil = ?mergeSort_rhs_rhss_0
+     mergeSort [x] | SplitRecOne = ?mergeSort_rhs_rhss_1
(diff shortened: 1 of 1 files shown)


docs/source/typedd/typedd.rst
```diff
@@ -281,28 +281,23 @@ in the type:
...
 
 It's no longer necessary to give ``{input}`` explicitly in the patterns for
 ``snocListHelp``, although it's harmless to do so.
 
-In ``IsSuffix.idr``, the matching has to be written slightly differently. The
-recursive with application in Idris 1 probably shouldn't have allowed this!
-Note that the ``Snoc`` - ``Snoc`` case has to be written first otherwise Idris
-generates a case tree splitting on ``input1`` and ``input2`` instead of the
-``SnocList`` objects and this leads to a lot of cases being detected as missing.
+``IsSuffix.idr`` also has to be adjusted to match the new ``Snoc`` arguments.
```
sagehane, line 289: The comment probably can be improved. I'm especially confused as to why the new code doesn't need a `| ysrec` at line 300.

sagehane:
Also, assuming this is the right place to have this file edited, the link at https://github.com/idris-lang/Idris2/blob/ecd6d47ea0ea5e2662cbb84fad0421cd250171a6/docs/source/typedd/typedd.rst?plain=1#L10-L11 should probably be updated to this repo, instead of edwinb's repo.

SlayerOfTheBad:
Whilst this solution works, it does not actually do what this example in Chapter 10 is meant to illustrate. This version does not recursively apply `ysrec`, which would lead to reduced performance, as the `SnocList` for `ys` needs to be constructed on each step.

CodingCellist:
This seems abandoned?  There also seems to have been a question as to whether the change preserved the pedagogy.  Closing in 1 week (Thursday 13 Nov UTC+0) unless there is new activity.

PR #3434, opened 2024-12-03
[ refactor ] Improve error handling in `Core`

### LINKED ISSUE #3427: Successful exit code after error in `--exec`

spcfox:
# Description

- Added utility functions to handle errors in `Core`.
- Added new error types: `NonZeroExitCode`, `SystemError`, `CodegenNotFound`, `UnsupportedOpertaion`.
- Added error handling in many places where they were ignored.
- All `coreLift_` calls replaced by explicit result processing.
- Removed `Maybe` from `compileExpr` and `incCompileFile` signatures. Instead of returning `Nothing`, `throw` is used. Earlier both variants were found in the code. Using `throw` allows you to pass a more clear error.
- fix #3427: the `--exec` command returns the program exit code. If several `--exec` options are passed, the result of the last execution will be returned.
- Removed `CompilationFailed` from `REPLResult`. More specific errors using `throw` are returned instead.

      `CONTRIBUTORS.md`).


src/Compiler/ES/Node.idr
```diff
@@ -54,8 +54,8 @@ compileExpr :
 compileExpr c s tmpDir outputDir tm outfile =
   do es <- compileToNode c s tm
      let out = outputDir </> outfile
-     Core.writeFile out es
-     coreLift_ $ chmodRaw out 0o755
+     writeFile out es
+     handleFileError out $ chmodRaw out 0o755
      pure (Just out)
```
gallais: This can never return `Nothing`, can it? So we may as well change the type (and it saves
us the boggy error handling in `executeExpr`).

src/Compiler/RefC/RefC.idr
```diff
@@ -979,45 +979,38 @@ generateCSourceFile defs outn =
...
      generateCSourceFile defs outn
-     Just _ <- compileCObjectFile outn outobj
-       | Nothing => pure Nothing
-     compileCFile outobj outexec
-
-compileExpr _ _ _ _ _ _ _ = pure Nothing
-
-
+     obj <- compileCObjectFile outn outobj
+     Just <$> compileCFile obj outexec
```
gallais: Same remark: why keep `Maybe` if we're only ever returning `Just`?

src/Compiler/Scheme/Gambit.idr
```diff
@@ -401,10 +400,8 @@ compileExpr c s tmpDir outputDir tm outfile
                  Nothing => gscBackend ++ ["-exe", "-cc-options", "-Wno-implicit-function-declaration", "-ld-options"] ++ libsfile
                  Just _ => ["-c"]
          let cmd = gsc ++ gscCompileOpts ++ ["-o", execPath, srcPath]
-         ok <- coreLift $ system cmd
-         if ok == 0
-            then pure (Just execPath)
-            else pure Nothing
+         safeSystem cmd
+         pure (Just execPath)
```
gallais: This is not logically equivalent. Why is it legitimate to drop the `Nothing` branch?

src/Compiler/Scheme/Racket.idr
```diff
@@ -455,23 +450,14 @@ compileExpr mkexec c s tmpDir outputDir tm outfile
...
-                    pure (Just outShRel)
-            else pure Nothing
+         when mkexec $ logTime 1 "Build racket" $
+             safeSystem $ raco ++ ["-o", outBinAbs, outRktAbs]
+
+         -- TODO: add launcher script
+         let outShRel = outputDir </> outfile
+         makeShPlatform isWindows (if mkexec then "" else racket ++ " ") outShRel appDirRel outRktFile
+         handleFileError outShRel $ chmodRaw outShRel 0o755
+         pure (Just outShRel)
```
gallais: Again we're also dropping the `Nothing` branch

src/Compiler/ES/Node.idr
gallais: Ugh, already fixed in a further commit

src/Compiler/RefC/RefC.idr
gallais: Ugh, already fixed in a further commit

src/Compiler/Scheme/Gambit.idr
spcfox: `safeSystem` throws an error if the exit code is not zero. This makes the errors more clear

src/Compiler/Scheme/Racket.idr
spcfox: I replaced all `Nothing` with `throw`. Now `Nothing` just wraps around to `CompilationFailed`. A more specific cause of the error is passed to `throw`.
https://github.com/idris-lang/Idris2/blob/ec74792a49bf4d22509172d1f03d153ffca1b95c/src/Idris/REPL.idr#L816

src/Compiler/ES/Node.idr
```diff
@@ -50,26 +50,23 @@ compileExpr :
...
+  (tmpDir : String) -> ClosedTerm -> Core ExitCode
 executeExpr c s tmpDir tm =
-  do let outn = tmpDir </> "_tmp_node.js"
-     js <- compileToNode c s tm
-     Core.writeFile outn js
+  do out <- compileExpr c s tmpDir tmpDir tm "_tmp_node.js"
      node <- coreLift findNode
-     quoted_node <- pure $ "\"" ++ node ++ "\"" -- Windows often have a space in the path.
-     coreLift_ $ system (quoted_node ++ " " ++ outn)
+     system $ "\"" ++ node ++ "\" " ++ out -- Windows often have a space in the path.
```
spcfox, line 69: Why does `escapeArg` not enclose in quotes in Windows?
https://github.com/idris-lang/Idris2/blob/ec74792a49bf4d22509172d1f03d153ffca1b95c/libs/base/System/Escape.idr#L10-L14

spcfox: Oh, it escapes spaces, okay
https://github.com/idris-lang/Idris2/blob/ec74792a49bf4d22509172d1f03d153ffca1b95c/libs/base/System/Escape.idr#L21

buzden: Maybe, it worth to add a comment nearby about this to make it clear for the future readers

spcfox: CI on windows fails if I replace quoting with `escapeArg`. It's looks like problem with `escapeArg`

spcfox: > CI on windows fails if I replace quoting with `escapeArg`. It's looks like problem with `escapeArg`

This looks like a Node.js problem https://github.com/nodejs/node/issues/56146

spcfox: I have tried to understand the reasons for this behavior and detailed it in #3436

src/Compiler/Scheme/Chez.idr
```diff
@@ -661,16 +647,15 @@ compileExpr makeitso c s tmpDir outputDir tm outfile
...
   Ref Syn SyntaxInfo ->
-  (tmpDir : String) -> ClosedTerm -> Core ()
-executeExpr c s tmpDir tm
-    = do Just sh <- compileExpr False c s tmpDir tmpDir tm "_tmpchez"
-            | Nothing => throw (InternalError "compileExpr returned Nothing")
-         coreLift_ $ system [sh]
+  (tmpDir : String) -> ClosedTerm -> Core ExitCode
+executeExpr c s tmpDir tm = do
+  out <- compileExpr False c s tmpDir tmpDir tm "_tmpchez"
+  system $ escapeArg out
```
spcfox, line 653: I think an explicit `escapeArg` is clearer than a single element list

spcfox:
This PR is ready for review. The lsp build is failing because I deleted `CompilationFailed`. I have not yet made a PR fixing this because I allow for the possibility of revising this decision during review.

spcfox:
If my refactoring is questionable, I can put everything related to #3427 in a separate PR to simplify the review

PR #3441, opened 2024-12-11
[ fix #3437 ] Add a check for multiple totality modifiers

### LINKED ISSUE #3437: Interaction of totality and coverage checking
Consider the following code snippets

```
%default total

covering
f : Int -> Int
f i = f i
```

```
total
covering
f : Int -> Int
f i = f i
```

# Steps to Reproduce

Using `idris2--0.7.0_3.arm64_sonoma` on MacOS Sonoma 14.7.1 installed via Homebrew.

# Expected Behavior

In both cases, I would expect to still get an error message about the function `f` not being total (being nonterminating):

```
Error: f is not total, possibly not terminating due to recursive path Main.f
```

# Observed Behavior

No error message about `f` not being total (being nonterminating), neither using CLI nor using the Idris Language VS Code plugin.

spcfox:
# Description

- Added check for multiple totality modifiers in `processFnOpt`.
- `elabImplementation` does not add a totality modifier from the interface if it is overridden in the implementation. This is to prevent multiple modifiers from appearing after desugaring.
- Added a check for totality modifiers when desugaring declarations, as they may not spawn functions (interfaces without implementations or without methods). I'm not sure if this is the right place.
- Added processing of totality modifier on forward-declared data (See https://github.com/idris-lang/Idris2/issues/3442#issuecomment-2536400351).

Checking when desugaring is enough for Idris, but I left the changes to `processFnOpt` as they also check TTImp.

      `CONTRIBUTORS.md`).


spcfox:
> "Multiple incompatible totality modifier were given for this function." And give multiple ways to solve it. Either by relying on the default totality or by overriding the default.

I think any use of multiple totality modifiers (including identical ones) on a single declaration should be a error. The default totality no matter.

andrevidela:
> The default totality no matter. (sic)

This is incorrect.

```idris
%default total

covering
main : IO ()
main = …
```

is a common pattern.

In the case of multiple definitions, like `total covering` the message should say to either remove `total` and keep `covering` or remove both and rely on `%default total`. Similarly if you get `covering covering`, just change the message explaining why it's bad.

spcfox:
I mean, the presence of `%default` does not affect the presence of this error, so it's strange to suggest the user to remove all totality modifiers from the declaration and rely on `%default`.

andrevidela:
Alright I'll do it myself

mattpolzin:
Regardless of the goals of this PR, I’d add that in my opinion it’s ok to have e.g. a `%default total` plus an explicit `total` on a function. Sure, the latter is redundant, but it’s harmlessly explicit and it could come into play if the default is modified so there’s a practical impact of it. 

That is, in my opinion the only thing that should be an error is if there are multiple totalities defined explicitly on the function.

spcfox:
I have tried to make correct error messages that take into account the presence of all totality modifiers. This commit does not affect [the ability to override totality in Data](https://github.com/idris-lang/Idris2/issues/3442#issuecomment-2536400351)

spcfox:
I'm sorry if my messages might have come across as rude. I don't always know the best way to express my thoughts in English

mattpolzin:
I think your messages are coming across just fine (no rude tone detected by me, at least).

tests/idris2/total/total027/expected
```diff
@@ -0,0 +1,79 @@
...
+ 5 | data X where
+ 6 |   MkX : (X -> X) -> X
+
+Possible solutions:
+ - Use the same totality modifiers
+ - Remove the totality modifier from the declaration
+ - Remove the totality modifier from the definition
+------
+1/1: Building recordDifferentTotality (recordDifferentTotality.idr)
+Error: Data Main.X has been forward-declared with totality `partial`, cannot change to `total`
```
spcfox, line 33: In this error, unlike the others, the name is printed along with the namespace. I'm not sure if this is the correct behavior, and how best to fix it

andrevidela: I think we can leave that be for now

src/TTImp/ProcessData.idr
```diff
@@ -480,12 +485,25 @@ processData {vars} eopts nest env fc def_vis mbtot (MkImpData dfc n_in mty_raw o
...
 
+                    let declTot = findSetTotal $ ndef.flags
+                    tot <- case (mbtot, declTot) of
+                      (Just oldTot, Just newTot) => do
+                        when (oldTot /= newTot) $ throw $ GenericMsgSol fc
+                          "Data \{show n_in} has been forward-declared with totality `\{show oldTot}`, cannot change to `\{show newTot}`"
+                          "Possible solutions"
+                          [ "Use the same totality modifiers"
+                          , "Remove the totality modifier from the declaration"
+                          , "Remove the totality modifier from the definition"
```
spcfox, line 496: Maybe?
```Remove `\{newTot}` from the definition```

File: src/Core/Context.idr [hunk 3]
```diff
@@ -1628,6 +1628,14 @@ hasFlag fc n fl
               | Nothing => undefinedName fc n
          pure (fl `elem` flags def)
 
+export
+hasSetTotal : {auto c : Ref Ctxt Defs} -> FC -> Name -> Core Bool
```
mattpolzin, line 1632: Why is this needed in addition to `hasFlag` (above)?

src/Idris/Desugar.idr
```diff
@@ -1033,6 +1033,39 @@ mutual
...
+      checkDuplicates : List TotalReq -> List TotalReq -> Maybe String
+      checkDuplicates totalities dedupTotalities =
+        toMaybe (totalities /= dedupTotalities) $ do
+          let repeated = filter (\x => count (x ==) totalities > 1) dedupTotalities
+          "Remove duplicates of " ++ showModifiers repeated
+
+      checkDefault : TotalReq -> List TotalReq -> Maybe String
+      checkDefault def totalities =
+        toMaybe (def `elem` totalities) $
+          "Remove modifiers " ++ showModifiers totalities
```
mattpolzin, line 1067: Small detail but would be really nice to spell this one out fully by saying something to the effect of "Remove modifiers [...], resulting in the default totality of <foo>"

src/TTImp/ProcessData.idr
mattpolzin: I'm fine either way on this one.

src/Idris/Elab/Implementation.idr
```diff
@@ -466,8 +466,10 @@ elabImplementation {vars} ifc vis opts_in pass env nest is cons iname ps named i
 
     mkTopMethDecl : (Name, Name, List (String, String), RigCount, Maybe TotalReq, RawImp) -> ImpDecl
     mkTopMethDecl (mn, n, upds, c, treq, mty)
-        = let opts = maybe opts_in (\t => Totality t :: opts_in) treq in
-              IClaim (MkFCVal vfc (MkIClaimData c vis opts (MkImpTy EmptyFC (NoFC n) mty)))
+        = do let opts = if isJust $ findTotality opts_in
+                          then opts_in
```
mattpolzin, line 470: Is this throwing away the explicit totality passed to this function quietly? Seems like that could result in surprising behavior.

src/Core/Context.idr
spcfox: Because totality can be set by three different flags (`SetTotal Total`, `SetTotal CoveringOnly`, `SetTotal PartialOK`). It is possible to implement this check with `hasFlag`, but 3 calls will be required.

src/Idris/Elab/Implementation.idr
spcfox: `treq` is a totality from a function declaration in the interface. If the interface implementation contains its own totality modifier, it is overrided. See https://github.com/idris-lang/Idris2/issues/3437#issuecomment-2529756651

mattpolzin: I think the only remaining question I have is whether the old behavior of having two modifiers attached (as you say, not useful) resulted in the same behavior of your new PR or if you are getting the _other_ modifier after your PR gets merged. I guess put another way, are you keeping the same modifier as would have "won" in the eyes of the totality checker before your change?

mattpolzin: I guess the answer to my question just be yes based on the examples you bring up in your comment on the linked issue, but want to make sure that couldn't be false.

spcfox: Yes, I'm keeping the modifier that would have won before changes. The implementation modifier is a higher priority now.

mattpolzin (approved):

mattpolzin:
I think we can go ahead and merge this. I'm less than 100% sure on the question of _where_ the check should be run in the compiler pipeline but the net impact of this PR is quite positive.

PR #3423, opened 2024-11-22
WithFC elab

andrevidela:
# Description

Update reflected TTImp to replicate the change that were made in the compiler's tree. 

This is also needed as part of a fix for #3417 

Unfortunately, I suspect this is going to break a number of elaboration scripts. Hopefully this is something #3421 can help with in the future.

## Should this change go in the CHANGELOG?

- [x] If this is a fix, user-facing change, a compiler change, or a new paper
      implementation, I have updated [`CHANGELOG_NEXT.md`](https://github.com/idris-lang/Idris2/blob/main/CHANGELOG_NEXT.md) (and potentially also
      `CONTRIBUTORS.md`).

PR #3443, opened 2024-12-12
Fix reflection of `MkFCVal`

mjustus:
Fixes small regression in #3423 discovered by @madman-bob: `MkFCVal` has an implicit type argument that is made explicit in Core so we have to insert a placeholder in the right position.


andrevidela (approved):
Looks great, thanks a lot!

PR #3446, opened 2024-12-17
[ ci ] Make pack-dependent jobs to use bleeding edge

buzden:
This would allow us to break the tie when we need green CI for the breaking compiler change AND simultaneously correct pack collection building with the previous compiler version, like it happened in #3415.


buzden:
This PR is ready. Once it's merged, I'm ready to merge both downstreams to make #3415 be rebased on this and then checked that CI is green. But in this case #3415 should be merged today (to not to create failing pack collection at night). @gallais are we good to synchronise?

Issue #3369, opened 2024-08-09
Parsing Performance & Maintainability.


andrevidela:
Long term users of the idris programming language will know that one of the weaknesses of the compiler is
it's most common interaction model: the parser. Because the parser is quite a complex piece of software, there
is not a single element that we can point out and fix to resolve all the symptoms we experience because of it.

The role of this issue is to keep track of problems related to the parser, and different efforts deployed to solve
them. 

This issue will be closed when the current state of the parser will be improve in at least two aspects: Performance
and maintainability. First, a non-exhaustive list of symptoms we experience and would like to address.

## Existing and known problems

- Performance is subpar
- Inconsistent error recovery messages
- Unexpected whitespace rules
- Code architecture regarding location is brittle
- Code architecture around desugaring and resugaring is brittle
- No tree sitter support
- Code synthesis generates unparsable code

## Next steps

Here are a list of smaller steps that are not necessarily related to code that we can push as a PR on the
project but we still need to do in order to make progress toward better identifying and scoping solutions.

- [ ] Measure performance in multiple scenarios: Large files, interactive modes, visually ambiguous code, etc.
- [ ] investigate options around tree sitter. Are there ways to generate idris-compatible code from tree-sitter? Is there a way to generate tree-sitter spec from the parser? What kind of custom lexer do we need to write?
- [ ] Investigate a better API for file location that is compatible with desugaring and resugaring, and does not require mindful use of `FC` in constructors but can be handled automatically

## Conclusion

There are definitely more things we could do but I thought those would be small, achievable targets in the currently underspecified
goal to address parser runtime performance and maintainability. Please feel free to add to this discussion and share ideas of
small goals we can achieve easily to better design issues. Please, also share your experience with the parser if it is not represented
here in a way that we might be able to address with this kind of project

andrevidela:
Some personal notes:
- Last year @stefan-hoeck  showed us a very efficient constant-time parser using his library [idris2-parser](http://github.com/stefan-hoeck/idris2-parser). My personal assessment was that it's essentially the target to hit in term of performance, but re-writing the entire parser with it might be too much work. 
[...]

PR #3450, opened 2024-12-20
Parser refactor

### LINKED ISSUE #3417: Invalid FC of Pi-types
FC of Pi-type with named argument points to the name of the argument instead of whole type.

# Steps to Reproduce

```idris
Main> getFC `((Nat) -> Nat)
MkFC (Virtual Interactive) (0, 8) (0, 20)
Main> getFC `((arg : Nat) -> Nat)
MkFC (Virtual Interactive) (0, 9) (0, 12)
```

This is the reason for the bad error messages. Example:
```idris
unnamed : ()
unnamed = (Nat) -> Nat

named : ()
named = (arg : Nat) -> Nat
```

```shell
Error: While processing right hand side of unnamed. When unifying:
    Type
and:
    ()
Mismatch between: Type and ().

BadMessages:4:11--4:23
 1 | module BadMessages
 2 | 
 3 | unnamed : ()
 4 | unnamed = (Nat) -> Nat
               ^^^^^^^^^^^^

Error: While processing right hand side of named. When unifying:
    Type
and:
    ()
Mismatch between: Type and ().

BadMessages:7:10--7:13
 3 | unnamed : ()
 4 | unnamed = (Nat) -> Nat
 5 | 
 6 | named : ()
 7 | named = (arg : Nat) -> Nat
              ^^^
```

# Expected Behavior

FC points to whole type

# Observed Behavior

FC points to the name of the argument

andrevidela:
# Description

Apologies, this PR blew up. Those changes aim to simplify some aspects of the parser, by moving some of its logic into the desugarer. It contains:

- Preliminary work for #3369
  - By updating the surface level syntax to avoid folding pi-types at parsing-time, there is still more to do but those are the ones I could do.
  - By removing FC from PDecl and relying on `fcBounds` to keep track of locations, rather than thread the locations manually
  - By documenting the grammar for each sub parser
- Preliminary work for  #3402
  - This is actually why I started cleaning up top-level declarations
- Fixes #3417
- Extracts all the tests from ideMode005 into their own tests so that they can run individually and in parallel
  - I could not find any motivation for the previous architecture which required a bespoke script to find the diff between expected and actual output. This solution makes the test suite uniform
- Updating semantic highlighting to keep track of the names that are highlighted, not just their positions.

If I had more time I would have written a shorter PR.

Merry Christmas


andrevidela:
Alrighty I guess I'm cleaning up those ideMode tests too because currently they're a nightmare to deal with.

andrevidela:
It turns out there was a very easy change (involving typecase) that adds a lot of information to semantic decorations, so the tests have been adapted.

mattpolzin:
I've just been reading some of this PR to see if I can make any constructive comments. Would it be feasible to pull the WithFC changes into a different PR to the extent they don't involve new code? That seems to make up a substantial chunk of the code in this PR st first blush.

andrevidela:
What changes in particular are you thinking of? It feels like it would be easiest to split off the changes of the test suite with the changes to the surface syntax tree.

mattpolzin:
At a glance it looked like the total number of changed lines could be reduced a fair bit by moving the (mostly trivial) `WithFC`/`MkFCVal` changes to another PR that could get merged without much fuss leaving the more interesting stuff for this PR.

andrevidela:
Alright. I'll see what I can do

andrevidela:
On hold until #3460

andrevidela:
Rebased and ready to review again

andrevidela:
Because I need this for #3402 , and I needed this 6 months ago, I'll get this merged now and hopefully finish #3402 soon

PR #3451, opened 2024-12-20
[ fix #3448 ] Fix constraint search

### LINKED ISSUE #3448: `%search` cannot find the parameter in the interface constraint
# Steps to Reproduce

```idris
X : Type
data Y : X -> Type

failing "Can't find an implementation for X"
  interface Y %search => Z (x : X) where

interface Y x => Z (x : X) where
```

# Expected Behavior

One of two: 
1. Both definitions are incorrect
2. `%search` can find `x`

# Observed Behavior

Interface parameters can be used in constraints, but `%search` cannot find them.

spcfox:
# Description

Added parameters to constraint hints to fix #3448

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Idris/Elab/Interface.idr
+++ b/src/Idris/Elab/Interface.idr
@@ -97,6 +97,15 @@ mkDataTy fc [] = IType fc
 mkDataTy fc ((n, (_, ty)) :: ps)
     = IPi fc top Explicit (Just n) ty (mkDataTy fc ps)
 
+jname : (Name, (RigCount, RawImp)) -> (Maybe Name, RigCount, RawImp)
+jname (n, rig, t) = (Just n, rig, t)
+
+mkTy : FC -> PiInfo RawImp ->
+       List (Maybe Name, RigCount, RawImp) -> RawImp -> RawImp
+mkTy fc imp [] ret = ret
+mkTy fc imp ((n, c, argty) :: args) ret
+    = IPi fc c imp n argty (mkTy fc imp args ret)
+
 mkIfaceData : {vars : _} ->
               {auto c : Ref Ctxt Defs} ->
               FC -> WithDefault Visibility Private -> Env Term vars ->
@@ -109,8 +118,9 @@ mkIfaceData {vars} ifc def_vis env constraints n conName ps dets meths
                     else [NoHints, UniqueSearch, SearchBy dets]
           pNames = map fst ps
           retty = apply (IVar vfc n) (map (IVar EmptyFC) pNames)
-          conty = mkTy Implicit (map jname ps) $
-                  mkTy AutoImplicit (map bhere constraints) (mkTy Explicit (map bname meths) retty)
+          conty = mkTy vfc Implicit (map jname ps) $
+                  mkTy vfc AutoImplicit (map bhere constraints) $
+                  mkTy vfc Explicit (map bname meths) retty
           con = MkImpTy vfc (NoFC conName) !(bindTypeNames ifc [] (pNames ++ map fst meths ++ vars) conty)
           bound = pNames ++ map fst meths ++ vars in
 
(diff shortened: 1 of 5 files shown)


src/Idris/Elab/Interface.idr
```diff
@@ -246,11 +246,13 @@ getConstraintHint : {vars : _} ->
...
+                    (params : List (Name, RigCount, RawImp)) ->
                     (Name, RawImp) -> Core (Name, List ImpDecl)
 getConstraintHint {vars} fc env vis iname cname constraints meths params (cn, con)
-    = do let ity = apply (IVar fc iname) (map (IVar fc) params)
-         let fty = IPi fc top Explicit Nothing ity con
+    = do let pNames = map fst params
+         let ity = apply (IVar fc iname) (map (IVar fc) pNames)
+         let fty = mkTy fc Implicit (map jname params) $
+                   mkTy fc Explicit [(Nothing, top, ity)] con
          ty_imp <- bindTypeNames fc [] (meths ++ vars) fty
```
spcfox, line 256: Do I need to add parameters?
```idris
ty_imp <- bindTypeNames fc [] (pNames ++ meths ++ vars) fty
```
like in `mkIfaceData`
https://github.com/idris-lang/Idris2/blob/457ca7c6effd66a3f49e721171b142f979bb7406/src/Idris/Elab/Interface.idr#L114
and `getMethDecl`
https://github.com/idris-lang/Idris2/blob/457ca7c6effd66a3f49e721171b142f979bb7406/src/Idris/Elab/Interface.idr#L152

tests/idris2/basic/basic049/expected
```diff
@@ -15,30 +15,6 @@ Warning: We are about to implicitly bind the following lowercase names.
...
- 77 |   constructor MkIface
- 78 |                                              -- constraints
- 79 |   data MyData : a -> Type                    -- and a data declaration.
- 80 |   someFunc : a -> a                          -- Constraints are now elaborated as auto implicits (as one would expect)
- 81 |   giveBack : {x : a} -> MyData x -> a        -- (previously as explicit arguments of the interface
-
-Warning: We are about to implicitly bind the following lowercase names.
-You may be unintentionally shadowing the associated global definitions:
-  a is shadowing Main.R2.R2.a, Main.R1.R1.a
-
```
spcfox, line 41: It looks like these warnings were occurring in hints for constraints. I'm not sure if they are useful, but it seems odd

buzden: > I'm not sure if they are useful, but it seems odd

To me these warning look completely okay, since both interface definition and interface implementations seem to shadow the top-level `a`s

spcfox: Yes, but in the following code `MyIface0`, `MyIface1` and `MyIface2` produce 0, 1 and 2 warnings respectively. This means that one warning per constraint is generated. This is because they do not arise from the data the interface is elabored into, but from the constraint hint. This is strange
```idris
a : Type

interface MyIface0 a where
  f0 : a -> Type

interface Show a => MyIface1 a where
  f1 : a -> Type

interface Show a => (num : Num a) => MyIface2 a where
  f2 : a -> Type
```

spcfox:
`%search` still can't find implicitly bound parameters:
```idris
X : Type
data Y : X -> Type

failing "While processing constructor MkZ. Can't find an implementation for X"
  interface Y %search => Z (y : Y x) where
    constructor MkZ

interface Y x => Z (y : Y x) where
  constructor MkZ

0 getX : Z y => X
getX = %search
```
I expected that [adding `pNames`](https://github.com/idris-lang/Idris2/pull/3451#discussion_r1893990896) would fix it, but no

spcfox:
Ok, it's not a problem with interfaces. This is probably the expected behavior
```idris
X : Type
data Y : X -> Type

failing "While processing constructor MkZ. Can't find an implementation for X"
  interface Y %search => Z (y : Y x) where
    constructor MkZ

-- The interface is elaborated into such data
failing "While processing constructor MkZ. Can't find an implementation for X"
  data Z : Y x -> Type where
    MkZ : {y : Y x} -> Y %search => Z y
```

src/Idris/Elab/Interface.idr
spcfox: It seems to me that this might be necessary since I added parameters to `fty`, but I can't think of an example where this would have an impact

tests/idris2/basic/basic049/expected
spcfox: A warning about shadowing is produced when `a` is an unbound implicit, but in the above examples `a` is explicitly bound
```idris
a : Type

-- Produce warning
interface Iface1 (x : a) where

-- No warning
interface Iface2 a (x : a) where
```

src/Idris/Elab/Interface.idr
gallais: I think so. Prior to that we were letting the bindTypeNames mechanism take care of adding them.

dunhamsteve:
This also fixes the issue that I was addressing in #3052, where this would not compile:
```idris
%unbound_implicits off
interface Foo a where
interface Foo a => Bar (a : Type) where
```
The issue there was that the constraints were relying on the unbound implicits mechanism to pick up the `a`, while all of the other generated code for interfaces was explicitly adding the parameters.

buzden (approved):

PR #3462, opened 2025-01-13
[ linear ] fork1, concurrently1

gallais:
Run a computation concurrently to the current thread.
This returns a receiver for the value. A typical usage
pattern is showcased by the implementation of `par1`:
in a do block start executing a series of actions
concurrently and then collect the results with a series
receiving steps.
```idris
do recva <- concurrently1 ioa
   recvb <- concurrently1 iob
   a <- recva
   b <- recvb
   pure1 (a # b)
```


mattpolzin (approved):
Cool, much more broadly usable.

PR #3464, opened 2025-01-15
deprecate old parameter-block syntax

### LINKED ISSUE #3447: RFC: Deprecate old parameter block syntax
I'd like to get the opinions of the community on the current parameter block syntax.

Currently, there are two syntaxes for parameter blocks:

```
parameters (n1 : t1, n2 : t2, n3 : t3, ...)
```

This syntax has been replaced by

```
parameters (0 n1 : t1) {n2 : t2} {auto n3 : t3} (n4, n5, n6 : t4) ...
```

The old syntax makes it easier to declare series of explicit arguments, but does not allow for implict arguments, or for quantities to be specified. It does not allow multiple names to be bound to the same type either. Finally, it's a syntax form that is not seen anywhere else.

The new syntax allows for quantities to be written, and for implicit arguments to be defined. It mimics the syntax for record parameters/indices.

The question is then: Should we keep supporting the old bespoke syntax, or should we push toward using the more flexible syntax?

andrevidela:
# Description

Fix #3447

      `CONTRIBUTORS.md`).

### DIFF
--- a/docs/source/tutorial/modules.rst
+++ b/docs/source/tutorial/modules.rst
@@ -301,15 +301,15 @@ Look at the following example.
     useNSFromOutside' : String
     useNSFromOutside' = Y.g "x" -- value is "xa"
 
-Parameterised blocks
-====================
+Parameterised blocks - `parameters`-blocks
+==========================================
 
 Groups of functions can be parameterised over a number of arguments
 using a ``parameters`` declaration, for example:
 
 .. code-block:: idris
 
-    parameters (x : Nat, y : Nat)
+    parameters (x : Nat) (y : Nat)
       addAll : Nat -> Nat
       addAll z = x + y + z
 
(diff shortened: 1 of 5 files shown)


docs/source/tutorial/modules.rst
```diff
@@ -301,15 +301,15 @@ Look at the following example.
     useNSFromOutside' : String
     useNSFromOutside' = Y.g "x" -- value is "xa"
 
-Parameterised blocks
+Parameterised blocks - `parameters`-blocks
 ====================
```
buzden, line 305: `rst` requires the underline to be extended too

andrevidela: should be fixed

CodingCellist:
Mostly out of curiosity given that the commas have disappeared for differently typed parameters (which I think is sensible): does it behave the same as type declarations? I.e. can you declare multiple parameters as `(a, b, c : Nat)`?  : )

andrevidela:
Yes you can!

gallais (approved):

PR #3465, opened 2025-01-15
[ re #3423 ] Make `IClaimData` to be a `data`, not `record`

buzden:
# Description

PR #3423 introduces a new data structure which added several top-level declarations with names `rig`, `vis`, `opts` and `type` which leads a lot of existing code to warnings of shadowing these top-level names with local variables. These names are pretty popular, especially for parameter names in certain contexts (even precisely this module has already several parameters named in this way).

I believe that this was not the intended behaviour, to I suggest to declare `IClaimData` not as a `record`, like it's done with the type `Record` in the same module.

Pinging @andrevidela as the original PR author.


andrevidela (approved):
Looks good

PR #3468, opened 2025-01-17
Fix parameter records

### LINKED ISSUE #3467: Record parameter cannot be used in implicit position
# Steps to Reproduce

```idris
record P (t : Type) where
  f : {x : t} -> Type
```

# Expected Behavior

Works fine

# Observed Behavior

Shits the bed with:

```
Error: Unsolved holes:
Bug.P.{x:851} introduced at: 
```

At position 0:0

# Comments

All of the following works
```idris
record P (t : Type) where
  f : (x : t) -> Type
```

```idris
record P (t : Type) where
  f : t -> Type
```

```idris
data P : Type -> Type where
  MkP : {t : Type} -> (f : {x : t} -> Type) -> P t
```

Possibly related? #1582 #2612

andrevidela:
# Description

fix #3467 

      `CONTRIBUTORS.md`).


dunhamsteve:
TL;DR - I think I got this sorted out, and now it's almost bed time for me.  Details below, my branch is at: https://github.com/dunhamsteve/Idris2/tree/parameter-records 

It looked like it it stumbling on args with internal names. It suffices to only eta-expand the leading implicits, anything past an explicit is fine.   If we stop at the first explicit everything builds and it remains to deal with tests.
```
diff --git a/src/TTImp/ProcessRecord.idr b/src/TTImp/ProcessRecord.idr
index b7377c0a7..b85bdcfa4 100644
--- a/src/TTImp/ProcessRecord.idr
+++ b/src/TTImp/ProcessRecord.idr
@@ -298,9 +298,8 @@ elabRecord {vars} eopts fc env nest newns def_vis mbtot tn_in params0 opts conNa
                    let piNames = collectPiNames ty_chk
 
                    let namesToRawImp : List (Bool, Name) -> (fn : RawImp) -> RawImp
-                       namesToRawImp [] fn = fn
-                       namesToRawImp ((False, nm) :: xs) fn = namesToRawImp xs (IApp fc fn (IVar fc nm))
                        namesToRawImp ((True,  nm) :: xs) fn = namesToRawImp xs (INamedApp fc fn nm (IVar fc nm))
+                       namesToRawImp _ fn = fn
 
                    -- Then apply names for each argument to the lhs
                    let lhs = namesToRawImp piNames lhs
```

A number of tests in interactive and ideMode are failing because of the FC additions, we can fix them by doing:
```idris
let fc' = virtualiseFC fc
```
and then using `fc'` where you added `fc`.

The following tests fail because the numbers in some machine names changed, they can be updated: 
```
idris2/basic044, base/deriving_*, interactive/interactive042, idris2/error/error018, idrist/evaluator/*
```
Interactive042 also changed by adding some "implicitly bound at ...", which is nice.

A number of tests are broken by the `Unelab.idr` change. It looks like your change wouldn't be needed if the unelab mode was `NoSugar _`.  So I changed the way that unelab was called from `ProcessRecord` to use that mode, and everybody else gets the same unelab that they always had.  I had to update the `record019` test to reflect that things have names now.

andrevidela:
```
diff --git a/src/TTImp/ProcessRecord.idr b/src/TTImp/ProcessRecord.idr
index b7377c0a7..b85bdcfa4 100644
--- a/src/TTImp/ProcessRecord.idr
+++ b/src/TTImp/ProcessRecord.idr
@@ -298,9 +298,8 @@ elabRecord {vars} eopts fc env nest newns def_vis mbtot tn_in params0 opts conNa
                    let piNames = collectPiNames ty_chk
 
                    let namesToRawImp : List (Bool, Name) -> (fn : RawImp) -> RawImp
-                       namesToRawImp [] fn = fn
-                       namesToRawImp ((False, nm) :: xs) fn = namesToRawImp xs (IApp fc fn (IVar fc nm))
                        namesToRawImp ((True,  nm) :: xs) fn = namesToRawImp xs (INamedApp fc fn nm (IVar fc nm))
+                       namesToRawImp _ fn = fn
 
                    -- Then apply names for each argument to the lhs
                    let lhs = namesToRawImp piNames lhs
```

I'm amazed this works, I assumed it would choke on `{x : t} -> (y : t) -> {z : t} -> Type` but it works just fine

dunhamsteve:
Yeah, with `{x : t} -> (y : t) -> {z : t} -> Type`, the compiler stops picking up implicits on its own when it hits the explicit `(y : t)`. It won't sneak in explicits. So if you write
```
foo : x -> {a : t} -> y -> {b : t} -> z
foo x = ?hole
```
The compiler will pick up the implicit `a` on the LHS (essentially `foo a {x} =`) and expects type `y -> {b : t} -> z` at `?hole`.  And for `foo = ?hole2`, the type of ?hole2 would be `x -> {a : t} -> y -> {b : t} -> z`.

andrevidela:
I tried it on the un-minimized program where I discovered this and it works so I'll merge this now

PR #3475, opened 2025-01-22
[ fix #3474 ] Fix implicit arguments in interface methods

### LINKED ISSUE #3474: An unused interface method argument produces an unresolved hole
# Steps to Reproduce

```idris
interface Iface where
  method : {x : Type} -> Type
```

# Expected Behavior

Successful compilation

# Observed Behavior

```
Error: Unsolved holes:
Main.{x:836} introduced at:
```

spcfox:
# Description

fix #3474

      `CONTRIBUTORS.md`).


src/Idris/Elab/Interface.idr
```diff
@@ -238,6 +232,20 @@ getMethToplevel {vars} env vis iname cname constraints allmeths params sig
...
+    collectImplicitNames (IPi _ _ Explicit _        _ ty) = []
+    collectImplicitNames (IPi _ _ _        (Just n) _ ty) = n :: collectImplicitNames ty
+    collectImplicitNames (IPi _ _ _        Nothing  _ ty) = collectImplicitNames ty
+    collectImplicitNames _                                = []
+
+    namesToRawImp : Bool -> List Name -> RawImp -> RawImp
+    namesToRawImp bind (nm@(UN{}) :: xs) fn =
+      namesToRawImp bind xs $ INamedApp vfc fn nm $
+        if bind
+           then IBindVar vfc $ nameRoot nm
```
spcfox, line 245: `IBindVar` is needed so that using capitalized names does not generate an error. Such names occur in the `interface010` and `interface011` tests.

To avoid a name conflict, I'd like to use machine generated names, but `IBindVar` accepts a string.

src/Idris/Elab/Interface.idr
```diff
@@ -238,6 +232,20 @@ getMethToplevel {vars} env vis iname cname constraints allmeths params sig
     methName : Name -> Name
     methName n = UN (Basic $ bindName n)
 
+    collectImplicitNames : RawImp -> List Name
+    collectImplicitNames (IPi _ _ Explicit _        _ ty) = []
+    collectImplicitNames (IPi _ _ _        (Just n) _ ty) = n :: collectImplicitNames ty
+    collectImplicitNames (IPi _ _ _        Nothing  _ ty) = collectImplicitNames ty
```
spcfox, line 238: Skipping implicit arguments without names can cause an error, but I'm not sure how to handle them properly

src/Idris/Elab/Interface.idr
```diff
@@ -238,6 +232,20 @@ getMethToplevel {vars} env vis iname cname constraints allmeths params sig
...
+    collectImplicitNames _                                = []
+
+    namesToRawImp : Bool -> List Name -> RawImp -> RawImp
```
spcfox, line 241: This function is based on [a similar function from `processRecord`](https://github.com/idris-lang/Idris2/blob/8e420d6b2e29c276f77b562955f01958664ccb58/src/TTImp/ProcessRecord.idr#L300-L302). It may be worth making a global function for it

src/Idris/Elab/Interface.idr
spcfox: Example with an error:
```idris
interface Iface where
  method : {_ : Type} -> Type
```

spcfox: As far as I know, an implicit argument without a name cannot be passed. It might be useful to add a “disable implicits” mode as [suggested in Discord](https://discord.com/channels/827106007712661524/827109161857712128/1329577616641888340) and use it when elaboring. It could also be useful in user code

src/Idris/Elab/Interface.idr
gallais: We can revisit this once #3526 is merged.

gallais (commented):
Looks good to me. Let's get the `IBindVar` change in so that we can have
the more principled solution and then move forward with this one too.

spcfox:
Now, implicits in methods and fields should be processed correctly regardless of their name

tests/idris2/data/record021/expected
```diff
@@ -1,2 +1,24 @@
...
+  x
+
+RecordParam:23:3--23:87
+ 19 | record Record7 where
+ 20 |   f : {_ : Type} -> Type
+ 21 | 
+ 22 | record Record8 where
+ 23 |   f : {x : Type} -> {x : Int} -> {_ : String} -> (x : Double) => {x : Integer} -> Type
+        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
spcfox, line 23: Two warnings, because two projections are generated: prefix and suffix

gallais:
I was already happy with this so also happy to merge once it's ready.

tests/idris2/data/record021/expected
spcfox: I think this is acceptable behaviour

PR #3478, opened 2025-01-25
[ fix ] False positives checking for conflicts in coverage checking

### LINKED ISSUE #3477: Not covering pattern compiles with auto proof on list length
# Steps to Reproduce

Enter the below function:

```haskell
test : (xs : List Int) -> {auto 0 prf : 2 = length xs} -> (Int, Int)
test (x1 :: x2 :: x3 :: [])  = (x1, x2)
```

# Expected Behavior

This function is not covering as no pattern can be matched.
Therefore the code should not compile and the below error should be reported:

```
Error: test is not covering.
```

# Observed Behavior

The function compiles.

dunhamsteve:
# Description

This fixes #3477.  Idris considers `2 = length xs` as empty while checking missing cases in coverage checking, and it incorrectly accepts this as covering:
```idris
test : (xs : List Int) ->  2 = length xs -> (Int, Int)
test (x1 :: x2 :: x3 :: []) pf = (x1, x2)
```

There are two changes here:

I removed the `clearDefs` in `conflictNF` to allow the `length xs` to expand when `xs` is known.  _Let me know if there is a reason we don't allow expansions here._ It passes tests and it seems to be necessary to enable Idris to automatically use the `2 = length xs`. 

Without this change, the following `impossible` clauses are required:
```idris
test : (xs : List Int) ->  2 = length xs -> (Int, Int)
test (x1 :: x2 :: []) pf = (x1, x2)
test (x1 :: Nil) Refl impossible
test (x1 :: x2 :: x3 :: _) Refl impossible
```

The second change addresses the root cause.  When checking for clashes between types in `conflictMatch`, idris checks if the heads clash, and if not it proceeds to check the arguments. This is a problem when `S (S Z)` is compared to `length xs`. `S` and `length` do not clash, but it is not correct to then compare `S Z` against `xs`.  To remedy this, it is now only checking the spines in the case of matching data or type constructors. The helper function returns `Nothing` to indicate the spine should be checked.

`Krivine.idr` was relying on this bug. The code was covering, but coverage checking was accepting it for the wrong reason. It was discounting the other case when looking at `Element (EvalContext{}) _` vs `(::) (Arr{}) ctx` (here `::` is a function, not a constructor). `Element` vs `(::)` was not clashing, so it proceeded to compare the arguments and saw that `EvalContext{}` and `Arr{}` clashed (values headed by different constructors).  After this change, it needed an impossible clause to unblock the `(::)`. 

I've also added a test case, fixed a typo in a comment, and added a `log` message that I use do diagnose these `impossible` issues. This will break code that relies on false positives in coverage, but given that the rest of papers and frex are uneffected, I suspect that is rare.

### DIFF
--- a/src/Core/Coverage.idr
+++ b/src/Core/Coverage.idr
@@ -24,20 +24,21 @@ conflictMatch : {vars : _} -> List (Name, Term vars) -> Bool
 conflictMatch [] = False
 conflictMatch ((x, tm) :: ms) = conflictArgs x tm ms || conflictMatch ms
   where
-    clash : Term vars -> Term vars -> Maybe Bool
+    data ClashResult = Distinct | Same | Incomparable
+
+    clash : Term vars -> Term vars -> ClashResult
     clash (Ref _ (DataCon t _) _) (Ref _ (DataCon t' _) _)
-        = if t /= t' then Just True else Nothing
+        = if t /= t' then Distinct else Same
     clash (Ref _ (TyCon t _) _) (Ref _ (TyCon t' _) _)
-        = if t /= t' then Just True else Nothing
-    clash (PrimVal _ c) (PrimVal _ c')
-      = Just $ c /= c'
-    clash (Ref _ t _) (PrimVal _ _) = Just $ isJust (isCon t)
-    clash (PrimVal _ _) (Ref _ t _) = Just $ isJust (isCon t)
-    clash (Ref _ t _) (TType _ _)   = Just $ isJust (isCon t)
-    clash (TType _ _) (Ref _ t _)   = Just $ isJust (isCon t)
-    clash (TType _ _) (PrimVal _ _) = Just True
-    clash (PrimVal _ _) (TType _ _) = Just True
-    clash _ _ = Just False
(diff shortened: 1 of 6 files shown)


src/Core/Coverage.idr
```diff
@@ -24,20 +24,20 @@ conflictMatch : {vars : _} -> List (Name, Term vars) -> Bool
 conflictMatch [] = False
 conflictMatch ((x, tm) :: ms) = conflictArgs x tm ms || conflictMatch ms
   where
-    clash : Term vars -> Term vars -> Bool
+    clash : Term vars -> Term vars -> Maybe Bool
```
gallais, line 27: Can you please document the meaning of this `Maybe`? Or should we even define
e.g. `data ClashingDecision = Distinct | Same | Incomparable`
with the idea that we are only checking argument if they're `Same`?

dunhamsteve: I was on the fence about whether to use an enum or not.  Coming back to this a week later and having to get my bearings again, I think it would be clearer if I did that.

Issue #3457, opened 2024-12-30
Functions with forward-declared data as an argument incorrectly accepted as covering


dunhamsteve:
Coverage checking appears to be broken when a forward declared data type is involved.

# Steps to Reproduce

```idris
data Three = A | B | C
data Bar : Type
data Foo : Type where
  MkFoo : Three -> Bar -> Foo

fun : Foo -> String
fun (MkFoo A y) = ""
fun (MkFoo B y) = ""

data Bar : Type where
  MkBar : Bar
```

# Expected Behavior

I expect Idris to call out the missing case of `C` in `fun` or complain that `Bar` hasn't been defined yet when checking `fun`.

What would be the ideal behavior in this case? Ignore the `y` slot and call out the missing `C` or make it an error to have undefined types when building case trees?

# Observed Behavior

The function `fun` is accepted as covering, despite missing the `C` case. I believe it's deciding that case is impossible because one of the pattern variables doesn't have a valid constructor. Failing to distinguish between empty and no constructors defined yet.

dunhamsteve:
Yeah, I'll add that test. It more subtle than the included tests.

Issue #1083, opened 2021-02-18
Scoping issue with mutual definitions


gallais:
New proof of `Void` thanks to @sysvinit 

```idris
%default total
 
data Bottom : Type -- where -- note that we don't even need the `where`!

boom : Not Bottom
boom arg impossible
 
data Bottom : Type where
  MkBottom : Bottom
 
argh : Void
argh = boom MkBottom
```

gallais:
Induction-reduction in general seems fairly broken:

```idris
module IR                                                                                                                       
 
%default total
 
data Sig : Type

Elt : Sig -> Type
Elt s impossible
 
data Sig : Type where
  NAT : Sig
  SIG : (a : Sig) -> (Elt a -> Sig) -> Sig
```

buzden:
Well, as far as I understand, the first example is not strictly a mutuality. I mean, `boom` should have no access to the definition of `Bottom` but definition of `Bottom` may have an access to the definition of `boom`. If they were literally mutual, none of both would have an access to definition of both.

Russoul:
Almost got caught by this in my "real" code today.
```
%default total

mutual
  data DataA : Type where
    MkDataA : Void -> DataA
    MkDataA' : DataB -> DataA

  notCovering : DataA -> Void
  notCovering (MkDataA empty) = empty

  data DataB : Type where
    MkDataB : DataB

boom : Void
boom = notCovering (MkDataA' MkDataB)

```
Was going to report that but found this issue :)

dunhamsteve:
Closing this because it has been fixed by #3480 (all three cases above report errors now).

PR #3480, opened 2025-01-30
Empty data declarations are confused with forward data declarations

dunhamsteve:
# Description

This addresses two issues related to confusing empty `data` declarations with forward `data` declarations.  

The following is accepted by Idris and I believe it should be rejected with an error that `Bar` is already defined.
```idris
data Bar : Type where

data Bar : Type where
    MkBar : Bar
```

The second issue is that this forward declaration is allowed to be used in an impossible clause and later is defined to not be empty:
```idris
data Bar : Type

total
foo : Bar -> a
foo x impossible

data Bar : Type where
    MkBar : Bar
```
Since forward declarations and empty data appear to be indistinguishable in the data, I added a type flag for `forwardDecl`.  There is an existing flag `external` that is used in similar cases. This affects serialization, so I updated the TTC version.

The test for `elab-util` is failing with this PR because it has an empty declaration instead of a forward declaration in one location. I'm filing a PR to address that.

### DIFF
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -585,9 +585,11 @@ HasNames Def where
       fullNamesPat (_ ** (env, lhs, rhs))
           = pure $ (_ ** (!(full gam env),
                           !(full gam lhs), !(full gam rhs)))
-  full gam (TCon t a ps ds u ms cs det)
+  full gam (TCon t a ps ds u ms Nothing det)
+      = pure $ TCon t a ps ds u !(traverse (full gam) ms) Nothing det
+  full gam (TCon t a ps ds u ms (Just cs) det)
       = pure $ TCon t a ps ds u !(traverse (full gam) ms)
-                                !(traverse (full gam) cs) det
+                                (Just !(traverse (full gam) cs)) det
   full gam (BySearch c d def)
       = pure $ BySearch c d !(full gam def)
   full gam (Guess tm b cs)
@@ -603,9 +605,11 @@ HasNames Def where
       resolvedNamesPat (_ ** (env, lhs, rhs))
           = pure $ (_ ** (!(resolved gam env),
                           !(resolved gam lhs), !(resolved gam rhs)))
-  resolved gam (TCon t a ps ds u ms cs det)
+  resolved gam (TCon t a ps ds u ms Nothing det)
+      = pure $ TCon t a ps ds u !(traverse (resolved gam) ms) Nothing det
+  resolved gam (TCon t a ps ds u ms (Just cs) det)
       = pure $ TCon t a ps ds u !(traverse (resolved gam) ms)
-                                !(traverse (resolved gam) cs) det
+                                (Just !(traverse (resolved gam) cs)) det
   resolved gam (BySearch c d def)
       = pure $ BySearch c d !(resolved gam def)
   resolved gam (Guess tm b cs)
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -585,11 +585,9 @@ HasNames Def where
       fullNamesPat (_ ** (env, lhs, rhs))
           = pure $ (_ ** (!(full gam env),
                           !(full gam lhs), !(full gam rhs)))
-  full gam (TCon t a ps ds u ms Nothing det)
-      = pure $ TCon t a ps ds u !(traverse (full gam) ms) Nothing det
-  full gam (TCon t a ps ds u ms (Just cs) det)
+  full gam (TCon t a ps ds u ms mcs det)
       = pure $ TCon t a ps ds u !(traverse (full gam) ms)
-                                (Just !(traverse (full gam) cs)) det
+                                !(traverseOpt (traverse (full gam)) mcs) det
   full gam (BySearch c d def)
       = pure $ BySearch c d !(full gam def)
   full gam (Guess tm b cs)
@@ -605,11 +603,9 @@ HasNames Def where
       resolvedNamesPat (_ ** (env, lhs, rhs))
           = pure $ (_ ** (!(resolved gam env),
                           !(resolved gam lhs), !(resolved gam rhs)))
-  resolved gam (TCon t a ps ds u ms Nothing det)
-      = pure $ TCon t a ps ds u !(traverse (resolved gam) ms) Nothing det
-  resolved gam (TCon t a ps ds u ms (Just cs) det)
+  resolved gam (TCon t a ps ds u ms mcs det)
       = pure $ TCon t a ps ds u !(traverse (resolved gam) ms)
-                                (Just !(traverse (resolved gam) cs)) det
+                                !(traverseOpt (traverse (full gam)) mcs) det
   resolved gam (BySearch c d def)
       = pure $ BySearch c d !(resolved gam def)
   resolved gam (Guess tm b cs)
--- a/src/Core/Termination/Positivity.idr
+++ b/src/Core/Termination/Positivity.idr
@@ -166,6 +166,7 @@ calcPositive loc n
          logC "totality.positivity" 6 $ do pure $ "Calculating positivity: " ++ show !(toFullNames n)
          case !(lookupDefTyExact n (gamma defs)) of
               Just (TCon _ _ _ _ _ tns dcons _, ty) =>
+                  let dcons = fromMaybe [] dcons in
                   case !(totRefsIn defs ty) of
                        IsTerminating =>
                             do log "totality.positivity" 30 $
--- a/src/TTImp/ProcessRecord.idr
+++ b/src/TTImp/ProcessRecord.idr
@@ -147,9 +147,9 @@ elabRecord {vars} eopts fc env nest newns def_vis mbtot tn_in params0 opts conNa
              let dataTy = IBindHere fc (PI erased) !(bindTypeNames fc [] vars (mkDataTy fc params0))
              defs <- get Ctxt
              -- Create a forward declaration if none exists
-             let dt = MkImpLater fc tn dataTy
-             log "declare.record" 10 $ "Pre-declare record data type: \{show dt}"
-             when (isNothing !(lookupTyExact tn (gamma defs))) $
+             when (isNothing !(lookupTyExact tn (gamma defs))) $ do
+               let dt = MkImpLater fc tn dataTy
+               log "declare.record" 10 $ "Pre-declare record data type: \{show dt}"
                processDecl [] nest env (IData fc def_vis mbtot dt)
              defs <- get Ctxt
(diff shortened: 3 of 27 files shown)


dunhamsteve:
This fixes #3457

buzden:
> It looks like this fixes #1083 and #3457

#3457 has more complex case than the added case. WDYT, should it also be added as test?

Also, does this change work well for records and their forward declarations? I believe, it should, but I recall some problems with translation of forward declarations of records some time ago.

dunhamsteve:
Yeah, I'd like to add that case. It's a little more subtle than the other two.

dunhamsteve:
The `record018` test checks that forward declarations of records still work.  The following reproduces the issue with `record` and is fixed by the PR:
```idris
record Bar
foo : Bar -> a
foo x impossible
record Bar where
  constructor MkBar
```

gallais:
I personally think a better fix in terms of modelising what is going on would be to
change the type of `TCon` like so:

```diff
  TCon : (tag : Int) -> (arity : Nat) ->
           (parampos : List Nat) -> -- parameters
           (detpos : List Nat) -> -- determining arguments
           (flags : TypeFlags) -> -- should 'auto' implicits check
           (mutwith : List Name) ->
-           (datacons : List Name) ->
+          (datacons : Maybe (List Name)) ->
           (detagabbleBy : Maybe (List Nat)) ->
                    -- argument positions which can be used for
                    -- detagging, if it's possible (to check if it's
                    -- safe to erase)
           Def
```

0r maybe even (using a yet-to-be-named record rather than the uninformative nested pair of course)

```
  TCon : (tag : Int) -> (arity : Nat) ->
-           (parampos : List Nat) -> -- parameters
-           (detpos : List Nat) -> -- determining arguments
           (flags : TypeFlags) -> -- should 'auto' implicits check
           (mutwith : List Name) ->
-           (datacons : List Name) ->
+          (datacons : Maybe (List Name, List Nat, List Nat, List Nat)) ->
-           (detagabbleBy : Maybe (List Nat)) ->
                    -- argument positions which can be used for
                    -- detagging, if it's possible (to check if it's
                    -- safe to erase)
           Def
```

and refuse to throw in extra constructors if that `Maybe` is a `Just`.

dunhamsteve:
I'd considered the first option, but went with the less invasive approach.  The `Maybe (List Name)` does model the situation better, forcing us to decide how we want to handle this not-yet-defined situation in each case.  I'm looking into that option 

I think I've got that working, but want to double check a couple of things after work.

While wiring that through, I saw that positivity checking also assumes the constructors are available. I haven't looked in to it yet, but we may have an issue where a forward declared data type is considered positive.

dunhamsteve:
I've updated this to use `Maybe (List Name)`.  To avoid breaking elab reflection, it will return an empty List in the case of Nothing for `getCons`.   I've given it a different `ttcVersion`.  Because `TCon` has changed, the lsp will need to be updated  after this patch.

gallais:
> While wiring that through, I saw that positivity checking also assumes the constructors are available. I haven't looked in to it yet, but we may have an issue where a forward declared data type is considered positive.

We seem to be okay:

```idris
%default total

data Negation : Type -> Type

failing "Absurd is not total, not strictly positive"

  data Absurd : Type -> Type where
    MkAbsurd : Negation (Absurd a) -> Absurd a
```

but it's probably worth adding as a test case.

dunhamsteve:
In case someone else needs it, the patch for LSP is fairly straightforward but can be found here: https://github.com/idris-community/idris2-lsp/compare/main...dunhamsteve:idris2-lsp:update-for-3480

In addition to that change, it would also need the submodule updated to point at the latest Idris after this PR is merged.

src/Core/Context.idr
```diff
@@ -585,9 +585,11 @@ HasNames Def where
       fullNamesPat (_ ** (env, lhs, rhs))
           = pure $ (_ ** (!(full gam env),
                           !(full gam lhs), !(full gam rhs)))
-  full gam (TCon t a ps ds u ms cs det)
+  full gam (TCon t a ps ds u ms Nothing det)
+      = pure $ TCon t a ps ds u !(traverse (full gam) ms) Nothing det
+  full gam (TCon t a ps ds u ms (Just cs) det)
       = pure $ TCon t a ps ds u !(traverse (full gam) ms)
-                                !(traverse (full gam) cs) det
+                                (Just !(traverse (full gam) cs)) det
```
gallais, line 592: Should this be
```idris
  full gam (TCon t a ps ds u ms mcs det)
      = pure $ TCon t a ps ds u !(traverse (traverse (full gam)) mcs) Nothing det
```
?

src/Core/Context.idr
```diff
@@ -603,9 +605,11 @@ HasNames Def where
       resolvedNamesPat (_ ** (env, lhs, rhs))
           = pure $ (_ ** (!(resolved gam env),
                           !(resolved gam lhs), !(resolved gam rhs)))
-  resolved gam (TCon t a ps ds u ms cs det)
+  resolved gam (TCon t a ps ds u ms Nothing det)
+      = pure $ TCon t a ps ds u !(traverse (resolved gam) ms) Nothing det
+  resolved gam (TCon t a ps ds u ms (Just cs) det)
       = pure $ TCon t a ps ds u !(traverse (resolved gam) ms)
-                                !(traverse (resolved gam) cs) det
+                                (Just !(traverse (resolved gam) cs)) det
```
gallais, line 612: Same here

src/Core/Termination/Positivity.idr
```diff
@@ -166,6 +166,7 @@ calcPositive loc n
          logC "totality.positivity" 6 $ do pure $ "Calculating positivity: " ++ show !(toFullNames n)
          case !(lookupDefTyExact n (gamma defs)) of
               Just (TCon _ _ _ _ _ tns dcons _, ty) =>
+                  let dcons = fromMaybe [] dcons in
```
gallais, line 169: Does it make sense to add a test case ensuring that this cannot be exploited
(e.g. by trying to have positivity be computed in one module & reused in another
after adding constructors):

```idris2
module OopsDef

%default total

public export
data Oops : Type

forcePosCheck : Oops -> Oops
forcePosCheck x = x
```

and

```idris
import OopsDef

%default total

failing "Oops is not total, not strictly positive"

  data OopsDef.Oops : Type where
    MkFix : Not Oops -> Oops
```

src/TTImp/ProcessRecord.idr
```diff
@@ -145,10 +145,12 @@ elabRecord {vars} eopts fc env nest newns def_vis mbtot tn_in params0 opts conNa
     preElabAsData tn
         = do let fc = virtualiseFC fc
              let dataTy = IBindHere fc (PI erased) !(bindTypeNames fc [] vars (mkDataTy fc params0))
-             -- we don't use MkImpLater because users may have already declared the record ahead of time
-             let dt = MkImpData fc tn (Just dataTy) opts []
+             defs <- get Ctxt
+             -- Create a forward declaration if none exists
+             let dt = MkImpLater fc tn dataTy
```
gallais, line 150: Should this move under the `when`?

src/Core/Context.idr
dunhamsteve: Yeah, I was trying to think of the which function did that and it was sitting there right in front of me. I'll take care of these after work.

dunhamsteve: But it turns out that `Core` is special and weird, so we need a `traverseOpt`:
```idris
  full gam (TCon t a ps ds u ms mcs det)
      = pure $ TCon t a ps ds u !(traverse (full gam) ms)
                                !(traverseOpt (traverse (full gam)) mcs) det
```

gallais:
Happy to merge. Should we get LSP up to speed first & merge on green?

dunhamsteve:
I've put a pr in at https://github.com/idris-community/idris2-lsp/pull/236 

That should fix the tests and unblock pack, but idris2-lsp also has a submodule `Idris2` that points to the revision that it works with. CI and pack ignore that submodule, but some users may may follow the instructions and use that. We can update that after the merge. Another option would be to point it directly to this PR.

mattpolzin:
I can confirm the PR up for the LSP project builds against this branch of Idris2. 

@dunhamsteve let's push a change to the Idris2 submodule in your LSP PR that points it at your branch so we can get that PR green in CI, merge it, and not have any commits of that project that don't build against the version of the compiler that the submodule points to. 

Then, as you suggest, we can fast follow the merge of this Idris2 PR with another LSP PR that switches back to the upstream compiler at latest `main`.

dunhamsteve:
@mattpolzin Sounds good, that PR is green now.  I pushed this branch to idris-lang/Idris2, so LSP would still be pointing at the official idris repository.

dunhamsteve:
I think pack is pinning the lsp to a specific commit, I didn't consider that bit.
```
[ build ] echo 'export idrisLSPVersion : ((Nat,Nat,Nat), String); idrisLSPVersion = ((0,1,0), "df322faeb")' >> src/Server/Generated.idr
```
I don't think we're going to get that test green before the merge.   Pack is going to pin to the old idris2-lsp and the old idris until it sees idris2-lsp#main and idris2#main that work together.

dunhamsteve:
Ok, I thought I saw a red build here, after the merge, but it's green now.

buzden:
> I don't think we're going to get that test green before the merge. Pack is going to pin to the old idris2-lsp and the old idris until it sees idris2-lsp#main and idris2#main that work together.

If you are talking about idris2-lsp's CI, we can implement [what we did for elab-utils](https://github.com/stefan-hoeck/idris2-elab-util/blob/ffbfe4f50900f45722035a16e6285f580047e846/.github/workflows/ci-lib.yml#L40): we run another CI job on the upstream's PR branch and all bleeding edge states of other libraries. The code is a bit clumsy and that's why I didn't suggest it anywhere, but if it's useful we still can do this

mattpolzin:
> If you are talking about idris2-lsp's CI, we can implement [what we did for elab-utils](https://github.com/stefan-hoeck/idris2-elab-util/blob/ffbfe4f50900f45722035a16e6285f580047e846/.github/workflows/ci-lib.yml#L40): we run another CI job on the upstream's PR branch and all bleeding edge states of other libraries. The code is a bit clumsy and that's why I didn't suggest it anywhere, but if it's useful we still can do this

This setup wouldn't be as useful for the LSP repo because you can't point the LSP at bleeding edge of the compiler if you're trying to merge a change to the LSP before the matching compiler change has merged. The other reason it would be less of a good fit in that repo is because it is expected that the Idris2 submodule will build the LS at any given point in time so we generally want to update that submodule anyway and once you do that the LSP repo's CI will go green with its current setup.

PR #3483, opened 2025-02-03
add a constructor to the Category interface

andrevidela:
# Description

Adding constructors to category interface

We should probably also move this outside contrib, maybe have other things that are categories to go along with it

      `CONTRIBUTORS.md`).


gallais (approved):

buzden:
> We should probably also move this outside contrib, maybe have other things that are categories to go along with it

I'm working on a proposition of a change in interfaces elaboration that would allow this interface and arrows interface to be more usable. I'm not suggesting waiting for it, but I definitely would for moving all these things out of contrib after my work is finished

PR #3484, opened 2025-02-04
[ new ] add --show-implicits flag

dunhamsteve:
# Description

Idris has command line flags `--show-machine-names` and `--show-namespaces`, but it is missing a flag for `--show-implicits`. This PR adds a `--show-implicits` flag.


buzden:
Is this practically needed only for `:t` command? I mean, we already have `:ti` for that. However, your change indeed looks in line with what's already present.

dunhamsteve:
Yeah, I should have included more motivation, and maybe a different test case. 

joelb had asked on discord how to get implicits to show in messages about missing cases. We have `:set showimplicits` in the repl, but that functionality isn't exposed on the command line. So you can't use that feature outside the repl. I believe it affects most error messages, but the missing cases message was what I had in mind.

mattpolzin (approved):

PR #3488, opened 2025-02-12
[ fix ] case splitting with multiple implicits

### LINKED ISSUE #3487: Case split on one implicit of a group omits variable name
If you write your implicit args as a group on the LHS, as in the example, below, then case split on one of them, it forgets to include implicit name in each case, which is invalid code

# Steps to Reproduce
Case split on `b` in 
```idris
foo : {a, b : Nat} -> () -> ()
foo {a, b} c = ?rhs
```

# Expected Behavior
```idris
foo : {a, b : Nat} -> () -> ()
foo {a, b = 0} c = ?rhs_0
foo {a, b = (S k)} c = ?rhs_1
```

# Observed Behavior
```idris
foo : {a, b : Nat} -> () -> ()
foo {a, 0} c = ?rhs_0
foo {a, (S k)} c = ?rhs_1
```

dunhamsteve:
# Description

This PR fixes #3487 where case splitting is broken for grouped implicits.  In addition to the case pointed out in the card, it also handles a similar issue with the initial implicit in a group. As a side effect, it fixes an issue where implicits that were not split were expanded to `{x = x}`.

Previously, case splitting on `b` in
```idris
foo : {a, b : Nat} -> () -> ()
foo {a, b} c = ?rhs
```
yields
```idris
foo : {a, b : Nat} -> () -> ()
foo {a, 0} c = ?rhs_0
foo {a, (S k)} c = ?rhs_1
```
and now it gives
```idris
foo : {a, b : Nat} -> () -> ()
foo {a, b = 0} c = ?rhs_0
foo {a, b = (S k)} c = ?rhs_1
```

### Details

Case splitting determines assignments for LHS variables and then walks through a tokenized version of the original line, substituting in the new values. It was looking for a name immediately surrounded by `{` and `}` to determine if it needed to use the `{a = Z}` syntax. This strategy failed for expressions like `{a,b}`.  We need to look for variables in braces, but not in `@{}` and also need to handle cases like `{a = Just b}`, `{a = (b,c)}`, or `{a = [b,c]}` where `b` and `c` are not implicit.

I changed the code to keep a stack of `Brace | Auto | Paren | Bracket` and insert the `x = ...` after `{` or after `,` if `Brace` is on top of the stack. The previous code was inserting `x =` even if the `x` wasn't split.  I had to adjust a few test cases, because that is now fixed.


dunhamsteve:
I'm transitioning this to draft. I want to address the cases discussed in #3490, which I have coded locally. But more importantly, I want to try to rework the case tree here, it's taking 49 seconds to check `CaseSplit.idr`.

We probably need something closer to parsing here, but I'm going to get it working with the existing approach before looking into how difficult that would be.

src/Idris/IDEMode/CaseSplit.idr
```diff
@@ -65,69 +65,75 @@ isWhitespace : SourcePart -> Bool
...
+  stk => pure (Other "]" :: !(doUpdates defs ups stk xs))
+doUpdates defs ups stk (Other "-" :: Other "-" :: xs) = pure (Other "--" :: xs)
+-- name after LBrace
+doUpdates defs ups stk (LBrace :: xs) =
+  let (ws, nws) = span isWhitespace xs in
+    case nws of
+      Name n :: rest => case lookup n ups of
+        Just up => pure (LBrace :: ws ++ Name n :: Whitespace " " :: Equal :: Whitespace " " :: Other up :: !(doUpdates defs ups stk rest))
+        Nothing => pure (LBrace :: !(doUpdates defs ups (Brace :: stk) xs))
+      _ => pure (LBrace :: !(doUpdates defs ups (Brace :: stk) xs))
```
gallais, line 104: Can we not still jump past the whitespace in these cases?
Same remark applies later on too.

src/Idris/IDEMode/CaseSplit.idr
```diff
@@ -65,69 +65,75 @@ isWhitespace : SourcePart -> Bool
...
-                  -- handle everything else as usual, preserving whitespace
-                  _ => pure (LBrace :: ws ++ Name n :: ws' ++
-                             !(doUpdates defs ups rest)
-                             )
-          -- not a special case: proceed as normal
-          _ => pure (LBrace :: [] ++ !(doUpdates defs ups xs))
+          Name n :: rest => case lookup n ups of
+            Nothing => pure (Other "," :: !(doUpdates defs ups stk xs))
+            Just up => case stk of
+              Brace :: _ => pure (Other "," :: ws ++ Name n :: Whitespace " " :: Equal :: Whitespace " " :: Other up :: !(doUpdates defs ups stk rest))
```
gallais, line 112: What if we have auto?

src/Idris/IDEMode/CaseSplit.idr
```diff
@@ -229,9 +240,15 @@ dropLast updChars with (snocList updChars)
 rtrim : String -> String
 rtrim = reverse . ltrim . reverse
 
+-- remove last paren and everything after it
+dropLastParen : SnocList Char -> SnocList Char
+dropLastParen Lin = Lin
+dropLastParen (cs :< ')') = cs
+dropLastParen (cs :< c) = dropLastParen cs :< c
```
gallais, line 247: Isn't this preserving everything after the last parens?
In which case you need to change the comment

tests/idris2/interactive/interactive048/Issue3487.idr
```diff
@@ -0,0 +1,18 @@
...
+three {a = Just x} c = ?rhs4
+
+four : { a : List Nat} -> () -> ()
+four {a = [x,y]} c = ?rhs5
+four {a = xs} c = ?rhs6
+
+five : Nat -> Nat
+five x = (case x of
+  y => ?rhs7 ) -- something y
+
```
gallais, line 18: Can we have test cases of the form `f {- this is f -} x = ?`?

src/Idris/IDEMode/CaseSplit.idr
dunhamsteve: That would hit the default case on line 113, since the `@{ ... }` notation doesn't allow named binding. (It looks like it doesn't allow `,` either, so we'd never hit this case with Auto on the top of the stack.)

tests/idris2/interactive/interactive048/Issue3487.idr
dunhamsteve: I left out that case, but I'll take care of it and add a test.

gallais (approved):

mattpolzin (approved):

PR #3489, opened 2025-02-12
[ refactor ] Use `List1` for determining parameters

spcfox:
# Description

The determining parameters are stored in `List`, but an empty list is never used. I suggest reflecting this in the signature with `List1`.

Also, the parser allows an empty list of determining parameters in interfaces, which is simply ignored during elaboration. I suggest making this a parser error:
```idris
interface Iface | where
```

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -1359,8 +1359,10 @@ dataOpt : OriginDesc -> Rule DataOpt
 dataOpt fname
     = (decorate fname Keyword (exactIdent "noHints") $> NoHints)
   <|> (decorate fname Keyword (exactIdent "uniqueSearch") $> UniqueSearch)
-  <|> (do decorate fname Keyword (exactIdent "search")
-          SearchBy <$> some (decorate fname Bound name))
+  <|> (do b   <- bounds $ decorate fname Keyword (exactIdent "search")
+          det <- mustWorkBecause b.bounds "Expected list of determining parameters" $
+                   some (decorate fname Bound name)
+          pure $ SearchBy det)
   <|> (decorate fname Keyword (exactIdent "external") $> External)
   <|> (decorate fname Keyword (exactIdent "noNewtype") $> NoNewtype)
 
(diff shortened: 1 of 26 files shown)


src/Idris/Parser.idr
```diff
@@ -1766,7 +1766,7 @@ parameters {auto fname : OriginDesc} {auto indents : IndentInfo}
             cons   <- constraints fname indents
             n      <- decorate fname Typ name
             params <- many ifaceParam
-            det    <- option [] $ decoratedSymbol fname "|" *> sepBy (decoratedSymbol fname ",") (decorate fname Bound name)
+            det    <- optional $ decoratedSymbol fname "|" *> sepBy1 (decoratedSymbol fname ",") (decorate fname Bound name)
```
andrevidela, line 1769: This is where we could detect that we need at least one argument right? Maybe we could use this to provide a helpful error message in case we get `Nothing` and say we need at least one parameter

spcfox: Done

src/Idris/Parser.idr
```diff
@@ -1359,8 +1359,10 @@ dataOpt : OriginDesc -> Rule DataOpt
 dataOpt fname
     = (decorate fname Keyword (exactIdent "noHints") $> NoHints)
   <|> (decorate fname Keyword (exactIdent "uniqueSearch") $> UniqueSearch)
-  <|> (do decorate fname Keyword (exactIdent "search")
-          SearchBy <$> forget <$> some (decorate fname Bound name))
+  <|> (do b   <- bounds $ decorate fname Keyword (exactIdent "search")
+          det <- mustWorkBecause b.bounds "Expected list of determining parameters" $
```
andrevidela, line 1363:
```suggestion
          det <- mustWorkBecause b.bounds "Expected at least one determining parameter" $
```

I think we can slightly improve the message, although I'm not thrilled with this change either, but due to the current error reporting framework we use I think that's the best we can do.

spcfox: I think a user is more likely to write something incorrect after `|` than leave the list blank. Therefore, I would leave the message more general.
```idris
Error: Expected list of determining parameters.

Test:1:17--1:18
 1 | interface Iface | "hello" where
                     ^
```

andrevidela: Oh I see, well so be it. One day someone's going to have to overhaul all those messages anyways

andrevidela:
looks good to me

PR #3494, opened 2025-02-18
[ base ] Add getThreadId

Matthew-Mosior:
# Description

This PR adds the `getThreadId` function to `System.Concurrency` (chez backend).

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -285,6 +285,8 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 
 * Added `System.Concurrency.channelGetWithTimeout` for the chez backend.
 
+* Added `System.Concurrency.getThreadId` for the chez backend.
+
 #### Contrib
 
 * `Data.List.Lazy` was moved from `contrib` to `base`.
--- a/libs/base/System/Concurrency.idr
+++ b/libs/base/System/Concurrency.idr
@@ -15,6 +15,8 @@ module System.Concurrency
 prim__setThreadData : {a : Type} -> a -> PrimIO ()
 %foreign "scheme:blodwen-get-thread-data"
 prim__getThreadData : (a : Type) -> PrimIO a
+%foreign "scheme:blodwen-get-thread-id"
+prim__getThreadId : PrimIO Int
 
 ||| Set the data stored in a thread's parameter to the given value.
 ||| Currently only supported under the scheme backends.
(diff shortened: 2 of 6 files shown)


Matthew-Mosior:
I think `channels008` is the issue here on the `windows-bootstrap-chez` workflow.

Matthew-Mosior:
Should be ready for review.

Matthew-Mosior:
> Should this function maybe rather return a value of type `PrimIO.ThreadID` (the same type as returned by `fork` when creating a new thread) instead of `Int`?

Hmm, it appears there isn't an implementation for `Show` on `ThreadID`, is this what we want here?  Thinking about use and testing here.

mattpolzin:
Yeah it looks like `ThreadId` is just a regrettable name for the opaque thing returned by `blodwen-thread` which is not related to the result of `get-thread-id`.

Why create a `blodwen-get-thread-id` instead of just using `get-thread-id` directly if there's no difference?

Matthew-Mosior:
> Yeah it looks like `ThreadId` is just a regrettable name for the opaque thing returned by `blodwen-thread` which is not related to the result of `get-thread-id`.
> 
> Why create a `blodwen-get-thread-id` instead of just using `get-thread-id` directly if there's no difference?

I guess I thought it would be good to create a `blodwen` for `get-thread-id`, but would be okay with just using `get-thread-id` directly.  Do you have a preference @mattpolzin?

dunhamsteve:
I think I would lean towards using `get-thread-id` directly, since the added blodwen function isn't doing anything in addition to calling `get-thread-id`. If this were implemented for `racket` too, and racket needed a different implementation, then I would use `blodwen-get-thread-id` and put `scheme:blodwen-get-thread-id`.

Matthew-Mosior:
> I think I would lean towards using `get-thread-id` directly, since the added blodwen function isn't doing anything in addition to calling `get-thread-id`. If this were implemented for `racket` too, and racket needed a different implementation, then I would use `blodwen-get-thread-id` and put `scheme:blodwen-get-thread-id`.

I like the idea of using `get-thread-id` directly. I believe I looked into this for Racket, but it doesn't have the idea of a numeric thread id, just an opaque thread object.

I'll try to rewrite this tonight to use `get-thread-id` directly.

Matthew-Mosior:
@mattpolzin @dunhamsteve 

I've updated the code to call `get-thread-id` directly, should be good to review again.

mattpolzin (approved):

Issue #1204, opened 2021-03-18
Mutual definitions trigger a linearity error


clayrat:
# Steps to Reproduce
```
data Tm = Foo Nat

mutual
  Env : Type
  Env = List Clos

  data Clos : Type where
    Cl : Tm -> Env -> Clos
```
# Expected Behavior

Typechecking succeeds.

# Observed Behavior
`Error: While processing right hand side of Env. Trying to use linear name Clos in non-linear context.`

Same error is triggered for a shorthand definition of `data Clos = Cl Tm Env`.

gallais:
Is this #586?

clayrat:
Yup, seems like. What's interesting is that an indexed version of the code doesn't trigger the error:

```
data Tm : List Unit -> Unit -> Type where
  Foo : Nat -> Tm [] ()

mutual
  Env : List Unit -> Type
  Env = All Clos

  data Clos : Unit -> Type where
    Cl : Tm g a -> Env g -> Clos a
```

PR #3495, opened 2025-02-19
[ fix ] Forward data declarations are linear

### LINKED ISSUE #586: Unexpected `use of linear name in non-linear context` for mutual declarations
# Steps to Reproduce

```idris
mutual
  public export
  ListT : (m : Type -> Type) -> (a : Type) -> Type 
  ListT m a = m (ListStruct m a) 
    
  public export
  data ListStruct : (m : Type -> Type) -> (a : Type) -> Type where
    Nil  : ListStruct m a
    (::) : a -> ListT m a -> ListStruct m a
```

# Expected Behavior
Type-check fine.

# Observed Behavior

```
     While processing right hand side of ListT. Trying to
     use linear name Control.Monad.List.ListStruct in non-linear context.
     
     Control/Monad/List.idr:9:18--9:32
        |
      9 |   ListT m a = m (ListStruct m a) 
        |                  ^^^^^^^^^^^^^^
     
```

This behaviour is unexpected especially as replacing the order of definitions in the mutual block type-checks fine:
```idris
mutual
  public export
  data ListStruct : (m : Type -> Type) -> (a : Type) -> Type where
    Nil  : ListStruct m a
    (::) : a -> ListT m a -> ListStruct m a

  public export
  ListT : (m : Type -> Type) -> (a : Type) -> Type 
  ListT m a = m (ListStruct m a) 
```

dunhamsteve:
# Description

This fixes #586 and #1204 (included as tests).  The occurs because forward data declarations are tagged as `linear`, I have changed it to `top`. I don't know why they were tagged as `linear` in the first place, so there is a chesterton's fence situation here, but if #586 is a bug then they probably should not be. If they were marked linear to distinguish forward declarations from empty types, that is now handled by #3480.

Previously, the following code would error with "Trying to use linear name Main.X in non-linear context." It now checks.
```idris
data X : Type

f : Type
f = Maybe X
```

### DIFF
--- a/src/TTImp/ProcessData.idr
+++ b/src/TTImp/ProcessData.idr
@@ -425,7 +425,7 @@ processData {vars} eopts nest env fc def_vis mbtot (MkImpLater dfc n_in ty_raw)
          arity <- getArity defs [] fullty
 
          -- Add the type constructor as a placeholder
-         tidx <- addDef n (newDef fc n linear vars fullty def_vis
+         tidx <- addDef n (newDef fc n top vars fullty def_vis
                           (TCon 0 arity [] [] defaultFlags [] Nothing Nothing))
          addMutData (Resolved tidx)
          defs <- get Ctxt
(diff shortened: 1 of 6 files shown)


gallais (approved):

gallais:
Nice!

PR #3496, opened 2025-02-20
[ cleanup ] Remove redundant `Ref` arguments

spcfox:
# Description

1. Removed `showInfo`, which simply returns `False`
2. Non mutual functions moved out of the mutual block in `LambdaLift.idr`
3. Label arguments made erased
4. Removed redundant `Ref` arguments

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Compiler/CompileExpr.idr
+++ b/src/Compiler/CompileExpr.idr
@@ -162,12 +162,10 @@ dconFlag n
 
 toCExpTm : {vars : _} ->
            {auto c : Ref Ctxt Defs} ->
-           {auto s : Ref NextMN Int} ->
            Name -> Term vars ->
            Core (CExp vars)
 toCExp : {vars : _} ->
          {auto c : Ref Ctxt Defs} ->
-         {auto s : Ref NextMN Int} ->
          Name -> Term vars ->
          Core (CExp vars)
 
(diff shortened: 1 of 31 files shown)


mattpolzin (approved):

spcfox:
I did it by hand. I had the an idea to make `ReadOnlyRef` since many functions read from `Ref`, but do not write to it. At the same time I found a lot of unused references.

After that I came to the conclusion that the idea of `ReadOnlyRef` is bad, because `Ref` can be contained in a closure and can still change.

For example, these functions are not equivalent to
``idris
f : Ref Ctxt Defs => Core () -> Core ()
f action = do
  defs <- get Ctxt
  log “elab” 10 “Do action”
  action
  put Ctxt defs

f' : ReadOnlyRef Ctxt defs => Core () -> Core ()
f' action = do
  log “elab” 10 “Do action”
  action
```
since `action` can have the effect of changing any reference.

An example of such a function in the compiler:
https://github.com/idris-lang/Idris2/blob/main/src/TTImp/Elab/Check.idr#L537

I dropped this idea because it can be confusing, giving the false impression that a function cannot change a reference. If you're interested, here's a commit with `ReadOnlyRef`:
https://github.com/spcfox/Idris2/commit/a480648315923fca051609c2c4d9c41f788171ce

spcfox:
I just went through the files in compilation order, and if the need for some `Ref` wasn't obvious, I tried making it `ReadOnly`. And if it worked, I checked to see if it was used at all.

spcfox:
`chez/channels008` still flaky

Matthew-Mosior:
> `chez/channels008` still flaky

Hmm I just bumped that test to use a timeout of `15000` instead of `5000`.  Wonder if it need to be increased again?

PR #3499, opened 2025-02-21
[new] Support literate Djot files

nmccarty:
[Djot](https://djot.net/) is a markup format that is derived from CommonMark, it's code block syntax is identical to CommonMark's, so we can support it as a CommonMark style literate formate by just adding the file extension.

# Description

      `CONTRIBUTORS.md`).


buzden:
Could you, please, add a test for this case?

nmccarty:
Added

buzden (approved):

PR #3504, opened 2025-02-28
[ ci ] add 45 minute timeout to jobs running tests

dunhamsteve:
# Description

We're having some CI runs freezing up during tests and being killed after five hours. This has occurred a few times in `macos-self-host-chez` and in `windows-bootstrap-chez`.  I don't have a root cause, but suspect a bug in chez or possibly the test runner is causing a deadlock. I see that there is some history of deadlocks on these platforms in the chez bug tracker.

As a workaround, I've added a 45 minute timeout to the jobs that run tests. The typically run in about half that time.  They will still fail when there is a deadlock, but will fail in 45 minutes instead of 5 hours.

### DIFF
--- a/.github/workflows/ci-idris2-and-libs.yml
+++ b/.github/workflows/ci-idris2-and-libs.yml
@@ -197,6 +197,7 @@ jobs:
   windows-bootstrap-chez:
     needs: [initialise, quick-check]
     runs-on: windows-latest
+    timeout-minutes: 45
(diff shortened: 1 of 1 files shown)


zenntenn:
I do know that timeouts will just happen sometimes on these agents, even if
the task itself doesn't lock out. So this timeout is a very good idea even
if all the Idris related bugs are fixed

PR #3506, opened 2025-03-08
Make `unrestricted` linear in the `(!*)`

joelberkeley:
# Description

It seems only natural for `unrestricted` to be linear in  `(!*)` .

      `CONTRIBUTORS.md`).

### DIFF
--- a/libs/linear/Data/Linear/Notation.idr
+++ b/libs/linear/Data/Linear/Notation.idr
@@ -21,6 +21,10 @@ public export
 export prefix 5 !*
 ||| Prefix notation for the linear unrestricted modality
 public export
-record (!*) (a : Type) where
-  constructor MkBang
-  unrestricted : a
+data (!*) : Type -> Type where
+  MkBang : (unrestricted : a) -> MkBang a
+
+||| Unpack an unrestricted value in a linear context
+public export
+unrestricted : !* a -@ a
+unrestricted (MkBang unr) = unr
--- a/libs/linear/Data/Linear/Notation.idr
+++ b/libs/linear/Data/Linear/Notation.idr
@@ -28,3 +28,10 @@ data (!*) : Type -> Type where
 public export
 unrestricted : !* a -@ a
 unrestricted (MkBang unr) = unr
+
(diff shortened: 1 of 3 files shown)


libs/linear/Data/Linear/Notation.idr
```diff
@@ -21,6 +21,10 @@ public export
 export prefix 5 !*
 ||| Prefix notation for the linear unrestricted modality
 public export
-record (!*) (a : Type) where
-  constructor MkBang
-  unrestricted : a
+data (!*) : Type -> Type where
+  MkBang : (unrestricted : a) -> MkBang a
```
joelberkeley, line 25: named for backwards compatibility

libs/linear/Data/Linear/Notation.idr
```diff
@@ -21,6 +21,10 @@ public export
...
-record (!*) (a : Type) where
-  constructor MkBang
-  unrestricted : a
+data (!*) : Type -> Type where
+  MkBang : a -> !* a
+
+||| Unpack an unrestricted value in a linear context
+public export
+unrestricted : !* a -@ a
+unrestricted (MkBang unr) = unr
```
joelberkeley, line 30: I've noticed we lose `(.unrestricted)`. Not sure what's the most favourable way to handle that.

andrevidela: yes, for this to be completely equivalent we need both `unrestricted` and `(.unrestricted)` might as well add both

joelberkeley: That would be backwards compatible, but it doesn't feel in the spirit of the stdlib

dunhamsteve: Would this be a precedent? (I searched for `\(\.\w`)

https://github.com/idris-lang/Idris2/blob/1678b4676f619bc28bf6eea25b789d7cf7f2f9d6/libs/base/Data/Singleton.idr#L13-L19

andrevidela:
Looks good

PR #3516, opened 2025-03-19
[ fix ] ensure scrutinee of case block is optimised

### LINKED ISSUE #3515: Miscompilation of `Nat` literals
```idris
toNat : Nat -> () -> Nat
toNat 0 () = 1
toNat _ _ = 1
```

# Steps to Reproduce

Load this file in `idris2`, then run `:di toNat`

# Expected Behavior

```
Compiled: \ {arg:0}, {arg:1} => case {arg:0} of
  { 0 => 1
  ; _ => 1
  }
```

or

```
Compiled: \ {arg:0}, {arg:1} => case {arg:0} of
  { 0 => 1 + 0
  ; _ => 1
  }
```

or similar

# Observed Behavior

```
Compiled: \ {arg:0}, {arg:1} => case {arg:0} of
  { 0 => Prelude.Types.S {tag = 1} [succ] 0
  ; _ => 1
  }
```

This produces the wrong thing, eg on chez you get `(vector 1 0)` instead of `1`

Z-snails:
Fortunately, at this point in compilation, scrutinees to case blocks are always variables that don't need to be optimised.
This assumption may be false in future, so I'm fixing it now.

~This doesn't fix #3515 unfortunately :(~
It does now, thanks to gallais (fixes #3515)

# Description

Probably not, this is not fixing a user facing issue and shouldn't change any behaviour.


gallais:
Should we have
```diff
    rewriteCExp exp = do
-        exp' <- try exp
-        rewriteSub exp'
+        exp' <- rewriteSub exp
+        try exp'
```
instead so that if `rewriteSub exp` triggers some simplification, `try` has then a chance to exploit them?

dunhamsteve:
It looks like gallais' proposal fixes the issue. I believe because `unitTree` is lifting a sub term up to the top, which is then ignored by `rewriteSub`.

PR #3531, opened 2025-04-11
Make the property argument to `replace` explicit

joelberkeley:
# Description

`replace` can almost never infer the property `p`, so I've made it explicit, after [@gallais ' agreement](https://discord.com/channels/827106007712661524/834611018775789568/1360384552421429351)

      `CONTRIBUTORS.md`).

### DIFF
--- a/docs/source/proofs/patterns.rst
+++ b/docs/source/proofs/patterns.rst
@@ -180,16 +180,16 @@ prelude:
 .. code-block:: idris
 
     Main> :t replace
-    Builtin.replace : (0 rule : x = y) -> p x -> p y
+    Builtin.replace : (0 p : _) -> (0 rule : x = y) -> p x -> p y
 
 Given a proof that ``x = y``, and a property ``p`` which holds for
 ``x``, we can get a proof of the same property for ``y``, because we
 know ``x`` and ``y`` must be the same. Note the multiplicity on ``rule``
 means that it's guaranteed to be erased at run time.
-In practice, this function can be
-a little tricky to use because in general the implicit argument ``p``
-can be hard to infer by unification, so Idris provides a high level
-syntax which calculates the property and applies ``replace``:
+In practice, this function can be a little tricky to use because in
+general the argument ``p`` can be hard to infer by unification, so
+Idris provides a high level syntax which calculates the property
+and applies ``replace``:
 
 .. code-block:: idris
 
--- a/tests/idris2/data/record021/Issue3501.idr
+++ b/tests/idris2/data/record021/Issue3501.idr
@@ -2,6 +2,6 @@ import Data.Vect
 record Foo (th : Vect n a) where
   nIsZero     : n === 0
   vectIsEmpty : (th ===)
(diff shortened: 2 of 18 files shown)


docs/source/proofs/patterns.rst
```diff
@@ -180,16 +180,16 @@ prelude:
...
 know ``x`` and ``y`` must be the same. Note the multiplicity on ``rule``
 means that it's guaranteed to be erased at run time.
-In practice, this function can be
-a little tricky to use because in general the implicit argument ``p``
-can be hard to infer by unification, so Idris provides a high level
-syntax which calculates the property and applies ``replace``:
+In practice, this function can be a little tricky to use because in
+general the argument ``p`` can be hard to infer by unification, so
+Idris provides a high level syntax which calculates the property
+and applies ``replace``:
```
joelberkeley, line 192: I don't quite follow the logic here, but I've kept it as similar as possible in wording. Do edit it if you know a better wording (also in the other file which duplicates this)

tests/idris2/data/record021/Issue3501.idr
```diff
@@ -2,6 +2,6 @@ import Data.Vect
 record Foo (th : Vect n a) where
   nIsZero     : n === 0
   vectIsEmpty : (th ===)
-                 $ replace {p = \ n => Vect n a} (sym nIsZero)
+                 $ replace (\ n => Vect n a) (sym nIsZero)
```
joelberkeley, line 5: not sure if I've broken this test, or what it tests

tests/idris2/error/perror021/Implicit.idr
joelberkeley, line 1: deleting this test is likely not the correct action, but I'm unclear what it's testing so waiting for someone to enlighten me

joelberkeley: perhaps that a space is expected in `{p=\n` ...

dunhamsteve: The original commit is #2781 (git has a hard time telling you because the files were moved around). Maybe we could find an example that doesn't use `replace`. Similar issue, I think it was no space after the `=`, but the error message itself should point to the correct location rather than saying "couldn't parse declaration".

docs/source/proofs/patterns.rst
dunhamsteve: I don't think there is any "infer by unification" here anymore, since `p` was implicit and this PR makes it explicit. Maybe something along the lines of "the argument `p` is cumbersome to write, so Idris provides ....".

In practice, I use `replace` when Idris can't figure out `p` in `rewrite`, especially in the case where I only want to replace a subset of the occurrences of a term. But others who have more experience with using Idris for proofs than I do.  I've always had to specify `p` when I use `replace`.

gallais:
I'd have expected this to be backwards compatible for all of the cases
where `p` is explicitly spelt out.

Not sure why named application is not successful here.

joelberkeley:
I'm not going to get round to this, so will close. Reopen if you want

PR #3553, opened 2025-05-14
Change List to List1 in ValidatedL doc comment

jmanuel1:
# Description

While I was using ValidatedL, I was confused for a few minutes about why it wouldn't work with a List and why List1 was being pulled in. That is, until I looked at the definition of ValidatedL. Hopefully this saves others a bit of their future time.

      `CONTRIBUTORS.md`).


andrevidela:
Looks good, thanks!

PR #3554, opened 2025-05-17
[ refactor ] Make `Binary` argument auto in `TTC`

spcfox:
# Description

In Yaffle, `Ref Bin Binary` is passed implicitly, reducing boilerplate slightly. I suggest adopting the same approach in Idris.

      `CONTRIBUTORS.md`).


andrevidela (approved):
Thank goodness for that it was long overdue

PR #3558, opened 2025-05-21
[ base ] Add default definitions for `zipWith3` and `unzipWith3`

spcfox:
# Description

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -304,6 +304,9 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 
 * `unrestricted`, for unpacking a `!* a`, now uses its argument once
 
+* Added default definitions for `zipWith3` and `unzipWith3` in `Zippable`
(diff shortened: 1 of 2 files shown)


andrevidela (approved):
Looks good

tests/idris2/basic/basic044/expected
```diff
@@ -117,33 +117,33 @@ Target type : ({arg:1} : (Data.Fin.Fin (Prelude.Types.S (Prelude.Types.S Prelude
 LOG elab.ambiguous:5: Ambiguous elaboration (kept 2 out of 2 candidates) (not delayed) at Vec:3:
 $resolved6
 $resolved7
-Target type : ?Vec.{a:4569}_[]
+Target type : ?Vec.{a:4704}_[]
```
gallais, line 120: In testutils.sh we have tools to get rid of these machine names which are changing for no good reason
when the libs are modified. Please use that rather than updating the golden file

PR #3559, opened 2025-05-22
[ base ] Reorganise, slightly extend and harmonise `Data.*.Quantities`

buzden:
# Description

Similar modules for different differ greatly in their fullness, I suggest harmonise them a bit. I didn't manage to harmonise them completely (e.g., there are no `length` and `index'` functions for several of those functions, thus functions that use them in `Quantifiers` are not present in every such module). Also, I propose to move couple of functions related only to `All` to the existing namespace `All`. Also, I suggest couple of new functions I found missing but very useful.

      `CONTRIBUTORS.md`).


libs/base/Data/List1/Quantifiers.idr
```diff
@@ -23,6 +23,11 @@ namespace Any
     ||| A proof that the satisfying element is in the tail of the `List1`
     There : {0 xs : List a} -> Any p xs -> Any p (x ::: xs)
 
+  export
+  {0 p : a -> Type} -> Uninhabited (p x) => Uninhabited (Any p xs) => Uninhabited (Any p $ x:::xs) where
+    uninhabited (Here y) = uninhabited y
+    uninhabited (There y) = uninhabited y
+
```
stefan-hoeck, line 30: Could this not be done via `All (Uninhabited . p) xs => Unihabited (Any p xs)`?

buzden: This definitely could be. I just copied this way from `List.Quantifier`. I would agree that the way using `All` seems to be slighly more universal, but then we need to change it in all modules (I'm ready to do this).

PR #3565, opened 2025-06-03
[ base ] Add a function connecting `All` and `Any` through `Alternative`

buzden:
# Description

      `CONTRIBUTORS.md`).

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -311,6 +311,8 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
   are harmonised among each other. Also, several existing functions related only to
   `All` were moved to appropriate namespace. Couple new functions for `Any` were added.
 
+* Add a function `altAll` connecting `All` to `Any` using `Alternative` to all `Quantifiers` modules.
+
 #### Contrib
 
 * `Data.Vect.Views.Extra` was moved from `contrib` to `base`.
--- a/libs/base/Data/List/Quantifiers.idr
+++ b/libs/base/Data/List/Quantifiers.idr
@@ -311,6 +311,11 @@ allAnies : All p xs -> List (Any p xs)
 allAnies [] = []
 allAnies (x::xs) = Here x :: map There (allAnies xs)
 
+export
+altAll : Alternative p => All (p . q) xs -> p $ Any q xs
+altAll []      = empty
+altAll (a::as) = Here <$> a <|> There <$> altAll as
+
 ------------------------------------------------------------------------
 -- Partitioning lists according to All
 
--- a/libs/base/Data/List1/Quantifiers.idr
+++ b/libs/base/Data/List1/Quantifiers.idr
@@ -211,6 +211,10 @@ public export
 allAnies : {0 xs : List1 a} -> All p xs -> List1 (Any p xs)
 allAnies (x:::xs) = Here x ::: map There (allAnies xs)
 
+export
+altAll : {0 xs : List1 a} -> Alternative p => All (p . q) xs -> p $ Any q xs
+altAll (a:::as) = Here <$> a <|> There <$> altAll as
+
 {-
 ------------------------------------------------------------------------
 -- Partitioning lists according to All
--- a/libs/base/Data/SnocList/Quantifiers.idr
+++ b/libs/base/Data/SnocList/Quantifiers.idr
@@ -276,6 +276,11 @@ allAnies : All p xs -> SnocList (Any p xs)
 allAnies [<] = [<]
 allAnies (xs:<x) = map There (allAnies xs) :< Here x
 
+export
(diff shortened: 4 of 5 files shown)


stefan-hoeck:
Sorry for chiming in here only after this was merged (but then, it was open for only four hours), but I think everything that goes into the standard libraries should be documented with doc strings. We keep talking about how documentation is important, but somehow I feel like nobody really cares when it comes to new pull requests.

Also, I'm a bit surprised that this was self-merged without being properly reviewed. If this is OK, then probably someone should change the contribution guidelines.

mattpolzin:
Chiming in here as a co-maintainer who thinks good points were raised. There's been an implicit acceptance of self-merged PRs given the tendency for PRs to get reviewed or not based on a combination of time and interest from other maintainers but I had forgotten there was explicit guidance on it in the contribution guidelines. Within the group that is allowed to merge PRs, the only rule to rule them all has been "use your best judgement" but again I do think it would be good for the documentation of merge processes to be in agreement with what is actually done. 

Personally I think self-merging is ok but not ideal and therefore if we do update the guidelines to make it explicitly ok instead of just "something that happens" then I'd vote for a policy of a minimum review window, something like "PRs will not be merged without review unless a minimum of 5 days has been given for the community to offer feedback. If affirmative code review is given faster than that, however, PRs do not need to be left open for 5 or more days just to solicit additional hypothetical feedback."

On the other meta point, I'm all for code documentation!

buzden:
> I think everything that goes into the standard libraries should be documented with doc strings. We keep talking about how documentation is important, but somehow I feel like nobody really cares when it comes to new pull requests.

I fully agree with this idea and I indeed feel guilty in not "really caring" about this. I'd suggest to change the PR text template to add a tick for "I added documentation for newly added/modified functions and data types" near to one mentioning `CHANGELOG_NEXT.md`, because existing tick is the only thing that really reminds me each time about non-pleasant but important thing.

> I'm a bit surprised that this was self-merged without being properly reviewed

> There's been an implicit acceptance of self-merged PRs given the tendency for PRs to get reviewed or not based on a combination of time and interest from other maintainers

I also was thinking about this before hitting the "merge" button, but I've seen self-merged PRs so many times, and I thought that this change is so small, that I decided it to be not so big deal (I was obviously wrong).

> I'd vote for a policy of a minimum review window, something like "PRs will not be merged without review unless a minimum of 5 days has been given for the community to offer feedback. If affirmative code review is given faster than that, however, PRs do not need to be left open for 5 or more days just to solicit additional hypothetical feedback."

I'd love this type of thing to be pronounced clearly and surely I would always follow this if it gets accepted. But I'd like some kind of mention of hotfixes (with some understanding if it), because I participated in some of them in the past, and I feel that something like fixing CI because of missed unintended breakage should be accepted without much waiting.

mattpolzin:
We can and should handle policy change elsewhere of course (probably an Issue) but I'm glad it was raised here and there's some initial high level agreement.

mattpolzin:
I was going to spin up an issue for this and I realized I am not spotting the part of the CONTRIBUTING document that touches on this. @stefan-hoeck can you point me to any prior documentation you've seen on requirements for a PR to be merged so I can reference them in my ticket?

stefan-hoeck:
The parts I was mainly thinking about when I saw this are:

> Many contributions will require accompanying tests and documentation updates. Bugfixes in particular should be accompanied by tests, to avoid future regressions.

> In all cases, a pull request must have a short description (one sentence is usually enough) that explains its purpose. However obvious you think it might be, it really helps when reviewing the changes.

> Changes to prelude and base libraries
'prelude' and 'base' are, in some sense, part of the language. There are a lot of trade offs to be made in the design of the prelude especially - such as interface hierarchies - and while you may not agree with the way it looks, by and large these decisions have already been made so there must be a compelling reason for them to be changed.

For me, the last point is the most important one because it often affects me directly: Changes to the standard libraries are often breaking changes and require additional work when maintaining the pack collection. #3559 , for instance, broke quantifiers-extra and potentially many of its dependencies. Fortunately, the fix was a simple one in this case. Other examples come to mind, where the fix was more time consuming. These things quickly lead to one or several hours of work on my end. And no, I'm neither opposed to these changes in general nor am I opposed to #3559 being self-merged, because it has been open long enough without anybody (including my lazy self) chiming in.

I think it is a very good idea to add a reminder for adding documentation (plus - if appropriate - tests!) as suggested by @buzden. I'd also kindly like to ask people with merge rights to consider that potentially breaking changes can lead to a non-trivial amount of work for package maintainers, so they should be treated with the necessary care.

andrevidela:
One thing I ought to finish is [this branch](https://github.com/andrevidela/Idris2/tree/ci-test-collection) that adds a CI step to compile the pack collection with the newest changes. This would have caught the breakage introduced by #3559.

mattpolzin:
This is what I put out there for further discussion on the subjects above: https://github.com/idris-lang/Idris2/issues/3566

PR #3567, opened 2025-06-10
[ fix ] Remove type constructor tags

spcfox:
# Description

Currently, the type constructor tag is always 100. In fact, they are not used. Therefore, I propose to remove them and also fix several related bugs.

Commits:
1. Fixed a bug that prevented the tag counter from increasing
2. Fixed irrefutable pattern with type in index
3. Fixed checking for emptiness of type with type in index. This is the only place where type constructors are compared by tags. But they are unique only within the module. Therefore, type constructors should be compared by name, as in all other places.
4. Type constructor tags have been removed because they are not suitable for comparison and are no longer used.
5. Remove some boilerplate.

      `CONTRIBUTORS.md`).

### DIFF
--- a/libs/base/Language/Reflection/TT.idr
+++ b/libs/base/Language/Reflection/TT.idr
@@ -108,7 +108,8 @@ data NameType : Type where
      Bound   : NameType
      Func    : NameType
      DataCon : (tag : Int) -> (arity : Nat) -> NameType
-     TyCon   : (arity : Nat) -> NameType
+     TyCon   : (tag : Int) -> (arity : Nat) -> NameType
+               -- TODO: remove type tag, it's is always 0
 
 %name NameType nty
 
--- a/libs/base/Language/Reflection/TT.idr
+++ b/libs/base/Language/Reflection/TT.idr
@@ -108,7 +108,7 @@ data NameType : Type where
      Bound   : NameType
      Func    : NameType
      DataCon : (tag : Int) -> (arity : Nat) -> NameType
-     TyCon   : (tag : Int) -> (arity : Nat) -> NameType
+     TyCon   : (arity : Nat) -> NameType
 
 %name NameType nty
 
--- a/src/Core/Metadata.idr
+++ b/src/Core/Metadata.idr
@@ -31,7 +31,7 @@ nameDecoration nm nt
   nameTypeDecoration Bound         = Bound
   nameTypeDecoration Func          = Function
   nameTypeDecoration (DataCon _ _) = Data
-  nameTypeDecoration (TyCon _ _  ) = Typ
+  nameTypeDecoration (TyCon _)     = Typ
 
 
 public export
--- a/src/Core/Metadata.idr
+++ b/src/Core/Metadata.idr
@@ -28,10 +28,10 @@ nameDecoration nm nt
   where
 
   nameTypeDecoration : NameType -> Decoration
-  nameTypeDecoration Bound         = Bound
-  nameTypeDecoration Func          = Function
-  nameTypeDecoration (DataCon _ _) = Data
-  nameTypeDecoration (TyCon _)     = Typ
+  nameTypeDecoration Bound        = Bound
+  nameTypeDecoration Func         = Function
+  nameTypeDecoration (DataCon {}) = Data
+  nameTypeDecoration (TyCon {})   = Typ
 
 
 public export
--- a/src/Core/Normalise/Eval.idr
+++ b/src/Core/Normalise/Eval.idr
@@ -182,8 +182,8 @@ parameters (defs : Defs) (topopts : EvalOpts)
         = evalMeta env fc n i args (args' ++ stk)
     applyToStack env (NDCon fc n t a args) stk
         = pure $ NDCon fc n t a (args ++ stk)
-    applyToStack env (NTCon fc n t a args) stk
-        = pure $ NTCon fc n t a (args ++ stk)
+    applyToStack env (NTCon fc n a args) stk
+        = pure $ NTCon fc n a (args ++ stk)
     applyToStack env (NAs fc s p t) stk
        = if removeAs topopts
             then applyToStack env t stk
@@ -278,10 +278,10 @@ parameters (defs : Defs) (topopts : EvalOpts)
         = do -- logC "eval.ref.data" 50 $ do fn' <- toFullNames fn -- Can't use ! here, it gets lifted too far
              --                             pure $ "Found data constructor: " ++ show fn'
              pure $ NDCon fc fn tag arity stk
-    evalRef env meta fc (TyCon tag arity) fn stk def
+    evalRef env meta fc (TyCon arity) fn stk def
         = do -- logC "eval.ref.type" 50 $ do fn' <- toFullNames fn
              --                             pure $ "Found type constructor: " ++ show fn'
-             pure $ ntCon fc fn tag arity stk
+             pure $ ntCon fc fn arity stk
     evalRef env meta fc Bound fn stk def
         = do -- logC "eval.ref.bound" 50 $ do fn' <- toFullNames fn
              --                              pure $ "Found bound variable: " ++ show fn'
@@ -358,7 +358,7 @@ parameters (defs : Defs) (topopts : EvalOpts)
               then evalConAlt env loc opts fc stk args (map snd args') sc
               else pure NoMatch
     -- Type constructor matching, in typecase
-    tryAlt {more} env loc opts fc stk (NTCon _ nm tag' arity args') (ConCase nm' tag args sc)
+    tryAlt {more} env loc opts fc stk (NTCon _ nm arity args') (ConCase nm' tag args sc)
          = if nm == nm'
               then evalConAlt env loc opts fc stk args (map snd args') sc
               else pure NoMatch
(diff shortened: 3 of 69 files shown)


spcfox:
Also, tags removed in Yaffle
https://github.com/edwinb/Yaffle/commit/f26c9a0e8463e04730bf1af63410e9c6ff8290e3#diff-ba25068d28a398c3bade02a547922670d11ca8845d0e54a9ed9b180e9d76b702

libs/base/Language/Reflection/TT.idr
```diff
@@ -108,7 +108,7 @@ data NameType : Type where
      Bound   : NameType
      Func    : NameType
      DataCon : (tag : Int) -> (arity : Nat) -> NameType
-     TyCon   : (tag : Int) -> (arity : Nat) -> NameType
```
andrevidela, line 111: Given how work intensive it is to make such changes to elab-reflection, what do you think of reverting that and keeping the dummy tag around until we collect more changes and update elab reflection later?

spcfox: This makes sense

src/Core/Termination.idr
```diff
@@ -113,7 +113,7 @@ checkTotal loc n_in
               Unchecked => do
                   mgdef <- lookupCtxtExact n (gamma defs)
                   case definition <$> mgdef of
-                       Just (TCon _ _ _ _ _ _ _ _)
+                       Just (TCon _ _ _ _ _ _ _)
```
andrevidela: BTW for those cases we can use `TCon {}` such that we dont need to change this line in the future if we change this constructor

src/TTImp/ProcessDecls/Totality.idr
```diff
@@ -24,7 +24,7 @@ checkTotalityOK n
          -- #524: need to check positivity even if we're not in a total context
          -- because a definition coming later may need to know it is positive
          case definition gdef of
-           (TCon _ _ _ _ _ _ _ _) => ignore $ checkPositive fc n
+           (TCon _ _ _ _ _ _ _) => ignore $ checkPositive fc n
```
andrevidela: Same here

src/TTImp/ProcessDef.idr
```diff
@@ -68,16 +68,16 @@ mutual
   mismatchNF defs (NDCon _ _ _ _ _) (NPrimVal _ _) = pure True
   mismatchNF defs (NPrimVal _ _) (NBind _ _ _ _) = pure True
   mismatchNF defs (NBind _ _ _ _) (NPrimVal _ _) = pure True
-  mismatchNF defs (NPrimVal _ _) (NTCon _ _ _ _ _) = pure True
-  mismatchNF defs (NTCon _ _ _ _ _) (NPrimVal _ _) = pure True
```
andrevidela, line 72: Here

src/TTImp/Elab/Check.idr
```diff
@@ -318,7 +318,7 @@ concrete defs env (NBind fc _ (Pi _ _ _ _) sc)
     = do sc' <- sc defs (toClosure defaultOpts env (Erased fc Placeholder))
          concrete defs env sc'
 concrete defs env (NDCon _ _ _ _ _) = pure True
-concrete defs env (NTCon _ _ _ _ _) = pure True
+concrete defs env (NTCon _ _ _ _) = pure True
```
andrevidela: Here

src/TTImp/Elab/Ambiguity.idr
```diff
@@ -125,7 +125,7 @@ expandAmbigName mode nest env orig args (IVar fc x) exp
     wrapDot : Bool -> EState vars ->
               ElabMode -> Name -> List RawImp -> Def -> RawImp -> RawImp
     wrapDot _ _ _ _ _ (DCon _ _ _) tm = tm
-    wrapDot _ _ _ _ _ (TCon _ _ _ _ _ _ _ _) tm = tm
+    wrapDot _ _ _ _ _ (TCon _ _ _ _ _ _ _) tm = tm
```
andrevidela: Here

src/Core/Termination/CallGraph.idr
```diff
@@ -213,8 +213,8 @@ mutual
     let (f, args) = getFnArgs t in
     let (g, args') = getFnArgs s in
     case f of
-      Ref _ (TyCon _ _) cn => case g of
-        Ref _ (TyCon _ _) cn' => if cn == cn'
+      Ref _ (TyCon _) cn => case g of
+        Ref _ (TyCon _) cn' => if cn == cn'
```
andrevidela: Here

src/Core/Primitives.idr
```diff
@@ -499,7 +499,7 @@ doubleCeiling = doubleOp ceiling
 -- Only reduce for concrete values
 believeMe : Vect 3 (NF vars) -> Maybe (NF vars)
 believeMe [_, _, val@(NDCon _ _ _ _ _)] = Just val
-believeMe [_, _, val@(NTCon _ _ _ _ _)] = Just val
+believeMe [_, _, val@(NTCon _ _ _ _)] = Just val
```
andrevidela: Here

src/Core/Normalise/Eval.idr
```diff
@@ -399,7 +399,7 @@ parameters (defs : Defs) (topopts : EvalOpts)
       where
         concrete : NF free -> Bool
         concrete (NDCon _ _ _ _ _) = True
-        concrete (NTCon _ _ _ _ _) = True
+        concrete (NTCon _ _ _ _) = True
```
andrevidela: Here

src/Core/Metadata.idr
```diff
@@ -31,7 +31,7 @@ nameDecoration nm nt
   nameTypeDecoration Bound         = Bound
   nameTypeDecoration Func          = Function
   nameTypeDecoration (DataCon _ _) = Data
-  nameTypeDecoration (TyCon _ _  ) = Typ
+  nameTypeDecoration (TyCon _)     = Typ
```
andrevidela: Here

src/TTImp/ProcessDef.idr
```diff
@@ -160,22 +160,22 @@ recoverable : {auto c : Ref Ctxt Defs} ->
...
 -- Type constructor vs. binder
-recoverable defs (NTCon _ _ _ _ _) (NBind _ _ _ _) = pure False
-recoverable defs (NBind _ _ _ _) (NTCon _ _ _ _ _) = pure False
+recoverable defs (NTCon _ _ _ _) (NBind _ _ _ _) = pure False
+recoverable defs (NBind _ _ _ _) (NTCon _ _ _ _) = pure False
 
-recoverable defs (NTCon _ _ _ _ _) _ = pure True
-recoverable defs _ (NTCon _ _ _ _ _) = pure True
+recoverable defs (NTCon _ _ _ _) _ = pure True
+recoverable defs _ (NTCon _ _ _ _) = pure True
```
andrevidela: Here

src/TTImp/ProcessDef.idr
```diff
@@ -114,16 +114,16 @@ impossibleOK defs (NPrimVal _ _) (NDCon _ _ _ _ _) = pure True
...
 
 -- NTCon is apart from NBind, and NType
-impossibleOK defs (NTCon _ _ _ _ _) (NBind _ _ _ _) = pure True
-impossibleOK defs (NBind _ _ _ _) (NTCon _ _ _ _ _) = pure True
-impossibleOK defs (NTCon _ _ _ _ _) (NType _ _) = pure True
-impossibleOK defs (NType _ _) (NTCon _ _ _ _ _) = pure True
+impossibleOK defs (NTCon _ _ _ _) (NBind _ _ _ _) = pure True
+impossibleOK defs (NBind _ _ _ _) (NTCon _ _ _ _) = pure True
+impossibleOK defs (NTCon _ _ _ _) (NType _ _) = pure True
+impossibleOK defs (NType _ _) (NTCon _ _ _ _) = pure True
```
andrevidela: Here too

src/TTImp/ProcessDef.idr
```diff
@@ -68,16 +68,16 @@ mutual
...
 
 -- NTCon is apart from NBind, and NType
-  mismatchNF defs (NTCon _ _ _ _ _) (NBind _ _ _ _) = pure True
-  mismatchNF defs (NBind _ _ _ _) (NTCon _ _ _ _ _) = pure True
-  mismatchNF defs (NTCon _ _ _ _ _) (NType _ _) = pure True
-  mismatchNF defs (NType _ _) (NTCon _ _ _ _ _) = pure True
+  mismatchNF defs (NTCon _ _ _ _) (NBind _ _ _ _) = pure True
+  mismatchNF defs (NBind _ _ _ _) (NTCon _ _ _ _) = pure True
+  mismatchNF defs (NTCon _ _ _ _) (NType _ _) = pure True
+  mismatchNF defs (NType _ _) (NTCon _ _ _ _) = pure True
```
andrevidela: And here

andrevidela (commented):
The change itself looks good to me. I noticed it's touching a number of places where we are not actually binding any values in the constructor of a pattern and for this I suggest we use the pattern `Con {}` so that is clear that we never want to match any values, that only the constructor tag matters, and that any constructor that we change in the future doesn't require changing all those pattern matches

andrevidela:
I just noticed the failing test but I'm not on The computer. I'll look into this later

andrevidela:
Ah right the fix is already availabe on the LSP side cool let's do it

PR #3572, opened 2025-06-24
[ fix ] Early evaluation of external functions

spcfox:
# Description

The following program prints `Error: Executed 'void'` when run.

```idris
import Data.So

myVoid : (0 _ : Void) -> ()
myVoid x = void x

foo : (y: Bool) -> (0 p : So y) => ()
foo False = myVoid (absurd p)
foo True  = ()

main : IO ()
main = pure (foo True)
```

This happens because `myVoid` has no runtime arguments and gets evaluated too early due to [constant optimisation](https://github.com/idris-lang/Idris2/pull/2817). However, external functions may have side effects and therefore should not be treated as constants during optimisation.

      `CONTRIBUTORS.md`).

### DIFF
--- a/src/Compiler/Opts/ToplevelConstants.idr
+++ b/src/Compiler/Opts/ToplevelConstants.idr
@@ -95,11 +95,12 @@ markProcessed n = do
 isProcessed : Ref SortTag SortST => Name -> Core Bool
 isProcessed n = map (contains n . processed) (get SortTag)
 
-checkCrash : Ref SortTag SortST => (Name, FC, NamedDef) -> Core ()
-checkCrash (n, _, MkNmError _) = update SortTag $ { nonconst $= insert n }
-checkCrash (n, _, MkNmFun args (NmCrash _ _)) = update SortTag $ { nonconst $= insert n }
-checkCrash (n, _, MkNmFun args (NmOp _ Crash _)) = update SortTag $ { nonconst $= insert n }
-checkCrash (n, _, def) = do
+checkNonPure : Ref SortTag SortST => (Name, FC, NamedDef) -> Core ()
+checkNonPure (n, _, MkNmError _) = update SortTag $ { nonconst $= insert n }
+checkNonPure (n, _, MkNmFun args (NmCrash _ _)) = update SortTag $ { nonconst $= insert n }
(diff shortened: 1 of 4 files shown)


src/Compiler/Opts/ToplevelConstants.idr
```diff
@@ -95,11 +95,12 @@ markProcessed n = do
...
-checkCrash : Ref SortTag SortST => (Name, FC, NamedDef) -> Core ()
-checkCrash (n, _, MkNmError _) = update SortTag $ { nonconst $= insert n }
-checkCrash (n, _, MkNmFun args (NmCrash _ _)) = update SortTag $ { nonconst $= insert n }
-checkCrash (n, _, MkNmFun args (NmOp _ Crash _)) = update SortTag $ { nonconst $= insert n }
-checkCrash (n, _, def) = do
+checkNonPure : Ref SortTag SortST => (Name, FC, NamedDef) -> Core ()
+checkNonPure (n, _, MkNmError _) = update SortTag $ { nonconst $= insert n }
+checkNonPure (n, _, MkNmFun args (NmCrash _ _)) = update SortTag $ { nonconst $= insert n }
+checkNonPure (n, _, MkNmFun args (NmOp _ Crash _)) = update SortTag $ { nonconst $= insert n }
+checkNonPure (n, _, MkNmFun args (NmExtPrim _ _ _)) = update SortTag $ { nonconst $= insert n }
```
spcfox, line 104: Top-level constants are functions without arguments. But external primitives such as `prim__void` can have side effects. However, some external primitives can be pure, such as `prim__os`. Maybe it makes sense to use pragmas to explicitly state that an external function is pure?

stefan-hoeck:
Without further tests, this is potentially devastating for several existing community projects that rely on the current handling of top-level constants for reasons of efficiency. Some libraries, *ilex* for instance, would be rendered unusable by this fix.

If an foreign function has side effects, it should return a `PrimIO`, in which case it could only be early evaluated via a call to `unsafePerformIO`. I don't think there's a need for a pragma in this case.

stefan-hoeck:
Personally, I think we should open an issue about this and discuss our options there.

spcfox:
Closed in favor of #3583

Issue #3570, opened 2025-06-17
Surface Syntax & TTImp improvement plans


andrevidela:
## Proposed Actions & Status

- [x] ✅ not breaking: Add WithData to the compiler #3573
- [x] ✅ not breaking: Replace WithFC by WithData without changing reflection. #3588
- [x] ✅ not breaking: Replace WithName by WithData without changing reflection. #3599
- [x]  ✅ not breaking: replace ImpTy #3615
- [x] ✅ not breaking: replace ImpParam #3599
- [x] ✅ not breaking: replace ImpRecord' #3612
- [x] ✅ not breaking: replace MkCon #3612
- [x] ✅ not breaking: replace `Method` #3619
- [x] ✅ not breaking: `conName` in `MkPRecord` should be `WithDoc Name` and not `(String, Name)` #3619
- [ ] ✅ not breaking: replace IClaimData
- [ ] ✅ not breaking: remove FC from PTerm
- [ ] ✅ not breaking: remove FC from RawImp
- [ ] ✅ Not breaking: remove FC from PDo
- [ ] :warning: breaking: Replace usage of WithFC and WithName by WithData. 
- [ ] :warning: breaking: Replace the redundant definitions by `WithData` in `TTImp` and `PTerm`. 

## The problem

Currently, any changes to RawImp create an enormous cascade of repetitive and mechanical changes. Adding a single field to a record can result in:
- Fixing every function that pattern matches on the record
- writing new TTC instances
- Write new traversal functions for `TTImp/Traversal`
- writing new Reflect and Reify instances
- Duplicating the definition in the reflection library & ensuring is matches the `Reflect` and `Reify` instances
- Due to the above: breaking backward compatibility with previous elab-reflection scripts

Here is an example of such change: https://github.com/idris-lang/Idris2/commit/13af07c2c66ccd89b69b61d8678b216c0b7a716c

Those changes aren't complicated but they are also unnecessary. One of the main design flaws of the compiler is that data that should stick together is allowed to be passed around freely. Functions routinely take 4 or more arguments for a piece of data that has no business in being split into 4 values. For example, `0 n : ty` is never stored as one single value but passed around in function as 4 arguments `FC -> Name -> Rig -> RawImp -> ...`. This split not only creates confusion about where the data comes from and how they are related to each other, it also creates a big maintenance burden when part of the tree needs to be annotated with extra information.

## Requirements for a solution

- Reduce boiler plate when adding record fields
- Automatically generate TTC instance
[...]

PR #3573, opened 2025-06-25
Add `WithData` to generalise over `WithName` and `WithFC`

andrevidela:
# Description

This is the first step towards #3570 it adds three modules:

- `Libraries.Data.Record` a generic labelled record with runtime nat-indexing for performance
- `Libraries.Data.WithData` A datastructure to attach metadata to a type by means of a record
- `Core.WithData` A series of helpers for existing known metadata that the compiler already uses

This design addresses the goals set out by #3570 by allowing us to refactor a number of existing records into either `Record` or `WithData`. This refactoring, in turn, allows us to make changes to the metadata without changing the fundamental structure of the surrounding program, since most tree operations are metadata agnostic. 

Once we add the Reify and Reflect instances for those types, they will also remove the need to write bespoke versions for types that use them in their definitions.

Finally, it allows some data that is currently spread out to be appropriately related, removing the uncertainty about what a piece of data does, and self-documenting the code.

### DIFF
--- a/src/Core/WithData.idr
+++ b/src/Core/WithData.idr
@@ -1,0 +1,185 @@
+module Core.WithData
+
+import Core.Name
+import Core.FC
+import Core.TT
+import public Libraries.Data.WithData
+import Libraries.Text.Bounded
+
+------------------------------------------------------------------------------------------
+-- Helpers for binding information
+------------------------------------------------------------------------------------------
+
+||| The "bind" label containing binding information for metadata records
+public export
+Bind' : KeyVal
+Bind' = "bind" :-: BindingModifier
+
+||| Obtain binding information from the metadata
+export
+(.bind) :
+    {n : Nat} ->
+    (0 inRange : NameInRange "bind" fields === Just (n, BindingModifier)) =>
+    WithData fields a -> BindingModifier
+(.bind) = WithData.get "bind"
+
+------------------------------------------------------------------------------------------
+-- Totality information
+------------------------------------------------------------------------------------------
+
+||| The "totalReq" label containing totality information for metadata records
+public export
+Tot' : KeyVal
+Tot' = "totalReq" :-: Maybe TotalReq
+
+||| Obtain totality information from the metadata
+export
+(.totReq) :
+    {n : Nat} ->
+    (0 inRange : NameInRange "totalReq" fields === Just (n, Maybe TotalReq)) =>
+    WithData fields a -> Maybe TotalReq
+(.totReq) = WithData.get "totalReq"
+
+------------------------------------------------------------------------------------------
+-- Helpers for FC information
+------------------------------------------------------------------------------------------
+
+||| The "fc" label containing file context information for metadata records
+public export
+FC' : KeyVal
+FC' = "fc" :-: FC
+
+||| Attach FC information to a type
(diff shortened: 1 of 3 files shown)


gallais:
> refactor a number of existing records into either `Record` or `WithData`.

Built-in records have constant time access to their fields but for `Record`  it's linear in the index. Isn't that problematic?

andrevidela:
yes this is worse, i'll benchmark my huge PR that refactors the trees using this and come back with a concrete number to measure the difference. 

I didn't notice anything in particular while working on it so my hope is that because most the records are 1 or 2 elements long the overhead is negligible. 

Do you know of a way to replace the heterogenous list by an IOArray? This should get us back O(1) indexing

andrevidela:
Here are the benchmark results

## Methodology

- bootstrap the compiler, the output binary has been compiled with the bootstrap compiler
- compile the compiler with the bootstrapped compiler, the new compiler binary is now compiled with itself
- install the libraries for this version of the compiler because of some TTC differences
- measure how long this compiler takes to typecheck `idris2.ipkg` on `main`
- repeat 5 times

essentially:
```
make clean
make bootstrap
make all
make install-libs
time ./build/exec/idris2 --typecheck idris2.ipkg
```

For benchmarking the branch with the refactored tree, I made sure to move back to `main` so that the program compiled is the same for both runs. I also double checked that the compiler version used was indeed the one being tested

Compiling the compiler currently:
```
Idris 2, version 0.7.0-562ef55a3
./build/exec/idris2 --typecheck idris2.ipkg  102.30s user 2.64s system 97% cpu 1:47.12 total
./build/exec/idris2 --typecheck idris2.ipkg  102.79s user 2.21s system 99% cpu 1:45.94 total
./build/exec/idris2 --typecheck idris2.ipkg  103.64s user 2.21s system 99% cpu 1:46.75 total
./build/exec/idris2 --typecheck idris2.ipkg  108.79s user 2.42s system 98% cpu 1:52.49 total
./build/exec/idris2 --typecheck idris2.ipkg  109.23s user 2.92s system 96% cpu 1:56.33 total

average: 105.35, median: 103.64
```

Compiling the compiler with some of the tree refactored to use WIthData
```
Idris 2, version 0.7.0-3e48e3e44
./build/exec/idris2 --typecheck idris2.ipkg  99.51s user 1.86s system 99% cpu 1:42.12 total
./build/exec/idris2 --typecheck idris2.ipkg  101.64s user 1.94s system 99% cpu 1:44.16 total
./build/exec/idris2 --typecheck idris2.ipkg  103.58s user 2.16s system 99% cpu 1:46.61 total
./build/exec/idris2 --typecheck idris2.ipkg  104.92s user 2.33s system 99% cpu 1:48.24 total
./build/exec/idris2 --typecheck idris2.ipkg  105.79s user 2.37s system 99% cpu 1:49.22 total

average: 103.09 median: 103.58
```

In conclusion, while the access time of the records and `WithData` is `O(n)` it does not seem to impact performance outside of run-to-run differences

(All tests run on MacOS 15.5 (24F74) with an M4 cpu and 16GB of ram)

andrevidela:
If nobody else objects in the meantime, I'll merge it in 2 days

PR #3578, opened 2025-07-05
[ fix ] blodwen-channel-get-with-timeout

### LINKED ISSUE #3577: `blodwen-channel-get-with-timeout` returns unboxed value. (`chez/channels008` test failure)
The `blodwen-channel-get-with-timeout` support function can return an invalid value when it fails to acquire the mutex. From what I can tell, the intent is that the function spins whilst waiting for the mutex, as indicated by having `loop` in the false branch of the `mutex-acquire`. Because of some missing parentheses, it instead returns the `loop` body as a procedure.

The quick fix is to wrap `loop` in [line 494](https://github.com/idris-lang/Idris2/blob/f85e6d4bf90ca93af62c711e6487e17bd9b8fceb/support/chez/support.ss#L494) in parentheses to make it a function call. Looking more long-term:
1. Why are we spinning? We don't even appear to be keeping track of how long we are spinning for. I would hope there is a better solution for acquiring the mutex than spinning. The Chez interface doesn't appear to accept a timeout for `mutex-acquire`, so we would either need to contribute to upstream or at the very least make sure we time how long we spin. Otherwise we might spin for longer than the timeout.
2. Can we find a more consistent way to test this? I think I stumbled into this by bad luck. I don't know enough about concurrency in Idris to work out how to cause the `mutex-acquire` to fail reliably.

Matthew-Mosior:
# Description

[Issue-3577](https://github.com/idris-lang/Idris2/issues/3577) details some issues with the existing implementation of `blodwen-channel-get-with-timeout`:
- Improper recursion on `loop`.
- Lack of tracking time spent trying to acquire mutex.

`blodwen-channel-get-with-timeout` has been strengthened to address the above issues.

`channelGetWithTimeout` has been clamped to a minimum of 25 ms on `windows` as prior `windows-bootstrap-chez` tests show issue with timeout smaller than 25 ms.

Added a test case to demonstrate that `channelGetWithTimeout` can indeed timeout.  We spawn multiple consumers and multiple producers, and observe that the first of the consumers times out.
- `channels009` 

Closes #3577.

      `CONTRIBUTORS.md`).

### DIFF
--- a/libs/base/System/Concurrency.idr
+++ b/libs/base/System/Concurrency.idr
@@ -236,13 +236,7 @@ channelGetWithTimeout : HasIO io => (chan : Channel a) -> (milliseconds : Nat) -
 channelGetWithTimeout chan milliseconds =
   case os of
     "windows" =>
-      let clampedms = case compare milliseconds 25 of
-                        LT =>
-                          25
-                        EQ =>
-                          25
-                        GT =>
-                          cast milliseconds
+      let clampedms = if milliseconds > 25 then cast milliseconds else 25
         in primIO (prim__channelGetWithTimeout chan clampedms)
     _         =>
       primIO (prim__channelGetWithTimeout chan (cast milliseconds))


tests/chez/channels009/Main.idr
yellowsquid, line 1: I don't follow the logic of this test. From what I can tell 99 threads are spawned and immediately write to the channel. We then hope that all of them take at least one second to do this so that `channelGetWithTimeout` gives up.

We could maybe simplify the test by having a single thread read with a one second timeout and then check it waited for between one and one + epsilon seconds. If we don't have any threads writing it will certainly time out. We use a forgiving epsilon to account for different testing environments. We might need multiple threads (two or three) trying to read to create contention for the channel mutex.

Matthew-Mosior: Yeah, I realized this test isn't sufficient for reliably timing out `channelGetWithTimeout`.

That's a good idea, yeah let me think through that and give it a go.

Matthew-Mosior: @yellowsquid 

`channels009` should now adequately test timeout with `channelGetWithTimeout`.

We fork `5` consumer threads, each of which calls `channelGetWithTimeout` with a short timeout of `100` ms.  With a long sleep after the first forked consumer, this is too short for it to be able to read a value from the channel after starting up `5` producer threads.

libs/base/System/Concurrency.idr
```diff
@@ -224,13 +226,26 @@ channelGetNonBlocking : HasIO io => (chan : Channel a) -> io (Maybe a)
...
+channelGetWithTimeout chan milliseconds =
+  case os of
+    "windows" =>
+      let clampedms = case compare milliseconds 25 of
+                        LT =>
+                          25
+                        EQ =>
+                          25
+                        GT =>
+                          cast milliseconds
```
gallais:
```suggestion
      let clampedms = if milliseconds > 25 then cast milliseconds else 25
```

support/chez/support.ss
```diff
@@ -460,38 +460,34 @@
   '()))
 
 (define (blodwen-channel-get-with-timeout ty chan timeout)
-  (let loop ()
-    (let* ([sec (div timeout 1000)])
+  ;; timeout is in milliseconds, convert to nanoseconds
+  (let* ([timeout-ns (* timeout 1000000)]
+         [sleep-ns 10000] ; 10 us step
+         [sleep-time (make-time 'time-duration (mod sleep-ns 1000000000)
+                                                (div sleep-ns 1000000000))])
```
gallais, line 467: Given that `sleep-ns` is a constant, aren't these values going to be exactly `10000` & `0`?

gallais: In #3577 you mentioned an "exponential backoff" but I guess it did not make it in the PR?

Matthew-Mosior: > In #3577 you mentioned an "exponential backoff" but I guess it did not make it in the PR?

Yeah, was having trouble implementing the exponential backoff correctly, so I scraped it.

Matthew-Mosior: > Given that `sleep-ns` is a constant, aren't these values going to be exactly `10000` & `0`?

Yeah, `sleep-ns` is a constant 10 us sleep, so yes I believe so

PR #3579, opened 2025-07-06
Grammar improvements to the beginning of the tutorial

Bersier:
# Description

I went over
1. starting.rst
2. windows.rst
3. typesfuns.rst
4. interfaces.rst

### DIFF
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -62,8 +62,8 @@ going to use it to convert each element to a ``String``:
             show' (x :: Nil) = show x
             show' (x :: xs)  = show x ++ ", " ++ show' xs
 
-Note that we need the explicit ``forall n .`` in the ``show'`` function;
-otherwise, ``n`` would refer to the top-level ``n`` already in scope.
+Note that we need the explicit ``forall n .`` in the ``show'`` function,
+because otherwise, ``n`` would refer to the top-level ``n`` already in scope.
 
 Default Definitions
 ===================
@@ -402,7 +402,7 @@ case back as follows:
 The effect of this version of ``readNumbers`` is identical to the first (in
 fact, it is syntactic sugar for it and directly translated back into that form).
 The first part of each statement (``Just x_ok <-`` and ``Just y_ok <-``) gives
-the preferred binding — if this matches, execution will continue with the rest
+the preferred binding; if this matches, execution will continue with the rest
 of the ``do`` block. The second part gives the alternative bindings, of which
 there may be more than one.
 
@@ -643,7 +643,7 @@ Named Implementations
 =====================
 
 It can be desirable to have multiple implementations of an interface for the
-same type — for example to provide alternative methods for sorting or printing
+same type, for example to provide alternative methods for sorting or printing
 values. To achieve this, implementations can be *named* as follows:
 
 .. code-block:: idris
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -262,11 +262,11 @@ abstracts the notion of function application:
 
 .. _monadsdo:
 
-Monads and ``do`` notation
+Monads and ``do``-notation
 ==========================
 
 The ``Monad`` interface allows us to encapsulate binding and computation,
-and is the basis of ``do`` notation introduced in Section
+and is the basis of ``do``-notation introduced in Section
 :ref:`sect-do`. It extends ``Applicative`` as defined above, and is
 defined as follows:
 
@@ -311,8 +311,8 @@ Using this, we can, for example, define a function that adds two
 
 This function will extract the values from ``x`` and ``y``, if they are both
 available, or return ``Nothing`` if one or both are not ("fail fast"). Managing
-the ``Nothing`` cases is achieved by the ``>>=`` operator, hidden by the ``do``
-notation.
+the ``Nothing`` cases is achieved by the ``>>=`` operator, hidden by the
+``do``-notation.
 
 ::
 
@@ -321,11 +321,11 @@ notation.
     Main> m_add (Just 82) Nothing
     Nothing
 
-The translation of ``do`` notation is entirely syntactic, so there is no
+The translation of ``do``-notation is entirely syntactic, so there is no
 need for the ``(>>=)`` and ``(>>)`` operators to be those defined in the
 ``Monad`` interface. In general, Idris will try to disambiguate which
 operators you mean based on their type. However, you can explicitly specify
-the source module using qualified ``do`` notation. For example:
+the source module using qualified ``do``-notation. For example:
 
 .. code-block:: idris
 
@@ -342,7 +342,7 @@ Pattern Matching Bind
 ~~~~~~~~~~~~~~~~~~~~~
 
 Sometimes we want to pattern match immediately on the result of a function
-in ``do`` notation. For example, let's say we have a function ``readNumber``
+in ``do``-notation. For example, let's say we have a function ``readNumber``
 which reads a number from the console, returning a value of the form
 ``Just x`` if the number is valid, or ``Nothing`` otherwise:
 
@@ -406,10 +406,10 @@ the preferred binding; if this matches, execution will continue with the rest
 of the ``do`` block. The second part gives the alternative bindings, of which
 there may be more than one.
 
-``!`` notation
+``!``-notation
 ~~~~~~~~~~~~~~
 
-In many cases, using ``do`` notation can make programs unnecessarily
+In many cases, using ``do``-notation can make programs unnecessarily
 verbose, particularly in cases such as ``m_add`` above where the value
 bound is used once, immediately. In these cases, we can use a
 shorthand version, as follows:
@@ -431,7 +431,7 @@ Note, however, that it is not really a function, but merely syntax! In
 practice, a subexpression ``!expr`` will lift ``expr`` as high as
 possible within its current scope, bind it to a fresh name ``x``, and
 replace ``!expr`` with ``x``. Expressions are lifted depth-first, left
-to right. In practice, ``!`` notation allows us to program in a more
+to right. In practice, ``!``-notation allows us to program in a more
 direct style, while still giving a notational clue as to which
 expressions are monadic.
 
@@ -480,7 +480,7 @@ qual``, using the following function:
 
     guard : Alternative f => Bool -> f ()
 
-Then the comprehension is converted to ``do`` notation:
+Then the comprehension is converted to ``do``-notation:
 
 .. code-block:: idris
 
@@ -514,7 +514,7 @@ programs, while still retaining direct access to ``IO`` primitives.
 Idiom brackets
 ==============
 
-While ``do`` notation gives an alternative meaning to sequencing,
+While ``do``-notation gives an alternative meaning to sequencing,
 idioms give an alternative meaning to *application*. The notation and
 larger example in this section is inspired by Conor McBride and Ross
 Paterson’s paper “Applicative Programming with Effects” [#ConorRoss]_.
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -517,7 +517,7 @@ Idiom brackets
 While ``do``-notation gives an alternative meaning to sequencing,
 idioms give an alternative meaning to *application*. The notation and
 larger example in this section is inspired by Conor McBride and Ross
-Paterson’s paper "Applicative Programming with Effects" [#ConorRoss]_.
+Paterson's paper "Applicative Programming with Effects" [#ConorRoss]_.
 
 First, let us revisit ``m_add`` above. All it is really doing is
 applying an operator to two values extracted from ``Maybe Int``. We
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -299,7 +299,7 @@ We can also define an implementation for ``Maybe``, as follows:
         Nothing  >>= k = Nothing
         (Just x) >>= k = k x
 
-Using this, we can, for example, define a function that adds two
+Using this, we can for example define a function that adds two
 ``Maybe Int``, using the monad to encapsulate the error handling:
 
 .. code-block:: idris
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -63,7 +63,7 @@ going to use it to convert each element to a ``String``:
             show' (x :: xs)  = show x ++ ", " ++ show' xs
 
 Note that we need the explicit ``forall n .`` in the ``show'`` function,
-because otherwise, ``n`` would refer to the top-level ``n`` already in scope.
+because n is already bound by the top-level definition.
 
 Default Definitions
 ===================
@@ -358,7 +358,7 @@ which reads a number from the console, returning a value of the form
          else pure Nothing
 
 If we then use it to write a function to read two numbers, returning
-``Nothing`` if either is invalid, we would like to pattern match
+``Nothing`` if any of the reads fail, we would like to pattern match
 on the result of ``readNumber``:
 
 .. code-block:: idris
@@ -667,7 +667,7 @@ Given the following list:
 
 We can sort it using the default ``Ord`` implementation by using the ``sort``
 function, available via ``import Data.List``. Then, we can try with the named
-implementation ``myord`` — both are shown below at the Idris prompt:
+implementation ``myord`` - both are shown below at the Idris prompt:
 
 ::
 
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -63,7 +63,7 @@ going to use it to convert each element to a ``String``:
             show' (x :: xs)  = show x ++ ", " ++ show' xs
 
 Note that we need the explicit ``forall n .`` in the ``show'`` function,
-because n is already bound by the top-level definition.
+because ``n`` is already bound by the top-level definition.
 
 Default Definitions
 ===================
@@ -470,7 +470,7 @@ qualn ]`` where ``quali`` can be one of:
 
 - A *guard*, which is an expression of type ``Bool``
 
-- A ``let`` binding ``let x = e``
+- A ``let``-binding ``let x = e``
 
 To translate a comprehension ``[exp | qual1, qual2, …, qualn]``, first
 any qualifier ``qual`` which is a *guard* is translated to ``guard
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -28,8 +28,8 @@ This generates a function of the following type (which we call a
 
     show : Show a => a -> String
 
-We can read this as: “under the constraint that ``a`` has an implementation
-of ``Show``, take an input ``a`` and return a ``String``.” An implementation
+We can read this as: "under the constraint that ``a`` has an implementation
+of ``Show``, take an input ``a`` and return a ``String``." An implementation
 of an interface is defined by giving definitions of the methods of the interface.
 For example, the ``Show`` implementation for ``Nat`` could be defined as:
 
@@ -517,7 +517,7 @@ Idiom brackets
 While ``do``-notation gives an alternative meaning to sequencing,
 idioms give an alternative meaning to *application*. The notation and
 larger example in this section is inspired by Conor McBride and Ross
-Paterson’s paper “Applicative Programming with Effects” [#ConorRoss]_.
+Paterson’s paper "Applicative Programming with Effects" [#ConorRoss]_.
 
 First, let us revisit ``m_add`` above. All it is really doing is
 applying an operator to two values extracted from ``Maybe Int``. We
@@ -684,7 +684,7 @@ the prelude defines the following ``Semigroup`` interface:
     interface Semigroup ty where
       (<+>) : ty -> ty -> ty
 
-Then it defines ``Monoid``, which extends ``Semigroup`` with a “neutral”
+Then it defines ``Monoid``, which extends ``Semigroup`` with a "neutral"
 value:
 
 .. code-block:: idris
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -44,7 +44,7 @@ For example, the ``Show`` implementation for ``Nat`` could be defined as:
     Main> show (S (S (S Z)))
     "sssZ" : String
 
-Only one implementation of an interface can be given for a type — implementations may
+Only one implementation of an interface can be given for a type - implementations may
 not overlap. Implementation declarations can themselves have constraints.
 To help with resolution, the arguments of an implementation must be
 constructors (either data or type constructors) or variables
--- a/docs/source/tutorial/interfaces.rst
+++ b/docs/source/tutorial/interfaces.rst
@@ -62,9 +62,8 @@ going to use it to convert each element to a ``String``:
             show' (x :: Nil) = show x
             show' (x :: xs)  = show x ++ ", " ++ show' xs
 
-Note that we need the explicit ``forall n .`` in the ``show'`` function
-because otherwise the ``n`` is already in scope, and fixed to the value of
-the top level ``n``.
+Note that we need the explicit ``forall n .`` in the ``show'`` function;
+otherwise, ``n`` would refer to the top-level ``n`` already in scope.
 
 Default Definitions
 ===================
@@ -92,7 +91,7 @@ of the methods. For example, for an implementation of ``Eq`` for ``Nat``:
 
         x /= y = not (x == y)
 
-It is hard to imagine many cases where the ``/=`` method will be
+It is hard to imagine many cases where the ``/=`` method would be
 anything other than the negation of the result of applying the ``==``
 method. It is therefore convenient to give a default definition for
 each method in the interface declaration, in terms of the other method:
@@ -137,26 +136,26 @@ defining some of its own:
 
 The ``Ord`` interface allows us to compare two values and determine their
 ordering. Only the ``compare`` method is required; every other method
-has a default definition. Using this we can write functions such as
+has a default definition. Using this, we can write functions such as
 ``sort``, a function which sorts a list into increasing order,
 provided that the element type of the list is in the ``Ord`` interface. We
-give the constraints on the type variables left of the fat arrow
+give the constraints on the type variables to the left of the fat arrow
 ``=>``, and the function type to the right of the fat arrow:
 
 .. code-block:: idris
 
     sort : Ord a => List a -> List a
 
-Functions, interfaces and implementations can have multiple
-constraints. Multiple constraints are written in brackets in a comma
-separated list, for example:
+Functions, interfaces, and implementations can have multiple
+constraints. Multiple constraints are written in brackets in a
+comma-separated list, for example:
 
 .. code-block:: idris
 
     sortAndShow : (Ord a, Show a) => List a -> String
     sortAndShow xs = show (sort xs)
 
-Constraints are, like types, first class objects in the language. You can
+Constraints, like types, are first-class objects in the language. You can
 see this at the REPL:
 
 ::
@@ -165,7 +164,7 @@ see this at the REPL:
     Prelude.Ord : Type -> Type
 
 So, ``(Ord a, Show a)`` is an ordinary pair of ``Types``, with two constraints
-as the first and second element of the pair.
+as the first and second elements of the pair.
 
 Note: Interfaces and ``mutual`` blocks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -180,7 +179,7 @@ default definitions on the second pass.
 Quantities for Parameters
 =========================
 
-By default parameters that are not explicitly ascribed a type in an ``interface``
+By default, parameters that are not explicitly ascribed a type in an ``interface``
 declaration are assigned the quantity ``0``. This means that the parameter is not
 available to use at runtime in the methods' definitions.
 
@@ -193,7 +192,7 @@ the type of the ``show`` method:
      Main> :t show
      Prelude.show : {0 a : Type} -> Show a => a -> String
 
-However some use cases require that some of the parameters are available at runtime.
+However, some use cases require that some of the parameters are available at runtime.
 We may for instance want to declare an interface for ``Storable`` types. The constraint
 ``Storable a size`` means that we can store values of type ``a`` in a ``Buffer`` in
 exactly ``size`` bytes.
@@ -201,8 +200,7 @@ exactly ``size`` bytes.
 If the user provides a method to read a value for such a type ``a`` at a given offset,
 then we can read the ``k`` th element stored in the buffer by computing the appropriate
 offset from ``k`` and ``size``. This is demonstrated by providing a default implementation
-for the method ``peekElementOff`` implemented in terms of ``peekByteOff`` and the parameter
-``size``.
+for the method ``peekElementOff`` in terms of ``peekByteOff`` and the parameter ``size``.
 
 .. code-block:: idris
 
@@ -217,14 +215,14 @@ for the method ``peekElementOff`` implemented in terms of ``peekByteOff`` and th
 
 
 Note that ``a`` is explicitly marked as runtime irrelevant so that it is erased by the
-compiler. Equivalently we could have written ``interface Storable a (size : Nat)``.
+compiler. Equivalently, we could have written ``interface Storable a (size : Nat)``.
 The meaning of ``| a`` is explained in :ref:`DeterminingParameters`.
 
 
 Functors and Applicatives
 =========================
 
-So far, we have seen single parameter interfaces, where the parameter
+So far, we have seen single-parameter interfaces, where the parameter
 is of type ``Type``. In general, there can be any number of parameters
 (even zero), and the parameters can have *any* type. If the type
 of the parameter is not ``Type``, we need to give an explicit type
@@ -264,11 +262,11 @@ abstracts the notion of function application:
 
 .. _monadsdo:
 
-Monads and ``do``-notation
+Monads and ``do`` notation
 ==========================
 
 The ``Monad`` interface allows us to encapsulate binding and computation,
-and is the basis of ``do``-notation introduced in Section
+and is the basis of ``do`` notation introduced in Section
 :ref:`sect-do`. It extends ``Applicative`` as defined above, and is
 defined as follows:
 
@@ -301,7 +299,7 @@ We can also define an implementation for ``Maybe``, as follows:
         Nothing  >>= k = Nothing
         (Just x) >>= k = k x
 
-Using this we can, for example, define a function which adds two
+Using this, we can, for example, define a function that adds two
 ``Maybe Int``, using the monad to encapsulate the error handling:
 
 .. code-block:: idris
@@ -324,10 +322,10 @@ notation.
     Nothing
 
 The translation of ``do`` notation is entirely syntactic, so there is no
-need for the ``(>>=)`` and ``(>>)`` operators to be the operator defined in the
-``Monad`` interface. Idris will, in general, try to disambiguate which
-operators you mean by type, but you can explicitly choose with qualified do
-notation, for example:
+need for the ``(>>=)`` and ``(>>)`` operators to be those defined in the
+``Monad`` interface. In general, Idris will try to disambiguate which
+operators you mean based on their type. However, you can explicitly specify
+the source module using qualified ``do`` notation. For example:
 
 .. code-block:: idris
 
@@ -360,7 +358,7 @@ which reads a number from the console, returning a value of the form
          else pure Nothing
 
 If we then use it to write a function to read two numbers, returning
-``Nothing`` if neither are valid, then we would like to pattern match
+``Nothing`` if either is invalid, we would like to pattern match
 on the result of ``readNumber``:
 
 .. code-block:: idris
@@ -388,7 +386,7 @@ we could try pattern matching on values of the form ``Just x_ok``:
            pure (Just (x_ok, y_ok))
 
 There is still a problem, however, because we've now omitted the case for
-``Nothing`` so ``readNumbers`` is no longer total! We can add the ``Nothing``
+``Nothing``, so ``readNumbers`` is no longer total! We can add the ``Nothing``
 case back as follows:
 
 .. code-block:: idris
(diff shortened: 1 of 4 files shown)


docs/source/tutorial/interfaces.rst
```diff
@@ -62,9 +62,8 @@ going to use it to convert each element to a ``String``:
             show' (x :: Nil) = show x
             show' (x :: xs)  = show x ++ ", " ++ show' xs
 
-Note that we need the explicit ``forall n .`` in the ``show'`` function
-because otherwise the ``n`` is already in scope, and fixed to the value of
-the top level ``n``.
+Note that we need the explicit ``forall n .`` in the ``show'`` function;
+otherwise, ``n`` would refer to the top-level ``n`` already in scope.
```
CodingCellist: Potentially, one could add a comma after "function", but otherwise I find that the text reads better as it was originally.

docs/source/tutorial/interfaces.rst
```diff
@@ -193,16 +192,15 @@ the type of the ``show`` method:
...
 We may for instance want to declare an interface for ``Storable`` types. The constraint
 ``Storable a size`` means that we can store values of type ``a`` in a ``Buffer`` in
 exactly ``size`` bytes.
 
 If the user provides a method to read a value for such a type ``a`` at a given offset,
 then we can read the ``k`` th element stored in the buffer by computing the appropriate
 offset from ``k`` and ``size``. This is demonstrated by providing a default implementation
-for the method ``peekElementOff`` implemented in terms of ``peekByteOff`` and the parameter
-``size``.
+for the method ``peekElementOff`` in terms of ``peekByteOff`` and the parameter ``size``.
```
CodingCellist, line 203: A comma after \`\`peekElementOff\`\` might more clearly convey the intended phrasing?

docs/source/tutorial/interfaces.rst
```diff
@@ -264,11 +262,11 @@ abstracts the notion of function application:
 
 .. _monadsdo:
 
-Monads and ``do``-notation
+Monads and ``do`` notation
```
CodingCellist: This should be reverted throughout. It might not look the prettiest, but writing "`do`-notation" is common in the literature.

docs/source/tutorial/interfaces.rst
```diff
@@ -301,7 +299,7 @@ We can also define an implementation for ``Maybe``, as follows:
         Nothing  >>= k = Nothing
         (Just x) >>= k = k x
 
-Using this we can, for example, define a function which adds two
+Using this, we can, for example, define a function that adds two
```
CodingCellist: This is a British vs American English thing, and given that Idris is a Scottish programming language, we tend to use the former.

docs/source/tutorial/interfaces.rst
```diff
@@ -324,10 +322,10 @@ notation.
...
 
 The translation of ``do`` notation is entirely syntactic, so there is no
-need for the ``(>>=)`` and ``(>>)`` operators to be the operator defined in the
-``Monad`` interface. Idris will, in general, try to disambiguate which
-operators you mean by type, but you can explicitly choose with qualified do
-notation, for example:
+need for the ``(>>=)`` and ``(>>)`` operators to be those defined in the
+``Monad`` interface. In general, Idris will try to disambiguate which
+operators you mean based on their type. However, you can explicitly specify
+the source module using qualified ``do`` notation. For example:
```
CodingCellist: > \`\`do\`\` notation

~> \`\`do\`\`-notation

docs/source/tutorial/interfaces.rst
```diff
@@ -360,7 +358,7 @@ which reads a number from the console, returning a value of the form
          else pure Nothing
 
 If we then use it to write a function to read two numbers, returning
-``Nothing`` if neither are valid, then we would like to pattern match
+``Nothing`` if either is invalid, we would like to pattern match
```
CodingCellist: Should potentially be "neither is valid", but it is clearer as "neither _ valid" rather than "either _ invalid".

docs/source/tutorial/interfaces.rst
```diff
@@ -404,14 +402,14 @@ case back as follows:
 The effect of this version of ``readNumbers`` is identical to the first (in
 fact, it is syntactic sugar for it and directly translated back into that form).
 The first part of each statement (``Just x_ok <-`` and ``Just y_ok <-``) gives
-the preferred binding - if this matches, execution will continue with the rest
+the preferred binding — if this matches, execution will continue with the rest
```
CodingCellist: This should possibly be a semicolon or simply a full stop? In any case, it should not be an em-dash, which are used to indicate an interruption of speech in British English.

docs/source/tutorial/interfaces.rst
```diff
@@ -404,14 +402,14 @@ case back as follows:
...
 of the ``do`` block. The second part gives the alternative bindings, of which
 there may be more than one.
 
-``!``-notation
+``!`` notation
```
CodingCellist: Should be reverted throughout; same comment as "`do`-notation". Apologies.

docs/source/tutorial/interfaces.rst
```diff
@@ -645,7 +643,7 @@ Named Implementations
 =====================
 
 It can be desirable to have multiple implementations of an interface for the
-same type, for example to provide alternative methods for sorting or printing
+same type — for example to provide alternative methods for sorting or printing
```
CodingCellist: Again, shouldn't be an em-dash. The original is fine here imo.

docs/source/tutorial/interfaces.rst
```diff
@@ -667,9 +665,9 @@ Given the following list:
     testList : List Nat
     testList = [3,4,1]
 
-We can sort it using the default ``Ord`` implementation, by using the ``sort``
-function available with ``import Data.List``, then we can try with the named
-implementation ``myord`` as follows, at the Idris prompt:
+We can sort it using the default ``Ord`` implementation by using the ``sort``
+function, available via ``import Data.List``. Then, we can try with the named
+implementation ``myord`` — both are shown below at the Idris prompt:
```
CodingCellist: I agree with most of these changes, but the original comma after "implementation" should definitely be preserved; it helps the flow of the sentence. Another em-dash.

docs/source/tutorial/typesfuns.rst
```diff
@@ -33,21 +33,20 @@ into the Idris interactive environment by typing ``idris2 Prims.idr``:
...
+collection of declarations and definitions. In this example, no imports
+have been specified. However, Idris programs can consist of several
+modules, each with its own namespace. This is discussed further in Section
+:ref:`sect-namespaces`. When writing Idris programs, both the order in which
 definitions are given and indentation are significant. Functions and
-data types must be defined before use, incidentally each definition must
-have a type declaration, for example see ``x : Int``, ``foo :
-String``, from the above listing. New declarations must begin at the
+data types must be defined before use. Incidentally, each definition must
+have a type declaration — for example, see ``x : Int`` or ``foo :
```
CodingCellist: "or" ~> "and", both provide a combined example of type declarations before definition.

docs/source/tutorial/typesfuns.rst
```diff
@@ -159,8 +157,8 @@ interfaces, meaning that we can also test the functions as follows:
 
 You may wonder, by the way, why we have unary natural numbers when our
 computers have perfectly good integer arithmetic built in. The reason
-is primarily that unary numbers have a very convenient structure which
-is easy to reason about, and easy to relate to other data structures
+is primarily that unary numbers have a very convenient structure that
+is easy to reason about, and easy to relate to other data structures,
```
CodingCellist, line 161: Please keep the "which"

docs/source/tutorial/typesfuns.rst
```diff
@@ -171,8 +169,8 @@ such as ``plus`` and ``mult``.
 -----------------
 
 Functions can also be defined *locally* using ``where`` clauses. For
-example, to define a function which reverses a list, we can use an
-auxiliary function which accumulates the new, reversed list, and which
+example, to define a function that reverses a list, we can use an
+auxiliary function that accumulates the new, reversed list, and which
```
CodingCellist: The first "that" is fine, the second should be reverted.

docs/source/tutorial/typesfuns.rst
```diff
@@ -188,9 +186,9 @@ indented further than the outer function.
 
 .. note:: Scope
 
-    Any names which are visible in the outer scope are also visible in
+    Any names that are visible in the outer scope are also visible in
     the ``where`` clause (unless they have been redefined, such as ``xs``
-    here). A name which appears in the type will be in scope in the
+    here). A name that appears in the type will be in scope in the
```
CodingCellist, line 191: Another British vs American English thing, I suspect. The original reads fine.

docs/source/tutorial/typesfuns.rst
```diff
@@ -263,7 +261,7 @@ You can override this with a ``partial`` annotation:
 
 However, this is not advisable, and in general you should only do this during
 the initial development of a function, or during debugging.  If you try to
-evaluate ``fromMaybe Nothing`` at run time you will get a run time error.
+evaluate ``fromMaybe Nothing`` at runtime, you will get a runtime error.
```
CodingCellist, line 264: I believe this is subtly different: I've seen "run time" (or "run-time") used to distinguish from "compile time", with "runtime" typically referring to the underlying language/system which implements the I/O etc.

docs/source/tutorial/typesfuns.rst
```diff
@@ -278,7 +276,7 @@ programs. For example, we could leave a hole for the greeting in our
     main = putStrLn ?greeting
 
 The syntax ``?greeting`` introduces a hole, which stands for a part of
-a program which is not yet written. This is a valid Idris program, and you
+a program that is not yet written. This is a valid Idris program, and you
```
CodingCellist, line 279: Please revert.

docs/source/tutorial/typesfuns.rst
```diff
@@ -329,7 +327,7 @@ For example, we could write a function which computes a type:
     isSingleton True = Nat
     isSingleton False = List Nat
 
-This function calculates the appropriate type from a ``Bool`` which flags
+This function calculates the appropriate type from a ``Bool`` that flags
```
CodingCellist, line 330: Please revert.

docs/source/tutorial/typesfuns.rst
```diff
@@ -438,8 +436,8 @@ importing ``Data.Fin``, or can be declared as follows:
        FS : Fin k -> Fin (S k)
 
 From the signature,  we can see that this is a type constructor that takes a ``Nat``, and produces a type.
-So this is not a set in the sense of a collection that is a container of objects,
-rather it is the canonical set of unnamed elements, as in "the set of 5 elements," for example.
+So this is not a set in the sense of a collection that is a container of objects;
+rather, it is the canonical set of unnamed elements, as in “the set of 5 elements”, for example.
```
CodingCellist: Literal unicode quotes accidentally left in from editing on MacOS?

docs/source/tutorial/typesfuns.rst
```diff
@@ -499,14 +497,14 @@ names, ``n`` and ``a``, which are not declared explicitly. These are
     index : forall a, n . Fin n -> Vect n a -> a
 
 Implicit arguments, given with the ``forall`` declaration,
-are not given in applications of ``index``; their values can be
+are not supplied in applications of ``index``; their values can be
```
CodingCellist: Maybe "provided", also fine to keep as "given", but "supplied" seems off in terms of what I've typically seen people use in this context.

docs/source/tutorial/typesfuns.rst
```diff
@@ -499,14 +497,14 @@ names, ``n`` and ``a``, which are not declared explicitly. These are
...
 inferred from the types of the ``Fin n`` and ``Vect n a``
-arguments. Any name beginning with a lower case letter which appears
+arguments. Any name beginning with a lowercase letter that appears
```
CodingCellist: "that" ~> "which"

docs/source/tutorial/typesfuns.rst
```diff
@@ -499,14 +497,14 @@ names, ``n`` and ``a``, which are not declared explicitly. These are
...
 as a parameter or index in a
-type declaration, which is not applied to any arguments, will
+type declaration, and is not applied to any arguments, will
```
CodingCellist: "and" ~> "and which" ?
Or possibly revert.

docs/source/tutorial/typesfuns.rst
```diff
@@ -526,7 +524,7 @@ help document a function by making the purpose of an argument more
 clear.
 
 The names of implicit arguments are in scope in the body of the function,
-although they cannot be used at run time. There is much more to say about
+although they cannot be used at runtime. There is much more to say about
```
CodingCellist, line 527: See previous comment on "run time" / "run-time" / "runtime"

docs/source/tutorial/typesfuns.rst
```diff
@@ -610,16 +608,16 @@ user or the system in some way. The difficulty in a pure language such
 as Idris — that is, a language where expressions do not have
 side-effects — is that I/O is inherently side-effecting. So, Idris provides
 a parameterised type ``IO`` which *describes* the interactions that the
-run-time system will perform when executing a function:
+runtime system will perform when executing a function:
```
CodingCellist, line 611: See previous comment on "run time" / "run-time" / "runtime"

docs/source/tutorial/typesfuns.rst
```diff
@@ -610,16 +608,16 @@ user or the system in some way. The difficulty in a pure language such
...
 .. code-block:: idris
 
-    data IO a -- description of an IO operation returning a value of type a
+    data IO a -- description of an I/O operation returning a value of type a
 
 We’ll leave the definition of ``IO`` abstract, but effectively it
 describes what the I/O operations to be executed are, rather than how
 to execute them. The resulting operations are executed externally, by
-the run-time system. We’ve already seen one I/O program:
+the runtime system. We’ve already seen one I/O program:
```
CodingCellist: See previous comment on "run time" / "run-time" / "runtime"

docs/source/tutorial/typesfuns.rst
```diff
@@ -969,7 +967,7 @@ If you like, you can write it out the long way; the two are equivalent:
 
 The type checker could infer the value of the first element
 from the length of the vector. We can write an underscore ``_`` in
-place of values which we expect the type checker to fill in, so the
+place of values that we expect the type checker to fill in, so the
```
CodingCellist, line 970: "that" ~> "which"

docs/source/tutorial/typesfuns.rst
```diff
@@ -1079,15 +1077,15 @@ updated):
     *Record> { firstName := "Jim", age $= (+ 1) } fred
     MkPerson "Jim" "Joe" "Bloggs" 31 : Person
 
-The syntax ``{ field := val, ... }`` generates a function which
+The syntax ``{ field := val, ... }`` generates a function that
```
CodingCellist, line 1080: "that" ~> "which"

docs/source/tutorial/typesfuns.rst
```diff
@@ -1174,7 +1172,7 @@ name. For example, a pair type could be defined as follows:
         snd : b
 
 Using the ``Class`` record from earlier, the size of the class can be
-restricted using a ``Vect`` and the size included in the type by parameterising
+restricted using a ``Vect``, and the size included in the type by parameterising
```
CodingCellist: I believe that comma should be after "type"? The class can be restricted by two things, which should be grouped and not comma-separated, with the second part of the sentence then clearly starting by having the comma after "type"?

docs/source/tutorial/typesfuns.rst
```diff
@@ -1281,14 +1279,14 @@ other things, avoid these ambiguities with propositional equality:
    Diag : a -> Type
    Diag v = let ty : Type := v = v in ty
 
-The code above can be read as "``ty`` has type ``Type`` and its value
-is ``v = v``".
+The code above can be read as “``ty`` has type ``Type`` and its value
+is ``v = v``”.
```
CodingCellist: More unicode quotes.

docs/source/tutorial/typesfuns.rst
```diff
@@ -1281,14 +1279,14 @@ other things, avoid these ambiguities with propositional equality:
...
 
-Local definitions can also be introduced using ``let``. Just like top level
-ones and ones defined in a ``where`` clause you need to:
+Local definitions can also be introduced using ``let``. Just like with
+top-level ones and those defined in a ``where`` clause, you need to:
 
 1. declare the function and its type
-2. define the function by pattern matching
+2. define the function via pattern matching
```
CodingCellist: Please revert. The expression "by pattern matching" is common.

docs/source/tutorial/typesfuns.rst
```diff
@@ -1352,7 +1350,7 @@ character:
     splitAt c x = case break (== c) x of
                       (l, r) => (l, strTail r)
 
-``break`` is a library function which breaks a string into a pair of
+``break`` is a library function that breaks a string into a pair of
```
CodingCellist, line 1353: "that" ~> "which"

docs/source/tutorial/typesfuns.rst
```diff
@@ -1392,16 +1390,16 @@ A total function is a function that either:
...
 
-+ If it's total, it will return a value of type ``String`` in finite time;
++ If it's total, it will return a value of type ``String`` in finite time.
 + If it's partial, then as long as it doesn't crash or enter an infinite loop,
   it will return a ``String``.
 
 Idris makes this distinction so that it knows which functions are safe to
 evaluate while type checking (as we've seen with :ref:`sect-fctypes`). After all,
-if it tries to evaluate a function during type checking which doesn't
+if it tries to evaluate a function during type checking that doesn't
```
CodingCellist, line 1402: "that" ~> "which"

docs/source/tutorial/windows.rst
```diff
@@ -6,21 +6,21 @@ Prerequisites for Windows
...
-       under there, for example)
+       under there, for example).
     3. In the directory where you installed MSYS2, find the file
        ``mingw64.ini`` and add the line ``MSYS2_PATH_TYPE=inherit``.
        This adds the normal windows PATH to the shell in MSYS2.
-    4. Start MSYS2 (click on mingw64.exe, as the icon in the start
-       menu won't pick up the MSYS2_PATH_TYPE from the ini, it can be
-       added to the system settings)
+    4. Start MSYS2 (click on mingw64.exe, as the icon in the Start menu
+       won’t pick up the MSYS2_PATH_TYPE from the .ini file — it can
```
CodingCellist: Unicode single quote. And another em-dash.

Bersier:
Thank you @CodingCellist for reviewing! I'm going to reply to your feedback with my thoughts. But if you still think your feedback should be applied as such, and don't think arguing is worth it, feel free to simply tell me to follow your feedback (or you can also edit the PR directly), and I'll be happy to do so without further comments. You can also do so on a per-comment basis, of course.

docs/source/tutorial/interfaces.rst
Bersier: "because otherwise the ``n`` is already in scope" is problematic, though. ``n`` is in scope regardless. You could argue that the top-level ``n`` is in scope only if it's not shadowed. But this subsentence doesn't even refer to the top-level ``n``. Even then, it doesn't feel right. Perhaps something like "because otherwise ``n`` is already bound by the top-level definition" would make sense.

docs/source/tutorial/interfaces.rst
Bersier: The phrase "in terms of ``peekByteOff`` and the parameter size" is a restrictive modifier; it directly specifies how the default implementation is provided. Restrictive modifiers generally do not take a comma in English (including British English).

docs/source/tutorial/interfaces.rst
Bersier: I think it makes sense to use a hyphen when not using a code font: "do-notation". But as code, ``do`` behaves like a noun, thus technically would not warrant hyphenation. I do get that it might look more familiar with a hyphen, though. Perhaps it would make more sense to simply write "do-notation", without code font?

docs/source/tutorial/interfaces.rst
Bersier: The comma here would also be more common in British English, although it's true it is more easily omitted in British English.

docs/source/tutorial/interfaces.rst
Bersier: "neither is valid" implies that we only return ``Nothing`` if both are invalid. But that's not the case here, right?

docs/source/tutorial/interfaces.rst
Bersier: The prepositional phrase "by using the sort function…” is restrictive, therefore the comma is technically incorrect. A variant could be "We can sort it using the default ``Ord`` implementation. This can be done by using the ``sort`` function, available via ``import Data.List``.

The em-dash was the only way I found to reconcile the fact that both prompts are shown after this paragraph, even though the first one relates to the first sentence. The original was a single sentence, but had its own issues.

docs/source/tutorial/typesfuns.rst
Bersier: "or" also implies that both are examples. Moreover, the list is not exhaustive. There's also bar : Char and quux : Bool. "and" would more so imply these are all the examples.

docs/source/tutorial/typesfuns.rst
Bersier: "that" is generally preferred for restrictive (defining) clauses, even in British English.

docs/source/tutorial/typesfuns.rst
Bersier: This is not a mainstream convention. "runtime" is less commonly used to refer to the underlying system in the first place. It's usually simply used to refer to what happens at runtime. "run time error" (without hyphen) is generally considered incorrect.

docs/source/tutorial/interfaces.rst
Bersier: Changed to semicolon.

docs/source/tutorial/interfaces.rst
Bersier: Reverted.

docs/source/tutorial/typesfuns.rst
Bersier: Done.

docs/source/tutorial/typesfuns.rst
Bersier: There were already lots of unicode quotes, so I assumed that is the preferred way to represent them outside of code. Should I replace all existing ones?

docs/source/tutorial/typesfuns.rst
Bersier: "provided" would suggest that the arguments are made available to the user by someone else (e.g., the compiler or library).

"given" is fine, except that it's slightly repetitive (and thus slightly off-putting), because it was just used in the line above, and with a different meaning.

docs/source/tutorial/typesfuns.rst
Bersier: "via pattern matching" is common as well. Reverted.

docs/source/tutorial/typesfuns.rst
Bersier: Reverted.

docs/source/tutorial/typesfuns.rst
Bersier: Changed to "and which".

docs/source/tutorial/windows.rst
Bersier: I think this qualifies as an interruption. Neither comma nor semicolon seem to fit here.

docs/source/tutorial/interfaces.rst
CodingCellist: > Perhaps something like "because otherwise `n` is already bound by the top-level definition" would make sense.

I think that is possibly better? Your call.

docs/source/tutorial/interfaces.rst
CodingCellist: TIL! Thanks  : )

docs/source/tutorial/interfaces.rst
CodingCellist: Technically, yes (I think). I would argue that it could be interpreted either ((heh)) way, but "neither is" implies combination (again, I think).

Maybe it should just be reworded entirely? "returning `Nothing` if any of the reads fail" (fail ~> are invalid ?). To me, "either is invalid" sounds a bit awkward, but I'm not sure how to address that...

docs/source/tutorial/interfaces.rst
CodingCellist: I think this might be a case of writing not strictly following the rules. Since `do` is a keyword, having it highlighted helps indicate that it is a part of the language; and this is then added on top of forming the compound noun (?) "do-notation".

docs/source/tutorial/interfaces.rst
CodingCellist: I more meant the "which". The comma looks out of place imo, but I can't formally argue it; it is more of a "there's too many commas there" kind of feeling...

docs/source/tutorial/interfaces.rst
CodingCellist: It just reads weird to me without it – too much uninterrupted flow/stream-of-words. :shrug:

I think the dash is fine, it should just be an en-dash rather than an em-dash (or the spaces should be removed on either side of the em-dash, but then that _is_ very American)

docs/source/tutorial/windows.rst
CodingCellist: In which case I think it should be an en-dash (the whole spaces/no-spaces thing again)  : )

CodingCellist (commented):
> Thank you @CodingCellist for reviewing! I'm going to reply to your feedback with my thoughts. But if you still think your feedback should be applied as such, and don't think arguing is worth it, feel free to simply tell me to follow your feedback (or you can also edit the PR directly), and I'll be happy to do so without further comments. 

Thanks for getting back so quickly on this  : )
I've replied to a couple of your comments, but please don't think of them as absolutes. I tend to err more on the side of what I feel reads "right" rather than grammar rules, but I also get that they exist for a reason  : )

docs/source/tutorial/interfaces.rst
Bersier: I switched to an en-dash.

docs/source/tutorial/interfaces.rst
Bersier: Done.

docs/source/tutorial/interfaces.rst
Bersier: Done.

docs/source/tutorial/interfaces.rst
Bersier: Added hyphens (including the ones I removed) to any such "notation" compound nouns (except "dot notation").

docs/source/tutorial/interfaces.rst
Bersier: I removed the commas around "for example" to make it less choppy.

docs/source/tutorial/interfaces.rst
Bersier: Done.

docs/source/tutorial/interfaces.rst
Bersier: Done.

docs/source/tutorial/windows.rst
Bersier: Done.

docs/source/tutorial/typesfuns.rst
Bersier: I tried another variant.

Bersier:
@CodingCellist Almost done! Below are the main remaining open points (I tried to address all the other ones):

1. "that" vs "which": I think it's more so a Scottish English thing (as less informal British English still favors "that" in restrictive clauses). Should I go ahead and revert the remaining introduced "which"->"that"s which I haven't addressed yet?
2. "runtime" vs "run-time": can you confirm that you want to use the convention you mentioned?
3. Unicode quotes: should I remove them all? Or just the ones I introduced?
4. Should I change other em-dashes to en-dashes? Just the ones I introduced?

docs/source/tutorial/interfaces.rst
```diff
@@ -62,9 +62,8 @@ going to use it to convert each element to a ``String``:
             show' (x :: Nil) = show x
             show' (x :: xs)  = show x ++ ", " ++ show' xs
 
-Note that we need the explicit ``forall n .`` in the ``show'`` function
-because otherwise the ``n`` is already in scope, and fixed to the value of
-the top level ``n``.
+Note that we need the explicit ``forall n .`` in the ``show'`` function,
+because n is already bound by the top-level definition.
```
jfdm: The markup indicated that `n` is code has been accidentally removed. It would be great to have the markup reinstated.

docs/source/tutorial/interfaces.rst
```diff
@@ -472,7 +470,7 @@ qualn ]`` where ``quali`` can be one of:
 
 - A *guard*, which is an expression of type ``Bool``
 
-- A let binding ``let x = e``
+- A ``let`` binding ``let x = e``
```
jfdm: In previous fixes, you inserted a hyphen in these instances, for example `do`-notation. A hyphen should be used here.

docs/source/tutorial/starting.rst
```diff
@@ -125,8 +125,8 @@ environment. You should see something like the following:
 
 .. literalinclude:: ../listing/idris-prompt-start.txt
 
-This gives a ``ghci`` style interface which allows evaluation of, as
-well as type checking of, expressions; theorem proving, compilation;
+This gives a ``ghci`` style interface which allows evaluation, as
+well as type checking, of expressions; theorem proving, compilation;
```
jfdm: It appears that a rogue comma is still there (between theorem proving and compilation) when a semi-comma should be inserted.

docs/source/tutorial/typesfuns.rst
jfdm: As there is no 'mainstream convention', the key thing here is that we are internally consistent within our documentation.

docs/source/tutorial/typesfuns.rst
```diff
@@ -1127,11 +1125,11 @@ Nested record fields can be accessed using the dot notation:
     x.a.b.c
     map (.a.b.c) xs
 
-For the dot notation, there must be no spaces after the dots but there may be
-spaces before the dots. The composite projection must be parenthesised,
-otherwise ``map .a.b.c xs`` would be understood as ``map.a.b.c xs``.
+For the dot notation, there must be no spaces after the dots, but there may be
```
jfdm: In other instances of `x`-notation, a hyphen has been inserted. We should be consistent within the documentation.

docs/source/tutorial/interfaces.rst
```diff
@@ -429,10 +427,10 @@ evaluated and then implicitly bound. Conceptually, we can think of
 
     (!) : m a -> a
 
-Note, however, that it is not really a function, merely syntax! In
+Note, however, that it is not really a function, but merely syntax! In
 practice, a subexpression ``!expr`` will lift ``expr`` as high as
 possible within its current scope, bind it to a fresh name ``x``, and
-replace ``!expr`` with ``x``. Expressions are lifted depth first, left
+replace ``!expr`` with ``x``. Expressions are lifted depth-first, left
 to right. In practice, ``!``-notation allows us to program in a more
```
jfdm, line 434: Whilst we are revisiting text, `!`-notation should probably be written as 'bang-notation'. 

Generally speaking, however, there is a larger debate to be had if the phrase 'x'-notation should include the name of the notation or the symbol itself.

Should it be do-notation or `dot`-notation? 

and referencing cases further on:

Should it be bang-notation or `!`-notation?

Should it be range-notation or `[A..B]`-notation?

docs/source/tutorial/typesfuns.rst
```diff
@@ -1322,22 +1320,22 @@ for building lists. The general form is:
...
 
 .. code-block:: idris
 
     pythag : Int -> List (Int, Int, Int)
     pythag n = [ (x, y, z) | z <- [1..n], y <- [1..z], x <- [1..y],
                              x*x + y*y == z*z ]
 
-The ``[a..b]`` notation is another shorthand which builds a list of
-numbers between ``a`` and ``b``. Alternatively ``[a,b..c]`` builds a
+The ``[a..b]``-notation is another shorthand which builds a list of
```
jfdm, line 1333: Having the form `x`-notation, where `x` is an example of the notation in use, may not be conducive to understanding what the notation is or what the notation is called. We should either use the name of the notation (i.e. range or enumerate) or rewrite as: "The notation, `[a..b]`,...".

docs/source/tutorial/interfaces.rst
Bersier: Of course. Sorry about this. I am planning to do a final pass at the end for these kinds of things.

docs/source/tutorial/starting.rst
Bersier: This sentence is weird, actually. I tried to improve it. Let me know if that works.

docs/source/tutorial/interfaces.rst
Bersier: Done.

docs/source/tutorial/typesfuns.rst
Bersier: In this case, "dot" is not in code font. And "dot notation" doesn't usually take a hyphen ("dot" is also a noun). Do you still want me to add a hyphen?

docs/source/tutorial/interfaces.rst
Bersier: I would argue that both are fine. Either using code font with the symbol used in code, or a name for the notation, not in code font. So

- do-notation and `do`-notation
- dot notation and `.`-notation
- bang notation and `!`-notation
- range notation and `[A..B]`-notation

(The latter three, in English, would usually not be hyphenated.)

The names are more pronounceable, of course, so defaulting to that seems reasonable. But it's not my call, of course. And it seems relatively minor.

docs/source/tutorial/typesfuns.rst
Bersier: @jfdm @CodingCellist Should I go ahead with this change?

docs/source/tutorial/interfaces.rst
jfdm: I think the key thing here is reader clarity, and consistency of use. I would advise that with first use in the text we *should* default to the names, or provide a clarifying statement after the symbol variant, and then be consistent. 

I would also suggest that for notations that involve many symbols, that the symbol version not be used as the name is not necessarily inferable from the symbol itself.

docs/source/tutorial/typesfuns.rst
jfdm: I would prefer that we change the writing to:

> The notation `[a..b]`, also known as range notation, supports construction of numbers in the range $(a,b)$.

docs/source/tutorial/interfaces.rst
CodingCellist: My personal preference would be to do both the first time a symbol-based notation is introduced, and then sticking to "how it would appear as code" for the rest of the document, for example:
"""
Idris provides `!`-notation ("bang"-notation) for [...]
Here, we can see `!`-notation used as `!expr` [...]
"""
(Modulo minor symbol/stylistic/phrasing choices)

This way the docs introduce the symbol, which is what the user will actually be writing in their code, while also immediately providing the way to pronounce and refer to it; a bit like introducing acronyms. Ideally I would also want that full version to be in the title of the page (if it is its own page).

docs/source/tutorial/typesfuns.rst
CodingCellist: The errors refer to it as "list range" notation, but I don't know if it can be used for things other than `List`?

CodingCellist:
> @CodingCellist Almost done! Below are the main remaining open points (I tried to address all the other ones):
> 
>     1. "that" vs "which": I think it's more so a Scottish English thing (as less informal British English still favors "that" in restrictive clauses). Should I go ahead and revert the remaining introduced "which"->"that"s which I haven't addressed yet?

Not too fussed at this point. I think it is probably fine in the grand scheme of things  : )

>     2. "runtime" vs "run-time": can you confirm that you want to use the convention you mentioned?

The [Idris2 paper](https://drops.dagstuhl.de/storage/00lipics/lipics-vol194-ecoop2021/LIPIcs.ECOOP.2021.9/LIPIcs.ECOOP.2021.9.pdf) uses "run-time", but the rest of the docs seem to use run-time/runtime interchangeably. I would _personally_ prefer if "run-time" meant "distinct from compile-/type-checking-time", and "runtime" meant "the system that runs the compiled code", but maybe switching to "runtime" for everything would lead to the fewest confusions/typos in future docs+contributions?

>     3. Unicode quotes: should I remove them all? Or just the ones I introduced?

If you could remove them all, that'd be fantastic! Thank you!

>     4. Should I change other em-dashes to en-dashes? Just the ones I introduced?

If it is trivial to do (a simple `sed` command or similar) then by all means. Otherwise, don't worry about other ones in the text.

docs/source/tutorial/typesfuns.rst
Bersier: Done.

Bersier:
> > ```
> > 2. "runtime" vs "run-time": can you confirm that you want to use the convention you mentioned?
> > ```
> 
> The [Idris2 paper](https://drops.dagstuhl.de/storage/00lipics/lipics-vol194-ecoop2021/LIPIcs.ECOOP.2021.9/LIPIcs.ECOOP.2021.9.pdf) uses "run-time", but the rest of the docs seem to use run-time/runtime interchangeably. I would _personally_ prefer if "run-time" meant "distinct from compile-/type-checking-time", and "runtime" meant "the system that runs the compiled code", but maybe switching to "runtime" for everything would lead to the fewest confusions/typos in future docs+contributions?
> 

I actually didn't find a single use of "runtime" to mean "the system that runs the compiled code" in the four files covered in this PR. Each time the runtime system is mentioned, the full term "runtime system" is used, in which the "runtime" adjective means the same thing as for the other uses, namely "distinct from compile-/type-checking-time".

docs/source/tutorial/interfaces.rst
Bersier: I'm not sure if there is consensus here. So I recommend a separate PR if someone wants to improve this.

Bersier:
@CodingCellist @jfdm Alright. I think I'm done. Let me know if I missed anything. If you still think some parts should be different, feel free to edit this PR directly. For additional "notation" changes, I think a new PR would be better.

jfdm:
I am good

CodingCellist (approved):
I was on holiday and only just got back, hence the slow reply. Thank you again @Bersier for doing all this and bearing with the numerous comments. LGTM!  : )

Bersier:
Thank you @CodingCellist and @jfdm!

Note: I cannot merge the PR.

jfdm:
Done! Thanks @Bersier and @CodingCellist

Issue #3400, opened 2024-10-19
Generate only es6


srghma:
- [ ] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

Let's allow to generate only es6 code

## Motivation

The purescript team decided moved from common js to es6, now purescript can generate only es6. Why? Because `require` is not pure, but `import export` is, allows to remove dead code

## The proposal

TODO

srghma:
## The proposal


Lets add optional field in `node:lambda:....:code`

this field can be any of 

```js
// will generate `import defaultExport_xxx from "module-name"`
// 
// in lambda all mentions of `defaultExport` will be
//  `code.replace(/\bdefaultExport\b/g, "defaultExport_xxx")` replaced with defaultExport_xxx 
// (so, be careful to not write `let defaultExport = ...`, though it will not lead to any errors)
import defaultExport from "module-name";

import * as name from "module-name";  // will be replaced with variable name_xxx
import { export1 } from "module-name";  // will be replaced with variable export1_xxx
import { export1 as alias1 } from "module-name"; // will be replaced with variable alias1_xxx
import { default as alias } from "module-name"; // will be replaced with variable alias_xxx
import { export1, export2 } from "module-name"; // will be replaced with variables export1_xxx and export2_xxx
import { export1, export2 as alias2, /* … */ } from "module-name"; // will be replaced with variables export1_xxx and alias2_xxx
import { "string name" as alias } from "module-name"; // will be replaced with variable alias_xxx
import defaultExport, { export1, /* … */ } from "module-name"; // will be replaced with variables export1_xxx
import defaultExport, * as name from "module-name"; // will be replaced with variable name_xxx
// import "module-name"; disallow, should be only in support.js
import styles from "https://example.com/styles.css" with { type: "css" }; // will be replaced with variable styles_xxx
```


to import multiple - separate them with `;`



### Examples

Right now we write

```idris
%foreign "C:idris2_readBufferData, libidris2_support, idris_file.h"
         "RefC:readBufferData"
         "node:lambda:(f,b,l,m) => require('fs').readSync(f.fd,b,l,m)"
prim__readBufferData : FilePtr -> Buffer -> Int -> Int -> PrimIO Int
```

After

```idris
%foreign "C:idris2_readBufferData, libidris2_support, idris_file.h"
         "RefC:readBufferData"
[...]

PR #3581, opened 2025-07-14
feat(#3400): generate only es6 -> implement

srghma:
# Description

## Self-check


andrevidela:
In accordance with the contributing guidelines, could you write down a description of what this is, what it addresses, why it should be accepted, why no other solution address that issue better, and what to look for during the review?

mattpolzin:
I'm going to close this PR due to inactivity (latest comment is a prompt for follow up).

In looking at the changes briefly, I see an opportunity to break things out into several PRs, each of which would still need to be well-described and justified. The changes to the FFI syntax for javascript alone are breaking and deserve discussion let alone the move to ES6, the introduction of Python scripts for linting, the renaming of "Javascript" to "Browser" in the backends, and more.

Not to say I don't see the benefits in some of the changes I am seeing here, but it's too much all at once with no accompanying discussion or written justification.

PR #3588, opened 2025-07-24
Replace WithFC by alias for WithData

andrevidela:
# Description

This change is part of #3570 

The goal is to gradually replace all constructor that tie meta-data to nodes with uses of `WithData` such that one can easily add metadata without changing the trees

### DIFF
--- a/src/Core/Core.idr
+++ b/src/Core/Core.idr
@@ -19,6 +19,9 @@ import Libraries.Data.WithData
 import public Data.IORef
 import System.File
 
+%hide Libraries.Data.Record.KeyVal.label
+%hide Libraries.Data.Record.LabelledValue.label
+
 %default covering
 
 public export
--- a/src/Core/Core.idr
+++ b/src/Core/Core.idr
@@ -3,6 +3,7 @@ module Core.Core
 import Core.Context.Context
 import Core.Env
 import Core.TT
+import public Core.WithData
 
 import Data.List1
 import Data.SnocList
@@ -13,6 +14,7 @@ import Libraries.Text.PrettyPrint.Prettyprinter
 import Libraries.Text.PrettyPrint.Prettyprinter.Util
 import Libraries.Text.PrettyPrint.Prettyprinter.Doc
 import Libraries.Data.Tap
+import Libraries.Data.WithData
 
 import public Data.IORef
 import System.File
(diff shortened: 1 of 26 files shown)


andrevidela:
I'd love to know why CI fails on a file I haven't touched

src/Core/Core.idr
```diff
@@ -857,9 +862,13 @@ namespace SnocList
   traverse_ : (a -> Core b) -> SnocList a -> Core ()
   traverse_ f xs = traverse_' f (reverse xs)
 
+%inline export
+traverseData : (ty -> Core sy) -> WithData fs ty -> Core (WithData fs sy)
```
gallais: I think `traverseX` are a wart of the current implementation so if we are introducing
new definitions, let's try to avoid that.

Can we please instead have `traverse` in a `namespace Data`? This way we can use the
short name `traverse` and only disambiguate with `Data.traverse` if Idris cannot figure
it out automatically.

src/Idris/Pretty.idr
```diff
@@ -185,16 +185,16 @@ mutual
...
-    prettyPrec d (NewPi (MkFCVal fc (MkPBinderScope (MkPBinder Explicit bind) scope))) =
+    prettyPrec d (NewPi (MkWithData fc (MkPBinderScope (MkPBinder Explicit bind) scope))) =
       lparen <+> pretty bind <+> rparen <++> arrow <++> prettyPrec d scope
-    prettyPrec d (NewPi (MkFCVal fc (MkPBinderScope (MkPBinder AutoImplicit bind) scope))) =
+    prettyPrec d (NewPi (MkWithData fc (MkPBinderScope (MkPBinder AutoImplicit bind) scope))) =
       lcurly <+> auto_ <++> pretty bind <+> rcurly <++> arrow <++> prettyPrec d scope
-    prettyPrec d (NewPi (MkFCVal fc (MkPBinderScope (MkPBinder (DefImplicit x) bind) scope))) =
+    prettyPrec d (NewPi (MkWithData fc (MkPBinderScope (MkPBinder (DefImplicit x) bind) scope))) =
       lcurly <+> default_ <++> prettyPrec appPrec x
       <++> pretty bind <+> rcurly <++> arrow <++> prettyPrec d scope
```
gallais: Is this a hint we should refactor this to have `NewPi` & a separate pretty instance for the `PBinderScope`
itself separating the `PBinder` from the constnatn `<++> prettyPrec d scope` bit?

src/Core/Core.idr
andrevidela: good point, will fix

src/Idris/Pretty.idr
andrevidela: I refactored it to its own Pretty instance

PR #3590, opened 2025-08-05
docs: started work on a type reference

wizard7377:
# Description

Created a reference for types


docs/source/reference/types.rst
```diff
@@ -0,0 +1,163 @@
...
+or
+
+.. code:: idris
+
+    vectorMap {n = m} f v
+
+These two examples show the power of Idris implicits: while they can be inferred by the compiler, if necessary (or desired) they may be specified by the user
+
+.. tip:: 
+    Idris *not have* a ``forall`` keyword, unlike Haskell.
```
joelberkeley: Idris does have a `forall`, but I don't know if it differs in meaning from that in Haskell

joelberkeley: `forall x . <stuff>` is `{0 x : _} -> <stuff>`

docs/source/reference/types.rst
```diff
@@ -0,0 +1,163 @@
...
+    vectorMap {n = m} f v
+
+These two examples show the power of Idris implicits: while they can be inferred by the compiler, if necessary (or desired) they may be specified by the user
+
+.. tip:: 
+    Idris *not have* a ``forall`` keyword, unlike Haskell. 
+    To get the equivalent of ``forall (x : k). A``, we use ``{x : k} -> A``. 
+    Not only is this more elegant, it more accurately highlights the connection between universal quantification, implication, and functions
+
+Idris has a *two* more types of implicits, however.
```
joelberkeley: Is it correct to say "type of implicit"? I always thought of them as modifiers on implicits

docs/source/reference/types.rst
```diff
@@ -0,0 +1,163 @@
...
+    This is because, as stated before, instead of ``forall``'s we have implicits. 
+    This means that if you want to use a type variable, say ``a`` in ``castOne : Cast a b => a -> [b]`` in the function, you must explicitly take it as an argument, e.g ``castOne : {a : Type} -> Cast a b => a -> [b]``, and then use it as ``castOne {a} x``
+
+^^^^^^^^^^^^^
+Quantities
+^^^^^^^^^^^^^
+
+Function types may also give quantities to their arguments.
+These are either ``0``, ``1``, or nothing (many), which are all placed before a named parameter, so, we have ``(# x : A)``.
+A value with a multiplicity of zero (erased) must only be used in the type signature (i.e, a Haskell type variable).
```
joelberkeley: this is not quite true. I can write
```idris
f : (0 x : Nat) -> ()
f x = let 0 y = x in ()
```
which is perhaps a silly example, cos `y` is never used, but it could be used in the types of what follows

docs/source/reference/types.rst
```diff
@@ -0,0 +1,163 @@
...
+
+The type system in Idris 2 is quite powerful, and is the main feature that sets it apart from simply typed languages like Haskell or OCaml.
+
+There are many ways to construct types, those being members of ``Type``, in Idris.
+
+-------------
+Functions
+-------------
+
+Function types, formed by ``(x : A) -> B``, are valid types, given that ``A`` is a valid type *or* kind and ``B`` is a valid type *or* kind.
```
andrevidela:
```suggestion
Function types, formed by ``(x : a) -> b``, are valid types, given that ``a``  and ``b`` are valid types.
```

protip: use lower case for generic names for types since they are automatically quantified in the compiler. It should probably be written in this file that `map : (a -> b) -> List a -> List b` is actually `map : {0 a, b : Type} -> (a -> b) -> List a -> List b` because `a` and `b` are lowercase but `List` is uppercase so it's not automatically added as an argument.

Also Idris does not have kinds, only types.

Otherwise good work, keep at it

docs/source/reference/types.rst
```diff
@@ -0,0 +1,163 @@
...
+    This means that if you want to use a type variable, say ``a`` in ``castOne : Cast a b => a -> [b]`` in the function, you must explicitly take it as an argument, e.g ``castOne : {a : Type} -> Cast a b => a -> [b]``, and then use it as ``castOne {a} x``
+
+^^^^^^^^^^^^^
+Quantities
+^^^^^^^^^^^^^
+
+Function types may also give quantities to their arguments.
+These are either ``0``, ``1``, or nothing (many), which are all placed before a named parameter, so, we have ``(# x : A)``.
+A value with a multiplicity of zero (erased) must only be used in the type signature (i.e, a Haskell type variable).
+A value with a multiplicity of one (linear) must be used exactly once in the definition of the value.
```
joelberkeley: what do you mean by "used exactly once in the definition of the value"?

docs/source/reference/types.rst
```diff
@@ -0,0 +1,163 @@
...
+
+---------------------------------------
+Interfaces and Implementations
+---------------------------------------
+
+Interfaces are the Idris equivalent of Haskell type-classes.
+They are modeled as a lot of sugar on implicit types.
+There syntax is as follows:
```
joelberkeley, line 81:
```suggestion
Their syntax is as follows:
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,163 @@
...
+
+Idris has a *two* more types of implicits, however.
+The first, while not that important to the type system of Idris, is *incredibly* useful, and it is the default implicit.
+Essentially "regular" implicits only try to figure out the type without adding any extra information.
+
+Default implicits, on the other hand, instead of trying to figure out a value, will instead just use a default, which means that unnecessary "simple" functions are unnecessary in Idris: you just use the default value.
+
+The syntax for default implicits is ``{default val x : A} -> B``, where ``val : A``, which is the default value for ``x``.
+
+The third kind of default are auto defaults.
```
joelberkeley: I think you've put "default" in this para where you meant "implicit"

docs/source/reference/types.rst
```diff
@@ -0,0 +1,163 @@
...
+Auto defaults are most useful in proofs, where they differ from regular defaults by irrelevance.
```
joelberkeley: is this true? I think there are cases where you'd want an auto implicit to be relevant, to pattern match on it

wizard7377:
Made some preliminary changes

docs/source/reference/types.rst
```diff
@@ -0,0 +1,233 @@
...
+The first, while not that important to the type system of Idris, is *incredibly* useful, and it is the default implicit.
+Essentially "regular" implicits only try to figure out the type without adding any extra information.
+
+Default implicits, on the other hand, instead of trying to figure out a value, will instead just use a default, which means that unnecessary "simple" functions are unnecessary in Idris: you just use the default value.
+
+The syntax for default implicits is ``{default val x : a} -> b``, where ``val : a``, which is the default value for ``x``.
+
+The third kind of implicits are auto implicits.
+Auto implicits are most useful in proofs, where they differ from regular implicits by irrelevance.
+That is, a regular default will never initiate *anything* that cannot otherwise be initiated, while an ``auto`` default will not change anything outside of the input that cannot be initiated.
```
joelberkeley: what does the second half of this sentence mean?

docs/source/reference/types.rst
```diff
@@ -0,0 +1,233 @@
...
+
+For instance, if we want to automatically pass a proof through some functions, we use ``auto`` defaults.
```
joelberkeley:
```suggestion
For instance, if we want to automatically pass a proof through some functions, we use ``auto`` implicits.
```

docs/source/reference/types.rst
joelberkeley:
```suggestion
That is, a regular implicit will never initiate *anything* that cannot otherwise be initiated, while an ``auto`` implicit will not change anything outside of the input that cannot be initiated.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,233 @@
...
+Very roughly, ``0`` corresponds to a value "being ignored" or *erased*, a value of ``1`` corresponds to a value being "used" once, or being *one-to-one*, and ω (which, in Idris syntax is ommited) corresponds to a type having an unrestricted number of uses.
+
+Note that this idea of "usage" is not always cut and dry.
+For instance, as we saw above, ``{0 a : Type} -> {0 b : Type} -> (a -> b) -> List a -> List b`` is a valid type, despite the fact it uses an erased arguement.
+This is because of the fact that erased arguements are only ignored at runtime, however, in the above, the usages of ``a`` and ``b`` are in a type, so they are fine.
+In addition, we can alias values without changing their multiplicity, so long as we treat them as the same value.
+For instance, if ``1 x : a``, ``let 1 y = x in EXPR`` is valid so long as exactly *one of* ``x`` or ``y`` is used in ``EXPR``.
+
+The syntax for specifying multiplicity in functions is ``(# x : a) -> B`` (or any variation on such), where ``#`` is the multiplicty.
+In addition, we may also declare ``let`` binding with a multiplictity, with the syntax ``let # x : a = val in EXPR`` or ``let # x = val in EXPR``
```
joelberkeley: I didn't quite mean to add the mention of `let`, since there are a number of other places where you'd do this (e.g. in a "type alias")
```idris
0 Numbers : Type
Numbers = List Int
```
I more mentioned it to point out that function parameters isn't the only place this happens

docs/source/reference/types.rst
```diff
@@ -0,0 +1,233 @@
...
+---------------------------------------
+Interfaces and Implementations
+---------------------------------------
+
+.. error::
+    Add info on interface constructors
+
+Interfaces are the Idris equivalent of Haskell type-classes.
+They are modeled as a lot of sugar on implicit types.
+There syntax is as follows:
```
joelberkeley:
```suggestion
Their syntax is as follows:
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,233 @@
...
+
+    [INAME] SPEC where
+        DEFS
+
+
+Where ``SPEC`` must be one of the following
+
+* ``... -> SPEC``, such that this would be valid type, and such that ``SPEC`` is itself a valid form
+* ``... => SPEC``, similar
+* For an ``interface`` deceleration, ``NAME ARG``, where ``NAME`` is the name of the interface being defined, and all of ``ARGS`` are valid types
```
joelberkeley:
```suggestion
* For an ``interface`` declaration, ``NAME ARG``, where ``NAME`` is the name of the interface being defined, and all of ``ARGS`` are valid types
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,233 @@
...
+
+----------------------
+Top Level Types
+----------------------
+
+The two main ways to declare types at the top level are ``record`` and ``data``
+
+.. note::
+    Type synonyms, type families, and the like from Haskell are all just normal functions from ``Type`` to something else.
+    In addition, this means that one can have anonymous instances, so, for instance, ``Monad (\x => x)`` is a perfectly valid form
```
joelberkeley: I believe that interface implementations over functions are discouraged, because type inference fails and Idris can't resolve the instances

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+As shorthand, for non-dependent types, ``a -> b`` is the same as ``(_ : a) -> b``.
+
+In addition, to do simple universal quantification, one can write ``forall x . a``, where ``x`` must be only a name (it can't have a type qualifier).
+This is desugared to ``{0 x : _} -> a``.
+
+^^^^^^^^^^^^^
+Implicits
+^^^^^^^^^^^^^
+
+Implicit function types, those where the value of a parameter can be inferred from the context, can take three forms.
```
andrevidela:
```suggestion
Implicit arguments, can take three forms.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+The first of these, ``{x : A} -> B``, forms the basis "basic" implicit types.
+That is, ``x`` should be inferred *syntactically*.
```
andrevidela:
```suggestion
The first one, ``{x : a} -> b`` is the most basic form of implicit arguments, that is
 ``x`` should be inferred *syntactically*.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+This is the equivalent of implicit ``forall`` in Haskell or Rocq, where we might state that ``vectorMap`` on a vector might be:
```
andrevidela: We should add a footnote here pointing people toward the difference between polymorphism and dependent types. I don't think it belongs in this document as a paragraph but it should probably be mentionned somewhere

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+
+    vectorMap f v
+
+or
+
+.. code:: idris
+
+    vectorMap {n = m} f v
+
+These two examples show the power of Idris implicits: while they can be inferred by the compiler, if necessary (or desired) they may be specified by the user
```
andrevidela:
```suggestion
These two examples show the power of Idris implicits: while they can be inferred by the compiler, if necessary (or desired) they may be specified by the user.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Idris has *two* additional modifiers of implicits, however.
+The first, while not that important to the type system of Idris, is *incredibly* useful, and it is the default implicit.
```
andrevidela:
```suggestion
The second form of implicits arguments is the default implicit.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+Essentially "regular" implicits only try to figure out the type without adding any extra information.
+
+Default implicits, on the other hand, instead of trying to figure out a value, will instead just use a default, which means that unnecessary "simple" functions are unnecessary in Idris: you just use the default value.
```
andrevidela:
```suggestion
Where "regular" implicits only try to resolve the type without adding any extra information, default implicit arguments, will instead use a default value when the value cannot be resolved syntactically.

This is *incredibly* useful to avoid repeating the same function definition multiple times with a different set of arguments when the most common case can be given via a default implicit and the other uses can give the argument by name.
```

I tried to combine you previous sentence about how useful it is with the one explaning how to use default implicits.

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+
+The syntax for default implicits is ``{default val x : a} -> b``, where ``val : a``, which is the default value for ``x``.
+
+The third kind of implicits are auto implicits.
+Auto implicits are most useful in proofs, where they differ from regular implicits by there purpose.
```
andrevidela: That's somewhat debatable. I suggest removing this sentence and instead show multiple uses for auto-implicits:

- proofs (like `Elem` or `IsJust`)
- interfaces
- effects (like how the compiler has tags to track allowable effects)

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+Regular implicits try to find a "necessary" match.
+``auto`` implicits, on the other hand, look for *any* possible construction of a value.
```
andrevidela:
```suggestion
Regular implicits try to find a "necessary" match that is in scope.
 ``auto`` implicits, on the other hand, look for *any* possible construction of a value, even if it is not in scope, the compiler will attempt to construct a value for it.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+This is useful for propositions, where we often don't need to reason about the specific value of a type member, merely that it exists.
+Essentially, auto implicits are for cases where we want the compiler to have more freedom to infer the value
```
andrevidela:
```suggestion
```
This sentence needs to go. This is actually a poor use for auto implicits since they are unpredictable and the compiler will generate different values for the same type depending on constructor definition order and other compiler internal details. It is much better to provide it with a decision procedure with a unary constructor than use it as a way to get more leeway out of the term-inference machinery.

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Default implicits, on the other hand, instead of trying to figure out a value, will instead just use a default, which means that unnecessary "simple" functions are unnecessary in Idris: you just use the default value.
+
+The syntax for default implicits is ``{default val x : a} -> b``, where ``val : a``, which is the default value for ``x``.
+
+The third kind of implicits are auto implicits.
+Auto implicits are most useful in proofs, where they differ from regular implicits by there purpose.
+Regular implicits try to find a "necessary" match.
+``auto`` implicits, on the other hand, look for *any* possible construction of a value.
+This is useful for propositions, where we often don't need to reason about the specific value of a type member, merely that it exists.
```
andrevidela:
```suggestion
This is useful for properties of types where we don't want to change the definition of the type but carry along some facts about it.
```

Not quite sure what the original sentence meant but I thought this was a more illuminating use-case for auto implicits. It's often the case that it's the best way to carry along a proof from a library type, because building your own is a hassle.

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+    If desired, this behavior can be toggled off by ``%unbound_implicits off`` (and back on by ``%unbound_implicits on``)
+
+
+
+^^^^^^^^^^^^^
+Quantities
+^^^^^^^^^^^^^
+
+Idris 2 uses Quantitative Type Theory (`QTT`_) as its theoretical basis.
+In QTT, every function type has an associated multiplicity, ``0``, ``1``, or ω.
```
andrevidela:
```suggestion
In QTT, every variable has an associated multiplicity, ``0``, ``1``, or ω.
```

Essentially every binder  `a : b` has a quantity attached to them: lambdas, pi-types, record fields, toplevel definitions, etc. probably important to explain that in some way. "Variable" is probably good enough for an entry-level document

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+These correspond with how many times the given input is "used" in the result type.
+
+Very roughly, ``0`` corresponds to a value "being ignored" or *erased*, a value of ``1`` corresponds to a value being "used" once, or being *one-to-one*, and ω (which, in Idris syntax is omitted) corresponds to a type having an unrestricted number of uses.
```
andrevidela:
```suggestion
Very roughly, ``0`` corresponds to a value existing only during typechecking but not at runtime. 
A value of ``1`` corresponds to a value being used exactly once, and ω (which, in Idris syntax is omitted) corresponds to a type having an unrestricted number of uses.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+
+Note that this idea of "usage" is not always cut and dry.
```
andrevidela: Usage is cut and dry. If your variable appears at runtime then it's used, erased quantities cannot appear at runtime, period.

The only loose rules are around subtyping of ω arguments and linearity but that shouldn't be something the audience of this text needs to worry about.

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+For instance, as we saw above, ``{0 a : Type} -> {0 b : Type} -> (a -> b) -> List a -> List b`` is a valid type, despite the fact it uses an erased argument.
```
andrevidela: There is some confusion here.

The _type_ itself `{0 a : Type} -> {0 b : Type} -> (a -> b) -> List a -> List b` is not bound to anything, so we cannot talk about its multiplicity. 

If you bind it to something like a top-level definition
```
fn : {0 a : Type} -> {0 b : Type} -> (a -> b) -> List a -> List b
```

Then nothing is wrong because the type is used in a `0` context

if you bind it as a return value
```
ty : Type
ty = {0 a : Type} -> {0 b : Type} -> (a -> b) -> List a -> List b
```
Then you're in trouble because you're instanciating variables `a` and `b` potentially at runtime (because of the ω multiplicity of `ty`) but you marked them as `0`. To fix this you have to mark the binder itself as `0`:

```
0 ty : Type
ty = {0 a : Type} -> {0 b : Type} -> (a -> b) -> List a -> List b
```

Now everything works again because `ty` is never going to appear at runtime.

I'll let you repharse this in a way that makes sense for you and the rest of the text.

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+This is because erased arguments are only ignored at runtime, however, in the above, the usages of ``a`` and ``b`` are in a type, so they are fine.
+In addition, we can alias values without changing their multiplicity, so long as we treat them as the same value.
+For instance, if ``1 x : a``, ``let 1 y = x in EXPR`` is valid so long as exactly *one of* ``x`` or ``y`` is used in ``EXPR``.
```
andrevidela: that is not true. Only `y` can be use in the body `EXPR`. `x` has been "spent" by being bound to `y`. And you can see this by writing a small program:

```
fn : (1 x : Nat) -> IO ()
fn x = let 1 y = x in ?hole
```

The hole has context:
```
 0 x : Nat
 1 y : Nat
hole : IO ()
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+This is one of many examples of how the quantitative notion of "usage" is not always exactly the same as the intuitive notion of "usage".
```
andrevidela: I think this sentence should stay becaues there are indeed many confusing uses for QTT but the example given doesn't illustrate that. I think we need more example where something is "spent" but can still be used in a `0` context. There is a couple of examples in the `linear` library.
It's probably not worth going on a tangent over this but there should be a way to mention it.

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+
+The syntax for specifying multiplicity in functions is ``(# x : a) -> B`` (or any variation on such), where ``#`` is the multiplicity.
```
andrevidela:
```suggestion
The syntax for specifying multiplicity in functions is ``(qty x : a) -> b`` (or any variation on such), where ``qty`` is the multiplicity either `0`, `1` or nothing for ω .
```

we should remain consistent with our use of letters and grammar placeholders. I suggest we use `qty` for quantities, `x` for variables `type` for types and `expr` for expressions, even if `type` and `expr` are technically the same, it should help the reader create a consistent mental model. And keep `a` `b` `c` as generic types

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+In addition, we may also declare ``let`` binding with a multiplicity, with the syntax ``let # x : a = val in EXPR`` or ``let # x = val in EXPR``.
```
andrevidela:
```suggestion
In addition, we may also declare ``let`` binding with a multiplicity, with the syntax ``let qty x : a = val in expr`` or ``let qty x = val in expr``.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+
+To declare (Agda style) Generalized Algebraic Data Types, we must write ``data NAME : KIND where CONS``.
+``KIND`` must be a kind (in Idris there isn't actually distinction between types and kinds) whose tail ultimately ends up returning the type ``Type``.
+So, for instance, ``data Vect : (n : Nat) -> Type where ...`` is valid, whilst ``data Vect : forall k. (n : Nat) -> k where ...`` is not.
+Each ``CONS`` must be a function declaration, except that the name being defined must start with an uppercase letter and the ultimate return type of it must be a fully applied instance of ``NAME``.
+
+The other way to define top level types, ``record``, is documented in detail on the `records`_ page.
+
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Existentials in Types
```
andrevidela:
```suggestion
Inferable Placeholders
```

I don't know what to call this class of terms but `existential` is extremely confusing because there is already something called `existential types` in programming and it's not what is exposed here. Existential types are usually an encoding of sigma-types for languages without dependent types.

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+While not restricted to types, existentials are so prevalent within types that there usage bears mentioning.
+Existentials are always of the form ``_``, ``?`` or ``?NAME``, where ``NAME`` is a valid name.
+Named existentials are scoped by *module*, however, they may only exist in one term.
```
andrevidela:
```suggestion
There is a class of expression that allow to call directly into Idris's inference machine, they are ``_``, ``?`` and ``?identifier``, where ``identifier`` is a valid variable name.

``_`` tells the compiler to wait for this expression to be resolved by another constraint, this is like a normal implicit argument.
``?`` tells the compiler to try to guess what the value is, this is like an auto-implicit.
``?identifiers`` is a **hole**. It allows you to query the type of the value at that point in the program.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+
+.. caution::
+
+    Existentials are scoped over modules.
```
andrevidela:
```suggestion
    Holes are scoped over modules.
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+    This means that a given hole must have a unique name in a module.
+
+Named existential forms an equivalent to "typed holes", that is, a value to be filled in later.
+These can occur in types, although typically occur in generated ones.
+Most often, this will occur in usages of generic types, where, for instance, given ``show : {a : Type} -> Show a => a -> String``, we transform the context ``show : {a : Type} -> Show a => a -> String, y = show x |- y : String`` to ``x : ?a, Show ?a |- y : String`` [#f2]_ .
+
+``?`` is similar except, while if ``?a``` cannot be inferred it is made it a typed hole, if ``?`` cannot be inferred, it raises an error
```
andrevidela:
```suggestion
```

`?` is completely different as it infers a value in addition to the type. Holes only infer the type.

docs/source/reference/types.rst
```diff
@@ -0,0 +1,286 @@
...
+Named existentials are scoped by *module*, however, they may only exist in one term.
+
+.. caution::
+
+    Existentials are scoped over modules.
+    This means that a given hole must have a unique name in a module.
+
+Named existential forms an equivalent to "typed holes", that is, a value to be filled in later.
+These can occur in types, although typically occur in generated ones.
+Most often, this will occur in usages of generic types, where, for instance, given ``show : {a : Type} -> Show a => a -> String``, we transform the context ``show : {a : Type} -> Show a => a -> String, y = show x |- y : String`` to ``x : ?a, Show ?a |- y : String`` [#f2]_ .
```
andrevidela: I think this needs to be reworked in light of the previous changes

andrevidela (changes requested):
I skipped the middle because I thought this would be quick but I spent way longer than I thought. There is enough to work with before tackling the middle anyways

docs/source/reference/types.rst
```diff
@@ -0,0 +1,297 @@
...
+-------------
+Functions
+-------------
+
+
+Function types, formed by ``(x : a) -> b``, are valid types, given that ``a`` and ``b`` are valid types
+As shorthand, for non-dependent types, ``a -> b`` is the same as ``(_ : a) -> b``.
+
+In addition, to do simple universal quantification, one can write ``forall x . a``, where ``x`` must be only a name (it can't have a type qualifier).
+This is desugared to ``{0 x : _} -> a``.
```
andrevidela, line 17: This should be moved to the next section, we haven't introduced implicits or quantities yet so there is no point in writing down what the desugaring is. Also there is no universal quantification in Idris although there is a `forall` keyword, calling it that might be misleading.I'll let you judge how to rephrase that,maybe there is another sentence that also uses the phrase "universal quantification" that isn't suggesting Idris has universal quantification

docs/source/reference/types.rst
```diff
@@ -0,0 +1,297 @@
...
+``_`` tells the compiler to wait for this expression to be resolved by another constraint, this is like a normal implicit argument.
+``?`` tells the compiler to try to guess what the value is, this is like an auto-implicit.
+``?identifiers`` is a **hole**. It allows you to query the type of the value at that point in the program.
+
+.. caution::
+
+    Holes are scoped over modules.
+
+Named holes occur most often in the context of type inference.
+Most often, this will occur in usages of generic types, where, for instance, given ``show : {a : Type} -> Show a => a -> String``, we transform the context ``show : {a : Type} -> Show a => a -> String, y = show x |- y : String`` to ``x : ?a, Show ?a |- y : String`` [#f2]_ .
```
andrevidela, line 264: I don't think the example is very insightful if all the steps are skipped. It's also a lot to introduce new syntax 90% of the way through the document

docs/source/reference/types.rst
```diff
@@ -0,0 +1,297 @@
...
+Other Forms
+----------------------
+
+The two main ways to declare types at the top level are ``record`` and ``data``
+
+.. note::
+    Type synonyms, type families, and the like from Haskell are all just normal functions from ``Type`` to something else.
+    In addition, this means that one can have anonymous instances, so, for instance, ``Monad (\x => x)`` is a perfectly valid form.
+
+    Even so, it is in general not a good idea to create such instances
```
andrevidela, line 234:
```suggestion
```

If it's not a good idea to do it, we should not encourage users to do it

docs/source/reference/types.rst
```diff
@@ -0,0 +1,297 @@
...
+    semiList = (MkSemigroup (++))
+
+----------------------
+Other Forms
+----------------------
+
+The two main ways to declare types at the top level are ``record`` and ``data``
+
+.. note::
+    Type synonyms, type families, and the like from Haskell are all just normal functions from ``Type`` to something else.
```
andrevidela, line 231:
```suggestion
    Type synonyms, type families, and the like from Haskell are all just normal functions ``Type -> a``
```

docs/source/reference/types.rst
```diff
@@ -0,0 +1,297 @@
...
+    Even so, it is in general not a good idea to create such instances
+
+``data`` declares (General) Algebraic Data Types ((G)ADTs).
+The Haskell style, simple ADT form is always a statement of the form ``data NAME ARGS = ALT {| ALT}``.
+``ARGS`` must be a valid list of names.
+In addition, each ``ALT`` must be of the form ``CONS PARAMS`` where each ``CONS`` is either an operator or a name (that will be declared as a constructor).
+Finally, each of ``PARAMS`` must be a valid type.
+
+To declare (Agda style) Generalized Algebraic Data Types, we must write ``data NAME : KIND where CONS``.
+``KIND`` must be a kind (in Idris there isn't actually distinction between types and kinds) whose tail ultimately ends up returning the type ``Type``.
```
andrevidela, line 243: There are no kinds in Idris those are types

docs/source/reference/types.rst
```diff
@@ -0,0 +1,297 @@
...
+
+The two main ways to declare types at the top level are ``record`` and ``data``
+
+.. note::
+    Type synonyms, type families, and the like from Haskell are all just normal functions from ``Type`` to something else.
+    In addition, this means that one can have anonymous instances, so, for instance, ``Monad (\x => x)`` is a perfectly valid form.
+
+    Even so, it is in general not a good idea to create such instances
+
+``data`` declares (General) Algebraic Data Types ((G)ADTs).
```
andrevidela, line 236:
```suggestion
``data`` declares (Generalized) Algebraic Data Types ((G)ADTs).
```

a bit weird to call them that because there is no difference between normal data types and gadts in Idris. I would reformulate as "data can also declare types with indicies, that is, where the parameter is a value that can vary depending on the constructor used." 


It's actually the other `data` that's a special compiler case and it desugars to a data type with constant indices

docs/source/reference/types.rst
```diff
@@ -0,0 +1,297 @@
...
+The Haskell style, simple ADT form is always a statement of the form ``data NAME ARGS = ALT {| ALT}``.
+``ARGS`` must be a valid list of names.
+In addition, each ``ALT`` must be of the form ``CONS PARAMS`` where each ``CONS`` is either an operator or a name (that will be declared as a constructor).
+Finally, each of ``PARAMS`` must be a valid type.
+
+To declare (Agda style) Generalized Algebraic Data Types, we must write ``data NAME : KIND where CONS``.
```
andrevidela, line 242: Saying that adga has gadts is also weird to me. adga has indexed types and you can use them to write down parameterised data types. it's only haskell that works the other way around

andrevidela (commented):
Okay I've only gone through a little bit of this but looking a the overall document I wonder where in the documentation this would fit.

it's definitely not a `reference` document, it's more like an informal beginner tutorial.
It also repeats information that's already in the idris crash course but in a less structured way, that's not bad in itself but it makes it hard to categorise. It also seems to be aimed at beginners but also makes references to and assumes Haskell knowledge which is already a small niche and it already addressed by the crash course. Finally, it contains many more things than types such as interfaces, quantities and holees.

My suggestion is that this should be broken down into a smaller document that would sit in front of the crash course and would be something like "an idris overview in 5 minutes" and that would allow to talk about all the different topics with only surface-level detail.

Another option would be to keep the same audience, surface-level interest people with some FP experience, and expand it into an actual tutorial,but that's a lot more work so I do not suggest it

docs/source/reference/types.rst
stefan-hoeck: Actually, aren't these functions that _result_ in a type? For instance, a type synonym is just

```idris
Label : Type
Label = String
```

Same for type families? In fact, we hardly ever write functions of the type `Type -> a`.

wizard7377:
Probably a quick start at the beginning of the tutorial. I don't  think there's an equivalent of that yet and that does seem rather important, so I'll get to work shortening it and adding the rest of basic Idris

docs/source/reference/types.rst
andrevidela: Type families both result in a type (kind) and depend on types given as argument, I've put `a` because you could be working with multiple kinds not just `*`. open type families is closest to pattern matching on types, data families are like an indexed interface `a -> Type`. None of those comparaisons actually make any sense if you think too hard about it. Maybe the paragraph should be scraped because there isn't much pedagogical benefit in relating dependent types to haskell-specific features that don't quite match

docs/source/reference/types.rst
dunhamsteve: Yeah, I know there are people (see Conor's recent mastodon post) get offended if you refer to an indexed inductive type as a GADT.

docs/source/reference/types.rst
dunhamsteve: I don't think ? is like an auto-implicit (you need %search for that). I'm pretty sure it's solved by inference. The difference between _ and ? is that ? can be solved later. If you use _ in a type signature it's an error if it is not solved at the end of a signature. If you use ?, it can be solved later in the file. (An error if not solved by the end of the file.)

docs/source/reference/types.rst
andrevidela: I found the post, quite amusing for us. Not amusing for them

wizard7377:
Because of the substantial differences both in content and purpose of the above changes, I have moved the development of this to a new pull request at #3618

PR #3591, opened 2025-08-06
update readme to mention pack

andrevidela:
# Description

update description to direct people toward pack and explain how to use external libraries

work toward https://github.com/andrevidela/idris-documentation-tracking/issues/1


dunhamsteve:
Should we update INSTALL.md? 

It briefly mentions pack, but doesn't encourage it over other methods. I think pack may be the best path for users (non-windows users at least), especially if they want the LSP.

andrevidela:
I update `INSTALL.md` with more pack info and a recommendation for it.

PR #3592, opened 2025-08-06
[ fix ] `mempty` is called `neutral`

mjustus:


gallais:
We should ideally have test cases showing the fix works

Issue #3440, opened 2024-12-10
Very large case tree for a simple function


spcfox:
[Gist with code](https://gist.github.com/spcfox/4bb08293986988d91c32e67e10009083)

# Steps to Reproduce

This code typechecks on my machine in about 50 seconds and generates a huge `.ttc`.

```idris
data X : () -> Type where
  A : X i
  B : X i
  C : X i

data Y : ((i ** X i), (i ** X i)) -> Type where
  MkY : Y x

f : (a, b : ()) -> (x : X a) -> (y : X b) -> Y ((a ** x), (b ** y)) -> ()
f a  () A A MkY = ()
f () b  A B MkY = ()
f a  () B A MkY = ()
f a  b  B B MkY = ()
f a  () C A MkY = ()
f a  b  C B MkY = ()
f a  b  _ _ _   = ()
```

I have prepared various modifications to find the influencing factors. Compilation time correlates with the size of `.ttc`, so I will give the file sizes for comparison:

```shell
$ du -h --apparent-size *.ttc | sort -hr
46M     ExampleFull.ttc
44M     Example.ttc
17M     ExampleNoMatchLastY.ttc
8.2M    ExampleSwap.ttc
4.8M    ExampleNoPair.ttc
1.4M    ExampleSwapNoPair.ttc
56K     ExampleNoMatchY.ttc
13K     ExampleNoMatchAB.ttc
13K     ExampleErased.ttc
12K     ExampleMatchAll.ttc
```

The full code for all modifications is available in [gist](https://gist.github.com/spcfox/4bb08293986988d91c32e67e10009083). Here I will give a brief description of them:

- [`Example`](https://gist.github.com/spcfox/4bb08293986988d91c32e67e10009083#file-example-idr) — the code presented above
- [`ExampleFull`](https://gist.github.com/spcfox/4bb08293986988d91c32e67e10009083#file-examplefull-idr) — all implicit arguments are explicitly specified
- [`ExampleNoMatchLastY`](https://gist.github.com/spcfox/4bb08293986988d91c32e67e10009083#file-examplenomatchlasty-idr) — removed last matching on `Y`
- [`ExampleSwap`](https://gist.github.com/spcfox/4bb08293986988d91c32e67e10009083#file-exampleswap-idr) — swap the first two cases
- [`ExampleNoPair`](https://gist.github.com/spcfox/4bb08293986988d91c32e67e10009083#file-examplenopair-idr) — `Y` has two indices instead of one pair
[...]

PR #3598, opened 2025-08-07
[ refactor ] reduce size of two case trees in the compiler

dunhamsteve:
# Description

This PR reduces the size of the scheme code of Idris by 2MB (19%) by changing the order of arguments in two functions. Previously `idris2.ss` was 11,736,341 bytes and now it is 9,505,951 bytes.

Also, the link step (doing `idris2 --build idris2.ipkg` after building all of the .ttc files) went from 45.9 seconds to 40.8 seconds (11% reduction).

The technique that I used here was to do `--dumpcases` and then sort by line length. There were two outliers at the end of the file with huge case trees. I managed to fix the issue by reordering arguments. I had a similar issue in my own code, but I had to break up the function in that case.

### DIFF
--- a/src/Compiler/CompileExpr.idr
+++ b/src/Compiler/CompileExpr.idr
@@ -459,44 +459,44 @@ getNArgs defs (NS _ (UN $ Basic "Struct")) [n, args]
 getNArgs defs n args = pure $ User n args
 
 nfToCFType : {auto c : Ref Ctxt Defs} ->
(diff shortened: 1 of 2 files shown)


dunhamsteve:
We'll have to keep an eye out for regressions or similar issues if we change the case builder.

spcfox:
I think this may be related to #3440. I will try to analyse that example to find the cause

andrevidela (approved):
Seems like a no-brainer, let's get this in. Thanks a lot

Issue #3409, opened 2024-10-30
Investigate suspicious location tracking


andrevidela:
Some places where code is elaborated the file context is a bit strange, it's probably hiding a bug or at least not explaining itself properly.

## `ProcessDef.idr`

```
       processType eopts nest env fc top Public []
          $ MkImpTy fc fc n $ holeyType (map snd args)
          -- Why does the name have the same location as the rest
          -- of the entire expression? That seems incorrect
          $ MkImpTy fc (MkFCVal fc n) $ holeyType (map snd args)
```

## `Idris/Elab/Interface.idr`

Elaborated clauses inheriting a strange location

```
         let fnclause = PatClause vfc
                                  (INamedApp vfc
                                             -- Another suspicious location
                                             -- why do we take the location of the signature
                                             -- rather than the location of the name?
                                             (IVar sig.location cn.val)
                                             (UN $ Basic "__con")
                                             conapp
                                             )
```

### `Idris/Syntax.idr`

Record field name locations are not tracked

```
 record RecordField' (nm : Type) where
    constructor MkRecordField
    doc : String
    rig : RigCount
    piInfo : PiInfo (PTerm' nm)
    names : List Name -- Those names should be `WithFC`
    type : PTerm' nm
```

andrevidela:
## Possibly related

- #3417 
- #3408 
- #3279

PR #3599, opened 2025-08-07
Unify binder data types using WithData

andrevidela:
# Description

This addresses a number of points from #3570 and #3409 namely:
- track location of names in binder with multiple names #3409
- Replace ImpParameter by its own data type using WithData #3570
- replace WithName by WithData #3570
- unify binder datatypes between syntax and TTImp via `PiBindData` (overdue)

We are starting to enjoy the benefits of `WithData` since we can now remove redundant TTC instances and have them properly behave using the default `WithData` instance. There is still some work to do around Reflect/reify instance but that should be resolved once we don't need to maintain backward compatibility with elab reflection.


andrevidela:
It's been a week so I'll merge this now

Issue #3407, opened 2024-10-30
CLI flags interpreted as files if mistyped


eayus:
Currently, the typing the following (mistyped) command:
```
idris2 --chec Test.idr
```
Will launch the repl saying it can't find the file `--chec`. 

```
     ____    __     _         ___
    /  _/___/ /____(_)____   |__ \
    / // __  / ___/ / ___/   __/ /     Version 0.7.0-2bd88b7d3
  _/ // /_/ / /  / (__  )   / __/      https://www.idris-lang.org
 /___/\__,_/_/  /_/____/   /____/      Type :? for help

Welcome to Idris 2.  Enjoy yourself!
File error in --chec: File Not Found
Main>
```

This is pretty unusual behaviour, as most applications will error saying that `--chec` is not a valid flag. From a user perspective I would say its undesirable as they most likely made a typo and didn't want to start the REPL anyway. (I think there is another argument to be made that the REPL should not start if given non existent files in general).

eayus:
I've been thinking about this a bit and I feel like any robust solution will basically implement 90% of a generic command line arguments parser. I'm not sure how we feel about pulling in external libraries for the compiler (which might rely on getting the compiler packaged with a package manger like `pack`). To my knowledge there are two CLI libraries, [collie](https://github.com/ohad/collie) and [sap](https://github.com/eayus/sap) (my own library, so I'm probably biased).

The other thing I was going to suggest is that the CLI for Idris could be broken down into subcommands, something like:

`idris2 check ...`
`idris2 repl ...`
`idris2 compile ...`

The main advantage for this is that we can segregate flags into the different subcommands depending on where they are relevant. For example, a user checking a file does not need to know about the flags to set the compiler backend. This would simplify the help dialogues, and (in my opinion) provide a more intuitive interface. I'm not sure whether this is something that other people agree on? It would obviously break backwards compatibility for scripts using the interface.

buzden:
> To my knowledge there are two CLI libraries, [collie](https://github.com/ohad/collie) and [sap](https://github.com/eayus/sap)

There is at least https://github.com/idris-community/idris2-getopts, which is a pulled-out version from `contrib` (https://github.com/idris-lang/Idris2/blob/main/libs/contrib/System/Console/GetOpt.idr)

PR #3602, opened 2025-08-07
Detect unknown command line flags and error appropriately

eayus:
# Description

This makes a few related changes:

- Detect when a command line argument is given that begins with `--`, and error if it is not a valid flag. This prevents annoying cases like in #3407 . Technically this is a breaking change if someone has named their file beginning with `--`, though honestly I think that is on them..
- When command line arguments parsing fails, do not print out the help dialogue. This part is potentially debatable, but I think printing out the entire help dialogue is not useful as the error message is hidden at the top. I tested some compilers for other languages and I didn't find any which print the full dialogue on an incorrect flag like we were doing.
- Print command line argument errors to `stdout` and return a proper exit code.

One point of note is that errors produced by the command line parser are not coloured by default, unlike error messages produced by the rest of the compiler. This seems to be because we support flags that disable coloured output, and if parsing flags fails then we don't know whether to colour the error messages or not. I'm not sure how critical disabling colour output is, but having coloured error messages by default for command line argument error messages would be nice imo -- not sure what other people think.

## Self-check

### DIFF
--- a/tests/idris2/pkg/pkg021/run
+++ b/tests/idris2/pkg/pkg021/run
@@ -15,7 +15,7 @@ cp -R "${INSTALL_DIR}" ./secondary_install/
 mv "${INSTALL_DIR}/${TTC_VERSION}" "${INSTALL_DIR}/12345"
 
 echo "expect error:"
-idris2 -p testpkg < input
+idris2 -p testpkg < input 2>&1
 append_package_path ./secondary_install
 echo ""
 echo "expect success:"


eayus:
Any ideas why the MacOS tests would fail specifically? I can't think of anything that I've changed that could affect the failing tests, especially only on Mac...

andrevidela:
I think this is a bug in the test suite. There must be a race on mac os. I've restarted the tests for it

eayus:
Ah ok that makes sense, looks like they're passing now.

gallais:
Thank you!

PR #3607, opened 2025-08-11
[ re #3603 ] Actually bother introducing VarSet

gallais:
We may as well define the well-scoped structure.

Note that it's all unsafe and so we have some (useless) checks
in `toList` when extracting the list. But then again there should
almost never be any use case for extracting a list.

I have also made a couple of quadratic functions linear and
removed redundant code.

### DIFF
--- a/src/Core/Unify.idr
+++ b/src/Core/Unify.idr
@@ -18,6 +18,8 @@ import Data.Maybe
 
 import Libraries.Data.List.SizeOf
 
+import Libraries.Data.VarSet
+
 import Libraries.Data.IntMap
 import Libraries.Data.NameMap
 import Libraries.Data.SnocList.SizeOf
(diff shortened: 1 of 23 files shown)


src/Core/Unify.idr
```diff
@@ -283,50 +285,35 @@ unifyArgs mode loc env _ _ = ufail loc ""
...
-           if anyFirst xs
-              then (_ ** Keep svs)
-              else (_ ** Drop svs)
-  where
-    anyFirst : List (Var (n :: ns)) -> Bool
-    anyFirst [] = False
-    anyFirst (MkVar First :: xs) = True
-    anyFirst (MkVar (Later p) :: xs) = anyFirst xs
+toThin : (vars : Scope) -> VarSet vars -> (newvars ** Thin newvars vars)
+toThin vars xs = if isEmpty xs then (_ ** Refl) else go vars xs where
```
gallais: Oops, that's very silly. The short-circuiting logic should be `if isFull xs`.
But that'd require knowing the size of `vars` which is annoying.

PR #3609, opened 2025-08-13
docs: added note about valid operators

wizard7377:
# Description

Added small note in the operators documentation describing all valid syntax

## Self-check

- [X] This is my first time contributing, I've carefully read [`CONTRIBUTING.md`](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md)


andrevidela:
This should definitely go somewhere else than at the end of the reference for binding operators

wizard7377:
> This should definitely go somewhere else than at the end of the reference for binding operators

Where would you suggest? Perhaps a start to a syntax reference?

andrevidela:
They should probably be in the `Basics` section of the document

PR #3613, opened 2025-08-13
[ refactor ] `NameSet` instead of `List Name` when possible

gallais:


gallais:
Was a lot slower than the original so abandoning this.

PR #3614, opened 2025-08-14
implement binding application

andrevidela:
# Description

Draft implementation for #3582 this is the prototype branch with main rebased and all commits squashed. 
to do:
- remove `<-` symbol for binding operators. It should be its own PR
- Add some error messages
- cleanup a bit
- update changelog
- update documentation

## Self-check

- [ ] If this is a fix, user-facing change, a compiler change, or a new paper
      implementation, I have updated [`CHANGELOG_NEXT.md`](https://github.com/idris-lang/Idris2/blob/main/CHANGELOG_NEXT.md)

PR #3615, opened 2025-08-14
replace ImpTy

andrevidela:
# Description

Prerequisite for #3614

### DIFF
--- a/src/Core/WithData.idr
+++ b/src/Core/WithData.idr
@@ -109,11 +109,6 @@ setFC : {n : Nat} ->
         WithData fields a -> WithData fields a
 setFC fc = WithData.set "fc" fc @{inRange}
 
-||| Attach binding and file context information to a type
-public export
-FCBind : Type -> Type
-FCBind = WithData [ Bind', FC' ]
-
 ||| A wrapper for a value with a file context.
 public export
 MkFCVal : FC -> ty -> WithFC ty


src/Core/WithData.idr
```diff
@@ -197,13 +197,13 @@ WithName = AddMetadata Name'
 ||| the "tyname" label containing a `FCBind Name` for metadata records
```
dunhamsteve: Comment should change from FCBind to WithFC.

src/TTImp/TTImp/TTC.idr
```diff
@@ -233,14 +233,6 @@ mutual
...
                _ => corrupt "AltType"
 
-  export
-  TTC ImpTy where
-    toBuf (MkImpTy fc n ty)
-        = do toBuf fc;  toBuf n; toBuf ty
-    fromBuf
-        = do fc <- fromBuf; n <- fromBuf; ty <- fromBuf
-             pure (MkImpTy fc n ty)
-
```
dunhamsteve, line 243: Do we need to change `ttcVersion`?  Would an existing TTC file have a ImpTy serialized into it, and if so, how would it deserialize?

andrevidela: oh yes we need to change the TTC version, I thought I did but it must have fallen through the rebase cracks.

The new instance is handled by https://github.com/idris-lang/Idris2/blob/main/src/Core/TTC.idr#L187-L194

dunhamsteve: I ran a quick test and it seems fine without incrementing ttcVersion.

dunhamsteve: The new code could load and link the TTC files from the previous revision.  I think they just happen to have the same serialized format. (The label is erased, etc.)

src/Core/WithData.idr
```diff
@@ -194,16 +194,16 @@ public export
 WithName : Type -> Type
 WithName = AddMetadata Name'
 
-||| the "tyname" label containing a `FCBind Name` for metadata records
+||| the "tyname" label containing a `WithFC Name` for metadata records. Typically used for type names.
 public export
 TyName' : KeyVal
-TyName' = "tyname" :-: FCBind Name
+TyName' = "tyname" :-: WithFC Name
```
buzden, line 195: Shouldn't `FCBind` be removed after this?

andrevidela: done

dunhamsteve:
Looks like a codeberg hiccup, rerunning failed tests.

PR #3616, opened 2025-08-15
[ cleanup ] Use `{}` in pattern matching

spcfox:
# Description

1. Following @andrevidela [comment](https://github.com/idris-lang/Idris2/pull/3567#pullrequestreview-3110052202), I replaced multiple `_` in pattern matching with `{}`. And minor cleanup.
2. Made `vars` erased in functions that do not use it.
3. Added `Dotted` processing in `HasNames`.

## Self-check


src/Compiler/LambdaLift.idr
```diff
@@ -521,8 +521,8 @@ dropUnused {vars} {outer} unused (LConstCase fc sc alts def) =
     dropConstCase (MkLConstAlt c val) = MkLConstAlt c (dropUnused unused val)
 
 mutual
-  makeLam : {auto l : Ref Lifts LDefs} ->
-            {vars : _} ->
+  makeLam : {vars : _} ->
+            {auto l : Ref Lifts LDefs} ->
```
spcfox, line 525: This is not very important, but other functions in this file use this order

spcfox:
Please let me know if this kind of PR is not appropriate. However, as André said, It will make someone's diff smaller.

andrevidela (approved):
Looks good to me!

Yeah that stuff ought to be handled by a linter, but in the absence of it, this kind of Pr is the best we can do

PR #3619, opened 2025-08-18
Method refactor

andrevidela:
# Description

Prerequisite for #3615

- replace `Method`
- harmonise and merge `Signature` and `Declaration`
- replace `(String, Name)` by `WithData [Doc', FC'] Name`

### DIFF
--- a/src/Core/Binary.idr
+++ b/src/Core/Binary.idr
@@ -30,7 +30,7 @@ import public Libraries.Utils.Binary
 ||| version number if you're changing the version more than once in the same day.
 export
 ttcVersion : Int
-ttcVersion = 2025_08_13_00
+ttcVersion = 2025_08_16_00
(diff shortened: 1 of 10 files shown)


andrevidela:
the `nameVal` helper was actually a necessary change because`.name.val` was too complicated to disambiguate in the cases where `.nameVal` is present. 
I'm not sure what to make of this in that I don't think it makes sense to change all the instances of `.name.val` but I also don't feel like `.nameVal` should even be necessary

mattpolzin (approved):
The changes look sane to me, but ideally this PR is left open a bit longer in case someone with more experience with the compiler internals than me sees something that could regress performance or be a breaking change downstream.

dunhamsteve (approved):
Looks good to me.

PR #3617, opened 2025-08-16
[ fix ] Fix channels009 test (windows chez bootstrap)

Matthew-Mosior:
# Description

## Self-check

- [ ] This is my first time contributing, I've carefully read [`CONTRIBUTING.md`](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md)
      and I've updated [`CONTRIBUTORS.md`](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTORS.md) with my name.
- [ ] If this is a fix, user-facing change, a compiler change, or a new paper
      implementation, I have updated [`CHANGELOG_NEXT.md`](https://github.com/idris-lang/Idris2/blob/main/CHANGELOG_NEXT.md)

PR #3622, opened 2025-08-18
[ test ] Use `clean_names` in tests by default

spcfox:
# Description

Often, when changing compilers, differences in machine names are detected in tests. I think it would be useful to apply `clean_names` to all tests by default.

## Self-check

### DIFF
--- a/tests/idris2/operators/operators002/expected
+++ b/tests/idris2/operators/operators002/expected
@@ -4,7 +4,7 @@ Error: Operator =@ is a type-binding (typebind) operator, but is used as an auto
 Errors:8:19--8:21
  4 | 0 (=@) : (a : Type) -> (a -> Type) -> Type
  5 | (=@) a f = (1 x : a) -> f x
- 6 |
+ 6 | 
  7 | data S : {ty : Type} -> (x : ty) -> Type where
  8 |   MkS : (x <- ty) =@ S x
                        ^^
@@ -18,10 +18,10 @@ Possible solutions:
 Error: Operator =@ is an automatically-binding (autobind) operator, but is used as a regular operator.
 
 Errors2:7:29--7:31
- 3 |
+ 3 | 
  4 | 0 (=@) : (a : Type) -> (a -> Type) -> Type
  5 | (=@) a f = (1 x : a) -> f x
- 6 |
+ 6 | 
  7 | wrongId : {0 a : Type} -> a =@ a
                                  ^^
 Explanation: regular, typebind and autobind operators all use a slightly different syntax, typebind looks like this: '(name : type) =@ expr', autobind looks like this: '(name <- expr) =@ expr'.
@@ -34,10 +34,10 @@ Possible solutions:
 Error: Operator =@ is a type-binding (typebind) operator, but is used as a regular operator.
 
 Errors3:7:29--7:31
- 3 |
+ 3 | 
  4 | 0 (=@) : (a : Type) -> (a -> Type) -> Type
  5 | (=@) a f = (1 x : a) -> f x
- 6 |
+ 6 | 
  7 | wrongId : {0 a : Type} -> a =@ a
                                  ^^
 Explanation: regular, typebind and autobind operators all use a slightly different syntax, typebind looks like this: '(name : type) =@ expr', autobind looks like this: '(name <- expr) =@ expr'.
@@ -51,8 +51,8 @@ Error: Operator =@ is a regular operator, but is used as a type-binding (typebin
 
 Errors4:9:18--9:20
  5 | (=@) a f = (1 x : a) -> f x
- 6 |
- 7 |
+ 6 | 
+ 7 | 
  8 | data S : {ty : Type} -> (x : ty) -> Type where
  9 |   MkS : (x : ty) =@ S x
                       ^^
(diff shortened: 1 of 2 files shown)


tests/idris2/error/error027/expected
```diff
@@ -1 +1 @@
-ERROR: Unhandled input for Prelude.Num.case block in mod at Prelude.Num:94:3--96:44
+ERROR: Unhandled input for Prelude.Num.case block in mod at Prelude.Num:1
```
spcfox, line 1: I think we should remove the location in the names of implementations. This will make the tests resistant to changes in Prelude.

tests/idris2/operators/operators002/expected
```diff
@@ -44,8 +44,8 @@ Explanation: regular, typebind and autobind operators all use a slightly differe
 
 Possible solutions:
  - Write the expression using typebind syntax: '(_ : a) =@ a'.
- - Change the fixity defined at Errors3:2:18--2:29 to 'private infixr 0 =@'.
- - Hide or remove the fixity at Errors3:2:18--2:29 and import a module that exports a compatible fixity.
+ - Change the fixity defined at Errors3:1 to 'private infixr 0 =@'.
```
spcfox: But perhaps we shouldn't clean the location here

tests/idris2/total/total012/expected
```diff
@@ -22,14 +22,14 @@ Issue1828:2
...
-Issue1828:1
+Issue1828:2:3--2:25
  1 | mutual
  2 |   caseTest : Nat -> Bool
        ^^^^^^^^^^^^^^^^^^^^^^
 
 Error: dummy is not total, possibly not terminating due to recursive path Main.caseTest -> Main.dummy
 
-Issue1828:2
+Issue1828:6:3--6:14
```
spcfox, line 32: Currently, I am not cleaning the locators at the beginning of the line in order to preserve the errors in their original form

spcfox:
Perhaps we should only clean locators after `(implementation | case block | with block) at`?

dunhamsteve:
If we wanted to be precise we could do after `(implementation | case block | with block) at` but also after `^LOG`. 

My only reservation with what you have is that we might loose tests for locations of errors (e.g. sometimes unification errors will name `a` and put a location after it), but we could choose to not use `clean_names` in tests where it is an issue.

tests/testutils.sh
```diff
@@ -75,9 +75,17 @@ _awk_clean_name='
 
     out = ""
     # the last one is FC
-    while (match($0, mainPat)) {
+    while (match($0, mainPat, matched)) {
```
spcfox: Not POSIX, supported by gawk

tests/idris2/error/error004/expected
```diff
@@ -22,8 +22,8 @@ Error2:13:38--13:45
                                            ^^^^^^^
 
 Possible correct results:
-    Show implementation at Error2:11:1--13:45
-    Show implementation at Error2:7:1--9:45
+    Show implementation at Error2:1
+    Show implementation at Error2:2
```
spcfox, line 26: Are locations important here?

spcfox:
I think it shouldn't affect locations in errors now

spcfox:
Windows hangs every time in this PR. Maybe `clean_names` is hanging there?

dunhamsteve:
That could be, it looks like the while loop expects the replacement to leave something that doesn't match the original regex. 

We need a termination checker for AWK. :)

tests/idris2/error/error004/expected
```diff
@@ -22,8 +22,8 @@ Error2:13:38--13:45
                                            ^^^^^^^
 
 Possible correct results:
-    Show implementation at Error2:11:1--13:45
```
gallais, line 25: I think here we do want to preserve the ranges as we sometimes have PRs improving such location information.

Is it possible to replaces numbers only if they're not part of a location (i.e. of the shape %d:%d)?

spcfox: I decided to clean up this range because it is not the FC in the error, but part of the implementation name
https://github.com/idris-lang/Idris2/blob/adfba5bf0fae917146325483dd25e1a95db033aa/src/Idris/Elab/Implementation.idr#L40-L42
Like this
https://github.com/idris-lang/Idris2/blob/adfba5bf0fae917146325483dd25e1a95db033aa/tests/idris2/misc/inlining001/expected#L17

spcfox: If you think this is unnecessary, maybe just clean the locations in the logs?

spcfox: Another option is to add special processing for tests in the `error` directory

spcfox:
> Windows hangs every time in this PR. Maybe `clean_names` is hanging there?

I've made many runs. It seems like it's just bad luck

UPD: Or not. It doesn't always hang, but it seems to happen more often than usual.

spcfox:
> > Windows hangs every time in this PR. Maybe `clean_names` is hanging there?
> 
> I've made many runs. It seems like it's just bad luck
> 
> UPD: Or not. It doesn't always hang, but it seems to happen more often than usual.

The reason is the `channels009` test. There were problems with it before. I suggest disabling it temporarily. PR #3617 is working on fixing it.

gallais:
Happy to merge this whenever it's ready to go.

PR #3625, opened 2025-08-20
[ fix #3623 ] Normalise lambda types during elaboration

### LINKED ISSUE #3623: Implicit binding forced to be brought into scope
Take below proof for the one-point rule of implication, used to eliminate a quantifier when the bound variable can effectively take only one value:
```idris
infix 0 <->
record (<->) (a, b : Type) where
  constructor Iff
  to  : a -> b
  from : b -> a

forall_one_point :
  (p : a -> Type) ->
  (forall x. x = t -> p x) <-> p t
forall_one_point p = Iff
  (\p_all => p_all Refl)
  (\p_t, x, p_eq => rewrite p_eq in p_t)
  -- here^ x is implicit, but needs to be brought into scope
```

On the very last line of the proof we need to bring `x` in scope although it is implicit by the `forall`. 

Mainly, this is confusing during proof writing. Having:
```idris
forall_one_point p = Iff
  (\p_all => p_all Refl)
  (\p_t => ?h)
```
Idris reports `h : x = t -> p x` as the goal. Which hints at adding an extra parameter. However, after doing exactly that:
```idris
forall_one_point p = Iff
  (\p_all => p_all Refl)
  (\p_t, p_eq => ?h)
```
Idris reports `h : p_eq = t -> p p_eq` as the goal, which is quite puzzling.

Question is: is it really needed to force implicits into scope? In Agda for example, `x` is left implicit:
```agda
record _↔_ (A B : Set) : Set where
  field
    to   : A → B
    from : B → A

∀1∙ : ∀ {α t} →
  (P : α → Set) →
  (∀ {x} → x ≡ t → P x) ↔ P t
∀1∙ P = record
  { to = λ P∀ → P∀ refl
  ; from = λ Pt x≡t → subst P (sym x≡t) Pt
  }
```

spcfox:
# Description

Fixes #3623. In example the lambda type is a metavariable.

## Self-check

### DIFF
--- a/src/TTImp/Elab/Binders.idr
+++ b/src/TTImp/Elab/Binders.idr
@@ -175,7 +175,7 @@ checkLambda rig_in elabinfo nest env fc rigl info n argTy scope (Just expty_in)
                     let env' : Env Term (n :: _) = Lam fc rigb info' tyv :: env
                     ignore $ convert fc elabinfo env (gnf env tyv) (gnf env pty)
                     let nest' = weaken (dropName n nest)
-                    pscnf <- normalise defs env' $ compat psc
+                    pscnf <- normaliseHoles defs env' $ compat psc
                     (scopev, scopet) <-
                        inScope fc env' (\e' =>
                           check {e=e'} rig elabinfo nest' env' scope


src/TTImp/Elab/Binders.idr
```diff
@@ -175,7 +175,7 @@ checkLambda rig_in elabinfo nest env fc rigl info n argTy scope (Just expty_in)
                     let env' : Env Term (n :: _) = Lam fc rigb info' tyv :: env
                     ignore $ convert fc elabinfo env (gnf env tyv) (gnf env pty)
                     let nest' = weaken (dropName n nest)
-                    pscnf <- normalise defs env' $ compat psc
+                    pscnf <- normaliseHoles defs env' $ compat psc
```
spcfox, line 178: With `normalise`, the frex build requires too many resources

buzden (approved):
I was initially concerned about moving `compat` out of `inScope`, but neither `normalise` nor `compat` depend on `EST` in this case, so looks good.

PR #3628, opened 2025-08-22
[ fix ] harmonised the extension as expected in the PR request.

jfdm:
The instructions ask one to update `CONTRIBUTORS.md`. 
Said file does not exist.
The file `CONTRIBUTORS`,
however, 
does.

Let's change that.


andrevidela:
I think the template is wrong actually, the file was probably setup without a .md extension so that it would render nicely on the github website. here is what it looks like rendered as a markdown page: https://github.com/idris-lang/Idris2/blob/f714661633e260e4a0a4a9dc151dbe8a7e85932d/CONTRIBUTORS.md

But as a plain text file, every name is properly rendered as a list instead of a paragraph https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTORS

Issue #3582, opened 2025-07-16
Binding Application Proposal & Design


andrevidela:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Table of Contents

- Summary
- Motivation
  - What's wrong with syntax blocks
  - Explicit goals
- Proposal
- Technical implementation
  - Edge cases
  - Interaction with short-circuit patterns
  - Interaction with `with`
  - Interaction with `with`-application
  - Overload with regular function application
  - Binding Misuse & Diagnosis
    - Not Binding a Type
    - Not a dependent binder
    - Forgotten Binder
    - Improper Type Entirely
    - Binding a Type as `autobind`
    - Incorrect Binding Symbol
    - Unexpected Use of Binding syntax
 - Example use-cases
   - For-loop syntax
   - Quantitative pairs
   - Data Description & Algebras
- Alternatives considered
  - Trailing lambda technology
  - How about a different separator?
  - I don't like `:` and `<-`
  - Just live without it
  - Why not allow multiple binders?
  - Elide the Type/name
- Future directions
  - Allow Arguments Before Binding
  - Handle Multiple Binding Sites
- Conclusion

## Summary

Implement `autobind` and `typebind` for function application, rather than just operators. Allowing users to write `Exists (n : Nat) | Vect n String`.

## Motivation

The motivation is the same as for `typebind` and `autobind` operators: Binding in dependently typed programming language is particularly important due to leading arguments appearing in the types of trailing ones. $\Pi : (x : Type) \to (x \to Type) \to Type$ being the leading examples. This feature could be done previously with syntax blocks, but those proved hard to maintain, slow, and difficult to debug. The ergonomic benefit for the user did not outweigh their negative impact. This feature recovers the ability to customise binder without introducing arbitrary syntax blocks.

### What's wrong with syntax blocks

Syntax blocks introduce a number of tradeoffs that were deemed unworthy of being ported into Idris2. They were:

- More maintenance burden on the compiler
- Impoverished developer experience due to unpredicatable errors. Is the syntax wrong because the user is wrong, because the syntax block is wrong, or because the syntax block is not imported? Impossible to know
- Higher learning curve for new users. Syntax blocks are not common in industrial programming languages and steepens the learning curve with a concept that's orthogonal to dependent types and with little benefit
[...]

PR #3631, opened 2025-08-25
replace binding symbol for autobind

andrevidela:
# Description

fine print of #3582 

## Self-check

(There is actually no changelog changes to make since there has not been a release since the previous implementation of binding operators)

### DIFF
--- a/docs/source/reference/operators.rst
+++ b/docs/source/reference/operators.rst
@@ -311,7 +311,7 @@ where the constructor for ``Pi`` takes terms on the left and lambdas on the righ
 We would like to use a custom operator to build values using ``VPi``, but its
 signature does not follow the pattern that ``typebind`` uses. Instead, we use
 ``autobind`` to tell the compiler that the type of the lambda must be inferred.
-For this we use ``:=`` instead of ``:``:
+For this we use ``<-`` instead of ``:``:
 
 .. code-block:: idris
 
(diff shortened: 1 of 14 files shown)


andrevidela:
I've not done a nice warning for the old syntax to move over the new one partly because I don't expect anything to break from that change but I guess we'll find out once we compile libraries from pack. 

If it turns out a library actually uses that, I'll add the warning to move over to the new syntax

mattpolzin (commented):
The changes look very straight forward. I can answer the question about other libraries and also show a use-case that doesn't give an autobind-specific error interestingly enough.

In the `telescope` library:
```
       > Error: Couldn't parse any alternatives:
       > 1: Expected 'case', 'if', 'do', application or operator expression.
       >
       > Data.Telescope.Telescope:77:33--77:34
       >  73 |   public export
       >  74 |   (\++) : {n : Nat} -> (gamma : Left.Telescope m) -> (Environment gamma -> Left.Telescope n) ->
       >  75 |          Telescope (plusAcc n m)
       >  76 |   (\++) {n = Z} gamma delta = gamma
       >  77 |   (\++) {n = S n} gamma delta = (env := gamma -. sigma) \++ uncurry theta
```

That appears to be the only library that has errors related to this PR.

andrevidela:
oh and that's my library, I can fix that quite easily

mattpolzin (approved):

mattpolzin:
I’ll leave this for self merge if you don’t mind so you can coordinate with yourself about the library fix :)

PR #3633, opened 2025-08-27
[ doc ] Expand help for fuzzy search

dunhamsteve:
# Description

This PR adds REPL documentation for `:fs` and corrects the parse error message (the `_` goes on the right side of the arrow).

```
Main> :help fs
Search for global definitions by sketching the names distribution of the wanted type(s).

The parameter must be in one of the forms A -> B, A -> _, or B, where A and B are lists of global names.

Idris will return all of the entries in the context that have all of the names in A
in some argument and all of the names in B within the return type.

For example:

  :fs List Maybe -> List

will match (among other things):

  Prelude.List.mapMaybe : (a -> Maybe b) -> List a -> List b
```

### DIFF
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -2285,8 +2285,8 @@ knownCommands =
 
       Prelude.List.mapMaybe : (a -> Maybe b) -> List a -> List b
 
-Note that the query 'List Nat -> String' does not describe the type 'List Nat',
-rather it describes both 'List a' and 'Nat' in the arguments.
+    Note that the query 'List Nat -> String' does not describe the type 'List Nat',
+    rather it describes both 'List a' and 'Nat' in the arguments.
 
     """
   where


src/Idris/Parser.idr
```diff
@@ -2269,7 +2269,23 @@ knownCommands =
       """
     )
   ] ++
-  explain ["fs", "fsearch"] "Search for global definitions by sketching the names distribution of the wanted type(s)."
+  explain ["fs", "fsearch"] """
+    Search for global definitions by sketching the names distribution of the wanted type(s).
+
+    The parameter must be in one of the forms A -> B, A -> _, or B, where A and B are lists of global names.
```
mattpolzin:
```suggestion
    The parameter must be in one of the forms A -> B, A -> _, or B, where A and B are space-delimited lists of global names.
```

src/Idris/Parser.idr
```diff
@@ -2269,7 +2269,23 @@ knownCommands =
...
+    in some argument and all of the names in B within the return type.
+
+    For example:
+
+      :fs List Maybe -> List
+
+    will match (among other things):
+
+      Prelude.List.mapMaybe : (a -> Maybe b) -> List a -> List b
+
```
mattpolzin, line 2287:
```suggestion

Note that the query 'List Nat -> String' does not describe the type 'List Nat',
rather it describes both 'List a' and 'Nat' in the arguments.

```

PR #3635, opened 2025-08-31
docs: added library documentation for `Elab`

wizard7377:
# Description

Added documentation for `Language.Reflection`


libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
+       ||| An 'as' pattern, valid on the LHS of a clause only
        IAs : FC -> (nameFC : FC) -> UseSide -> Name -> TTImp -> TTImp
-       -- A 'dot' pattern, i.e. one which must also have the given value
-       -- by unification
+       ||| A 'dot' pattern, i.e. one which must also have the given value
+       ||| by unification
        IMustUnify : FC -> DotReason -> TTImp -> TTImp
 
-       -- Laziness annotations
+       ||| Laziness annotations
```
buzden: This particular line is not a doc string for `IDelayed`, it is a comment to the next three constructors

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
+       ||| Quasi-quotation of expression (`( ... ))
        IQuote : FC -> TTImp -> TTImp
+       ||| Quasi-quotation of a name (`{ ... })
        IQuoteName : FC -> Name -> TTImp
+        ||| Quasi-quotation of a list of declarations (`[ ... ])
        IQuoteDecl : FC -> List Decl -> TTImp
+       ||| Unquote of an expression (~e)
        IUnquote : FC -> TTImp -> TTImp
-
+       ||| A primitive value, such as an integer or string constant
```
buzden: I would add here "or a primitive type itself" since expressions like `Integer` or `String` are also `IPrimVal $ PrT IntegerType` and `IPrimVal $ PrT StringType` correspondingly

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
+       ||| A named hole
        IHole : FC -> String -> TTImp
 
-       -- An implicit value, solved by unification, but which will also be
-       -- bound (either as a pattern variable or a type variable) if unsolved
-       -- at the end of elaborator
+       ||| An implicit value, solved by unification, but which will also be
+       ||| bound (either as a pattern variable or a type variable) if unsolved
+       ||| at the end of elaborator
        Implicit : FC -> (bindIfUnsolved : Bool) -> TTImp
```
buzden, line 112: I would add here to the doc that `Implicit True` is `_` in the surface language, and `Implicit False` is `?`

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
        IWithUnambigNames : FC -> List (FC, Name) -> TTImp -> TTImp
```
buzden: This can be also documented, since this is `with Vect.(::) [1, 2, 3]` in the surface language, where `Vect.(::)`, i.e. disambiguation names, goes to the list and `[1, 2, 3]` in this example goes to the `TTImp`

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -105,17 +150,17 @@ mutual
...
        Deprecate : FnOpt
        TCInline : FnOpt
-       -- Flag means the hint is a direct hint, not a function which might
-       -- find the result (e.g. chasing parent interface dictionaries)
+       ||| Flag means the hint is a direct hint, not a function which might
+       ||| find the result (e.g. chasing parent interface dictionaries)
        Hint : Bool -> FnOpt
-       -- Flag means to use as a default if all else fails
+       ||| Flag means to use as a default if all else fails
        GlobalHint : Bool -> FnOpt
```
buzden, line 144: The boolean flag should be explained here. `True` is `%defaulthint` in the surface language, and means that it applies only when types matches, `False` is `%globalhint` in the surface language, and means "try to apply anyway", which allows to refine holes at the place of application

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -129,10 +174,15 @@ mutual
 
   public export
   data DataOpt : Type where
+       ||| Determing arguements
```
buzden:
```suggestion
       ||| Determing arguments
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -105,17 +150,17 @@ mutual
...
+       ||| Defined externally, list calling conventions
        ForeignFn : List TTImp -> FnOpt
-       -- Mark for export to a foreign language, list calling conventions
+       ||| Mark for export to a foreign language, list calling conventions
        ForeignExport : List TTImp -> FnOpt
-       -- assume safe to cancel arguments in unification
+       ||| assume safe to cancel arguments in unification
        Invertible : FnOpt
        Totality : TotalReq -> FnOpt
        Macro : FnOpt
```
buzden, line 154: It is `%macro` in the surface language and means that `%runElab` would be added before the call of this function at each call site

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
-       -- A name which should be implicitly bound
+       ||| A name which should be implicitly bound
        IBindVar : FC -> Name -> TTImp
-       -- An 'as' pattern, valid on the LHS of a clause only
+       ||| An 'as' pattern, valid on the LHS of a clause only
        IAs : FC -> (nameFC : FC) -> UseSide -> Name -> TTImp -> TTImp
-       -- A 'dot' pattern, i.e. one which must also have the given value
-       -- by unification
+       ||| A 'dot' pattern, i.e. one which must also have the given value
+       ||| by unification
```
buzden: You can add that this is a `.(...)` in the surface language

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
-       -- Any implicit bindings in the scope should be bound here, using
-       -- the given binder
+       ||| Any implicit bindings in the scope should be bound here, using
+       ||| the given binder
        IBindHere : FC -> BindMode -> TTImp -> TTImp
-       -- A name which should be implicitly bound
+       ||| A name which should be implicitly bound
        IBindVar : FC -> Name -> TTImp
-       -- An 'as' pattern, valid on the LHS of a clause only
+       ||| An 'as' pattern, valid on the LHS of a clause only
```
buzden: You can add that this is `@expr` in the surface language

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
-
+      ||| A function application
        IApp : FC -> TTImp -> TTImp -> TTImp
+      ||| A named function application (for named parameters), e.g, `f {x=3}`
        INamedApp : FC -> TTImp -> Name -> TTImp -> TTImp
+      ||| An explicitly inserted auto implicit, `f @{x}`
        IAutoApp : FC -> TTImp -> TTImp -> TTImp
+      ||| A view in a `with` rule
        IWithApp : FC -> TTImp -> TTImp -> TTImp
-
```
buzden, line 51: Please, don't remove this newline, it's purpose is a logical separation

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
                List Clause -> TTImp
+      ||| A list of full declerations local to a term
+      ||| @ (List Decl) the list of declerations
+      ||| @ TTImp the body of the local declaration
        ILocal : FC -> List Decl -> TTImp -> TTImp
+      ||| An update to a record value
+      ||| @ (List IFieldUpdate) the list of field update descriptions
+      ||| @ TTImp the record value being updated
        IUpdate : FC -> List IFieldUpdate -> TTImp -> TTImp
-
```
buzden, line 46: Please don't remove this newline, it's for logical separation

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
+      ||| A named function application (for named parameters), e.g, `f {x=3}`
        INamedApp : FC -> TTImp -> Name -> TTImp -> TTImp
+      ||| An explicitly inserted auto implicit, `f @{x}`
        IAutoApp : FC -> TTImp -> TTImp -> TTImp
+      ||| A view in a `with` rule
        IWithApp : FC -> TTImp -> TTImp -> TTImp
-
+      ||| `%search`
        ISearch : FC -> (depth : Nat) -> TTImp
        IAlternative : FC -> AltType -> List TTImp -> TTImp
```
buzden, line 69: This operation is for expression that tries to typecheck expression in the list according to the given `AltType` and return either one of them, or default one (if present in the `AltType`), or fails

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
   public export
   data TTImp : Type where
+      ||| A variable reference, by name
        IVar : FC -> Name -> TTImp
+      ||| A function type
+      ||| @ Count the multiplicity of the arguement
+      ||| @ PiInfo If the arguement is implicit, explicit, auto implicit or a default implicit
+      ||| @ (Maybe Name) the optional index, such as `x` in `(x : Type) -> Type`
+      ||| @ argTy the type of the argument
+      ||| @ retTy the return type
```
buzden: I really like the style of tacit docstrings you added in the second half of `TTImp`, also mentioning expressions in the surface langauge that correspond to constructors of `TTImp`. I think this style should be appropriate also for these constructors in the beginning of `TTImp`.

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
   %name DotReason dr
 
+  ||| The elaborator representation of an Idris term
+  ||| All of these take a file context `FC` as their first arguement
   public export
   data TTImp : Type where
+      ||| A variable reference, by name
        IVar : FC -> Name -> TTImp
+      ||| A function type
+      ||| @ Count the multiplicity of the arguement
```
buzden: Maybe, this should go into the description of the `Count` type itself, instead of repeating it in each constructor that uses it

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,59 +29,104 @@ mutual
...
+      ||| @ PiInfo If the arguement is implicit, explicit, auto implicit or a default implicit
```
buzden: Maybe, this should go into the description of the `Count` type itself, instead of repeating it in each constructor that uses it

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -129,10 +174,15 @@ mutual
 
   public export
   data DataOpt : Type where
+       ||| Determing arguements
        SearchBy : List1 Name -> DataOpt -- determining arguments
+       ||| Don't generate search hints for constructors
        NoHints : DataOpt -- Don't generate search hints for constructors
```
buzden: Maybe, you should delete the comments after each constructor, since you added exactly same text to the docstring

buzden (changes requested):

libs/base/Language/Reflection/TTImp.idr
wizard7377: Because I'm not familiar with this, I'm just going to copy paste your description if that's alright with you

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,66 +29,96 @@ mutual
...
 
+  ||| The elaborator representation of an Idris term
+  ||| All of these take a file context `FC` as their first argument
   public export
   data TTImp : Type where
+       ||| A variable reference, by name
        IVar : FC -> Name -> TTImp
-       IPi : FC -> Count -> PiInfo TTImp -> Maybe Name ->
+       ||| A function type, of the form `(mult binder : argTy) -> retTy`, with implicitness determined by `info`
+       IPi : FC -> (mult : Count) -> (info : PiInfo TTImp) -> (binder : Maybe Name) ->
```
buzden, line 39: Could you please change `info` to `piinfo`, because `info` seems to be too generic

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,66 +29,96 @@ mutual
...
+       IUpdate : FC -> (updates : List IFieldUpdate) -> (val : TTImp) -> TTImp
+
+       ||| A function application, `f x`
+       IApp : FC -> (f : TTImp) -> (x : TTImp) -> TTImp
+       ||| A named function application (for named parameters), e.g, `f {arg=x}`
+       INamedApp : FC -> (f : TTImp) -> (arg : Name) -> (x : TTImp) -> TTImp
+       ||| An explicitly inserted auto implicit, `f @{x}`
+       IAutoApp : FC -> (f : TTImp) -> (x : TTImp) -> TTImp
+       ||| A view in a `with` rule
        IWithApp : FC -> TTImp -> TTImp -> TTImp
```
buzden: This is not a view. View with `with`-rule is done through appropriate `WithClause` consturctor of the `Clause` data type. `IWithApp` is an application in form of `f | arg` which can be used only inside `with`-clauses.

libs/base/Language/Reflection/TTImp.idr
buzden: Yes, it's okay

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,66 +29,96 @@ mutual
...
        IAlternative : FC -> AltType -> List TTImp -> TTImp
-       IRewrite : FC -> TTImp -> TTImp -> TTImp
+       ||| A rewrite expression, `rewrite eq in exp`
+       IRewrite : FC -> (eq : TTImp) -> (exp : TTImp) -> TTImp
 
-       -- Any implicit bindings in the scope should be bound here, using
-       -- the given binder
+       ||| Any implicit bindings in the scope should be bound here, using
+       ||| the given binder
        IBindHere : FC -> BindMode -> TTImp -> TTImp
```
buzden, line 75: I'd say "A bound for implicit bindings to be automatically bound using the given bind mode"

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,66 +29,96 @@ mutual
...
+       IAs : FC -> (nameFC : FC) -> UseSide -> (group : Name) -> (pat : TTImp) -> TTImp
+       ||| A 'dot' pattern, i.e. one which must also have the given value
+       ||| by unification, `.(...)`
        IMustUnify : FC -> DotReason -> TTImp -> TTImp
 
        -- Laziness annotations
-       IDelayed : FC -> LazyReason -> TTImp -> TTImp -- the type
-       IDelay : FC -> TTImp -> TTImp -- delay constructor
+       ||| The delay type, `Delay t`
+       IDelayed : FC -> LazyReason -> TTImp -> TTImp
```
buzden, line 86: Delay types are `Lazy` and `Inf` (depending on the `LazyReason` argument). `Delay` is a constructor of a lazy value, it is of type `a -> Lazy a` or `a -> Inf a` depending on the context

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,66 +29,96 @@ mutual
...
+       ||| The constructor of `Delay`, `delay t`
+       IDelay : FC -> TTImp -> TTImp
```
buzden, line 88: `delay` is just a library function, the constructor-like thing in the surface language is `Delay` (with an upper-case letter`

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,66 +29,96 @@ mutual
...
+       ||| `force`
        IForce : FC -> TTImp -> TTImp
```
buzden, line 90: The same for above, `IForce` corresponds to the `Force` pseudo-constuctor, `force` is just a function

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,66 +29,96 @@ mutual
...
        IType : FC -> TTImp
+       ||| A named hole
        IHole : FC -> String -> TTImp
 
-       -- An implicit value, solved by unification, but which will also be
-       -- bound (either as a pattern variable or a type variable) if unsolved
-       -- at the end of elaborator
+       ||| An implicit value, solved by unification, but which will also be
+       ||| bound (either as a pattern variable or a type variable) if unsolved
+       ||| at the end of elaborator.
```
buzden, line 110: I don't understand what is said about the bound. I mean, if `Implicit False` is used, unsolvedness of this expression is a compilation error. But when it is unsolved and it is `Implicit True`, then this expression is bound to a fresh variable.

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -29,66 +29,96 @@ mutual
...
+       ||| Note that `Implicit False` is `?`, while `Implicit True` is `_`
        Implicit : FC -> (bindIfUnsolved : Bool) -> TTImp
-       IWithUnambigNames : FC -> List (FC, Name) -> TTImp -> TTImp
+       ||| `with name exp`
+       IWithUnambigNames : FC -> (name : List (FC, Name)) -> (exp : TTImp) -> TTImp
```
buzden: I'd add that except it is `with name exp`, this is a disambiguation expression, i.e. when a name matches with one of names given in the list, it disambiguates to these names instead of searching for all applicable names. Also, please, rename the `name` argument, since it is a list. I think `names` should apply

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -105,22 +135,23 @@ mutual
...
-       -- find the result (e.g. chasing parent interface dictionaries)
+       ||| Flag means the hint is a direct hint, not a function which might
+       ||| find the result (e.g. chasing parent interface dictionaries)
        Hint : Bool -> FnOpt
-       -- Flag means to use as a default if all else fails
+       ||| `%globalhint` if true, `%defaulthint` if false
        GlobalHint : Bool -> FnOpt
        ExternFn : FnOpt
-       -- Defined externally, list calling conventions
+       ||| Defined externally, list calling conventions
```
buzden:
```suggestion
       ||| Defined externally, takes a list of calling conventions
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -105,22 +135,23 @@ mutual
...
        ForeignFn : List TTImp -> FnOpt
-       -- Mark for export to a foreign language, list calling conventions
+       ||| Mark for export to a foreign language, list calling conventions
```
buzden:
```suggestion
       ||| Mark for export to a foreign language, takes a list of calling conventions
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -105,22 +135,23 @@ mutual
        NoInline : FnOpt
        Deprecate : FnOpt
        TCInline : FnOpt
-       -- Flag means the hint is a direct hint, not a function which might
-       -- find the result (e.g. chasing parent interface dictionaries)
+       ||| Flag means the hint is a direct hint, not a function which might
+       ||| find the result (e.g. chasing parent interface dictionaries)
        Hint : Bool -> FnOpt
-       -- Flag means to use as a default if all else fails
+       ||| `%globalhint` if true, `%defaulthint` if false
```
buzden:
```suggestion
       ||| A hint that is searched if direct hints search failed.
       ||| `%globalhint` if the argument is `True`, `%defaulthint` if `False`.
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -165,15 +201,18 @@ mutual
 
   public export
   data WithFlag = Syntactic
-
+  ||| A clause in a function definition
   public export
   data Clause : Type where
+       ||| A simple pattern
        PatClause : FC -> (lhs : TTImp) -> (rhs : TTImp) -> Clause
+       ||| A pattern with views
```
buzden, line 211: Maybe, we could add the surface syntax here, like
```idris
<lhs> with <rig> (<wval>) proof <prf>
  <subclauses>
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
                    (type : ITy) ->
                    IClaimData
-
+  ||| A top-level declaration
   public export
   data Decl : Type where
-       IClaim : WithFC IClaimData -> Decl
-       IData : FC -> WithDefault Visibility Private -> Maybe TotalReq -> Data -> Decl
-       IDef : FC -> Name -> (cls : List Clause) -> Decl
+       ||| A type ascription, `a : b`.
```
buzden, line 256: Type ascription term is used then you try to refine the type of an expression. I think, this should be called "a function signature"

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
+       ||| Called a claim because of Curry Howard, the statement `x : p` is equivalent to `x` is a proof of `p`.
```
buzden, line 257: I'm not sure this line is a valid one. I thought it is a claim because it starts definition of a function, but does not define its body

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
+       IClaim : (claim : WithFC IClaimData) -> Decl
+       ||| A data declaration
+       IData : FC -> (vis : WithDefault Visibility Private) -> Maybe TotalReq -> (cons : Data) -> Decl
+       ||| A function definition by its clauses
```
buzden:
```suggestion
       ||| A function body definition
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
+       ||| A record declaration
+       ||| @ ns Nested namespace
        IRecord : FC ->
-                 Maybe String -> -- nested namespace
-                 WithDefault Visibility Private ->
-                 Maybe TotalReq -> Record -> Decl
+                 (ns : Maybe String) ->
+                 (vis : WithDefault Visibility Private) ->
+                 (totality : Maybe TotalReq) -> (rec : Record) -> Decl
+       ||| A namespace declaration, `namespace ns where decls`
```
buzden, line 272: In namespace syntax where is not `where`, a newline with a indent or `{ ... }` should go there

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
        INamespace : FC -> Namespace -> (decls : List Decl) -> Decl
+       ||| A transformation declaration
        ITransform : FC -> Name -> TTImp -> TTImp -> Decl
+       ||| A top-level elaboration run
        IRunElabDecl : FC -> TTImp -> Decl
+       ||| A logging directive
```
buzden:
```suggestion
       ||| A directive for enabling compile-time logging, `%logging "<topic>" <level>`
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
-                 WithDefault Visibility Private ->
-                 Maybe TotalReq -> Record -> Decl
+                 (ns : Maybe String) ->
+                 (vis : WithDefault Visibility Private) ->
+                 (totality : Maybe TotalReq) -> (rec : Record) -> Decl
+       ||| A namespace declaration, `namespace ns where decls`
        INamespace : FC -> Namespace -> (decls : List Decl) -> Decl
+       ||| A transformation declaration
        ITransform : FC -> Name -> TTImp -> TTImp -> Decl
+       ||| A top-level elaboration run
```
buzden:
```suggestion
       ||| A top-level elaborator script run, `%runElab`
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
        IRecord : FC ->
-                 Maybe String -> -- nested namespace
-                 WithDefault Visibility Private ->
-                 Maybe TotalReq -> Record -> Decl
+                 (ns : Maybe String) ->
+                 (vis : WithDefault Visibility Private) ->
+                 (totality : Maybe TotalReq) -> (rec : Record) -> Decl
+       ||| A namespace declaration, `namespace ns where decls`
        INamespace : FC -> Namespace -> (decls : List Decl) -> Decl
+       ||| A transformation declaration
```
buzden:
```suggestion
       ||| A transformation rule declaration
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
        IParameters : FC -> (params : List (Name, Count, PiInfo TTImp, TTImp)) ->
                      (decls : List Decl) -> Decl
+       ||| A record declaration
+       ||| @ ns Nested namespace
        IRecord : FC ->
-                 Maybe String -> -- nested namespace
-                 WithDefault Visibility Private ->
-                 Maybe TotalReq -> Record -> Decl
+                 (ns : Maybe String) ->
+                 (vis : WithDefault Visibility Private) ->
```
buzden, line 270: Is there any need for names here? I suppose, their types are self expanatory and they are not used on the documentation either, so these names, I think, just clutter reading. What do you think?

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
+  ||| A top-level declaration
   public export
   data Decl : Type where
-       IClaim : WithFC IClaimData -> Decl
-       IData : FC -> WithDefault Visibility Private -> Maybe TotalReq -> Data -> Decl
-       IDef : FC -> Name -> (cls : List Clause) -> Decl
+       ||| A type ascription, `a : b`.
+       ||| Called a claim because of Curry Howard, the statement `x : p` is equivalent to `x` is a proof of `p`.
+       IClaim : (claim : WithFC IClaimData) -> Decl
+       ||| A data declaration
```
buzden:
```suggestion
       ||| A data type declaration
```

libs/base/Language/Reflection/TTImp.idr
```diff
@@ -209,22 +248,34 @@ mutual
...
+       IData : FC -> (vis : WithDefault Visibility Private) -> Maybe TotalReq -> (cons : Data) -> Decl
+       ||| A function definition by its clauses
+       IDef : FC -> (f : Name) -> (cls : List Clause) -> Decl
+       ||| A parameter block
```
buzden: I'd suggest giving an example here, something like
```suggestion
       ||| A parameters block, e.g. `parameters {0 m : _} {auto _ : Monad m} (level : Nat)
```

buzden (dismissed):

wizard7377:
My bad, I really need to keep better track of my pr's

PR #3636, opened 2025-09-01
std: created `Control.Monad`

wizard7377:
# Description

Added `Control.Monad`, which re-exports all the `Control.Monad.*` modules

## Self-check


buzden:
Could you show a case where one needs to import such a module?

PR #3637, opened 2025-09-02
[ error ] check for similar names in let and where

dunhamsteve:
# Description

At the top level of a file, we report an error for definitions that have no claim, if there is an unused claim with a similar name.  This PR updates this to also work for functions defined in `let` and `where`.

At the top level, the following code:
```idris
fox : ()
foo = ()
```
gives an error:
```
Error: No type declaration for Main.foo.

Main:2:1--2:9
 1 | fox : ()
 2 | foo = ()
     ^^^^^^^^
Did you mean: fox?
```
but inside a `let` or `where`, this doesn't happen:
```idris
blah : Nat
blah = 
   let fox : ()
       foo = ()
   in 3
```

After this change we get:
```
Error: While processing right hand side of blah. No type declaration for foo.

LetError:4:8--4:16
 1 | blah : Nat
 2 | blah =
 3 |    let fox : ()
 4 |        foo = ()
            ^^^^^^^^
Did you mean: fox?
```

### DIFF
--- a/src/Core/Context.idr
+++ b/src/Core/Context.idr
@@ -1145,6 +1145,30 @@ getSimilarNames : {auto c : Ref Ctxt Defs} ->
                   -- Sometime we might want to hide names that we know make no sense.
                   {default Nothing keepPredicate : Maybe ((Name, GlobalDef) -> Core Bool)} ->
                   Name -> Core (Maybe (String, List (Name, Visibility, Nat)))
+getSimilarNames (NS ns (Nested (outer,_) inner)) = case show <$> userNameRoot inner of
+  Nothing => pure Nothing
+  Just str => if length str <= 1 then pure (Just (str, [])) else
+    do defs <- get Ctxt
+       let threshold : Nat := max 1 (assert_total (divNat (length str) 3))
+       let test : Name -> Core (Maybe (Visibility, Nat)) := \ nm => do
+               let (Just str') = checkName nm
+                   | _ => pure Nothing
+               dist <- coreLift $ Levenshtein.compute str str'
+               let True = dist <= threshold
+                   | False => pure Nothing
+               Just def <- lookupCtxtExact nm (gamma defs)
+                   | Nothing => pure Nothing -- should be impossible
+               let predicate = fromMaybe (\_ => pure True) keepPredicate
+               True <- predicate (nm, def)
+                   | False => pure Nothing
+               pure (Just (collapseDefault $ visibility def, dist))
+       kept <- NameMap.mapMaybeM @{CORE} test (resolvedAs (gamma defs))
+       pure $ Just (str, toList kept)
+    where
+      checkName : Name -> Maybe String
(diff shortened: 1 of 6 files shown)


mattpolzin:
I'm not sure if we want this change. At the top level Idris2 does not attempt to infer the type of a definition (save for the experiment done for trivial constants that I think may have been reverted since then) so an error feels appropriate. On the other hand, unification does allow for some expressions found in a `let` to omit a type and this error would disallow that any time names were similar. An edge case, perhaps, but the following contrived example shows the situation I am describing:
```idris

test : Maybe Bool
test =
  let too : String
      far : String = "hello \{too}"
      foo = Just True
  in foo

```
That code works with the compiler on `main` and represents a theoretically desirable workflow where `too` is left as a hole for now but it gives an error for this branch.

dunhamsteve:
The request came from users on discord, but merits discussion. I think they were expecting let/where to behave like top level. If we don't want it, we'll have a record of that for the next time someone asks.

That experiment (#1952) was never removed, and it is what is generating these errors. At the top level, if you have an undeclared function and a declaration _with no definition_ that has a similar name, you will get this error.  Definitions without claims are currently accepted by Idris. The complexity of the RHS doesn't seem to matter, but it will fail to infer in a lot of cases.

If we did remove that experiment (it might be a good idea, but I doubt it will happen because a lot of code depends on it), your example would not compile. The reason is subtle, but the `foo = Just True` in your example is a local function with no declaration. It is not a `let` binding and behaves differently from `let foo = Just True in ...`.  It would, for example, be substituted into types.

With a `:=` the code would be accepted and `foo` would be a proper `let` binding:
```idris
test : Maybe Bool
test =
  let too : String
      far : String = "hello \{too}"
      foo := Just True
  in foo
```

The reason is our rather unfortunate (IMO) overloading of `let`.  It will try to parse `let`-bindings, and when it fails, switch to parsing top-level declarations, until that fails, and then it switches back. Here the `:=` causes a failure to parse as a top level declaration, and it switches back to `let` mode.  (As a matter of hygiene, I try not to mix the two types in a single `let` statement, but that is not enforced.)

mattpolzin:
That's interesting context that I definitely didn't entirely see initially. Thanks for the explanation. 

I do still think that this is the wrong next step (personal opinion) because the new error can potentially show up simply because a name is changed to be too similar to another name. If the intention was to type the other name, that's a nice error. If the intention was not to type the other name, the error is more confusing than ever. If the proposed solution to the problem is to use `:=` then we probably need to take a step toward that being more obvious before introducing errors that can occur because of not using `:=` but provide no indication of that fact.

dunhamsteve:
Fair enough. With the situation as-is, this change removes one potential confusion and adds another. The issue with `let` should be solved first, if anything.

I think the solution there is one of:
- Don't allow local function definitions in `let`
- Use a different keyword for local function definitions
- Turn undeclared definitions back into a normal `let`.
- Deprecate `=` for `let` in favor of `:=`

PR #3640, opened 2025-09-05
Ide index

hakanaras:
# Description

Previously IdeMode was not able to suggest completions for symbols outside of what is already imported.

This PR adds a new command line option `--ide-index`, which:
1. Implies `--ide-mode`, if neither `--ide-mode` nor `--ide-mode-socket` is set.
2. Reads all TTC files from all package_dirs and places their public definitions into `ROpts` before starting the IdeMode loop.
3. When providing completions, in addition to symbols from the current context, now also returns public symbols from this new index along with the module identifier that has to be imported to make the symbol accessible.

I've also tried out the new functionality in a fork of [meraymond2/idris-vscode](https://github.com/meraymond2/idris-vscode), and will be creating a pull request there if this gets merged. I've not done anything for the emacs-mode or other tools.

The auto-import completions are just the first of multiple IDE features that could benefit from such an index.

Behaviour should be unchanged as long as `--ide-index` is not set.

Demo:

https://github.com/user-attachments/assets/aa70af86-4f45-4ce5-8f05-8631e78f3701

Things I'm unsure of:

- This will almost certainly impact performance of IdeMode. On my MacBook Pro I felt no difference, but on older machines this might be prohibitively expensive. It is opt-in though.
- ~~I had to export some previously private functions in `Core.Binary` to help with loading the TTC files. Might be better to use the function `readFromTTC`, which was already exported?~~
- I call it an index, but really it's just a list of `GlobalDef`s. Not sure how much memory that's guzzling.
- Currently only indexes definitions from package_dirs, not from the current package.
- This is my first time contributing. I might generally be doing some things catastrophically wrong.

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -32,6 +32,11 @@ This CHANGELOG describes the merged but unreleased changes. Please see [CHANGELO
 * The `idris2 --exec` command now takes an arbitrary expression, not just the
   function name.
 
+* A new `idris2 --ide-index` option reads all modules from dependencies and
+  uses them to provide completion for symbols that aren't yet imported.
+  If neither `--ide-mode` nor `--ide-mode-socket` is set, this option
+  automatically enables `--ide-mode`.
+
 * Command-line arguments beginning with `--` which are not a known flag now
   produce an error.
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -32,6 +32,7 @@ George Pollard
 GhiOm
 Giuseppe Lomurno
 Guillaume Allais
+Hakan Aras
 Hiroki Hattori
 Ilya Rezvov
 Jacob Walters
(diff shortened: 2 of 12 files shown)


hakanaras:
> I had to export some previously private functions in Core.Binary to help with loading the TTC files. Might be better to use the function readFromTTC, which was already exported?

~~Actually not as simple as just using `readFromTTC`. We need the `currentNS` from the `TTCFile`, otherwise we can't get the module to import from the `Name`/`GlobalDef` later it seems.~~

hakanaras:
Made some changes as recommended by dunham on Discord.
- Now uses the already exported `readFromTTC` from `Core.Binary`. (The module identifier/namespace is now determined based on the file path of the ttc file that is being read.)
- If one module re-exports another, now both modules are suggested as auto-imports. The list of re-exports is also added to the `IDEIndex` for other features to use.


https://github.com/user-attachments/assets/2f32fd2f-6049-46c4-81b1-e3846e573554

PR #3643, opened 2025-09-10
[ fix #399 ] Thread-safe way to invoke GC at program end

### LINKED ISSUE #399: Make test gives an error on Windows 10 msys2
This issue comes up in the tests for the latest HEAD running on MinGW 64-Bit, as a part of make test, with idris2/perf003

Error message:

`idris2/perf003: succesException in collect: cannot collect when multiple threads are active
rm: cannot remove 'libmkalloc.dll': Device or resource busy
make[2]: *** [Makefile:25: clean] Error 1`

However, it still says at the end `264/264 tests successful`

cypheon:
# Description

According to Chez's docs, "the thread that invokes collect must be the only active thread".  This implies that the only safe way to manually trigger GC in a (potentially) multi-threaded program is by `(collect-rendezvous)`.  That function will pause all threads and make one of them call the `collect-request-handler`.

`collect-request-handler` in turn will also call
`(blodwen-run-finalisers)` so it doesn't need to be called explicitly.

## Self-check

- [✓] If this is a fix, user-facing change, a compiler change, or a new paper


cypheon:
Looks like the custom `collect-request-handler` does not always invoke `(blodwen-run-finalisers)` and does not always perform a full collection either, so I will rework this PR to fix that.

Matthew-Mosior:
This also address the issues seen in the `channels009` test.

buzden:
> so I will rework this PR to fix that

@cypheon Have you had a chance to look at this?

My understanding is that the current state of PR does strictly better, despite not being ideal. Is it correct? I mean, if this PR fixes some problems but not all of them, wouldn't it be good to merge it anyway?

Matthew-Mosior:
> My understanding is that the current state of PR does strictly better, despite not being ideal. Is it correct? I mean, if this PR fixes some problems but not all of them, wouldn't it be good to merge it anyway?

I think this might have been addressed via [e877066](https://github.com/idris-lang/Idris2/pull/3643/commits/e8770665cb980adfeb8586af641d62ae68494b0a) right @cypheon?

cypheon:
> @cypheon Have you had a chance to look at this?

@buzden: yes, I fixed this remaining issue with my push of commit e877066. I should have commented on the PR to explicitly mark this as resolved...

So yes, this PR is an improvement on the current status quo. To my knowledge, there are no issues with the current state of this patch. So I'd be happy if it could be merged soon :)

PR #3645, opened 2025-09-16
[ fix ] Render only first paragraph of module docstring in `--mkdoc` documentation's index

UARTman:
# Description

Currently, `--mkdoc` dumps the entire module docstring into the module description on `--mkdoc`-generated documentation's `index.html` page. However, the relevant [documentation](https://idris2.readthedocs.io/en/latest/reference/documenting.html) states that:
> The first paragraph of the documentation is taken to be an overview, and in some contexts, only this overview will be shown.

While it does not specify "some contexts", the default assumption when viewing the documentation's module index is that it should indeed display only the overviews of the module docstrings. The documentation tools for other languages (i.e. Rust's `rustdoc`) also take this approach when it comes to module overview.

This patch alters the `renderDocIndex` so that it cuts off the module docstring at the first blank line, thus leaving only the first paragraph of the docstring (since, as per the documentation, the docstrings are supposed to be written in markdown).

## Self-check


tests/idris2/mkdoc/mkdoc001/run
```diff
@@ -0,0 +1,4 @@
+. ../../../testutils.sh
+
+idris2 --mkdoc test.ipkg
+head -n -1 build/docs/index.html
```
buzden: I'd suggest to add a small filter to remove the header from the `expected` file:

```suggestion
cat build/docs/index.html | sed '1,/<\/header>/d'
```

buzden:
Looks good! I'd suggest also to test that the full package description is present in the description of the whole module (i.e., that existing behaviour hasn't broken), i.e. e.g. adding something like
```shell
for i in `ls build/docs/docs/Test*.html | sort`; do
  echo "---------"
  echo "File: $i"
  cat $i | sed '1,/<\/header>/d'
done
```

gallais (approved):

PR #3646, opened 2025-09-22
docs: copied Idris 1 elab reflection docs

wizard7377:
# Description

Copied the Idris 1 docs on elab reflection


mattpolzin:
I don't think the caveat "this may be inaccurate because it was copied from Idris 1" quite does enough to make these documents worth having without any modifications. There's good stuff in there no doubt, but the docs you've copied mention Haskell a lot, spend a fair amount of energy on tactics which is not at all a focus of elab reflection in Idris 2, and contain broken links to images (unless I'm not understanding how the links resolve).

There is a world where these documents are a starting place, but I don't think we should merge them as is.

buzden:
Well, elaborator reflection in Idris 2 is completely different from one in Idris 1, so I think it's even harmful to just copy this part of doc.

Moreover, there is a good [introduction tutorial](https://github.com/stefan-hoeck/idris2-elab-util/blob/main/src/Doc/Index.md) by Stefan which contains basic description of main concepts of elaborator reflection of Idris 2. Probably docs could have a reference to it, or at least one can look for stuff that is really present in elaborator reflection of Idris 2

wizard7377:
I was doing this as a demonstration of the more general idea of porting some Idris 1 docs to the Idris 2 docs with a note.

PR #3656, opened 2025-10-28
[ release ] Prepare v0.8.0

CodingCellist:
# Description

It has been a while since we last did a release, which is getting increasingly silly (and problematic) given that there have been _numerous_ improvements to Idris2 in the past two years (!!).

I had a chat with Edwin today who said he was happy if a release was just put together to "officially" launch all the stuff which has made it in by now. I was thinking end-of-week might be nice, just to have a Hallowe'en release? (Idris being a Scottish PL and all that)  ^^

## Aside: Release cycle

Stefan pointed out on the Discord that we really ought to have a more stable release schedule, and that is absolutely true. I had an in-person chat with Edwin today, who is more than happy for a quarterly-or-so release schedule, especially as it would help convey that Idris2 is very much actively being worked on.

I propose this go in an issue or PR for now, so that we can discuss the details there, and put it in some parts of the documentation. Probably the readme or contributing file? And the website and/or readthedocs should probably also mention whichever schedule we decide on?.

## Self-check


gallais:
> quarterly-or-so release schedule

Unless we have someone on a somewhat stable contract willing to do this; I don't think we should commit
to any kind of regularity.

andrevidela:
With the right setup, the release could be entirely automated beside clicking a button. This is the entire point of the `D` in "CI/CD". I think we should do it.

CodingCellist:
> With the right setup, the release could be entirely automated beside clicking a button.

Or running a script (I may or may not have a decent amount of WIP towards this)

jfdm:
I agree with @gallais here, we should avoid *over committing*. That being said, if we can make the process reasonably straightforward, then I think it is reasonable to have six monthly releases where someone can find time to push the required buttons. 

Selfishly, having a release in late Spring would help, as this is when our lab images for the next academic year are first pressed. A corresponding one in Autumn would provide balance.

andrevidela:
If that's too much of a commitment we can also setup a cron job that runs the release script every 4 months (or at key dates list April 1st) and literally have nothing to do

mattpolzin:
This is beginning to be more of a conversation than a PR review; that said, I'm going to continue conversing 😀.

Fully automated may be a step too far; we run the risk of cutting a release with a bug in it that gets fixed the next day and then we need to decide to manually publish a patch release or not anyway.

CodingCellist:
I believe the `pack` CI failures are due to the same hitch we hit on the previous release: `idris2-pack-db/collections/HEAD.toml` has the compiler version fixed to the previous release, causing CI to fail.

Previous discussion: https://github.com/idris-lang/Idris2/pull/3155#issuecomment-1859048229

CodingCellist:
Confirmed with Stefan, and it is indeed the same issue as last time.  I'm going to go ahead with the release, then pack-db's version can be bumped, and then everything should be happy again  : )

PR #3657, opened 2025-10-29
[ fix ] Allow capitalised package names with dashes

spcfox:
# Description

The package name can start with a capital letter or contain a hyphen. However, due to an error in the lexer, it cannot do both. `Package` and `package-test` are valid names, but `Package-test` is not. Fixed

The problem was that the lexer in the string `"Ab-cd"` recognised `"Ab"` as `namespacedIdent`.

## Self-check

### DIFF
--- a/src/Libraries/Text/Lexer/Core.idr
+++ b/src/Libraries/Text/Lexer/Core.idr
@@ -167,7 +167,7 @@ tokenise pred line col acc tmap str
                Just (tok, rest) =>
                  let line' = line + cast (countNLs tok)
                      col' = getCols tok col in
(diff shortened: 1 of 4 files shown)


stefan-hoeck:
Personally, I'd advise against favoring `pack` over `fastPack` (same for `unpack`) *unless* you are sure the function needs to reduce during unification. Transform rules are not the most stable aspect of Idris, so I'd rather be on the safe side in performance critical code.

In particular I advise exchanging the `fast...` versions for the other ones unless you are absolutely sure they were not written out explicitly on purpose.

PR #3663, opened 2025-11-02
[ make ] Pass `IDRIS2_PREFIX` in `install-libs`

spcfox:
# Description

The command `make install-libs IDRIS2_PREFIX=/some/path` installs libraries in `/some/path`,
but `make install-libs PREFIX=/some/path` installs them in the default directory (`yprefix`, by default `~/.idris2`)

This difference is not obvious from the description of the variables:
https://github.com/idris-lang/Idris2/blob/72864ac9407a03c0dd3814b2f184013261198b25/docs/source/reference/envvars.rst?plain=1#L14-L15

I suggest passing `IDRIS2_PREFIX` when installing libraries, to fix this. Then both commands will install libraries to the specified path.

## Self-check


spcfox:
The `pack` build failed due to an bug I made in a recent PR https://github.com/stefan-hoeck/idris2-pack/pull/353. I am working on it.

spcfox:
If I understand correctly, `PREFIX` is the path where Idris and libraries should be installed, and `IDRIS2_PREFIX` is the path to the libraries of the current Idris. This is important when calling, for example, `idris2-exec`.

It seems that the current `install-libs` violates this semantics and uses `IDRIS2_PREFIX` as the installation path. Perhaps we should pass `PREFIX`?
```diff
- ${MAKE} -C libs/prelude install IDRIS2=${TARGET} IDRIS2_PATH=${IDRIS2_BOOT_PATH} IDRIS2_INC_CGS=${IDRIS2_CG}
+ ${MAKE} -C libs/prelude install IDRIS2=${TARGET} IDRIS2_PATH=${IDRIS2_BOOT_PATH} IDRIS2_INC_CGS=${IDRIS2_CG} IDRIS2_PREFIX=${PREFIX}
```

buzden:
Well, historically, `PREFIX` variable wasn't used at all, despite it's common usage overall. Only `IDRIS2_PREFIX` was controlling place where the compiler was installed. Support for `PREFIX` variable was added as an attempt to be compatible with system package managers in pre-pack era. So, as far as I understand, `IDRIS2_PREFIX` should act as a fallback for `PREFIX`, if it is not set.

spcfox:
> The `pack` build failed due to an bug I made in a recent PR [stefan-hoeck/idris2-pack#353](https://github.com/stefan-hoeck/idris2-pack/pull/353). I am working on it.

Already fixed

PR #3665, opened 2025-11-03
Fix #3662

### LINKED ISSUE #3662: Regression in 0.8.0: build fails due to unconditional usage of `aligned_alloc`: `memoryManagement.c:48:28: error: implicit declaration of function 'aligned_alloc'`
```
/opt/local/var/macports/build/idris2-b99237ae/work/compwrap/cc/opt/local/bin/gcc-mp-14 -Os -arch ppc -Wall -fPIC -Wall -fPIC -Wall -fPIC -O2 -I/opt/local/include  -c -o memoryManagement.o memoryManagement.c
memoryManagement.c: In function 'idris2_newValue':
memoryManagement.c:48:28: error: implicit declaration of function 'aligned_alloc' [-Wimplicit-function-declaration]
   48 |   Value *retVal = (Value *)aligned_alloc(
      |                            ^~~~~~~~~~~~~
memoryManagement.c:6:1: note: include '<stdlib.h>' or provide a declaration of 'aligned_alloc'
    5 | #include "runtime.h"
  +++ |+#include <stdlib.h>
    6 | 
memoryManagement.c:48:28: warning: incompatible implicit declaration of built-in function 'aligned_alloc' [-Wbuiltin-declaration-mismatch]
   48 |   Value *retVal = (Value *)aligned_alloc(
      |                            ^~~~~~~~~~~~~
memoryManagement.c:48:28: note: include '<stdlib.h>' or provide a declaration of 'aligned_alloc'
make[2]: *** [memoryManagement.o] Error 1
make[2]: Leaving directory `/opt/local/var/macports/build/idris2-b99237ae/work/Idris2-0.8.0/support/refc'
make[1]: *** [build-refc] Error 2
```

`aligned_alloc` is not guaranteed to be available and should be checked at configure time.

seagull-kamome:
# Description

This PR aims to resolve issue #3662 by using posix_memalign instead of aligned_alloc on older versions of macOS.

## Self-check

### DIFF
--- a/support/refc/memoryManagement.c
+++ b/support/refc/memoryManagement.c
@@ -42,22 +42,25 @@ void idris2_dumpMemoryStats(void) {
 void idris2_dumpMemoryStats() {}
 #endif
 
-Value *idris2_newValue(size_t size)
-{
+Value *idris2_newValue(size_t size) {
   /* Try to get memory aligned to pointer-size. Prefer C11 aligned_alloc
-    (not available on some platforms like older macOS), then posix_memalign,
-    and finally fall back to malloc which typically returns pointer-aligned
-    memory suitable for our needs. */
-#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) && \
+     (not available on some platforms like older macOS), then posix_memalign,
+     and finally fall back to malloc which typically returns pointer-aligned
+     memory suitable for our needs. */
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) &&              \
     !defined(__APPLE__) && !defined(_WIN32)
-  Value *retVal = (Value *)aligned_alloc(sizeof(void *),
-                                        ((size + sizeof(void *) - 1) / sizeof(void *)) * sizeof(void *));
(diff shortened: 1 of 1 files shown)


andrevidela:
@barracuda156 Can we confirm this fix works for you?

barracuda156 (approved):
Fixes the build for me on macOS, thank you!

barracuda156:
P. S. For the record – I am not sure this is needed, since it will be more complicated – there is a way to use macOS system macros to specify exactly what is or is not available.
```
#ifdef __APPLE__
#include <AvailabilityMacros.h>

#if MAC_OS_X_VERSION_MIN_REQUIRED >= 101300
/* aligned_alloc is supported */
#elif MAC_OS_X_VERSION_MIN_REQUIRED >= 1060
/* posix_memalign is supported */
#else
/* fallback to malloc */
#endif
#endif
```
I do not know whether there is any advantage with `aligned_alloc` as compared to `posix_memalign`. Also, while strictly speaking 10.5 does not have `posix_memalign`, this is redundant in MacPorts, since `posix_memalign` is supplied by `legacy-support` library, which is used with `idris2` anyway now for another reason.
So quite likely that going with `posix_memalign` is fine.

andrevidela:
thanks for checking! I think this will do for now

PR #3666, opened 2025-11-04
[ cleanup ] Post v0.8.0 cleanup

spcfox:
# Description

Need to merge PR #3659 first for green CI

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -7,6 +7,28 @@ should target this file (`CHANGELOG_NEXT`).
 
 ## [Next version]
 
+### Compiler changes
+
+* Removed modules and functions moved to `base`:
+  - `Libraries.Data.Fin` → `Data.Fin`
+  - `Libraries.Data.IOArray` → `Data.IOArray`
+  - `Libraries.Data.List.Extra.minimum` → `Data.List.minimum`
+  - `Libraries.Data.List.Lazy` → `Data.List.Lazy`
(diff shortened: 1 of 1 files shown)


spcfox:
@Z-snails, is this comment still relevant? What should we do?
https://github.com/idris-lang/Idris2/blob/8bfe6b92fc0f9f28c4922f0aeb21020084212c74/src/Compiler/Opts/Constructor.idr#L82-L83

Z-snails:
> @Z-snails, is this comment still relevant? What should we do?
> 
> https://github.com/idris-lang/Idris2/blob/8bfe6b92fc0f9f28c4922f0aeb21020084212c74/src/Compiler/Opts/Constructor.idr#L82-L83

I think integerToNat and natToInteger can be removed from there (although it's worth checking that the generated code is the same).
I don't think `%transform` is reliable enough to remove the other functions (plus, times, compare, etc).

spcfox:
Since we are not sure that this is reliable, I suggest not including it in the cleanup PR.

CodingCellist (changes requested):
There appears to be quite a lot of removals in this one, and while I appreciate that some of them were marked in comments as "remove after next release", could we please have one or two bullet-points in the `CHANGELOG_NEXT.md` summarising what has been removed and why (and what to use instead, if there is anything).

spcfox:
@CodingCellist, I'm not sure how best to handle minor fixes in the CHANGELOG. I've tried to keep it brief.

CodingCellist (approved):

Issue #2325, opened 2022-02-16
Case splitting - not covering


MilanKral:
The following code typechecks
```idris
fnc : (a : Nat) -> (b : Nat) -> (a > b) = True -> Type
fnc 0 0 prf = ?fnc_rhs_2
fnc 0 (S k) prf = ?fnc_rhs_3
fnc (S k) b prf = ?fnc_rhs_1
```

When case splitting on prf in line 2 and 3 I get the following code:

```idris
fnc : (a : Nat) -> (b : Nat) -> (a > b) = True -> Type
fnc 0 0 Refl impossible
fnc 0 (S _) Refl impossible
fnc (S k) b prf = ?fnc_rhs_1
```

Typechecking the code, I get following error about not covering:
```
"c.idr" 4L, 136C written
Error: fnc is not covering.

c:1:1--1:55
 1 | fnc : (a : Nat) -> (b : Nat) -> (a > b) = True -> Type
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Missing cases:
    fnc 0 _ _    
```

gallais:
Try using `absurd prf` on the RHS instead of `impossible` on the LHS.
This will keep the pattern-matches in the case tree and help the coverage checker see that everything is alright.

dunhamsteve:
The issue here is that we silently ignore `impossible` clauses with `fromInteger` on the left side. So essentially, you can't have numbers in impossible clauses and Idris doesn't tell you that.  If you put `Z` instead of `0`, it would work.

I have now made a PR that emits a warning when an impossible clause is not accepted.

PR #3675, opened 2025-11-09
[ warning ] Warn user when impossible clauses are ignored

dunhamsteve:
# Description

Idris sometimes silently ignores `impossible` clauses, which leads to confusing coverage errors. This PR warns the user when that happens.

Cases where it can happen are:

- A name is ambiguous and Idris can't resolve the ambiguity via the `match` heuristics in `Impossible.idr`
- Functions like `fromInteger` or `fromString` are used in an `impossible` clause
- A constructor of the wrong type is used (e.g. `Z` when a `Vect _ _` is expected)

These cases would throw an error which is caught, and then the `impossible` clause is ignored.  I have changed the code to emit a warning that the clause is being ignored. For compatibility and because the ignored clause is harmless, I chose not to make it an error.

As an example, in #2325 a `0` appears in the LHS of an `impossible` clause and it is ignored because `fromInteger` is not handled in impossible clauses. The function is reported as not covering because of the ignored clauses. With the warning, the user would know that the clause is being ignored and they could use `Z` instead.

And in frex-project/idris-frex#74, `ConsUlt` and `Ultimate` were ambiguous, causing the `impossible` clause to be silently ignored. The function was reported as not covering, and I made a work-around that didn't use `impossible`. If I had known the root cause of the coverage error, I would have simply qualified the ambiguous names.

I removed some clauses in `libs` that generate these warnings, since they had been ignored anyway. If we want to retain them as documentation, another option is to change `0` to `Z`. I also removed a bad `impossible` clause from `Krivine.idr` that generated a warning. It referenced `Machine.Beta`, but the scrutinee has type `Refocus.Trace`, so the impossible constructor is `Step`.

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -13,6 +13,12 @@ should target this file (`CHANGELOG_NEXT`).
 
 ### Backend changes
 
+### Compiler changes
+
+* The compiler now warns the user when `impossible` clauses are ignored. This
+  typically happens when a numeric literal or an ambiguous name appears in an
+  `impossible` clause.
+
 #### RefC Backend
 
 * Fixed an issue to do with `alligned_alloc` not existing on older MacOS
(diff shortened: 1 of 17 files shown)


spcfox:
Looks good to me!
I've been experimenting with improving recognition of impossible cases, and this is what I've been missing. It's hard to tell now whether the impossible case had any impact.

PR #3684, opened 2025-11-21
[ docs ] Update look 'n feel

jfdm:
Traditionally, Sphinx required authors to use reST and provided basic markdown support. Now, myST is a variant of markdown that incorporates many reST features but in a markdown many are familiar with.

This PR does the following:

1. Introduce myST as a dependency, enabling authors to use myST for input;
2. Change the style to `sphinx-book-theme` a more featured html style that enables more featured documentation to be written.
   Including, light/dark modes.
3. Pull in bibtex tooling and 'copy button' for code.
4. Provide instructions for managing python projects using the `uv` tool.

# Description

## Self-check


jfdm:
I will update the changelog later.

gallais:
Do you know if there's anywhere we can see the rendered result?

Files like `uv.lock`  seem autogenerated (unless you're the SHA master who can see
through them like through matrix code :D); should we document how/when to
regenerate them or are we not expecting them to change?

jfdm:
> Do you know if there's anywhere we can see the rendered result?

urgh, I thought there would be something from my github. Alas no.

So I have dumped [something temporary here](https://personal.cis.strath.ac.uk/jan.de-muijnck-hughes/temp-idris/)

> Files like uv.lock seem autogenerated (unless you're the SHA master who can see
> through them like through matrix code :D); should we document how/when to
> regenerate them or are we not expecting them to change?

We can add a note. For `uv`, the lock file [is generated automatically](https://docs.astral.sh/uv/concepts/projects/sync/) when `uv` is invoked.

dunhamsteve:
Am I correct in assuming that `uv.lock` only updates when you add or update a python dependency? (Rather than every time you invoke uv.)

jfdm:
Let me clarify, [locking and syncing are automatic in `uv`](https://docs.astral.sh/uv/concepts/projects/sync/). Thus, each time `uv` is called the lock file (i.e. the dependencies) are checked for freshness and updated if required. Given the dependencies for the documentation tooling, I suspect updates will be minimal and far between.

Note, having lock files in the code base for this project is not new. We keep the `flake.lock` for nix. We should be consistent in accepting them.

jfdm:
It looks like the linter is being overly pedantic on the `conf.py` file, most of which I did not touch. I will fix these issues, but it is kinda annoying seeing them. If there is a chance to configure the linter to be less pedantic, that should be discussed elsewhere but something to consider.

jfdm:
Merci! I was going to look at this later this afternoon.

PR #3688, opened 2025-11-26
[ ci, test ] Test compiler builded with Racket and some fixes in tests

spcfox:
# Description

The tests `idris/basic/case001` and `codegen/enum` fail when the compiler is built using Racket. We could have caught this earlier if `ubuntu-self-host-racket` were being run in CI. It looks like it was disabled many years ago, and since it is currently running without errors, I propose re-enabling it. I've set the timeout to 60 minutes because Racket runs more slowly than Chez (the other tests use 45 minutes).

Some additional fixes to the tests:
1. Tests for specific backends have been moved to the corresponding directories.
2. Removed the codegen specification from `run` for tests located in backend directories, as their code generator is specified in `tests/Main.idr`. Specifying the code generator in `run` makes a test less portable, because it causes the test to ignore the code generator set in `Main.idr`. For example, if such a test is copied from chez to racket, it would still use Chez. Therefore, I suggest avoiding duplication of the code generator specification in `run`.
3. Added debug output in `codegen/fix3515`. Issue #3515 is not backend-specific, so it’s useful to check that `compexpr` has not changed.
5. Removed unused argument of the `findTests` function in Golden.

## Self-check


buzden:
I think this re-enabling is reasonable, despite bigger time values, because as far as I remember, we removed those tests prior we had "quick-check" job **and** prior to decision to stop jobs of the same branch on new commits. None of this arguments are relevant now.

Also, putting things in order in backend-related stuff seems to good pretty good.

spcfox:
In #3691 I added a test `case003`. This test is not for case but for casetree, so I moved it

Issue #3357, opened 2024-07-26
Erased type-casing function seen as total when it's not


gallais:
I suspect the issue is that we see `{0 a : Type}` and assume for coverage
checking purposes that `a = Nat` is a forced pattern rather than a type case.

```idris
%default total

0 oops : {0 a : Type} -> a
oops {a = Nat} = Z

boom : Void
boom = void oops
```

Issue #3690, opened 2025-11-27
Does not accept a simple impossible case


spcfox:
# Steps to Reproduce

```idris
data NatEq : Nat -> Nat -> Type where
  Z : NatEq Z Z
  S : NatEq n m -> NatEq (S n) (S m)

foo : NatEq 1 2 -> Void
foo _ impossible
```

# Expected Behavior

Successful type checking

# Observed Behavior

```idris
Error: foo _ is not a valid impossible case.

Main:6:1--6:17
 2 |   Z : NatEq Z Z
 3 |   S : NatEq n m -> NatEq (S n) (S m)
 4 | 
 5 | foo : NatEq 1 2 -> Void
 6 | foo _ impossible
     ^^^^^^^^^^^^^^^^
```

# Description

This is not a bug, because we do not expect compiler to be able to infer that a case is impossible in all situations. However, I think this example is sufficiently trivial that it would be nice if the compiler could handle it.

For `NatEq 0 1` it cannot find a matching constructor, but for `NatEq 1 2` it does find one, and it does not check whether its arguments are inhabited.

buzden:
My opinion is when, roughly speaking, compiler is able to `%search` `Nat 2 1`, then it should be able to prove impossibility of `Nat 1 2`

dunhamsteve:
It seems like it could be exponential if we start recursing on these arguments, unless we've got some heuristics to help (e.g. return "don't know" if there are multiple candidates and maybe limit depth). 

Given other issues we have with `impossible`, I suspect Agda may have a better approach by asking the user to expand the LHS and point out the empty bit with `()`. If we do drill down in Idris it does sort this case out:
```idris
foo : NatEq 1 2 -> Void
foo (S _) impossible
```
But doing that would be painful with something like `NatEq 42 51 -> Void`.

gallais:
> I suspect Agda may have a better approach by asking the user to expand the LHS and point out the empty bit with (). 

💯

> But doing that would be painful with something like `NatEq 42 51 -> Void`.!

Which is a good hint that one should prove a general theorem and deploy it rather
than trying to insert more (costly) automation.

PR #3691, opened 2025-11-27
[ fix ] Improvements and fixes to the coverage checker

### LINKED ISSUE #1800: Incorrect coverage checking with pattern containing %search
# Steps to Reproduce
```idris
import Data.List
import Data.List.Elem
import Decidable.Decidable
import Decidable.Equality

%default covering

toWitness : (prf1 : Dec a) -> IsYes prf1 -> a
toWitness (Yes prf2) x = prf2

data Singleton : {0 k : String} -> k `Elem` ks -> Type where
  Is : (k : String) -> {0 ks : List String} -> {0 prf' : IsYes (k `isElem` ks)}
    -> Singleton (toWitness (isElem k ks) prf')

Example : List String
Example = ["String1", "LongerString", "LastString"]

total
mycase : {0 s : String} -> (0 prf : s `Elem` Example) -> (single : Singleton prf) -> Nat
mycase .(Here) (Is "String1"     ) = 0
--mycase .(There Here) (Is "LongerString") = 1
mycase .(There $ There Here) (Is "LastString"  ) = 2
mycase %search n impossible
```
# Expected Behavior

Report a coverage error, since the commented-out case is a valid missing case.

# Observed Behavior

Type-checks fine.

Notes: 
1. I'm sorry this is a bit convoluted, we tried to minimise it further but couldn't easily. Ideas welcome.
2. This isn't too convoluted an example, we want to use this technique in [collie](http://www.github.com/ohad/collie) to make it nice to match on user-defined records.

### LINKED ISSUE #1998: Proving false is true in "0" space with implicit vars
https://gist.github.com/redfish64/203dddfe8e3dc532b8c7fb2da374177b
```
0 proveit : c = True
proveit {c=True} = Refl

0 falseIsTrue : False = True
falseIsTrue = proveit {c=False}
```
# Steps to Reproduce

Typecheck

# Expected Behavior

Failure

# Observed Behavior

OK!

### LINKED ISSUE #2159: covering checker has a bug
I want to write a proof that \nexists r \in N : r+r = 1. Rewriting this, gives me

`plusRRNever1 : {r:Nat} -> plus r r = 1 -> Void`

which shouldn't be too hard. However, I ran into the following issue, trying to do a case-split on `r`

```
plusRRNever1 : {r:Nat} -> plus r r = 1 -> Void
plusRRNever1 {r = 0} Refl impossible
plusRRNever1 {r = (S 0)} Refl impossible
plusRRNever1 {r = (S (S k))} prf = ?somethingWrong
```

return the following error message

```
plusRRNever1 is not covering.

errorCase:5:1--5:47
 1 | 
 2 | import Data.Nat
 3 | 
 4 | 
 5 | plusRRNever1 : {r:Nat} -> plus r r = 1 -> Void
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Missing cases:
    plusRRNever1 _
```

which is clearly wrong

### LINKED ISSUE #2318: Inconsistency due to pattern matching error
I've found an inconsistency bug:
```idris
%default total

A : Type
A = ((), Nat)

Z : A
Z = ((), 0)

S : A -> A
S a = ((), S (snd a))

data P : A -> Type where
    C : P (S a)

data Q : A -> Type where
    U : Q Z
    V : Q (S a)

oopsie : P a -> Q a -> Void
oopsie C U impossible

bad : Void
bad = oopsie (C {a = Z}) (V {a = Z})
```
It seems that `oopsie C V` is not recognized as a valid case, even though it clearly is, therefore `oopsie` is considered total. Furthermore, when I try to add the `oopsie C V` clause, Idris gives the following error on the pattern `C`:

```
While processing left hand side of oopsie.
Can't solve constraint between: ?a [no locals in scope] and ?a [no locals in scope].
```

# Steps to Reproduce
Compile code.
# Expected Behavior
Should not compile, since `oopsie` is not total.
# Observed Behavior
Compiles.

### LINKED ISSUE #2690: Proof of Void due to a coverage-checking bug
# Steps to Reproduce
```Haskell
%default total

trythis : (mb : Maybe Bool) -> mb === Just True -> Void
trythis Nothing ab = absurd ab

boom : Void
boom = trythis (Just True) Refl
```

# Expected Behavior

`trythis` must be seen as partial

# Observed Behavior

Whole program is seen as total

Found by @reswatson

### LINKED ISSUE #2822: Proof of void when number of args changes depending on pattern match
# Steps to Reproduce

    total
    fn : (x : Bool) -> if x then (Nat -> Nat) else Void
    fn True x = x

    total
    f2 : Void
    f2 = fn False

# Expected Behavior

Shouldn't typecheck

# Observed Behavior

OK!

### LINKED ISSUE #3679: [ regression ] misleading compiler warning about impossible case
# Steps to Reproduce
Typecheck the following:

```idris
module Impossible

%default total

public export
interface Cogen a where
  constructor MkCogen
  perturb : a -> Nat -> Nat

export
Cogen Void where
  perturb _ impossible
```

# Observed Behavior

Generates a warning about the `impossible` clause but at the location of the definition of `perturb`:

```
Warning: Badly formed impossible clause ((Hedgehog.Internal.Function.perturb {imp:0}), ([], ([], [])))

 34 | ||| generator and use its output as an output for a function.
 35 | public export
 36 | interface Cogen a where
 37 |   constructor MkCogen
 38 |   perturb : a -> StdGen -> StdGen
                  ^^^^^^^^^^^^^^^^^^^^^
```

Seems related to #3675

spcfox:
# Description

Fixes various issues in the coverage checker, primarily those that lead to a proof of false. Also improves the checking of impossible cases (both user-defined and compiler-inferred).

1. Added test for #2159. It seems to have already been fixed in main branch. Closes #2159.
1. Added test for #2690. It seems to have already been fixed in main branch. Closes #2690.
1. Fixed #1800
1. Fixed #1998
1. Fixed #2318
1. Fixed #2822
1. Fixed #3679
1. Fixed several previously unknown proofs of false and bugs (see tests)
1. ~~Fixed (reverted) #3357~~
    The [fix](https://github.com/spcfox/Idris2/commit/c8fb2f529a96e6a1e176d0c49f83c0220adf8090) significantly [worsens](https://github.com/spcfox/Idris2/commit/a70ea4bc2f88707fb7b6277574c332e2821010bc) coverage checking for erased functions. I’ll return to this later.

I do not use the `total` modifier in tests because coverage should be checked by default.

## Self-check

### DIFF
--- a/src/Core/TT/Var.idr
+++ b/src/Core/TT/Var.idr
@@ -35,6 +35,10 @@ data IsVar : a -> Nat -> List a -> Type where
 
 %name IsVar idx
 
+-- `vs` is available in the erased fragment, and the case builder
+-- pattern-matches on it. To simplify the case tree and help the
+-- coverage checker, we use an explicit dot pattern here.
+-- TODO: remove `{vs = .(_)}` once the compiler generates more optimal case trees.
 export
 0 Last : HasLength (S n) vs -> Exists (\ nm => IsVar nm n vs)
 Last {vs = .(_)} (S Z) = Evidence _ First


src/Core/TT/Var.idr
```diff
@@ -37,9 +37,8 @@ data IsVar : a -> Nat -> List a -> Type where
 
 export
 0 Last : HasLength (S n) vs -> Exists (\ nm => IsVar nm n vs)
-Last {vs = []} p impossible
-Last (S Z) = Evidence _ First
-Last (S (S p)) = bimap id Later (Last (S p))
+Last {vs = .(_)} (S Z) = Evidence _ First
+Last {vs = .(_)} (S (S p)) = bimap id Later (Last (S p))
```
spcfox, line 45: Without dotting, the case trees become complex, and the coverage checker rejects them due to #3690.

src/TTImp/Impossible.idr
```diff
@@ -115,6 +107,15 @@ mutual
...
   processArgs fn ty [] [] [] = pure fn
+  processArgs fn (NBind fc x (Pi _ _ Explicit ty) sc) [] autos named
+     = do defs <- get Ctxt
+          case findNamed x named of
+            Just ((_, e), named') =>
+               do e' <- mkTerm e (Just ty) [] [] []
+                  processArgs (App fc fn e') !(sc defs (toClosure defaultOpts Env.empty e'))
+                              [] autos named'
+            Nothing => -- Expected an explicit argument, but only implicits left
+                       badClause fn [] autos named
```
dunhamsteve: I think it's a little more subtle than this - the top level being short of arguments is fine (it will be compared for number of arguments later, in case tree building), but we need to be sure that constructors are fully applied on the LHS.  With this change, it now incorrectly accepts this as covering:
```idris
foo : Nat -> Nat
foo S impossible
foo Z = Z
```

spcfox: Fixed

src/Core/TT/Var.idr
spcfox: I don't think that's a problem. I just explained why the need for this change arose.

src/TTImp/Impossible.idr
dunhamsteve: Thanks. I also wonder if we should change the error message to something like "Too few arguments" in this case. The existing error message isn't very clear.

spcfox: Good idea! I’ve added a clearer error message

src/TTImp/Impossible.idr
```diff
@@ -88,9 +88,13 @@ mutual
                               !(sc defs (toClosure defaultOpts Env.empty e'))
                               [] autos named'
             Nothing => -- Expected an explicit argument, but only implicits left
-                       if not con && null autos && null named
-                          then pure fn
-                          else badClause fn [] autos named
+                       do let False = con
+                            | True => throw $ GenericMsg (getLoc fn) $
+                                                "Cannot match on a partialy applied constructor: "
```
Matthew-Mosior: I think we want `partially` here right?

tests/idris2/coverage/impossible003/expected
```diff
@@ -1,5 +1,5 @@
 1/1: Building Main (Main.idr)
-Warning: Main:2:5--2:6:Badly formed impossible clause (Prelude.Types.S, ([], ([], [])))
+Warning: Main:2:5--2:6:Cannot match on a partialy applied constructor: Prelude.Types.S
```
Matthew-Mosior: Same as above?

Matthew-Mosior (approved):

spcfox:
> After the last change regarding to the management of `Type`s, I think it's all okay now

I’ve found that there’s still an issue with matching on primitive types. I’m already finishing up the fix.
```idris
matchInt : () -> Void
matchInt Int impossible
```

spcfox:
Done!

src/Core/TT/Var.idr
dunhamsteve: Should we add a comment to the code explaining why? It might help if somebody has to look at this code again.

Matthew-Mosior (approved):

PR #3704, opened 2025-12-19
[ perf ] Optimize Chez, Racket and Gambit BitsXY operations using fixnums

Matthew-Mosior:
# Description

The idea for this work came from a few discussions on discord (see below) regarding truncation/bounds checking for operations on `BitsXY` numbers.

Currently, we define operations over `BitsXY` in `chez` as follows:

```
...

(define (blodwen-toUnsignedInt x bits)
  (logand x (sub1 (ash 1 bits))))
  
...

(define bu+ (lambda (x y bits) (blodwen-toUnsignedInt (+ x y) bits)))
(define bu- (lambda (x y bits) (blodwen-toUnsignedInt (- x y) bits)))
(define bu* (lambda (x y bits) (blodwen-toUnsignedInt (* x y) bits)))
(define bu/ (lambda (x y bits) (blodwen-toUnsignedInt (quotient x y) bits)))

...
```

Addition, subtraction and multiplication, regardless of the width (`8`, `16`, `32`, or `64`), goes through a truncation via `blodwen-toUnsignedInt`.

@stefan-hoeck had the idea that we could use chez's `fixnum` instead of `bignum` for the following:
- addition/subtraction <-> `Bits8`, `Bits16`, `Bits32`
- multiplication <- > `Bits8`, `Bits16`

The PR adds the following specialized functions (for each bit-width up to `32`) that utilize some of `chez`s (and extended to `racket` and `gambit`) `fixnum` specific functions, such as `fx-`, `fx+`, `fx*`.

Below shows new `chez`-specialized functions that make heavy use of `fixnum`s internally (similar for `gambit` and `racket`):

```
(define (blodwen-toUnsignedInt8 x)
  (fxlogand x 255))

(define (blodwen-toUnsignedInt16 x)
  (fxlogand x 65535))

(define (blodwen-toUnsignedInt32 x)
  (fxlogand x 4294967295))

(define (bu8+ x y) (blodwen-toUnsignedInt8 (fx+ x y)))
(define (bu16+ x y) (blodwen-toUnsignedInt16 (fx+ x y)))
(define (bu32+ x y) (blodwen-toUnsignedInt32 (fx+ x y)))
(define (bu8- x y) (blodwen-toUnsignedInt8 (fx- x y)))
(define (bu16- x y) (blodwen-toUnsignedInt16 (fx- x y)))
(define (bu32- x y) (blodwen-toUnsignedInt32 (fx- x y)))
(define (bu8* x y) (blodwen-toUnsignedInt8 (fx*/wraparound x y)))
(define (bu16* x y) (blodwen-toUnsignedInt16 (fx*/wraparound x y)))
```

Using `fixnums` should bring some inherent performance benefits compared to the original non-specific calls to `blodwen-toUnsignedInt` (when it is safe to do so).

After numerous attempts to optimize `blodwen-toSignedInt` (for `IntXY`), it became clear that it was already fast, and any specialized `fixnum` functions added no tangible performance benefit.

# Why `fx*/wraparound` and not `fx*`?

The reason `bu8*`/`bu16*` were implemented via `fx*/wraparound` is because it demonstrated a noticeable performance gain (implementing them via `fx*` did not however).

It is important to note that since the Idris type checker proves that values are within correct bounds, it is safe to use `fx*/wraparound`.

Below is a table summarizing `fx*` and `fx*/wraparound`:

| Operation | Overflow handling | Raises exception | Typical use
|---------------- | -------------------------- | ------------------------- | --------------- |
fx* | Checked | Yes | Safe arithmetic
fx*/wraparound | Modular wraparound | No | Low-level / bitwise math

# Benchmark

To get hard data on whether or not the `fixnum` specific functions shown above provide a tangible performance benefit, benchmarking was done.

Each of the sizes in question (`8`, `16` and `32`) were given numeric operation heavy tasks in order to stress test the competing implementations:
- `addition` <-> Adds sequential numbers starting from 0 up to n-1
- `subtraction` <-> Subtracts sequential numbers starting from 0 up to n-1
- `multiplication` <-> Multiplies sequential numbers starting from 1 up to n (factorial style)

## Architecture

CPU: AMD Ryzen 9 5900X 12-Core Processor
RAM: 127345212 kB
OS: Ubuntu 25.04

#### Time per run

| test       | original  |  fixnum |
| ----------  |------------- | ------------ |
| `Bits8`   | 51.42 ns | 37.32 ns |
| `Bits16` | 502.1 ns | 353.4 ns |
| `Bits32` | 123.3 μs | 105.3 μs |

#### Mean

| test       | original  |  fixnum |
| ----------  |------------- | ------------ |
| `Bits8`   | 71.42 ns |  58.75 ns |
| `Bits16` |  610.4 ns | 468.8 ns |
| `Bits32` | 162.4 μs | 142.0 μs |

#### Time per run

| test       | original  |  fixnum |
| ----------  |------------- | ------------ |
| `Bits8`   | 35.05 ns | 28.07 ns |
| `Bits16` | 502.4 ns |426.8 ns |
| `Bits32` | 124.0 μs | 105.6 μs |

#### Mean

| test       | original  |  fixnum |
| ----------  |------------- | ------------ |
| `Bits8`   | 52.27 ns |  47.31 ns |
| `Bits16` | 611.9 ns | 537.1 ns |
| `Bits32` | 161.2 μs | 145.3 μs |

#### Time per run

| test       | original  |  fixnum |
| ----------  |------------- | ------------ |
| `Bits8`   | 11.18 ns | 10.20 ns |
| `Bits16` | 15.87 ns | 14.56 ns |

#### Mean

| test       | original  |  fixnum |
| ----------  |------------- | ------------ |
| `Bits8`   | 24.39 ns |  23.25 ns |
| `Bits16` | 30.03 ns | 28.44 ns |

## Analysis

The `addition`, `subtraction`, and `multiplication` benchmarks show that the `fixnum` specific functions provide a clear and consistent performance advantage over the originals.

## Source code

You can find the source code that was used to generate the above results [here](https://github.com/Matthew-Mosior/idris2-fixnum-benchmarks).

# Discord discussions 
- [1](https://discord.com/channels/827106007712661524/838742669953204224/1344833359490322462)
- [2](https://discord.com/channels/827106007712661524/1364269409727676536/1415388060623044688)

## Self-check


dunhamsteve:
Chez fixnums are 61-bit signed integers on my platform.  I don't think that's sufficient to multiply 32-bit numbers
```
> (log (most-positive-fixnum) 2)
60.0
```

dunhamsteve:
Ok, I see you're looking at the `fixnum-width` - the code was truncated to the right on my phone. :)

You can run those tests locally with:
```
make test only=chez/newints
make test only=chez/integers
```
(or `only=chez` for all of the chez tests)

I see in that first thread that Stefan has a benchmark. It might be worth trying that or a similar benchmark to see what the improvement is.

stefan-hoeck:
Thanks for doing this. The idea was, to use *exactly* the same code as for `bu+` and the others, but to use fixnum ops in the implementation instead (*where it is safe*). In their current version, the `fast` operations add additional checks, which will make it hard to verify that they indeed perform better. Since we know the bit size of the primitives at compile time, there should be no need to check it again at runtime.

In addition, this should be done for *all* primitive operations working on these numbers, in particular, also for the comparison functions and bitwise operations.

dunhamsteve:
We should keep an eye out for this too:

> Fixnum-specific procedures normally check their inputs and outputs (where appropriate), but at optimization level 3 the compiler generates, in most cases, code that does not perform these checks.

If you change `(optimize-level 3)` in `Chez.idr` to 2. I believe it will throw an exception in any cases that might cause trouble. I don't know what the behavior is at level 3.

stefan-hoeck:
I don't know if this is already ready for review, but here are a couple of thoughts:

- If we want maximum performance for small-bit integer operations, we should avoid *all* unnecessary runtime tests and conversions. Every little piece of unnecessary work will add to the overall cost in performance-critical parts of the code.
- In particular: We know already that the numbers are fixnums (due to the Idris elaborator), so calls to `fixnum?` should not be necessary, should they?
- `blodwen-toSignedInt` and `blodwen-toUnsignedInt` do *not* use fixnum operations internally, so we should re-implement them for fixnums.
- Since the bit size is known at compile time, the bit mask used for truncation is likewise known at compile time. It should therefore *not* be computed from a "number of bits" runtime argument. This means, that we probably should have specialized `bu8+`, `bu16+`, and so on, operations for all numeric computations that might overflow.
- We need hard data demonstrating that these changes are necessary and beneficial.

Given all of the above: This is a non-trivial task and requires quite a bit of testing, both, when it comes to performance but also when it comes to safety (see Steve's comments about fixnum ops and runtime errors at optimization level 3). Personally, I'm quite certain that we will be safe without additional runtime checks, because Idris did all the work for us at compile-time, but I haven't verified this.

As a final thought: This should probably be marked as a "draft", because I see a non-trivial amount of work ahead before this is ready for merging. If possible, I'd like to review this once it is ready, but I won't be able to follow along with every additional commit.

Matthew-Mosior:
> I don't know if this is already ready for review, but here are a couple of thoughts:
> 
>     * If we want maximum performance for small-bit integer operations, we should avoid _all_ unnecessary runtime tests and conversions. Every little piece of unnecessary work will add to the overall cost in performance-critical parts of the code.
> 
>     * In particular: We know already that the numbers are fixnums (due to the Idris elaborator), so calls to `fixnum?` should not be necessary, should they?
> 
>     * `blodwen-toSignedInt` and `blodwen-toUnsignedInt` do _not_ use fixnum operations internally, so we should re-implement them for fixnums.
> 
>     * Since the bit size is known at compile time, the bit mask used for truncation is likewise known at compile time. It should therefore _not_ be computed from a "number of bits" runtime argument. This means, that we probably should have specialized `bu8+`, `bu16+`, and so on, operations for all numeric computations that might overflow.
> 
>     * We need hard data demonstrating that these changes are necessary and beneficial.
> 
> 
> Given all of the above: This is a non-trivial task and requires quite a bit of testing, both, when it comes to performance but also when it comes to safety (see Steve's comments about fixnum ops and runtime errors at optimization level 3). Personally, I'm quite certain that we will be safe without additional runtime checks, because Idris did all the work for us at compile-time, but I haven't verified this.
> 
> As a final thought: This should probably be marked as a "draft", because I see a non-trivial amount of work ahead before this is ready for merging. If possible, I'd like to review this once it is ready, but I won't be able to follow along with every additional commit.

Thank you for the feedback here.

I agree with your points, and will work through these.  In the meantime, I have converted to draft, but will re-open (hopefully soon) when this is in a place where I'd like some more feedback from the community and can hopefully get this merged in if it shows promise from a performance-standpoint.

Matthew-Mosior:
> Ok, I see you're looking at the `fixnum-width` - the code was truncated to the right on my phone. :)
> 
> You can run those tests locally with:
> 
> ```
> make test only=chez/newints
> make test only=chez/integers
> ```
> 
> (or `only=chez` for all of the chez tests)
> 
> I see in that first thread that Stefan has a benchmark. It might be worth trying that or a similar benchmark to see what the improvement is.

I like this idea, will benchmark this and report back the results.  This should give us a good idea as to how much faster (hopefully) this is than what we are currently doing.

Matthew-Mosior:
@stefan-hoeck @dunhamsteve

This should be ready for review, although the `fx*/wraparound` fixnum op utilized for `multiplication` is only available from chez scheme and racket v10.0.0 on-wards, and since at least gambit 4.0.

I'm going to update the `.github/workflows/ci-idris2-and-libs.yml` workflows so that they pull and build the latest version of chez and racket.

dunhamsteve:
Does it improve performance significantly?

Matthew-Mosior:
> Does it improve performance significantly?

Yes, significantly for `addition`/`subtraction` and noticeably for `multiplication` (see the the benchmarking section in the description above).

stefan-hoeck:
This looks now very promising. The new backend code is clean and has been shown to be more performant than what we currently have. Only one small nitpick: I'd drop the `fast` from the names of new backend functions. For instance: `bu8+` instead of `bu8+fast`. This slightly reduces the generated code size and seems to be easier on the eye.

Matthew-Mosior:
@stefan-hoeck @dunhamsteve 

Okay, this PR should now be in good shape for a proper review.

stefan-hoeck:
The Idris and Scheme parts look very good and clean now. However, I'm not sure it is feasible to enforce a lower bound of 10.3.0 on the Chez version. On the other hand: The new `wraparound` functions should not be needed. They are currently used for 8-bit and 16-bit multiplication, both of which are completely safe without the wraparound safeguard (we have a proof from the Idris type checker, that the values are within the correct bounds).

Matthew-Mosior:
> The Idris and Scheme parts look very good and clean now. However, I'm not sure it is feasible to enforce a lower bound of 10.3.0 on the Chez version. On the other hand: The new `wraparound` functions should not be needed. They are currently used for 8-bit and 16-bit multiplication, both of which are completely safe without the wraparound safeguard (we have a proof from the Idris type checker, that the values are within the correct bounds).

@stefan-hoeck 

The reason I chose to implement `bu8*`/`bu16*` via  `fx*/wraparound` is because it demonstrated a noticeable performance gain.  Implementing them via `fx*` did not however.

Below is a summary of what I found of `fx*` vs `fx*/wraparound`:

`fx*`:

`Semantics` <-> Checked fixnum multiplication.

`Behavior on overflow` <-> If the result does not fit in a fixnum, an exception is raised.

`Use case` <-> When you want correctness and safety, and overflow indicates a bug or invalid input.

`Typical lowering` -> Compiles to a machine multiply plus overflow check.

`fx*/wraparound`:

`Semantics` <-> Unchecked fixnum multiplication with modular arithmetic.

`Behavior on overflow` <-> The result wraps around modulo the fixnum word size (two’s complement behavior).

`Use case` <-> When you explicitly want wraparound arithmetic, e.g. hashing, bit-level algorithms, ring arithmetic.

`Typical lowering` <-> Compiles to a raw machine multiply without overflow checks.

| Operation | Overflow handling | Raises exception | Typical use
|---------------- | -------------------------- | ------------------------- | --------------- |
fx* | Checked | Yes | Safe arithmetic
fx*/wraparound | Modular wraparound | No | Low-level / bitwise math

stefan-hoeck:
I'm not against using the wraparound function, but if this means that Idris can only be used on Chez 10.3.0 or later, it most likely is a blocker for this PR. However, I'm not the person who has to decide that, so probably some of the Idris maintainers should have a say here.

Matthew-Mosior:
> I'm not against using the wraparound function, but if this means that Idris can only be used on Chez 10.3.0 or later, it most likely is a blocker for this PR. However, I'm not the person who has to decide that, so probably some of the Idris maintainers should have a say here.

If it makes a difference, we also have the option of only bumping to v10.0.0 instead of v10.3.0 (since v10.0.0 is the earliest version that the `fx*/wraparound` is available on Chez Scheme and v7.9 for Racket).

dunhamsteve:
Another PR is proposing to pin the version to 9.6.4 in CI (https://github.com/idris-lang/Idris2/pull/3513#issuecomment-3645887650), which would conflict with requiring 10.0.0. So could be an issue here until that bug is sorted out. (I think the issue may be because `readdir` is not thread safe.)

I personally wouldn't mind a minimal required version of 10.0.0, since older versions will not run on modern macs, but I think that might require some discussion and sorting out that `readdir` issue.

buzden:
Another consideration may be the presence of the proposed requirement in the probably most used ubuntu-latest container

GulinSS:
@dunhamsteve 

> (I think the issue may be because readdir is not thread safe.)

I think your assumption is right, going to apply a mutex fix at https://github.com/idris-lang/Idris2/pull/3513

GulinSS:
https://github.com/idris-lang/Idris2/pull/3513 Should be fine now

GulinSS:
@Matthew-Mosior my changes only these: https://github.com/idris-lang/Idris2/pull/3513/commits/7d9bc1469fa586b35a3c43335797bcd2fee51437

Your changes https://github.com/idris-lang/Idris2/pull/3704/commits/9b071c269d5bc7e69f42221c39aea57a762bee76 seemed to me are larger than mine.

If you want to try my changes, use only them. I change only macos sections to pin versions, nothing else. So, I am not sure what was cause a failure at your CI.

----

Howbeit, its macos part had [all green](https://github.com/idris-lang/Idris2/actions/runs/20834192381/job/59856881643?pr=3704) at the CI, so, the issue of red CI might be at something else, besides my changes.

Matthew-Mosior:
> @Matthew-Mosior my changes only these: [7d9bc14](https://github.com/idris-lang/Idris2/commit/7d9bc1469fa586b35a3c43335797bcd2fee51437)
> 
> Your changes [9b071c2](https://github.com/idris-lang/Idris2/commit/9b071c269d5bc7e69f42221c39aea57a762bee76) seemed to me are larger than mine.
> 
> If you want to try my changes, use only them. I change only macos sections to pin versions, nothing else. So, I am not sure what was cause a failure at your CI.
> 
> Howbeit, its macos part had [all green](https://github.com/idris-lang/Idris2/actions/runs/20834192381/job/59856881643?pr=3704) at the CI, so, the issue of red CI might be at something else, besides my changes.

@GulinSS 

Ah, looks like I'm still getting failures with 7d9bc14.

I also updated the Ubuntu, Windows and MacOS sections to v10.3.0 in d0b3098 (which is probably the key difference between 7d9bc14 and d0b3098 since everything (each and every section) for this PR needs to be at least v10.0.0 for Chez Scheme and v7.9 for Racket, in order to have access to `fx*/wraparound`).

P.S.

The `macos-self-host-chez (pull_request)` check has definitely passed with this code before no problem, but is cancelled now via timeout after 45 mins).

GulinSS:
> check has definitely passed with this code before no problem, but is cancelled now via timeout after 45 mins).

@Matthew-Mosior 

Yeah, that's what should be fixed by [mutex patch](https://github.com/idris-lang/Idris2/pull/3513/commits/64560d659fb97b78de241235a52b5a186d5d0314). It will hang randomly with chez version at least `10.2.0` and `10.3.0` without that patch. The most safe version of chez is `9.x.x`. But I hope that @dunhamsteve's idea is right.

Matthew-Mosior:
> > check has definitely passed with this code before no problem, but is cancelled now via timeout after 45 mins).
> 
> @Matthew-Mosior
> 
> Yeah, that's what should be fixed by [mutex patch](https://github.com/idris-lang/Idris2/pull/3513/commits/64560d659fb97b78de241235a52b5a186d5d0314). It will hang randomly with chez version at least `10.2.0` and `10.3.0` without that patch. The most safe version of chez is `9.x.x`. But I hope that @dunhamsteve's idea is right.

Ah okay, let me apply the patch and watch the tests again.

Matthew-Mosior:
@GulinSS 

Update:

I reverted the mutex patch and I reverted back to the version of the CI yml that worked for this PR. It seems that they wouldn't work for the this PR.  Could it still be an issue with both the mutex patch and/or the CI from #3513?

GulinSS:
To be honest, I am not sure what is happening here or what is actually breaking your build. If you are using Chez Scheme 10.3.0, you most likely need the readdir mutex patch. I have reviewed your CI attempts, and here are my observations.

It seems to me that these build cancellations are unrelated to the mutex problem because they occur very late in the process:
https://github.com/idris-lang/Idris2/actions/runs/20863471564/job/59951324256
https://github.com/idris-lang/Idris2/actions/runs/20865633666/job/59958695035
https://github.com/idris-lang/Idris2/actions/runs/20866987835/job/59962465788
https://github.com/idris-lang/Idris2/actions/runs/20869439498/job/59969362975

My original issue manifested as a hang during test execution right at the beginning. I saw no output below:

```
./build/exec/runtests /Users/runner/work/Idris2/Idris2/build/exec/idris2  --timing --failure-file failures --threads 3 --only  --except
```

However, in your attempts, the tests are actually executed. I suspect that we are dealing with an additional issue here. It might be that my readdir mutex patch is incomplete and you see another edge of this change.

I also encountered an unknown error (`nonrecoverable invalid memory reference`): https://github.com/idris-lang/Idris2/actions/runs/20868077671/job/59965583766
This error is unrelated to my patch and my issue with Chez 10.3.0.

----

Howbeit, I've rerun by PR there https://github.com/idris-lang/Idris2/pull/3513 with `Chez 10.3.0` and enabled `readdir mutex patch`: MacOS is green.

----

https://github.com/idris-lang/Idris2/pull/3704/commits/f899a0fc8f1c957c434fbe9698f9652f76e6c720 very probablly this green status with chez 10.3.0 means only a lucky attempt.

Matthew-Mosior:
> To be honest, I am not sure what is happening here or what is actually breaking your build. If you are using Chez Scheme 10.3.0, you most likely need the readdir mutex patch. I have reviewed your CI attempts, and here are my observations.
> 
> It seems to me that these build cancellations are unrelated to the mutex problem because they occur very late in the process: https://github.com/idris-lang/Idris2/actions/runs/20863471564/job/59951324256 https://github.com/idris-lang/Idris2/actions/runs/20865633666/job/59958695035 https://github.com/idris-lang/Idris2/actions/runs/20866987835/job/59962465788 https://github.com/idris-lang/Idris2/actions/runs/20869439498/job/59969362975
> 
> My original issue manifested as a hang during test execution right at the beginning. I saw no output below:
> 
> ```
> ./build/exec/runtests /Users/runner/work/Idris2/Idris2/build/exec/idris2  --timing --failure-file failures --threads 3 --only  --except
> ```
> 
> However, in your attempts, the tests are actually executed. I suspect that we are dealing with an additional issue here. It might be that my readdir mutex patch is incomplete and you see another edge of this change.
> 
> I also encountered an unknown error (`nonrecoverable invalid memory reference`): https://github.com/idris-lang/Idris2/actions/runs/20868077671/job/59965583766 This error is unrelated to my patch and my issue with Chez 10.3.0.
> 
> Howbeit, I've rerun by PR there #3513 with `Chez 10.3.0` and enabled `readdir mutex patch`: MacOS is green.
> 
> [f899a0f](https://github.com/idris-lang/Idris2/commit/f899a0fc8f1c957c434fbe9698f9652f76e6c720) very probablly this green status with chez 10.3.0 means only a lucky attempt.

Hmm, looking back at it, I've always gotten green with my CI updates (c7f77dc and 17ebdab). Im pretty sure you'll need to incorporate these changes (again, since I made sure to update every section that I could to v10.3.0).

As far as the mutex patch, I'm thinking now that some change in your PR requires it (since I get consistent greens without it).

GulinSS:
> I'm thinking now that some change in your PR requires it (since I get consistent greens without it).

However, we now know that the build may succeed for your PR without the patch, while the patch is necessary to make it work for my PR. Unfortunately, the exact reasons for this inconsistent behavior remain unclear. Therefore, gaining a clearer understanding of the issue is still preferred in order to proceed with greater confidence in the changes.

These errors — the late cancellations, readdir hangs and the invalid memory reference — look like a separate story. They seem to be outside the scope of our current pull requests, in my opinion.

Matthew-Mosior:
> > I'm thinking now that some change in your PR requires it (since I get consistent greens without it).
> 
> However, we now know that the build may succeed for your PR without the patch, while the patch is necessary to make it work for my PR. Unfortunately, the exact reasons for this inconsistent behavior remain unclear. Therefore, gaining a clearer understanding of the issue is still preferred in order to proceed with greater confidence in the changes.
> 
> These errors — the late cancellations, readdir hangs and the invalid memory reference — look like a separate story. They seem to be outside the scope of our current pull requests, in my opinion.

Agreed, I think the underlying issues you mention here deserve deeper investigation in order to resolve them, and that they are outside of the scope of #3704 and #3513.

GulinSS:
Can't wait until we have all versions of backends pinned and aligned for all supported platforms at our CI! 🎉 Pinning MacOS versions for chez/racket/coreutils was a really pain step for me in a deal with homebrew. 😅

Matthew-Mosior:
> Can't wait until we have all versions of backends pinned and aligned for all supported platforms at our CI! 🎉 Pinning MacOS versions for chez/racket/coreutils was a really pain step for me in a deal with homebrew. 😅

Yeah, this was the hardest part of this PR.  It think now with this PR, should be pinned to v10.3.0.

Below outlines essentially the same basic steps for each OS and back-end (that worked in this PR):

```
      - name: Install Chez Scheme dependencies
        run: |
          brew update
          brew install libx11
          brew install ncurses
      - name: Build Chez Scheme (v10.3.0)
        run: |
          git clone --depth 1 --branch v10.3.0 https://github.com/cisco/ChezScheme
          cd ChezScheme
          ./configure --threads --prefix="$HOME/chez"
          make
          make install
          cd ..
      - name: Add Chez Scheme to PATH
        run: |
          echo "$HOME/chez/bin" >> "$GITHUB_PATH"
      - name: Test Scheme
        run: |
          scheme --version
```

GulinSS:
>          brew install libx11
>          brew install ncurses

better to pin them, homebrew is a very fluid thing

Matthew-Mosior:
> > ```
> >      brew install libx11
> >      brew install ncurses
> > ```
> 
> better to pin them, homebrew is a very fluid thing

Great idea to pin these, I think I've achieved that via 8b17ea3.

Matthew-Mosior:
@stefan-hoeck @dunhamsteve 

Should be ready for another round of review whenever you have time.

cypheon:
Shouldn't it be safe to use the `fx*/wraparound` operation also for `Bits32` as well?

[Chez documentation says](https://cisco.github.io/ChezScheme/csug10.3.0/numeric.html#./numeric:s21):
> [...] when the result is too large to fit in a fixnum, as many high bits of the result as necessary are discarded to make the result representable as a fixnum.

Since it only truncates bits off the high end (let's say it truncates down to 60ish), it shouldn't change the final result given we truncate even further down to 32 bits afterwards anyway (via `blodwen-toUnsignedInt32`).

Matthew-Mosior:
> Shouldn't it be safe to use the `fx*/wraparound` operation also for `Bits32` as well?
> 
> [Chez documentation says](https://cisco.github.io/ChezScheme/csug10.3.0/numeric.html#./numeric:s21):
> 
> > [...] when the result is too large to fit in a fixnum, as many high bits of the result as necessary are discarded to make the result representable as a fixnum.
> 
> Since it only truncates bits off the high end (let's say it truncates down to 60ish), it shouldn't change the final result given we truncate even further down to 32 bits afterwards anyway (via `blodwen-toUnsignedInt32`).

We could use `fx*/wraparound` on `Bits32` as well, however, time profiling showed a lack of performance improvement.

Issue #3705, opened 2025-12-31
infinite loop when unifying codata with eta-equivalent functions


spikedoanz:
## steps to reproduce

create a file with the following code:

```idris
tabulate : (Nat -> a) -> Stream a
tabulate f = f 0 :: tabulate (f . S)

index : Stream a -> Nat -> a
index (x :: _)  Z     = x
index (_ :: xs) (S n) = index xs n

lemma : (f : Nat -> Nat)
        -> (k : Nat)
        -> index (tabulate (f . S)) k = index (tabulate (\m => f (S m))) k
lemma f k = Refl
```

then run:
```bash
idris2 --check file.idr
```
## expected behavior

The file should compile successfully. The two terms `(f . S)` and `(\m => f (S m))` are definitionally equal (eta-equivalent), so the `Refl` proof should be accepted.

## observed behavior

the compiler hangs indefinitely (infinite loop). it never produces output or terminates.

## analysis

the issue is in `src/Core/Unify.idr` in the `unifyNoEta` function. When unifying two `NDelay` values with `LInf` (codata/`Inf`), the code calls `unifyArgs` which fully evaluates the delayed values:

```idris
unifyNoEta mode loc env (NDelay xfc _ xty x) (NDelay yfc _ yty y)
    = unifyArgs mode loc env [xty, x] [yty, y]
```

for recursive codata like `Stream`, this causes infinite unfolding when the function arguments are eta-equivalent but syntactically different (e.g., `(f . S)` vs `(\m => f (S m))`).

the conversion check in `src/Core/Normalise/Convert.idr` handles this correctly by clearing definitions for `LInf` to prevent infinite unfolding:

```idris
convGen q i defs env (NDelay _ r _ arg) (NDelay _ r' _ arg')
    = if compatible r r'
         then do adefs <- case r of
                               LLazy => pure defs
                               _ => clearDefs defs
                 convGen q i adefs env arg arg'
         else pure False
```

but the unification code lacks this protection.

PR #3706, opened 2026-01-01
[ test ] add regression test for codata unification infinite loop

spikedoanz:
## description

adds a regression test `idris2/reg/reg060` that reproduces the infinite loop from #3705 when unifying codata with eta-equivalent functions.

**note: this test currently fails/times out.** it documents the issue and will pass once #3705 is fixed.

## test details

- test `idris2/reg/reg060` reproduces the exact issue from #3705
- includes 5-second timeout to detect hangs
- currently expected to timeout/fail until the fix is implemented

run the test:
```bash
make test only=idris2/reg/reg060
```

## Self-check

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -73,6 +73,7 @@ Rodrigo Oliveira
 Rohit Grover
 Rui Barreiro
 Ruslan Feizerahmanov
+spikedoanz
 Sam Phillips
 Simon Chatterjee
 Stefan Höck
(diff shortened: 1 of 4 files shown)


gallais:
As it stands, it will simply make the test suite fail

spikedoanz:
I can pull this back to draft in that case if there isn't an obvious good fix. I had a patch that passed regression but was a bit ad hoc.

PR #3708, opened 2026-01-08
fix(refc): Fix integer comparison on WASM32

shogochiai:
## Summary

- Fix integer comparison operators returning incorrect results on WASM32
- Add regression test for the fix

## Problem

When using the RefC backend compiled to WASM32 via Emscripten, all integer comparison operators (`==`, `<`, `>`, `<=`, `>=`) return incorrect results:

```idris
testIntEq : Int -> Int -> Bool
testIntEq a b = a == b

main : IO ()
main = do
  let r = testIntEq 5 5
  putStrLn $ if r then "True" else "False"  -- Outputs "False" on WASM32!
```

## Root Cause

In `support/refc/runtime.c`, the `idris2_extractInt` function used `idris2_vp_to_Int32` for unboxed values:

```c
int idris2_extractInt(Value *v) {
  if (idris2_vp_is_unboxed(v))
    return (int)idris2_vp_to_Int32(v);  // BUG
```

The `idris2_vp_to_Int32` macro in `_datatypes.h` has a fallback path when `UINTPTR_WIDTH` is not defined (common in Emscripten):

```c
#if !defined(UINTPTR_WIDTH)
#define idris2_vp_to_Int32(p) \
    ((idris2_vp_int_shift == 16) \
        ? (((Value_Int32*)(p))->i32) \  // Dereferences unboxed pointer!
        : ((int32_t)((uintptr_t)(p) >> idris2_vp_int_shift)))
```

On WASM32, `idris2_vp_int_shift == 16`, so the macro dereferences the unboxed pointer (created by `idris2_mkBool`) as a `Value_Int32*`, which reads garbage memory.

## Fix

For unboxed values, always use shift extraction:

```c
int idris2_extractInt(Value *v) {
  if (idris2_vp_is_unboxed(v))
    return (int)((uintptr_t)(v) >> idris2_vp_int_shift);  // Always correct
```

## Test plan

- [x] All existing RefC tests pass (21/21 including new test)
- [x] Verified fix on native (64-bit)
- [x] Verified fix on WASM32 (Emscripten + Node.js)
- [x] Added regression test: `tests/refc/wasm32cmp001`

### DIFF
--- a/support/refc/runtime.c
+++ b/support/refc/runtime.c
@@ -159,10 +159,6 @@ void idris2_removeReuseConstructor(Value_Constructor *constr) {
 
 int idris2_extractInt(Value *v) {
   if (idris2_vp_is_unboxed(v))
-    // FIX: For unboxed values, always extract via shift operation.
-    // The original idris2_vp_to_Int32 macro may incorrectly dereference
-    // the unboxed pointer as a Value_Int32* on 32-bit platforms (WASM32)
-    // when UINTPTR_WIDTH is not defined.
     return (int)((uintptr_t)(v) >> idris2_vp_int_shift);
 
   switch (v->header.tag) {


support/refc/runtime.c
```diff
@@ -159,7 +159,11 @@ void idris2_removeReuseConstructor(Value_Constructor *constr) {
 
 int idris2_extractInt(Value *v) {
   if (idris2_vp_is_unboxed(v))
-    return (int)idris2_vp_to_Int32(v);
+    // FIX: For unboxed values, always extract via shift operation.
```
mattpolzin: There's no need for a comment that describes the fix; the new change looks correct at first blush anyway IMO and anyone looking for historical context need only blame the line and/or otherwise view the git history.

shogochiai: Thanks for the review! I've removed the comment as suggested.

mattpolzin (approved):

Issue #3711, opened 2026-01-14
Primitave functions should be resugared when inspecting types


mattpolzin:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

It would be nice if type inspection yielded a result more similar to what is almost certainly written in-code when primitive functions are involved. We already resugar some select few primitives like `=` but many primitive functions are left with their primitive names.

## Motivation

Here is a hole for which the type involves primitive string concatenation:
```idirs
t : (name : String) -> "hi" ++ name ++ "!" === "hi \{name}!"
t name = ?h
```

Here is what the compiler produces when asked what the type of `?h` is:
```
   name : String
------------------------------
h : prim__strAppend "hi" (prim__strAppend name "!") = prim__strAppend (prim__strAppend "hi " name) "!"
```

## The proposal

I propose that binary primitive operations get resugared. This is a subset of all primitive operations but a much larger one than currently gets resugared. There's no reason not to expand this beyond binary operations, but I propose this as a starting place in part because it's what I have implemented so far and in part because once we merge this resugaring (if we do) then additional resugaring (e.g. of primitive prefix operators) can take binary operators as precedent.

### Examples

Take the above example Idris2 code again:
```idirs
t : (name : String) -> "hi" ++ name ++ "!" === "hi \{name}!"
t name = ?h
```

Before resugaring:
```
   name : String
------------------------------
h : prim__strAppend "hi" (prim__strAppend name "!") = prim__strAppend (prim__strAppend "hi " name) "!"
```

After resugaring of primitive binary operations is implemented, this will be the result when the compiler is asked for the type of `?h`:
```
   name : String
------------------------------
h : "hi" ++ (name ++ "!") = ("hi " ++ name) ++ "!"
```

---

A second similar example involves numeric operations:
```idris
t : (n : Integer) -> 1 + n - 3 * n === 1 - 2 * n
t n = ?h
```

Before resugaring:
```
   n : Integer
------------------------------
h : prim__sub_Integer (prim__add_Integer 1 n) (prim__mul_Integer 3 n) = prim__sub_Integer 1 (prim__mul_Integer 2 n)
```

After resugaring:
```
   n: Integer
------------------------------
h : (1 + n) - (3 * n) = 1 - (2 * n)
```

### Technical implementation

[...]

PR #3712, opened 2026-01-15
Resugar binary primitive functions

mattpolzin:
# Description

This PR is an implementation of the proposal found in issue [#3711](https://github.com/idris-lang/Idris2/issues/3711).

In short, this PR resugars primitives like `prim__strAppend` to their almost ubiquitous in-code names (e.g. the infix form of `++` instead of `prim__strAppend`). Like all resugaring, this only impacts inspecting code as can be done via the REPL, IDE protocol, or LSP.

## Self-check


buzden:
I like this. The only concern is the consistency with the current `SynInfo`-based resugaring. Say, for user function operation `(+) 1 2` would be resugared to `1 + 2` or `(+) 1 2` depending on whether fixity for `+` is present (unhidden) in the current syntactic context. How would primitives be resugared in the case of, say, hiding fixity declarations from `Prelude.Ops`?

mattpolzin:
Thanks for taking a look.

> The only concern is the consistency with the current SynInfo-based resugaring.

As far as I know, `SyntaxInfo` resugaring is done in the same place (well, the same `Resugar` module, anyway). Can you tell me what end-user experience you are thinking of where primitive operations might be resugared differently than in the code I wrote here? As far as I know, they were never resugared before (though other user-defined operators are).

> Say, for user function operation (+) 1 2 would be resugared to 1 + 2 or (+) 1 2 depending on whether fixity for + is present (unhidden) in the current syntactic context.

This refers to resugaring of an operator that is already normal form right? The current work I am doing only applies to operators that have been evaluated all the way to the point of no longer being e.g. `+` by name, instead it would be `prim__`-something by name. My only point being that my code will never impact the resugaring of user-defined operators, just the primitive ones.

That all just means that I don't think there's any conflict with other resugaring. On the other hand, if you are saying that my new code _should_ resugar e.g. `prim__strAppend` differently depending on whether the prelude fixity of `++` has been hidden/redefined or not, that is an interesting consideration. I'm not sure if I think it should resugar differently or not; it sounds very edge-casey.

mattpolzin:
Marking this ready for review. It was a draft while I waited for feedback on the proposal Issue but that proposal has not generated discussion so I'll officially consider this PR up for review.

mattpolzin:
I'm going to get this merged. It is pretty unintrusive and even if we want to rip this resugaring back out again that'll be trivial to do.

PR #3715, opened 2026-01-15
Change idris-lang.org to www.idris-lang.org in README and docs

### LINKED ISSUE #3745: Please update the github link to the idris homepage
Currently it points at https://idris-lang.org/ 

The correct working url is https://www.idris-lang.org/

Z-snails:
The DNS records/github pages config for `idris-lang.org` seems to be wrong, so the wrong TLS certificate is served for `idris-lang.org`. Using `www.idris-lang.org` avoids this issue. Also `www.idris-lang.org` is the canonical domain, since `idris-lang.org` redirects to `www.idris-lang.org`.

The ideal solution would be to fix `idris-lang.org`, but I don't know how to do that, so in the mean time this will help avoid some confusion for new-comers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![Documentation Status](https://readthedocs.org/projects/idris2/badge/?version=latest)](https://idris2.readthedocs.io/en/latest/?badge=latest)
 [![Build Status](https://github.com/idris-lang/Idris2/actions/workflows/ci-idris2-and-libs.yml/badge.svg?branch=main)](https://github.com/idris-lang/Idris2/actions/workflows/ci-idris2-and-libs.yml?query=branch%3Amain)
 
-[Idris 2](https://idris-lang.org/) is a purely functional programming language
+[Idris 2](https://www.idris-lang.org/) is a purely functional programming language
 with first class types.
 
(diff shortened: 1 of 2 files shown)


dunhamsteve:
I believe both `www.idris-lang.org` and `idris-lang.org` would be handled if the github config were changed to say `idris-lang.org` instead of `www.idris-lang.org`.  (Based on [this github documentation](https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site#configuring-an-apex-domain-and-the-www-subdomain-variant))

Also, to get ip6 working the AAAA records for `idris-lang.org` at mythic-beasts.org should be changed to:
````
2606:50c0:8000::153
2606:50c0:8001::153
2606:50c0:8002::153
2606:50c0:8003::153
```
(from the documentation at the above link).  Currently it is pointed at a mythic-beasts address.

CodingCellist:
Edwin has _just_ updated the DNS records, so fingers crossed this is resolved over the next 24-48 hours?

CodingCellist:
Fixed by @edwinb updating the DNS records.  Thanks @dunhamsteve for the comment with how to!

Issue #3719, opened 2026-01-23
macOS Chez Scheme 10.x.x testsuite running crash: "Cannot allocate memory"


GulinSS:
I've decided to extract [this](https://github.com/idris-lang/Idris2/pull/3513#issuecomment-3645887650) one into a separate issue because no reasons to think about it as a part of [Snoc list refactor](https://github.com/idris-lang/Idris2/pull/3513). It is observed during the development of that refactor but that refactor takes 1.5 years already, so, who knows what ever might be happened during such timeline. For now we [100% sure](https://github.com/spcfox/Idris2/actions/runs/21269950377/job/61219199811#step:6:40) that it is a separate issue.

-----

The macOS build pipeline intermittently fails with Chez Scheme compilation errors during the Idris2 test suite execution. The error manifests as:

```
failed to list linear: Cannot allocate memory
failed to list contrib: Cannot allocate memory
```

1. **Version Independence**: The issue appears in both Chez Scheme versions 10.3.0 and 10.2.0, and is independent of the binary source (locally compiled vs. Homebrew distributed).

2. **Error Details**: When `idris2_strerror` is modified to output the error number:
   ```c
   char *idris2_strerror(int errnum) { 
     fprintf(stderr, "idris2_strerror errnum = %d\n", errnum);
     return strerror(errnum); 
   }
   ```
   The output shows:
   ```
   idris2_strerror errnum = 12
   failed to list XXX: Cannot allocate memory
   ```
   Where error code 12 corresponds to `ENOMEM` (Out of memory).

3. **Timing Pattern**: The error appears to occur more frequently in the afternoon (GMT+3 timezone) than in the morning, suggesting potential resource contention or environmental factors.

4. **Stable Alternative**: Chez Scheme version 9.6.4 appears stable enough to run CI on macOS without this issue.

[...]

PR #3720, opened 2026-01-23
A fix for "Cannot allocate memory" at macOS with CI pinning

GulinSS:
Just to track the attempt to fix of macOS `readdir` [issue](https://github.com/idris-lang/Idris2/issues/3719)

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -255,9 +255,17 @@ cCall fc cfn clib args ret collectSafe
          argTypes <- traverse (cftySpec fc . snd) args
          retType <- cftySpec fc ret
          let callConv : Builder = if collectSafe then " __collect_safe" else ""
-         let call = "((foreign-procedure" ++ callConv ++ " " ++ showB cfn ++ " ("
-                      ++ sepBy " " argTypes ++ ") " ++ retType ++ ") "
-                      ++ sepBy " " !(traverse buildArg args) ++ ")"
+         let builded_args = !(traverse buildArg args)
+         let call = if cfn == "idris2_nextDirEntry"
+                      then
+                        "(with-interrupts-disabled "
+                          ++ "((foreign-procedure" ++ callConv ++ " " ++ showB cfn ++ " ("
+                          ++ sepBy " " argTypes ++ ") " ++ retType ++ ") "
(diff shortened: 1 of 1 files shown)


GulinSS:
An attempt to fix via using `with-interrupts-disabled` as a wrapper for `idris2_nextDirEntry` foreign calls.

src/Compiler/Scheme/Chez.idr
```diff
@@ -255,9 +255,17 @@ cCall fc cfn clib args ret collectSafe
          argTypes <- traverse (cftySpec fc . snd) args
          retType <- cftySpec fc ret
          let callConv : Builder = if collectSafe then " __collect_safe" else ""
-         let call = "((foreign-procedure" ++ callConv ++ " " ++ showB cfn ++ " ("
-                      ++ sepBy " " argTypes ++ ") " ++ retType ++ ") "
-                      ++ sepBy " " !(traverse buildArg args) ++ ")"
+         let builded_args = !(traverse buildArg args)
```
dunhamsteve, line 258: It might be better to use a scheme FFI call and do this in scheme, if we need it.  Hardcoding the name here is a big fragile. (If this is just for experimenting, that's fine.) Although we would have to reconcile the other backends.

Also, I don't know if this is going to do the trick here. Wouldn't a single `with-interrupts-disabled` need to be wrapped around both the call to `nextdir` and an fetch of the `errno`.

If we fetched the errno within the nextDirEntry, we might be good. The question is how to return the equivalent of `Either Error AnyPtr`. I don't know that Idris allows returning inductive types and if we hard code assumptions about data shape, that too might be fragile - PureScript passes in constructors to accomplish this, but Idris doesn't like those types in FFI calls.

Ideally we could return `Either Int (Ptr String)` for nextDirEntry, but I can't think of a good way to make that happen.

One option that comes to mind is to make our own `errno` - create a structure that has both `DirInfo *` and `errno` in it, and a call to ask it for its `errno`. Return that for `opendir` and free it in `closedir`. We update its errno value inside `idris2_nextDirEntry` and the idris side looks up the `errno` from this `Directory` object rather than the global one. (If RefC, Javascript, etc are not concerned with this issue, they could simply look at errno to implement `getDirError`.)

GulinSS:
Yeah, it is just a draft/sample of attempts over our existing CI at `idris-lang/Idris2`

Issue #3721, opened 2026-01-27
Inferred patterns complicate case trees


spcfox:
# Description

Consider the following function:

```idris
data IsBool : Bool -> Type where
  IsTrue : IsBool True
  IsFalse : IsBool False

foo : {b : Bool} -> IsBool b -> Int
foo IsTrue = 1
foo IsFalse = 2
```

It has the following case tree:

```
case {arg:0} of
  True => case {arg:1} of
    IsTrue => 1
  False => case {arg:1} of
    IsFalse => 2
```

One can notice that matching on `b` is redundant. The reason is that during type checking of the LHS concrete values are inferred, and it is transformed into:

```idris
foo {b = True} IsTrue = 1 
foo {b = False} IsFalse = 2
```

We can avoid this extra matching by explicitly using dot patterns:

```idris
foo {b = .(_)} IsTrue = 1 
foo {b = .(_)} IsFalse = 2
```

It would be nice if the compiler could do this automatically and construct simpler case trees without user involvement.

## Erased functions

This is most noticeable in erased functions, since all arguments are available in them. Consequently, any inferred pattern in the LHS makes the case tree more complex.

For example, this function:

```idris
0 foo : Vect n a -> Int
foo [x, y] = 1
foo _ = 2
```

matches on all `n` through the constructors `(::)` and has the following case tree:

```
case {arg:1} of
  S {e:0} => case {e:0} of
    S {e:1} => case {e:1} of
      Z => case {arg:2} of
        (::) {e:2} {e:3} {e:4} {e:5} => case {e:2} of
          S {e:6} => case {e:6} of
            Z => case {e:5} of
              (::) {e:7} {e:8} {e:9} {e:10} => case {e:7} of
                Z => case {e:10} of
                  Nil {e:11} => 1
                  _ => 2
                _ => 2
              _ => 2
            _ => 2
          _ => 2
        _ => 2
      _ => 2
    _ => 2
  _ => 2
```

For comparison, this is the case tree produced for `foo` without the `0`.

```
case {arg:2} of
  (::) {e:0} {e:1} {e:2} {e:3} => case {e:3} of
    (::) {e:4} {e:5} {e:6} {e:7} => case {e:7} of
      Nil {e:8} => 1
      _ => 2
    _ => 2
  _ => 2
```

[...]

PR #3722, opened 2026-01-27
[ new ] Dot inferred patterns in erased functions

### LINKED ISSUE #3357: Erased type-casing function seen as total when it's not

spcfox:
# Description

This PR addresses the issue of excessively complex case trees in erased functions, as described in #3721. As described there, this solution is limited to erased functions and does not apply to runtime functions.

## Implementation

The algorithm implemented in `TTImp.ProcessDef.Dot` compares the raw term with the checked term in order to identify patterns that were inferred during type checking. All such patterns are wrapped in `Dotted`, which prevents the compiler from generating case splits for them.

The effectiveness of the algorithm can be seen by examining the [changes to the `expected`](https://github.com/idris-lang/Idris2/compare/3494df83f3ca5bec926ba2e46310c40c404cd508..0be11175f0612fdf8bbab48c6513b379607353a0#diff-a8788e82394633d57481fe3334a45283a26158bc9a1620874770b6634de7884c) in `idris2/casetree/dotInferred001`:

```diff
 1/1: Building Main (Main.idr)
 Main> 0 Main.foo
 Arguments [{arg:1}, {arg:2}, {arg:3}]
-Compile time tree: case {arg:2} of
-  S {e:1} => case {e:1} of
-    S {e:2} => case {e:2} of
-      S {e:3} => case {e:3} of
-        Z => case {arg:3} of
-          (::) {e:4} {e:5} {e:6} {e:7} => case {e:4} of
-            S {e:8} => case {e:8} of
-              S {e:9} => case {e:9} of
-                Z => case {e:7} of
-                  (::) {e:10} {e:11} {e:12} {e:13} => case {e:10} of
-                    S {e:14} => case {e:14} of
-                      Z => case {e:13} of
-                        (::) {e:15} {e:16} {e:17} {e:18} => case {e:15} of
-                          Z => case {e:18} of
-                            Nil {e:19} => 1
-                            _ => 2
-                          _ => 2
-                        _ => 2
-                      _ => 2
-                    _ => 2
-                  _ => 2
-                _ => 2
-              _ => 2
-            _ => 2
-          _ => 2
+Compile time tree: case {arg:3} of
+  (::) {e:1} {e:2} {e:3} {e:4} => case {e:4} of
+    (::) {e:5} {e:6} {e:7} {e:8} => case {e:8} of
+      (::) {e:9} {e:10} {e:11} {e:12} => case {e:12} of
+        Nil {e:13} => 1
         _ => 2
       _ => 2
     _ => 2
```

This also allowed fixing #3357 without any negative impact on coverage checking.

Close #3357

## Alternative solution

Instead of comparing the raw term with the checked term, we could attempt to identify inferred terms based on the fact that they originate from metavariables. However, this approach has two drawbacks that significantly complicate the implementation:

1. Metavariables may be reduced during unification, and without eagerly marking them we would have no way of knowing that this has happened. Idris 1 appears to have had a dedicated [constructor](https://github.com/idris-lang/Idris-dev/blob/fc50b6fd9f67a0f499d08b5b81fff139a080cb2d/src/Idris/Core/TT.hs#L1003-L1005) for such terms, but for now I decided not to introduce such a fundamental change in TT.

2. In some cases, metavariables can appear in positions where the source code originally contained explicit arguments. For example, this occurs in [`checkRtoL`](https://github.com/idris-lang/Idris2/blob/2be43760d947b6d315de909bc7f44404e3e9de7a/src/TTImp/Elab/App.idr#L462).

For these reasons, I chose not to further complicate the unification algorithm and instead implemented dotting as a separate, independent pass.

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -65,9 +65,10 @@ should target this file (`CHANGELOG_NEXT`).
   typically happens when a numeric literal or an ambiguous name appears in an
   `impossible` clause.
 * Do not inline `Core.sequence`, because it is recursively defined.
-* Fixed coverage checker issues (#1800, #1998, #2318, #2822, #3679).
+* Fixed coverage checker issues (#1800, #1998, #2318, #2822, #3357, #3679).
 * Fixed totality checking in namespace and mutual blocks (#2868, #3692).
 * Fixed incorrect argument multiplicity when using an as-pattern (#3687).
+* Do not match on inferred patterns in erased functions (#3721).
 
 ### Building/Packaging changes
 


src/Core/TT/Var.idr
```diff
@@ -35,10 +35,7 @@ data IsVar : a -> Nat -> List a -> Type where
 
 %name IsVar idx
 
--- `vs` is available in the erased fragment, and the case builder
--- pattern-matches on it. To simplify the case tree and help the
--- coverage checker, we use an explicit dot pattern here.
--- TODO: remove `{vs = .(_)}` once the compiler generates more optimal case trees.
+-- TODO: remove `{vs = .(_)}`
```
spcfox, line 38: I can't remove it in this PR because the coverage checker in `main` branch will reject this function.

spcfox:
I'm not sure this PR should close #3721. The issue also describes similar problems with runtime functions that aren't addressed here. Perhaps we should open a separate issue to discuss redundant matching in runtime functions?

PR #3727, opened 2026-02-03
[ fix ] Fix infinite loop on "generate definition" command

buzden:
# Description

Currently compiler hangs when being given code like

```idris
module Nats

import Data.Nat

%default total

nats : (m : Nat) -> (n : Nat) -> (prf : NonZero n) -> m = n
```

it is asked in the REPL (or through LSP) for, e.g.

```idris
:gd 7 nats
```

This was due to occasional passing an empty list of possibilities during the search of unique name to generate. I've added a case for handling this particular case, preserving existing behaviour as status quo as possible.

This PR contains two commits, the main change in the second one, the first is just a cleaning up.


mattpolzin (approved):
Looks reasonable to me!

mattpolzin:
Would probably be nice to just guard against the infinite loop condition on `unique` but this fix is a net win regardless and its just a local function not something with an exported interface.

PR #3731, opened 2026-02-11
[ fix ] Remove trailing whitespaces in errors

spcfox:
# Description

We normally use `(<++>)` for non-empty string concatenation. If one string happens to be empty, it introduces a space at the end or beginning. This creates trailing spaces in errors, so I’m proposing a change to how it works


andrevidela:
oh my god yesss

PR #3734, opened 2026-02-12
[ fix ] Use correct FC for impossible term

### LINKED ISSUE #3733: Invalid FC in warnings about bad Impossible case
# Steps to Reproduce

```idris
data X = MkX Int String

foo : X -> Int
foo (MkX x) impossible
foo (MkX x s) = x
```

# Expected Behavior

```
Warning: Main:4:6--4:11:Cannot match on a partially applied constructor: (Main.MkX {imp:0})

Main:4:6--4:11
1 | data X = MkX Int
2 |
3 | foo : X -> Int
4 | foo (MkX x) impossible
         ^^^^^
```

# Observed Behavior

```
Warning: Main:1:14--1:17:Cannot match on a partially applied constructor: (Main.MkX {imp:0})

Main:1:14--1:17
 1 | data X = MkX Int String
                  ^^^
```

spcfox:
# Description

Fixes #3733

### DIFF
--- a/src/TTImp/Elab/App.idr
+++ b/src/TTImp/Elab/App.idr
@@ -15,6 +15,7 @@ import TTImp.TTImp
 import Data.List
 import Data.Maybe
 
+import Libraries.Data.List.Extra
 import Libraries.Data.NatSet
 import Libraries.Data.VarSet
 import Libraries.Data.WithDefault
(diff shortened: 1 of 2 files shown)


src/TTImp/Elab/App.idr
```diff
@@ -560,7 +560,7 @@ mutual
                           fntm fnty (n, 1 + argpos) expargs autoargs namedargs kr expty
 
   export
-  findNamed : Name -> List (Name, RawImp) -> Maybe ((Name, RawImp), List (Name, RawImp))
+  findNamed : Name -> List (Name, a) -> Maybe ((Name, a), List (Name, a))
   findNamed n l = case partition ((== n) . fst) l of
```
gallais: Is it worth changing this `partition` to something that's more akin to `

gallais: Is it worth defining a library function

```
search : (a -> Bool) -> List a -> Maybe (a, List a)
```

that stops as soon as it finds said value and replacing `partition`
with a call to `search`?

spcfox: Good point. Maybe we should call it `findAndDelete`?

PR #3737, opened 2026-02-13
[ fix ] Handle type synonyms at derivings

GulinSS:
This PR improves the deriving mechanism to support a broader range of type expressions in derived instances, especially for `Functor`. Previously, deriving worked when the type was represented as a bare `Name` applied to parameters. Now it handles partially applied types, type aliases in point-free style, and type expressions (for example, containing primitives and arithmetic).

**Commits**

https://github.com/idris-lang/Idris2/pull/3737/commits/565afc4e100b9c1f6fafcf5d0b4cd37cd81ce1c3
Added tests covering partially applied types (D Int), type aliases (WithD = D Int), type families with parameters (TypedContainer n). Tests include both related working cases and cases that are expected to fail but planned to be fixed during this PR.

https://github.com/idris-lang/Idris2/pull/3737/commits/c614a45f8ae0336eb6ad35e3b3eed03b4823a9e8
Previous implementation only handled type parameters that were simple Names. Introduced `TypeParameter` to represent parameters as local variables, primitives, type applications, or `IType`. Renamed `isType` to `isFamily` and extended it to parse and handle arbitrary type expressions like `TypedContainer 2` or `TypedContainer $ a + b`.

https://github.com/idris-lang/Idris2/pull/3737/commits/3bac7297baabe1361ba9a718c06f465bc80ea7b7
Added `normaliseName` function that resolves function names to their underlying type constructor. This enables support for type aliases in point-free style: `WithD = D Int` is now normalised and handled correctly. When a name is a function (not a type constructor), the code normalises it by checking `IVar fc n` and quotting the result.

https://github.com/idris-lang/Idris2/pull/3737/commits/8b0863d9fa14691882f4bd53a32777ec8ec614c7
Added `checkAccessToDefinition` to verify that type aliases have public export visibility, providing clear error messages when they don't. Also added `leftMost` helper to `Data/SnocList.idr` to retrieve the first element of a snoc-list for extracting the current function name during elaboration.

---

The following cases are not supported and are intentionally out of scope. These will be addressed (very probably) in follow-up issues/PRs:

- Data types with dependent constructors where the index is a concrete value or expression — For example:

```idris
data TypedContainer : Nat -> Type -> Type where
  Nil : TypedContainer Z v
  (::) : v -> TypedContainer p v -> TypedContainer (S p) v
```

Deriving Functor `(TypedContainer 2)` or `Functor (TypedContainer $ a + b)` is not supported. The issue is that constructors `Nil` and `(::)` are defined for specific indices (`Z` and `S p`), and the deriving mechanism cannot determine which constructors are valid for an arbitrary index expression.

- Type functions with explicit parameters (not in point-free style) — For example:

```idris
A2 : Type -> Type
A2 a = A Int a
```

Constructor arguments normalisation only works for point-free definitions like `A1 = A Int`. When parameters are written explicitly, the current implementation cannot extract and normalise the underlying type application. Note that the `point-free` case `(A1 = A Int)` is now supported by this PR.


libs/base/Deriving/Common.idr
```diff
@@ -59,24 +65,36 @@ isTypeCon fc ty = do
...
+      _ => go (S idx) acc t
     _ => go (S idx) acc t
-  go idx acc (IAutoApp _ t (IVar _ nm)) = case nm of
-    -- Unqualified: that's a local variable
-    UN (Basic _) => go (S idx) ((AutoArg emptyFC nm, idx) :: acc) t
+  go idx acc (IAutoApp _ t arg) = case arg of
+    IVar _ nm' => case nm' of
+      -- Unqualified: that's a local variable
+      UN (Basic _) => go (S idx) ((AutoArg emptyFC nm', idx) :: acc) t
+      _ => go (S idx) acc t
```
gallais: I'm not super chuffed about this. IMO we should be able to rebuild the original
type from its `isType` (which should probably be called `isFamily` instead as there
is no guarantee the type constructor is fully applied) view.

And you can indeed see in `Deriving.Functor` that we indeed sometimes use this
parameters to generate some types.

GulinSS: Thanks for the review! Looking into it

libs/base/Deriving/Common.idr
```diff
@@ -48,35 +49,142 @@ wording Func = "a function name"
...
+  where
+    getNS : TT.Name -> TT.Namespace
+    getNS (NS ns nm) = ns
+    getNS nm = TT.MkNS []
+
+    isParentOf : (given, candidate : TT.Namespace) -> Bool
+    isParentOf (MkNS ms) (MkNS ns)
+      = List.isSuffixOf ms ns
+
+normalizeTypeFunction : Elaboration m => FC -> Name -> m Name
```
spcfox: We typically use British English in our codebase, so please use `normaliseTypeFunction`

GulinSS:
@gallais, now it is possible to rebuild the original type from its `isFamily`. For example:

```
LOG derive.common.isFamily:100: Attempt to run over: D (D P P)
...
LOG derive.common.isFamily:100: Result received for D (D P P): MkIsFamily (D) ([(Arg _ (MkTPType MkIsFamily (D) ([(Arg _ (MkTPType MkIsFamily (P) ([]) ([(MkP, P)])), 0), (Arg _ (MkTPType MkIsFamily (P) ([]) ([(MkP, P)])), 1)]) ([(MkD, {0 a, b : Type} -> ({arg:3634} : b) -> D a b)])), 0)]) ([(MkD, {0 a, b : Type} -> ({arg:3634} : b) -> D a b)])
```

```
LOG derive.common.isFamily:100: Attempt to run over: D Type
...
LOG derive.common.isFamily:100: Result received for D Type: MkIsFamily (D) ([(Arg _ (MkTPIType), 0)]) ([(MkD, {0 a, b : Type} -> ({arg:3634} : b) -> D a b)])
```

GulinSS:
Tried to [compile](https://github.com/stefan-hoeck/idris2-pack-db/blob/7797cb28aaa41ae0ef5039c7fed1e1619324dfb5/.github/workflows/ci-db.yml#L69C31-L69C94) pack collection.

[errors.txt](https://github.com/user-attachments/files/25472116/errors.txt)
[nightly-260208-testdb.toml.txt](https://github.com/user-attachments/files/25472122/nightly-260208-testdb.toml.txt)
[STATUS.md](https://github.com/user-attachments/files/25472123/STATUS.md)

Seems all is fine. Existing errors are about aarch64 macos, not PR.

libs/base/Deriving/Common.idr
```diff
@@ -36,49 +38,199 @@ isFreeOf x ty
...
+             *> failAt fc "\{show n} is ambiguous"
+    pure (MkDataConName n, ty')
+
+analyzeFunctionName : Elaboration m => FC -> Name -> m (List (DataConName, TTImp))
+analyzeFunctionName fc fullName = do
+  currentFn <- getCurrentFn
+  let (Just currentName) = first currentFn
+    | _ => failAt fc "Deriving requires a function declaration, not a top level"
+
+  when (not !(checkAccessToDefinition fullName currentName)) $
```
gallais:
```suggestion
  unless !(checkAccessToDefinition fullName currentName) $
```

libs/base/Deriving/Common.idr
```diff
@@ -36,49 +38,199 @@ isFreeOf x ty
...
+    failAt fc "Make sure \{show fullName} has public export visibility"
+
+  normalisedTypeName <- normaliseTypeFunction fc fullName
+
+  when (not !(checkAccessToDefinition normalisedTypeName fullName)) $
```
gallais:
```suggestion
  unless !(checkAccessToDefinition normalisedTypeName fullName) $
```

libs/base/Deriving/Common.idr
```diff
@@ -36,49 +38,199 @@ isFreeOf x ty
...
+    failAt fc "Make sure \{show normalisedTypeName} has public export visibility"
+
+  getDCons fc normalisedTypeName
+  where
+    first : SnocList Name -> Maybe Name
+    first [<]       = Nothing
+    first [<x]      = Just x
+    first (xs :< _) = first xs
```
gallais: I think first/last is ambiguous once we start considering lists & snoclists

```suggestion
    leftMost : SnocList Name -> Maybe Name
    leftMost [<]       = Nothing
    leftMost [<x]      = Just x
    leftMost (xs :< _) = first xs
```
Also, this should go in a library.

libs/base/Deriving/Common.idr
```diff
@@ -36,49 +38,199 @@ isFreeOf x ty
...
+
+    getDCons fc ty
+
+mutual
+  toTypeParameter : Elaboration m => TTImp -> m TypeParameter
+  toTypeParameter (IType fc) = pure MkTPIType
+  toTypeParameter (IPrimVal fc c) = pure (MkTPPrim c)
+  -- Unqualified: that's a local variable
+  toTypeParameter (IVar fc nm@(UN (Basic _))) = pure (MkTPLocal nm)
+  toTypeParameter arg = MkTPType <$> isFamily arg
```
gallais: Why do we insist that the parameter should be a family?

libs/base/Deriving/Common.idr
GulinSS: I'll move to lib and update the call

libs/base/Deriving/Common.idr
GulinSS: I'm not sure I fully understand your question, probably due to my lack of theoretical background. 😇 Could you explain it in more detail? That would help me attempt an answer.

gallais: If the type parameter is an arbitrary expression, is that problematic?
E.g. in `Vec 4`, `4` is not a type family, it's a natural number but `Vect 4`
is still a `Functor`.

GulinSS: > If the type parameter is an arbitrary expression, is that problematic?

This is indeed a bug. We had a test with `Vect n`, but it didn't catch the issue seems because `n` there is a type parameter. A test with `Vect 4` would have caught it (I've verified it locally), but it is actually missing. I'll add this case to the test suite and provide a fix.

I am still having doubts regarding good naming here:

```idris
  toTypeParameter arg = MkTPType <$> isFamily arg
```

It is actually like something `isFamilyOrType` 🙂

Thanks for your patience!

----

It seems `Vect (k + m)` is a similar case of this issue.

GulinSS:
@gallais I'm still not sure whether we should leave a type parameter truly arbitrary, but I've made it more flexible than our `IsFamily`. I think keeping it at least somewhat shaped is better than leaving it as raw `TTImp`.

However, your `Vect 4` example raised another round of questions. A Haskell-like deriving algorithm doesn't expect dependent parameters like `S n` in `Vect (S n) ...` . So there's still room for improvement. May be for another turn? 😊 (See the `TODO: Not Yet Implemented` comments in the tests).

GulinSS:
I've found a case which ruins my PR :pensive:, mark it as a draft till I will find out how to resolve that.

libs/base/Deriving/Common.idr
```diff
@@ -36,48 +38,203 @@ isFreeOf x ty
...
+
+export
+DecEq DataConName where
+  decEq (MkDataConName n1) (MkDataConName n2) = decEqCong $ decEq n1 n2
+
+export
+Show DataConName where
+  show (MkDataConName n) = "MkDataConName \{show n}"
+
+mutual
```
buzden: Seems that no `mutual` is needed here

GulinSS:
Dear reviewers! The summary has been updated. Commits were recombined to make it easier to follow along and show improvements step-by-step. The first commit sets up tests which will transition to working state as commits progress.

Glad to get more of your 👀 (and hope for ❤️ )

libs/base/Deriving/Common.idr
```diff
@@ -36,10 +36,24 @@ isFreeOf x ty
...
-  constructor MkIsType
+data TypeParameter
+  = MkTPLocal Name
+  | MkTPPrim Constant
+  | MkTPApp (Name, SnocList (Argument TypeParameter))
+  | MkTPIType
+
+export
+Show TypeParameter where
+  show (MkTPLocal a) = "MkTPLocal \{show a}"
```
gallais: This should probably be using `showCon`

libs/base/Deriving/Common.idr
```diff
@@ -48,38 +62,105 @@ wording Func = "a function name"
 wording (DataCon tag arity) = "a data constructor"
 wording (TyCon tag arity) = "a type constructor"
 
-isTypeCon : Elaboration m => FC -> Name -> m (List (Name, TTImp))
-isTypeCon fc ty = do
+checkAccessToDefinition : Elaboration m => (given, candidate : Name) -> m Bool
+checkAccessToDefinition g c =
+  pure $ isParentOf (getNS g) (getNS c) || !(isPublic g)
+  where
+    getNS : Name -> Namespace
```
gallais: This (and other definitions below) should probably be moved `Language.Reflection.TT`

gallais (commented):
That looks great; sorry it took so long to get back to it.
I have suggested minor changes and I am happy to merge after that.

gallais:
Ugh, I did not have a working installation of Idris handy so, of course, my patch is broken.

PR #3740, opened 2026-02-16
[ fix ] Backticked names cannot be used as prefix operators

### LINKED ISSUE #3550: Encountered `INTERNAL ERROR: Invalid input to shunting` when using infix `Elem` and `.` together

# Steps to Reproduce

```idris
CharacterExcept : Type
CharacterExcept = `Elem` _ . fst
```

Note that this code is not expected to type check, but it produces an internal compiler error.

# Expected Behavior

Idris reports type errors and undefined names instead.

# Observed Behavior

The following error message is printed:

```
Error: INTERNAL ERROR: Invalid input to shunting 
```

spcfox:
# Description

Fixes #3550

## Self-check


tests/idris2/error/perror007/expected
```diff
@@ -15,7 +15,7 @@ StrError2:2:19--2:20
  1 | foo : String
  2 | foo = "nested: \{ " \{ 1 +  } " }"
                        ^
-... (31 others)
+... (30 others)
```
spcfox: Is it important to check this number in tests? Perhaps we should sanitise it in `clean_names`?

gallais:
Is there any chance we could have a more helpful message suggesting the use of
a section? The following is valid:

```
(`minus` 5)
```

spcfox:
> Is there any chance we could have a more helpful message suggesting the use of a section? The following is valid:
> 
> ```
> (`minus` 5)
> ```

Not sure if this error can be easily added during parsing. I couldn't get it to work.

spcfox:
@andrevidela, could we add this kind of error to the parser? Perhaps we shouldn't modify the parser itself and instead throw the error during desugaring?

I think it’s better to make it a parsing error, but I couldn't figure out how to add an informative error message.

andrevidela:
In my experience, it's easier to catch this kind of error during desugaring, we have more contextual information than during parsing and it keeps the parser simple.

 It also keeps the parse tree semantically simple, if we catch the backtick misuse at parsing time, we still produce a value of type `OpStr` which could be backticked at desugaring time, but since the error would have already been thrown by then it would be an impossible value to have. This would force us to handle a case that we know cannot happen. I'd rather have this handling code in the desugarer such that we're not saddled with a value of `OpStr` that doesn't faithfully represent the space of possible values accurately.

spcfox:
Ok, now the error is handled during desugaring

tests/idris2/error/perror007/expected
dunhamsteve: I don't find the "xx others" particularly useful. Trying to surface the proper error with `mustWork` and `commit` is a little awkward and fragile. I had an idea that if we're taking one error out of 31, we should take the one that starts (FC) latest in the file. The idea is that the branch that made the most progress would have the best error. But when I tried that a while ago, I found that the error that I wanted was often already discarded and not in the pile of errors (you can view them all with `--alt-error-count 50`).  

In this case there is a `mustWork $ expr ...`, but it's successfully parsing `1` as an expression. This leaves the `+` token, which is not a `}`, and then error for that is later dropped.  PR #3741 moves the `interpEnd` inside the `mustWork` to get a better error message:
```
1/1: Building StrError2 (StrError2.idr)
Error: Expected string interp end.

StrError2:2:26--2:27
 1 | foo : String
 2 | foo = "nested: \{ " \{ 1 +  } " }"
                              ^
```

spcfox: Thanks. As it turns out, this PR doesn't change the expected files, so I think it's better to leave it for a separate PR.

tests/idris2/error/error035/expected
```diff
@@ -0,0 +1,22 @@
+1/1: Building Issue3550 (Issue3550.idr)
+Error: Cannot use backticked name as a prefix operator
+
+Issue3550:2:19--2:25
+ 1 | CharacterExcept : Type
+ 2 | CharacterExcept = `Elem` _ . fst
+                       ^^^^^^
+
+Possible solutions:
+ - Use section: (`Elem` _ . fst)
```
gallais, line 10: Should the section not be
```
(`Elem` _) . fst
```
?

tests/idris2/error/error035/expected
```diff
@@ -0,0 +1,22 @@
...
+ - Use application: Elem _ . fst
```
gallais, line 11: This application has a different meaning as `_` is now for the first argument rather than the second one.

spcfox: We don’t know what the user intended. In the first case, I'm suggesting the most likely option — a section. But it’s also possible that the backticks were just a typo.

tests/idris2/error/error035/expected
spcfox: Should we be guessing the parenthesis placement here? Both cases could potentially be correct

spcfox: The internal error occurred in the shunting algorithm, which reorders operators according to their precedence. That’s why I added this check before it. We could potentially run it, but I’m not sure it would be helpful

tests/idris2/error/error035/expected
andrevidela: yeah we can't run the typechecker at this stage to know which one is the one that the user meant. I think this is fine as the user will probably try both and pick the one that works

PR #3741, opened 2026-02-18
[ error ] improve errors in string interpolation

dunhamsteve:
# Description

Some a parse error in a string interpolation is lost and a generic "Expected 'case', 'if', 'do', application or operator expression" is reported at an earlier location.

We already have a `mustWork` on the `expr` inside the interpolation, but in cases where there are extra tokens after the expression, the error is lost. One case where this happens is for a trailing operator, which occurs in the perror007 tests.

Currently we see:
```
1/1: Building StrError2 (StrError2.idr)
Error: Couldn't parse any alternatives:
1: Expected 'case', 'if', 'do', application or operator expression.

StrError2:2:19--2:20
 1 | foo : String
 2 | foo = "nested: \{ " \{ 1 +  } " }"
                       ^
... (31 others)
```
And with this patch, we get:
```
1/1: Building StrError2 (StrError2.idr)
Error: Expected string interp end.

StrError2:2:26--2:27
 1 | foo : String
 2 | foo = "nested: \{ " \{ 1 +  } " }"
                              ^
```

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -59,7 +59,7 @@ should target this file (`CHANGELOG_NEXT`).
   - `Libraries.Data.List.Quantifiers.Extra.tabulate`.
   - `Libraries.Utils.Binary.nonEmptyRev`
   - `Libraries.Utils.String.dotSep`
-* Fixes an issue when unifying labmda terms with implicits (#3670)
+* Fixes an issue when unifying lambda terms with implicits (#3670)
 * The "With clause does not match parent" error now points to the correct location
 * The compiler now warns the user when `impossible` clauses are ignored. This
   typically happens when a numeric literal or an ambiguous name appears in an
@@ -68,6 +68,7 @@ should target this file (`CHANGELOG_NEXT`).
 * Fixed coverage checker issues (#1800, #1998, #2318, #2822, #3679).
 * Fixed totality checking in namespace and mutual blocks (#2868, #3692).
 * Fixed incorrect argument multiplicity when using an as-pattern (#3687).
+* Better messages for errors inside string interpolation.
 
 ### Building/Packaging changes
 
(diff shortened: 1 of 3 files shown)


spcfox:
I would expect the error to be similar to this case:
```idris
foo = 1 +
```
Shouldn't `mustWork (expr …)` act exactly like this? Why does the FC point to the quote?

UPD: Ah, it's because `expr` consumes the `1` and then looks for `}`. Yeah, we should put them both inside `mustWork` then.

dunhamsteve:
The `foo = 1 +` case also has a poor error message. It is reading the 1, ending the `expr` and then trying to say "I thought the block was done, but I see a `+`".  

I don't think we can push this one down to desugaring because `POp` expects a right hand side. We could put an error in if there is an outdent after the operator:
```diff
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -371,14 +371,16 @@ mutual
              <|>
              (do b <- bounds $ do
                         continue indents
-                        op <- fcBounds iOperator
+                        op <- bounds iOperator
                         e <- case op.val of
                                OpSymbols (UN (Basic "$")) => typeExpr q fname indents
-                               _ => expr q fname indents
+                               _ => do
+                                continue indents <|> fatalLoc op.bounds {c=False} "trailing operator"
+                                expr q fname indents
                         pure (op, e)
                  (op, r) <- pure b.val
                  let fc = boundToFC fname (mergeBounds l b)
-                 pure (POp fc (map NoBinder l.withFC) op r))
+                 pure (POp fc (map NoBinder l.withFC) op.withFC r))
                <|> pure l.val
 
   opExpr : ParseOpts -> OriginDesc -> IndentInfo -> Rule PTerm

```

Edit: I used the bounds of `op` instead of `mustContinue` because the pointer for the error was ending up at the beginning of the next line.

PR #3744, opened 2026-03-08
[ log ] Log timing for elaboration scripts

spcfox:
# Description

Elaborator scripts can sometimes become quite complex and time-consuming to execute. To improve debuggability and help identify performance bottlenecks, this PR adds `logTime` instrumentation to capture the execution duration of these scripts.

## Self-check


andrevidela:
chilling on a Sunday and I found this gem. Thanks!

PR #3746, opened 2026-03-09
Fix pretty pkg fields

tzemanovic:
# Description

When I used `pack new` it got my name from git, but it inserted it into ipkg file with the non-ascii characters escaped. Pack uses pretty printing provided by idris, which uses `show` over string fields (similar to the issue fixed by #3699).

## Self-check

### DIFF
--- a/.github/workflows/ci-idris2-and-libs.yml
+++ b/.github/workflows/ci-idris2-and-libs.yml
@@ -500,8 +500,11 @@ jobs:
       - name: Build API
         run: make install-api
         shell: bash
-      - name: Test API
+      - name: Test API 001
         run: cd tests/idris2/api/api001 && ./run "$HOME/.idris2/bin/idris2"
+        shell: bash
+      - name: Test API 002
+        run: cd tests/idris2/api/api002 && ./run "$HOME/.idris2/bin/idris2"
         shell: bash
 
   ######################################################################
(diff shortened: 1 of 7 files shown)


.github/workflows/ci-idris2-and-libs.yml
```diff
@@ -500,9 +500,12 @@ jobs:
       - name: Build API
         run: make install-api
         shell: bash
-      - name: Test API
+      - name: Test API 001
         run: cd tests/idris2/api/api001 && ./run "$HOME/.idris2/bin/idris2"
         shell: bash
+      - name: Test API 002
+        run: cd tests/idris2/api/api002 && ./run "$HOME/.idris2/bin/idris2"
+        shell: bash
```
tzemanovic, line 508: this is kinda meh - it would be nice to re-use the golden tests setup, but not sure if there's maybe a simpler way to test this with dep on idris2 api?

CodingCellist: > it would be nice to re-use the golden tests setup, but not sure if there's maybe a simpler way to test this with dep on idris2 api?

Based on the [README](https://github.com/idris-lang/Idris2/blob/main/tests/idris2/api/README) in the `tests/idris2/api` directory, it doesn't seem like it.  As you say, it's annoying, but it is what it is.

CodingCellist (approved):

PR #3751, opened 2026-03-14
Fix RefC negation function headers

joelberkeley:
# Description

I've not thought this through deeply, but it seems extremely likely esp. given this error
```
build/exec/tinker.c: In function ‘Main_main’:
build/exec/tinker.c:201:24: error: implicit declaration of function ‘idris2_negate_Double’; did you mean ‘idris2_nagate_Double’? [-Werror=implicit-function-declaration]
  201 |     Value *primVar_4 = idris2_negate_Double(var_1);
      |                        ^~~~~~~~~~~~~~~~~~~~
      |                        idris2_nagate_Double
build/exec/tinker.c:201:24: error: initialization of ‘Value *’ from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
cc1: all warnings being treated as errors
```
for `main = printLn (-1.0)`

## Self-check


buzden:
Since it wasn't revealed for so long, we should definitely have a test for this, then

joelberkeley:
I don't have the energy for that (I'm only using refc to try to debug another issue and don't want to go down a rabbit hole here). You're welcome to close this PR.

BTW I anticipate there are other bugs in refc, given I've already found three in a week and my code still isn't compiling.

dunhamsteve:
Thanks Joel for sharing your fixes.

I added a case to the refc/doubles test. It fails prior to your fix and passes afterwards.

Issue #3753, opened 2026-03-17
Proposal: Add `getFC` to elaborator reflection for call-site source location


bio-aeon:
- [x] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [x] I have checked that there is no existing PR/issue about my proposal.

## Summary

Add a `GetFC : Elab FC` constructor to the elaborator reflection API, exposing the call-site source location to user-level macros. This restores functionality that Idris 1 had (`getSourceLocation : Elab SourceLocation`) but was dropped in Idris 2.

## Motivation

There's currently no way for an Idris 2 macro to get the source location of its call site. The compiler knows it - `elabScript` receives the `fc` parameter, but it's not exposed to user code through the `Elab` monad.

This makes it hard to write things like test assertion macros, logging utilities or debugging helpers that need to report *where* they were called from.

The possible workaround today is accepting a dummy `TTImp` argument and pulling the FC off it:

```idris
%macro
here : TTImp -> Elab SrcLoc
here t = pure (fcToSrcLoc (getFC t))

-- at the call site, you have to write:
loc = here `(())
```

This works, but it's awkward:
- Every call site needs a meaningless `` `(()) `` argument
- It can't appear as the last expression in a do-block (macro argument splitting issue)
- Users unfamiliar with quasiquoting find it confusing

With `getFC`, this becomes a clean zero-argument macro:

```idris
%macro
here : Elab SrcLoc
here = do
  fc <- getFC
  pure (fcToSrcLoc fc)

-- clean call site:
loc = here
```

### Examples

A test assertion macro that reports failure location:

```idris
%macro
assert : Bool -> Elab ()
assert b = do
  fc <- getFC
  let loc = fcToSrcLoc fc
  check `(when (not ~(quote b))
            (putStrLn ("Assertion failed at " ++ show ~(quote loc))))
```

A logging helper:

```idris
%macro
logHere : String -> Elab (IO ())
logHere msg = do
  fc <- getFC
  let loc = fcToSrcLoc fc
  check `(putStrLn (~(quote (show loc)) ++ ": " ++ ~(quote msg)))
```

### Technical implementation

**In `libs/base/Language/Reflection.idr`:**
- Add `GetFC : Elab FC` constructor to `data Elab`
- Add `getFC : m FC` to the `Elaboration` interface
- Wire up the `Elab` instance and `MonadTrans` lift

[...]

PR #3762, opened 2026-04-02
[ elab ] Add `getFC` for call-site source location

bio-aeon:
# Description

This PR adds `GetFC : Elab FC` to the elaborator reflection API so macros can get the source location of their call site. Discussed in #3753.

The `fc` is already in scope in `elabScript` - this change just exposes it via `scriptRet`.

## Self-check


stefan-hoeck:
If I understand correctly what this does, it would certainly be nice to have. Would be great, if you could add an example (as a comment to this issue as well as a test case to show that this works as expected) how this is supposed to be used.

stefan-hoeck:
Never mind my comment above. Just saw the corresponding issue. There still should be a test for this.

bio-aeon:
@stefan-hoeck Thanks for the feedback!
I've added a test in `reflection036` (a `%macro` where `getFC` is being called and we get the call-site line and column).

gallais (approved):

PR #3729, opened 2026-02-07
[ ci ] Update `ci-bootstrap.yml`

spcfox:
# Description

[ci-bootstrap](https://github.com/idris-lang/Idris2/blob/main/.github/workflows/ci-bootstrap.yml) currently runs on the deprecated `ubuntu-20.04` runner, which [was retired some time ago](https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/).

The original purpose of this job was to ensure we could build with older Chez Scheme versions from before support for `__collect_safe`. That check is no longer relevant: `__collect_safe` landed in Chez Scheme 9.5.1, and the same release introduced `collect-rendezvous` ([Chez release notes](https://cisco.github.io/ChezScheme/release_notes/v10.3.0/release_notes.html), 2.81 and 2.82), which Idris 2 now relies on.
https://github.com/idris-lang/Idris2/blob/b714fcaeaa7a91fae36ccee31bde250fa4858150/src/Compiler/Scheme/Chez.idr#L109

This PR updates the workflow to run on a supported runner and repurposes it to test bootstrapping against the minimum supported Chez Scheme version.

PR #3765, opened 2026-04-05
Prefix RefC internal names to disambiguate

joelberkeley:
# Description

`Value` is sufficiently vague that it can collide with names from other support libs, such as `Value` in my MLIR lib. I've prefixed everything with `Idris2_`.

c.f. zulip discussion [#help > Rename &#96;Value&#96; in RefC?](https://idris-lang.zulipchat.com/#narrow/channel/556789-help/topic/Rename.20.60Value.60.20in.20RefC.3F/with/583615015)

## Self-check

### DIFF
--- a/support/refc/_datatypes.h
+++ b/support/refc/_datatypes.h
@@ -44,10 +44,10 @@ typedef struct {
 
 typedef struct {
   Value_header header;
-  // `Value` is an "abstract" struct,
+  // `Idris2_Value` is an "abstract" struct,
   // `Value_Xxx` structs have the same header
   // followed by type-specific payload.
-} Value;
+} Idris2_Value;
 
 /*
 We expect at least 4 bytes for `Value_header` alignment, to use bit0 and bit1 of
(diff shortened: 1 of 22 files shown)


bootstrap/idris2_app/idris2.rkt
joelberkeley, line 1: should I have updated this and other autogenerated files?

support/refc/_datatypes.h
```diff
@@ -38,42 +38,42 @@ typedef struct {
   uint16_t refCounter;
   uint8_t tag;
   uint8_t reserved;
-} Value_header;
+} Idris2_header;
```
joelberkeley, line 41: not sure about this rename

joelberkeley:
Do I need to do anything else here? Looks like CI timed out

dunhamsteve:
Should we be updating the bootstrap image? I'm not sure what the policy is here, I think at one point we liked to have it be the previous version of Idris.

It looks like bootstrap-ci is wedging waiting for an ubuntu-20 runner, probably because [they're deprecated](https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/). The like says they'd be down on the 8th to "raise awareness".

```
Evaluating initialise.if
Evaluating: success()
Result: true
Requested labels: ubuntu-20.04
Job defined at: idris-lang/Idris2/.github/workflows/ci-bootstrap.yml@refs/pull/3765/merge
Waiting for a runner to pick up this job...
```
If updating bootstrap is ok, we probably need to move bootstrap-ci.yml to a newer ubuntu.

joelberkeley:
> Should we be updating the bootstrap image?

I have absolutely no idea. I don't know how it works (I thought the bootstrap image would be built from whatever changes I make elsewhere), in which case the answer would be "no, we shouldn't update the bootstrap image". But I have extremely low confidence in my knowledge here.

I only updated it because it appeared in the find/replace.

dunhamsteve:
Yeah, the bootstrap files are usually only updated for releases. I wouldn't update it unless something is broken and we're forced to update it.  I believe it is only used to build a scheme target (either chez or racket), so it should be fine as-is with these changes, and we should revert those two files. That should fix the build.

Aside from that does anyone have objections to this change?

joelberkeley:
Reverted bootstrap

dunhamsteve:
Looks good to me, and I don't see any objections.

joelberkeley:
p.s. bootstrap version handled in this PR https://github.com/idris-lang/Idris2/pull/3729

PR #3769, opened 2026-05-01
[ide-mode] Make `(:interpret ":version" and ":ops")` return output as string instead of S-expression.

keram:
# Description

Previously, the IDE commands  returned a structured S-expression 
which wasn't handled well by `idris-mode` as it expects for `interpret` always return a string.

See: https://github.com/idris-hackers/idris-mode/issues/661 
Although we have "hotfix" in `ide-mode` https://github.com/idris-hackers/idris-mode/pull/662
 I think better to make the change in Idris itself.

This PR supersede https://github.com/idris-lang/Idris2/pull/3768
Relates also to https://idris-lang.zulipchat.com/#narrow/channel/556701-general/topic/LLMs.20and.20Idris.202.20programming/near/590513106

### Before:

```
idris2 --ide-mode
((:interpret ":opts") 3)
00009b(:return (:ok ((:show-implicits :False) (:show-machinenames :False) (:show-namespace :False) (:show-types :False) (:eval "normalise") (:editor "vim"))) 3)
((:get-options) 3)
00009b(:return (:ok ((:show-implicits :False) (:show-machinenames :False) (:show-namespace :False) (:show-types :False) (:eval "normalise") (:editor "vim"))) 3)
```

In Emacs

before-fix-version-opts

### After:

```
$ idris2 --ide-mode
((:interpret ":opts") 3)
00008d(:return (:ok "showimplicits = False
showmachinenames = False
shownamespace = False
showtypes = False
eval = normalise
editor = \"vim\"") 3)
((:get-options) 3)
00009b(:return (:ok ((:show-implicits :False) (:show-machinenames :False) (:show-namespace :False) (:show-types :False) (:eval "normalise") (:editor "vim"))) 3)
```

Emacs:

opts-idris-repl-emacs-after

## Self-check

- [X]  I've carefully read [`CONTRIBUTING.md`](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [X] I confirm that this contribution did not involve GenerativeAI nor Large Language Models.

(used LLM to review and correct solution in commit https://github.com/idris-lang/Idris2/pull/3769/changes/1163504b169e) Otherwise human, written contribution.

### DIFF
--- a/src/Idris/IDEMode/REPL.idr
+++ b/src/Idris/IDEMode/REPL.idr
@@ -127,6 +127,7 @@ data IDEResult
   | Term String   -- should be a PTerm + metadata, or SExp.
   | TTTerm String -- should be a TT Term + metadata, or perhaps SExp
   | NameLocList (List (Name, FC))
+  | VersionSExp Version
 
 replWrap : Core REPLResult -> Core IDEResult
 replWrap m = pure $ REPL !m
@@ -233,7 +234,7 @@ process (EnableSyntax b)
     = do setSynHighlightOn b
          pure $ REPL $ Printed (reflow "Syntax highlight option changed to" <++> byShow b)
 process Version
-    = replWrap $ Idris.REPL.process ShowVersion
+    = pure $ VersionSExp version
 process (Metavariables _)
     = FoundHoles <$> getUserHolesData
 process GetOptions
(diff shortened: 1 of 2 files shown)


src/Idris/IDEMode/REPL.idr
```diff
@@ -127,6 +127,23 @@ data IDEResult
   | Term String   -- should be a PTerm + metadata, or SExp.
   | TTTerm String -- should be a TT Term + metadata, or perhaps SExp
   | NameLocList (List (Name, FC))
+  | VersionSExp Version
+  | OptionsSExp (List REPLOpt)
+
+getOptions : {auto c : Ref Ctxt Defs} ->
```
keram: This is copy-paste from `Idris/REPL.idr` https://github.com/idris-lang/Idris2/blob/main/src/Idris/REPL.idr#L191 as when trying reusing it here got an error.
```
Error: While processing right hand side of process. 
Name Idris.REPL.getOptions is private.
...
Suggestion: add an explicit export or public export modifier. 
````
I don't know which is the right suggested solution or some other better way to get the opts.

keram: Added `export` to Idris/REPL.idr version and removed the copy from here.

PR #3773, opened 2026-05-11
Check the type of typed lambdas

### LINKED ISSUE #3771: typed lambda in type constructor using pair syntax sugar creates Void

# Steps to Reproduce

```
record Test where
  constructor K
  a : Pair Type Type
  b : fst a -> Void

t : Test
t = K (Nat, Bool) (\ n : Void => n)

TVoid : Void
TVoid = t.b 3
```

# Expected Behavior

```
Diagnostics:
While processing right hand side of t. When unifying:
    Void
and:
    fst (Nat, Bool)
Mismatch between: Void and Nat.

Y.idr:8:32: Error while unification
Y.idr:5:7: Error while unification
```

# Observed Behavior

Typechecks fine

Note that this doesn't happen if you use `MkPair` instead of `(,)`

andrevidela:
# Description

fixes #3771

## Self-check

- [x] I confirm that this contribution did not involve Generative AI nor Large Language Models.

### DIFF
--- a/src/TTImp/Elab/Binders.idr
+++ b/src/TTImp/Elab/Binders.idr
@@ -12,6 +12,8 @@ import TTImp.Elab.Check
 import TTImp.Elab.Delayed
 import TTImp.TTImp
 
+import Libraries.Data.IntMap
+
 %default covering
 
 -- Drop the name from the nested function declarations. We do this when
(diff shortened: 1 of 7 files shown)


src/TTImp/Elab/Binders.idr
```diff
@@ -190,8 +190,14 @@ checkLambda rig_in elabinfo nest env fc rigl info n argTy scope (Just expty_in)
                         throw (CantConvert fc (gamma defs) env
                                   (Bind fc n (Pi fc' rigb info' tyv) !(getTerm scopet))
                                   (Bind fc bn (Pi fc' c info' pty) psc))
-                    pure (Bind fc n (Lam fc' rigb info' tyv) scopev,
-                          gnf env (Bind fc n (Pi fc' rigb info' tyv) !(getTerm scopet)))
+                    let lamTerm = Bind fc n (Lam fc' rigb info' tyv) scopev
+                    let lamType = gnf env (Bind fc n (Pi fc' rigb info' tyv) !(getTerm scopet))
+                    let lambdaConstraints = constraints lambdaUnifyResult
```
gallais, line 209: I suppose some of these constraints may have been solved by elaborating the body of the lambda.
So it may be worth pruning them before checking if the the constraints are `null`.

andrevidela: Should be fixed

gallais (approved):

PR #3779, opened 2026-05-19
Extension to elaborator reflection interface to enable automated derivation of decidable equality 

bhaktishh:
# Description

This PR adds a function to the elaborator reflection interface, to enable automated derivation of decidable equality (in library [idris2-elab-util](https://github.com/stefan-hoeck/idris2-elab-util)). 

The function `getDecEqConPairs` expects a type, and returns a pair `(List Nat, List (String, String))`, which represent the indices of the type arguments that are parameters, and pairs of constructor names (as strings) that can have the same type.

For example, for the type `Vect`, the function would return `([1], [("Nil", "Nil"), ("::", "::")])`. 

## Self-check


gallais:
As I said, my main issue with this PR is that it is very ad-hoc for this specific use-case.
However, Bhakti did claim it would help derive equality for not just ADTs but *families*
so given the proper test cases (see e.g. [DeriveFunctor.idr](https://github.com/idris-lang/Idris2/blob/main/tests/base/deriving_functor/DeriveFunctor.idr) for an extensive test suite
documenting both where it works and where it fails), I'd be OK merging it in.

I still think there is a better way (via [a view isolating the diagonal](https://github.com/gallais/potpourri/blob/main/agda/poc/LinearDec.agda)) but if this lands
features today it's better than an ideal solution in X years.

Edit: FTR and to my own surprise, I was not actually unsub'd from this PR.

PR #3794, opened 2026-06-07
[ new ] add support for setting a datadir for a particular package in its ipkg file

mattpolzin:
# Description

This PR adds support for specifying one (additional) data directory for a package in its `ipkg` file if desired.

Idris2 supports the concept of "data directories" which can be used in different ways by different backends to augment Idris2 code during compilation or installation. Data directories can be set via ENV var currently but this mode of adding data directories is not well suited to projects that introduce a package meant to be consumed by other projects and need to set a new data directory up as well -- such packages should not need to rely on custom Makefile invocations to set an ENV var prior to compilation; after all, the package itself _always_ needs that data directory defined, not just sometimes, and that data directory is likely part of the project and so its relative path is always known and static.

To put a concrete example out there, a number of my projects provide `support/js` directories that get used via NodeJS FFI calls (much like the compiler project does) so I would benefit from being able to specify `datadir = "support"` for those projects.

It becomes particularly important to be able to set the data directory for a package when that package is consumed via a package manager like `pack` -- Such a package manager may manage ENV vars itself and regardless it may not use a Makefile provided by the package to build the package so there is no reliable way to set a data directory up as of today.

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -86,6 +86,8 @@ should target this file (`CHANGELOG_NEXT`).
 * Change `flake.nix` to point at `idris-community/idris2-mode` as the URL for
   `inputs.idris-emacs-src` (from the user fork `redfish64/idris2-mode`).
 * Fix UTF-8 character handling in package description fields.
+* A project's data directory can now be specified with the `datadir` entry in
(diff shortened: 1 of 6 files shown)


andrevidela (approved):
Looks good!

PR #3795, opened 2026-06-08
[Zurihack] Small Driver refactor

andrevidela:
# Description

This change is addressing technical debt related to the architecture of the driver. Those improvements
are driven by [Containers as APIs](https://arxiv.org/abs/2407.16713) as an experiment carried during Zurihack. While this PR itself does not introduce any containers or dependent lenses, the nature of changes were made evident by porting the driver to an API-first paradigm and refactoring the interaction model over there, and then backporting the changes to the main version.

## Changes
- `postOptions` was reading command line opts long after it became irrelevant. Instead we introduce a `PostSession` record that stores all the options that are relevant after a compiler session. To avoid polluting the global state (REPLOpts) it is stored in its own variable. It is also now written sequentially instead of recusively to better see the control flow of the application. There is more to do in that area in the future, but this is a good first step.
- The compiler has three different entry points depending on the flags `--ttm` and `--yaffle` (and nothing). The previous version handled this via a side effect, here we declaratively generate a value of `Entrypoint` carrying all the necessary data to run the matching entry point
- The imperative implementation of entrypoints would allow both `--ttm` and `--yaffle` flags and silently ignore the first one if the second was present. Because choosing the entry point is an exclusive choice, we now detect those two flags as being incompatible if they are present at the same time.
- Some boolean flags were used for control flow, sometimes `True` would stand for "continue" and sometimes for "abort", instead we now have `ControlFlow` type with two inhabitants `Continue` and `Abort` which clearly state intent.

There is a lot more opportunities to improve but there is a balance to strike between a small reviewable PR and one with too many unjustified changes. I think those changes are beneficial to the overall architecture without making large sweeping changes that are hard to review.

## Self-check

### DIFF
--- a/src/Idris/Driver.idr
+++ b/src/Idris/Driver.idr
@@ -38,7 +38,6 @@ public export
 CustomBackends : Type
 CustomBackends = List (String, Codegen)
 
-
 findInputs : List CLOpt -> Maybe (List1 String)
 findInputs [] = Nothing
 findInputs (InputFile f :: fs) =
@@ -70,7 +69,7 @@ updateEnv
          cg <- coreLift $ idrisGetEnv "IDRIS2_CG"
          whenJust cg $ \ e => case getCG (options defs) e of
            Just cg => setCG cg
-           Nothing => throw (UserError ("Unknown code generator " ++ show e))
+           Nothing => throw (InternalError ("Unknown code generator " ++ show e))
          inccgs <- coreLift $ idrisGetEnv "IDRIS2_INC_CGS"
          whenJust inccgs $ \ cgs =>
            traverseList1_ (setIncrementalCG False) $
--- a/src/Idris/Driver.idr
+++ b/src/Idris/Driver.idr
@@ -34,6 +34,11 @@ import Yaffle.Main
 
 %default covering
 
+public export
+CustomBackends : Type
+CustomBackends = List (String, Codegen)
+
+
 findInputs : List CLOpt -> Maybe (List1 String)
 findInputs [] = Nothing
 findInputs (InputFile f :: fs) =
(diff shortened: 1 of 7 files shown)


andrevidela:
One hint for future improvements is that every use of `ControlFlow` is a indicator that we could be splitting apart the code into two choices and that could be represented with a rich datatype rather than an early-return in the middle of imperative code

mattpolzin:
After a first pass, I commend you for picking such a pragmatic size of change. It really is very easy to review. All looks reasonable and beneficial to me.

andrevidela:
Thanks! I also briefly looked at your PR and was quite happy to see that it was on a part of the codebase that I just combed through. I'll take a proper look once I make the tests here pass

src/Idris/Driver.idr
```diff
@@ -95,24 +100,35 @@ updateREPLOpts
     = do ed <- coreLift $ idrisGetEnv "EDITOR"
          whenJust ed $ \ e => update ROpts { editor := e }
 
-tryYaffle : List CLOpt -> Core Bool
-tryYaffle [] = pure False
-tryYaffle (Yaffle f :: _) = do yaffleMain f []
-                               pure True
-tryYaffle (c :: cs) = tryYaffle cs
+-- There are three ways to run the compiler
+-- Either run normall, or run in yaffle mode, or dump TTM
```
gallais:
```suggestion
-- Either run normal, or run in yaffle mode, or dump TTM
```

src/Idris/SetOptions.idr
```diff
@@ -441,10 +445,11 @@ preOptions (SetCG e :: opts)
             Just cg => do setCG cg
                           preOptions opts
             Nothing =>
-              do coreLift $ putStrLn "No such code generator"
-                 coreLift $ putStrLn $ "Code generators available: " ++
-                                 showSep ", " (map fst (availableCGs (options defs)))
-                 coreLift $ exitWith (ExitFailure 1)
+              do throw $ InternalError $ """
```
gallais: I don't think this should be an internal error; it's a user error to request
a backend that does not exist.

src/Idris/SetOptions.idr
```diff
@@ -476,9 +481,10 @@ preOptions (DebugElabCheck :: opts)
 preOptions (AltErrorCount c :: opts)
     = do setSession ({ logErrorCount := c } !getSession)
          preOptions opts
-preOptions (RunREPL _ :: opts)
+preOptions (RunREPL expr :: opts)
     = do setOutput (REPL VerbosityLvl.ErrorLvl)
-         setSession ({ nobanner := True } !getSession)
+         setSession ({ nobanner := True} !getSession)
+         update PostS {runRepl := Just expr}
```
gallais, line 487: Should these destructive updates check they're not overwriting prior changes
and raise an error if there are conflicting options (I realise this is a change in
semantics but we may as well be more principled if we are cleaning code up)?

andrevidela: Two things are happening here:
- should destructive update make sure that they don't override a previously set option?  should we have some way to avoid destructive updates once we've set the values we want?
- Should flags be checked for consistency at this stage?

The _new_ part of the code does not perform any destructive updates since the "PostSession" record is constructed once just before we modify it and to keep the same programming pattern as the rest of the code it is done imperatively by looping over every flag one by one. The PostSession value is then left untouched until it is read again. This is also the case for all the other flags, I don't think I found one that it written to twice.

This is reassuring because it removes the risk of misuse via overriding. Something we could do is have another type `ReadOnlyRef` that only has a read function with no `set` or `modify` and  convert our option type to `ReadOnlyRef` after we are done mutating it.

Doing this would also open the opportunity to split up the compiler's global state into smaller chunks, right now `REPLOpts` contains many unrelated settings that are not relevant to most of the operations of the compiler. 

Either way, this should only be a temporary measure over the real fix: a proper CLI parser that generates an option record in one go, reports errors, and informs the compiler's control flow declaratively, instead of relying on for-loops with early returns.

This is a big architectural changes and it will likely break shell scripts and CI since currently we allow all possible inconsistent uses of flags (except `--ttm` and `--yaffle` after this PR), and therefore we should do it conscientiously. 

My conclusion is that, for now, that's the best we can do: we're already splitting the state into smaller chunks, we're already providing clear semantics for our flags via a record with typed fields, we are already starting to recover from some mild flag misuse. 

We could do more, we could always do more, but if we were to do everything, this PR would be inrecognisable and would be a giant -5000+5000 rewrite of the driver. This is the goal eventually, but it's not the goal of this particular PR

andrevidela:
Everyone ok with merging this?

mattpolzin (approved):
I did not trace the possible paths through the old recursive postOptions to determine that no functionality is lost by the new implementation, but it looks reasonable to me and to the extent we don't have all the possible interactions with the CLI under test I guess at some point we just handle every case we can think of as being useful and see what happens :)

andrevidela:
I know for a fact that the new code is not equivalent but they are quite close. It's also the first step toward a more maintainable way to handle command line options across the compiler. 

The recursive implementation from before was really hard to reason about so them being different in that case is desirable.

PR #3799, opened 2026-06-16
[ fix ] update chez version in windows ci

dunhamsteve:
# Description

Windows CI is broken because it is using v9.5.8a which expects an older version of visual studio. It was pinned in #3040 because the dev branch occasionally failed to build.  I've updated it to the latest release, v10.4.1.

In a test build, this failed once in channels009 and then later worked, so we may need to revisit that test at some point in time.

### DIFF
--- a/.github/workflows/ci-idris2-and-libs.yml
+++ b/.github/workflows/ci-idris2-and-libs.yml
@@ -216,7 +216,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Get Chez Scheme
         run: |
-          git clone --depth 1 --branch v9.5.8a https://github.com/cisco/ChezScheme
+          git clone --depth 1 --branch v10.4.1 https://github.com/cisco/ChezScheme
           c:\msys64\usr\bin\bash -l -c "pacman -S --noconfirm tar make mingw-w64-x86_64-gcc"
           echo "PWD=$(c:\msys64\usr\bin\cygpath -u $(pwd))" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
       - name: Configure and Build Chez Scheme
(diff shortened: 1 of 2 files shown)


Matthew-Mosior:
@dunhamsteve 

I wonder if we could bump up the CI on all platforms, since #3704 alters the ci-idris2-and-libs GitHub workflow so that Chez Scheme v10.3.0 is explicitly built against (instead of using the Chez Scheme available via the package manager). Chez Scheme v10.3.0 provides the fx*/wraparound function that is utilized for Bits8/Bits16 multiplication in #3704.

dunhamsteve:
Which version would we want are are there any objections to updating chez on all of the platforms?  At least with those changes in place, we would be in control, and it will be easy to change the version.

Matthew-Mosior:
> Which version would we want are are there any objections to updating chez on all of the platforms? At least with those changes in place, we would be in control, and it will be easy to change the version.

No objections from me to updating chez to at least v10.3.0 (fine with anything after that) on all of the platforms.

dunhamsteve:
I've backed off to 9.6.4 for now. There seems to be an issue with the chez/channels009 test when using newer versions of chez on windows, consistent with `channelGetWithTimeout` not timing out properly, but I'm not sure that is what is happening.

buzden:
Is there something preventing merging this now? I think fixing CI failing on other PRs and the main branch is important

dunhamsteve:
I'm done with this for now. Chez 10 would be nice at some point, but that requires further investigation.

andrevidela:
the tests passed here but `main` fails again?What is going on

buzden:
Looks like `channel009` has become (more) flaky after this

dunhamsteve:
Yeah, I thought I had backed off to version 9.6.4, but it looks like that I didn't push that change to this branch. My test build with 9.6.4 on another branch succeeded, but it failed when I tried again just now. So we might be out of luck with channels009 on windows unless we can sort out what is going wrong. (Possibly that sleep doesn't work as expected.)

PR #3802, opened 2026-06-17
[ proposal ] fixed size C arrays a-la Struct

ear7h:
# Description

We have a blessed `Struct` type for passing around c structs using FFI, but I could not hack fixed size arrays. Idris seems to be okay with basic newtypes, but it complains for dependent types like:

```
data CArray : Nat -> Type -> Type where
```

I think what's happening is that the index types (the thing left of `:` right?) are being subject to the FFI restrictions of being basic types. I think a more radical fix here would be to drop the restrictions on types and only enforce them on constructors.

---

Some notes:
* the blessed type is called `CArray` in idris code to avoid clashing with the common `Array` type
* I'm using the 99 tag as placeholder

## Self-check

### DIFF
--- a/src/Compiler/Scheme/Chez.idr
+++ b/src/Compiler/Scheme/Chez.idr
@@ -193,9 +193,16 @@ cftySpec fc CFGCPtr = pure "void*"
 cftySpec fc CFBuffer = pure "u8*"
 cftySpec fc (CFFun s t) = pure "void*"
 cftySpec fc (CFIORes t) = cftySpec fc t
-cftySpec fc (CFStruct n t) = pure $ "(* " ++ fromString n ++ ")"
+cftySpec fc (CFArray n t) = do
+  inner <- cftySpec fc t
+  pure $ "(array n " ++ inner ++ ")"
+cftySpec fc (CFStruct n t) = pure $ fromString n
 cftySpec fc t = throw (GenericMsg fc ("Can't pass argument of type " ++ show t ++
                          " to foreign function"))
+
+cftySpecArg : FC -> CFType -> Core Builder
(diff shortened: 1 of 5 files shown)


src/Compiler/Scheme/Chez.idr
```diff
@@ -193,10 +193,17 @@ cftySpec fc CFGCPtr = pure "void*"
...
+cftySpec fc (CFArray n t) = do
+  inner <- cftySpec fc t
+  pure $ "(array n " ++ inner ++ ")"
+cftySpec fc (CFStruct n t) = pure $ fromString n
 cftySpec fc t = throw (GenericMsg fc ("Can't pass argument of type " ++ show t ++
                          " to foreign function"))
 
+cftySpecArg : FC -> CFType -> Core Builder
+cftySpecArg fc t@(CFStruct _ _) = (\x => "(& " ++ x ++ ")") <$> cftySpec fc t
+cftySpecArg fc t = cftySpec fc t
```
ear7h, line 205: This is actually a fix to an existing compilation bug, I should break it into a different PR.

`cftySpec` outputs the scheme form of the type within a struct, and `cftySpecArg` outputs the scheme form of the type within function arguments.

Relevant documentation:

> (* ftype-name): This type allows a pointer to a foreign type (ftype) to be passed. The argument must be an ftype pointer of the type identified by ftype-name, and the actual argument is the address encapsulated in the ftype pointer.
> ...
> (& ftype-name): This type allows a foreign type (ftype) to be passed as a value, but represented on the Scheme side as a pointer to the foreign-type data. That is, a (& ftype-name) argument is represented on the Scheme side the same as a (* ftype-name) argument, but a (& ftype-name) argument is passed to the foreign procedure as the content at the foreign pointer's address instead of as the address.
>
> https://cisco.github.io/ChezScheme/csug/foreign.html

PR #3825, opened 2026-08-07
docs: add documentation for `curry` and `uncurry` in `Prelude.Basics`

user470750:
# Description
Added documentation with examples for `curry` and `uncurry` in `Prelude.Basics`.

## Self-check

### DIFF
--- a/libs/prelude/Prelude/Basics.idr
+++ b/libs/prelude/Prelude/Basics.idr
@@ -82,12 +82,25 @@ public export %tcinline
 apply : (a -> b) -> a -> b
 apply f = \a => f a
 
+||| Converts a function on pairs to a curried function that takes arguments separately.
+||| The curried form allows partial application.
+||| @ f the function to curry
+|||
+||| ```idris example
+||| (curry swap) x y = (y, x)
+||| ```
 public export
-curry : ((a, b) -> c) -> a -> b -> c
+curry : (f : (a, b) -> c) -> a -> b -> c
 curry f a b = f (a, b)
 
+||| Converts a curried function that takes arguments separately to a function on pairs.
+||| @ f the function to uncurry
+|||
+||| ```idris example
+||| (uncurry min) (x, y) = min x y
+||| ```
 public export
-uncurry : (a -> b -> c) -> (a, b) -> c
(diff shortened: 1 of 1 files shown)


andrevidela:
Hi, thanks for this I just noticed that the Idris example code are actually not valid Idris code, what's the intention here?

user470750:
> Hi, thanks for this I just noticed that the Idris example code are actually not valid Idris code, what's the intention here?

Hi, thanks for catching this! You're right, the example doesn't compile. I based the format/style on the documentation for `on` in the same module (just as a template for how examples are written there):
https://github.com/idris-lang/Idris2/blob/3e8882813e4135e425cc248497a59c3dd5ef116e/libs/prelude/Prelude/Basics.idr#L55-L70

user470750:
Concrete and valid examples would be:
```idris
(curry swap) 1 0 = swap (1, 0)
```
and
```idris
(uncurry min) (1, 0) = min 1 0
```
It's probably a good idea to make the examples valid code so I'll update them

libs/prelude/Prelude/Basics.idr
```diff
@@ -82,12 +82,25 @@ public export %tcinline
...
+|||
+||| ```idris example
+||| (curry swap) 1 0 = swap (1, 0)
+||| ```
 public export
-curry : ((a, b) -> c) -> a -> b -> c
+curry : (f : (a, b) -> c) -> a -> b -> c
 curry f a b = f (a, b)
 
+||| Converts a curried function that takes arguments separately to a function on pairs.
```
joelberkeley: If you don't mind me mentioning, I was told by tech writers it's best to use the tense `Convert`

user470750: Thank you for the review, you're right. Unrelated to this PR, but it looks like `flip`'s documentation in this module has the same issue ("Takes in the first two arguments in reverse order.").

joelberkeley: The whole code base is inconsistent in this regard, which is kind of expected

andrevidela (approved):
Let's get this in as it's sorely needed

PR #3831, opened 2026-08-11
[ docs ] Improved documentation for `Prelude.Interfaces`

user470750:
# Description
* Add documentation for the `Applicative` interface and the `pure`, `<*>`, `*>`, and `<*` functions, including code examples. Examples use `==` in accordance with the documentation style in Prelude.Interfaces.
* Add documentation for `mapHom` function following the style of the documentation for `bimap`, `mapFst`, `mapSnd` documentation.
* Add documentation for `Traversable` and `Bitraversable` intefaces.
* Add documentation for named interfaces implementations.
* Add documentation for `Bifoldable` interface following the style of the documentation for `Foldable`.

## Self-check

### DIFF
--- a/libs/prelude/Prelude/Interfaces.idr
+++ b/libs/prelude/Prelude/Interfaces.idr
@@ -183,21 +183,37 @@ namespace Bifunctor
 public export
 interface Functor f => Applicative f where
   constructor MkApplicative
-  ||| Lift a value into the structure.
+  ||| Lifts a value into the structure.
   ||| @ f the parameterised type
   pure : a -> f a
 
-  ||| Sequential application.
+  ||| Sequential application. Applies an `Applicative` of functions to
+  ||| a second `Applicative`.
+  |||
+  ||| ```idris example
+  ||| ((Just \x => x + 1) <*> Just 5) == Just 6
+  ||| ```
   ||| @ f the parameterised type
   (<*>) : f (a -> b) -> f a -> f b
 
 ||| Sequence actions, discarding the value of the second argument.
+|||
+||| ```idris example
+||| (Just 5 <* Just 10) == Just 5
+||| ```
 ||| @ f the parameterised type
 public export %tcinline
 (<*) : Applicative f => f a -> f b -> f a
 a <* b = map const a <*> b
 
 ||| Sequence actions, discarding the value of the first argument.
+|||
+||| ```idris example
+||| (Just 5 *> Just 10) == Just 10
+||| ```
+||| ```idris example
+||| (Nothing *> Just 2) == Nothing
+||| ```
 ||| @ f the parameterised type
 public export %tcinline
 (*>) : Applicative f => f a -> f b -> f b
--- a/libs/prelude/Prelude/Interfaces.idr
+++ b/libs/prelude/Prelude/Interfaces.idr
@@ -491,27 +491,40 @@ public export %tcinline
 for_ : Applicative f => Foldable t => t a -> (a -> f b) -> f ()
 for_ = flip traverse_
 
+||| `Applicative` structures over a `Semigroup` type form a semigroup
+||| by lifting the semigroup operation through the `Applicative` structure.
 public export
 [SemigroupApplicative] Applicative f => Semigroup a => Semigroup (f a) where
   x <+> y = [| x <+> y |]
 
+||| `Applicative` structures over a `Monoid` type form a monoid
+||| by lifting the monoid operation and the neutral element
+||| through the `Applicative` structure.
 public export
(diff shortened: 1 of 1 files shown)


libs/prelude/Prelude/Interfaces.idr
```diff
@@ -169,16 +176,45 @@ namespace Bifunctor
     mapFst = map . mapFst
     mapSnd = map . mapSnd
 
+||| A functor with application, providing operations to embed
+||| pure expressions (`pure`) and sequence computations (`<*>`).
+||| Abstracts the notion of function application.
+||| @ f a parameterised type
```
joelberkeley, line 182: IMO the `@ f`s are unnecessary here, and busy the docs. They're also not reflected by docs for `a` and `b` later

libs/prelude/Prelude/Interfaces.idr
```diff
@@ -169,16 +176,45 @@ namespace Bifunctor
...
+  ||| ```idris example
+  ||| ((Just \x => x + 1) <*> Just 5) == Just 6
+  ||| ```
+  ||| @ f the parameterised type
   (<*>) : f (a -> b) -> f a -> f b
 
+||| Sequence actions, discarding the value of the second argument.
+|||
+||| ```idris example
+||| (Just 5 <* Just 10) == Just 5
```
joelberkeley, line 202: Note this example doesn't convey that the second effect is used. The converse of the `(*>)`  example might help
```
Main> Just 2 <* Nothing
Nothing
```
It's a shame sth like `State` wouldn't be suitable here.

libs/prelude/Prelude/Interfaces.idr
```diff
@@ -169,16 +176,45 @@ namespace Bifunctor
...
     mapSnd = map . mapSnd
 
+||| A functor with application, providing operations to embed
+||| pure expressions (`pure`) and sequence computations (`<*>`).
+||| Abstracts the notion of function application.
+||| @ f a parameterised type
 public export
 interface Functor f => Applicative f where
   constructor MkApplicative
+  ||| Lift a value into the structure.
```
joelberkeley, line 186: Is "structure" always appropriate? What about `IO` or `State`? You use `Applicative` further down, though I don't think that's strictly true either. Is the `Applicative` not the `pure` and `(<*>)`, rather than e.g. the `Maybe`? I'm unsure if `Maybe` "_is_ an applicative functor" (small "a").

I'm aware that genericity of these interfaces will make it difficult to document them accurately, without it becoming impractically abstract.

libs/prelude/Prelude/Interfaces.idr
user470750: I based my choice of which arguments to document on the existing documentation for `Functor`.

libs/prelude/Prelude/Interfaces.idr
user470750: For example, the word "structure" is used in the tutorial for Functor:
https://github.com/idris-lang/Idris2/blob/811e065a3ed89af0abf22bb1d8b8c6073ae1401a/docs/source/tutorial/interfaces.rst?plain=1#L238-L239
As for the wording "Applicative", I'm using it by analogy with the documentation for [`Biapplicative`](https://github.com/idris-lang/Idris2/blob/811e065a3ed89af0abf22bb1d8b8c6073ae1401a/tests/idris2/interface/interface006/Biapplicative.idr) though I'm not sure how reliable a source the tests are for documentation wording.
It might make sense to use the same wording ("Applicative") for pure as well, for consistency.
In some of the phrasing, I also looked at Haskell for reference (e.g. "functor with application").
Also for example, the documentation for the functions related to Functor (`map`, `<$>`, `<&>`, etc.) uses the wording "a parameterised type".

Do you have any suggestions for alternative wording instead of "structure" and "Applicative"?

libs/prelude/Prelude/Interfaces.idr
user470750: Thank you for pointing this out, I'll add such an example.

user470750: Added the second example for `<*`.

libs/prelude/Prelude/Interfaces.idr
joelberkeley: ah, well I guess I disagree with Edwin there. But yeah it makes sense to copy `Functor`'s style.

libs/prelude/Prelude/Interfaces.idr
joelberkeley: I would guess the haskell docs would be the standard to follow. And they call it a structure, so seems sensible.

PR #3832, opened 2026-08-12
[ refc ] Fix casts from String failing to compile

### LINKED ISSUE #3812: RefC: casts from `String` fail to compile
# Steps to Reproduce

```idris
module Main

main : IO ()
main = printLn (cast {to = Double} "1.5")
```

```sh
idris2 --cg refc -o test Main.idr
```

# Expected Behavior

The program compiles and prints `1.5`, as it does on the chez, racket and node backends.

# Observed Behavior

```
build/exec/repro.c:201:31: error: implicit declaration of function 'idris2_cast_string_to_Double'; did you mean 'idris2_cast_String_to_Double'? [-Werror=implicit-function-declaration]
build/exec/repro.c:201:31: error: initialization of 'Idris2_Value *' from 'int' makes pointer from integer without a cast [-Werror=int-conversion]
cc1: all warnings being treated as errors
```

`-Werror` is hardcoded in `Compiler.RefC.CC`, so the mismatch is always fatal.

## Probable root cause

`Compiler.RefC.RefC.cOp` builds cast helper names via `cPrimType`, which maps `StringType` to lowercase `"string"`, while `support/refc/casts.h` / `casts.c` export the capitalised `idris2_cast_String_to_Int64`, `idris2_cast_String_to_Integer`, `idris2_cast_String_to_Double`, etc. So all String-source casts miscompile.

Looks like it's a regression after #3181. It introduced the capitalised support helpers without updating `cPrimType`.

bio-aeon:
# Description

Fixes #3812.

Every cast with a `String` source type fails to compile on the RefC, because codegen builds cast helper names like `idris2_cast_string_to_*`, but after #3181 the support library has exported these helpers capitalised (`idris2_cast_String_to_*`).

This PR renames the cast helpers to the lowercase form the codegen expects.

## Self-check


joelberkeley:
I fixed an almost identical issue wrt float a while back. In my own project I avoid issues like this by compiling the project with refc, which involves the C compiler trying and failing to match the `%foreign` strings with non-existent names in C headers, and throwing an error. Perhaps the problem here is that there is nothing in the repo calling these functions, so no such problem is encountered.

I mention this to hopefully come up with some way to catch all such typos once and for all.

bio-aeon:
> Perhaps the problem here is that there is nothing in the repo calling these functions, so no such problem is encountered.

I found this particular issue when running `log4types` test suite across backends. This forced me to rewrite some tests without string-casts.
Agreed that it deserves a proper catch-them-all solution, but that feels like its own piece of work.

joelberkeley:
> but that feels like its own piece of work

oh for sure, wasn't suggesting you do that here

PR #3834, opened 2026-08-15
Fix %transform docs

Carrotstrip:
Example usage of %transform was incorrect (returned j + j instead of j + k).

## Self-check


mattpolzin:
@gallais this PR is going to need admin-override since `overall-success` requires several CI steps that only run when code is changed.

PR #3839, opened 2026-08-20
[ cleanup ] Lexer code using SnocList

gallais:


src/Libraries/Text/Lexer/Core.idr
```diff
@@ -137,51 +145,55 @@ public export
...
+                     Just (MkBounded (fn (fastPack (tok <>> []))) False (MkBounds line col line' col'),
                            line', col', rest)
                Nothing => getFirstToken ts str
 
 export
 lexTo : (a -> Bool) ->
         TokenMap a -> String -> (List (WithBounds a), (Int, Int, String))
 lexTo pred tmap str
-    = let (ts, (l, c, str')) = tokenise pred 0 0 [] tmap (unpack str) in
+    = let (ts, (l, c, str')) = tokenise pred 0 0 [<] tmap (fastUnpack str) in
```
mattpolzin, line 190: Does the transform rule for unpack fail here or is this change just force of habit?

gallais: Just habit.

I had originally changed the code so as *not* to unpack and instead
use `String` & `strM` but that turned out to be a lot slower than unpacking
the whole file and working on `List Char`.
When manually reverting these (interleaved with `SnocList`) changes, I
reached for `fastUnpack`.

src/Libraries/Text/Lexer/Tokenizer.idr
```diff
@@ -66,68 +66,61 @@ Pretty Void StopReason where
...
-                    assert_total $ tokenise end middle line' col' [] rest
+              (acc', (reason, line'', col'', rest'')) =
+                    assert_total $ tokenise end middle line' col' (acc :< beginTok'') rest
            in case reason of
                    ComposeNotClosing {} => Left reason
                    _ => let Just (endTok', lineEnd, colEnd, restEnd) =
                                 getNext end line'' col'' rest''
                               | _ => Left $ ComposeNotClosing (line, col) (line', col')
                             endTok'' = MkBounded (mapEnd endTok') False (MkBounds line'' col'' lineEnd colEnd)
-                         in Right ([endTok''] ++ reverse midToks ++ [beginTok''], lineEnd, colEnd, restEnd)
```
gallais, line 125: Found this horror show when investigating the slow lexer reported by
@JacquesCarette and @TOTBWF in https://jacquescarette.github.io/panbench/#Newlines

Still, it's not gotten any faster :(

JacquesCarette: "horror show" is a rather good description of that particular line!

Issue #3223, opened 2024-03-03
Overly eager auto implicit in GADT ctor field's type results in typechecking failure


0xd34df00d:
# Steps to Reproduce

Try typechecking
```idris
module Foo

interface SomeMonad (s : Type) (m : Type -> Type) where

data Operation : (res : Type) -> Type where
  Op : (opFun : forall s, m. SomeMonad s m => m r1) -> (cont : r1 -> Operation r2) -> Operation r2

Functor Operation where
  map f (Op opFun cont) = Op opFun (map f . cont)
```

# Expected Behavior

Typechecks fine.

# Observed Behavior

Fails with
```
Error: While processing right hand side of map. Can't find an implementation for SomeMonad ?s ?m.

Foo:9:30--9:35
 5 | data Operation : (res : Type) -> Type where
 6 |   Op : (opFun : forall s, m. SomeMonad s m => m r1) -> (cont : r1 -> Operation r2) -> Operation r2
 7 | 
 8 | Functor Operation where
 9 |   map f (Op opFun cont) = Op opFun (map f . cont)
                                  ^^^^^
```

If we replace the `opFun` at the rhs with a hole and look at its type, it'd be
```
rhs : m r1
```
instead of the (expected) `SomeMonad s m => m r1`, so surely it fails to find the implementation.

Note 1: removing `s` from both `SomeMonad` definition and from `forall` in the `Op` definition makes it compile, albeit the hole type is still funny and lacks the `SomeMonad m =>` constraint.

Note 2: adding a dummy argument like
```
data Operation : (res : Type) -> Type where
  Op : (opFun : () -> forall s, m. SomeMonad s m => m r1) -> (cont : r1 -> Operation r2) -> Operation r2
```
also makes things compile.

joelberkeley:
Note the type shown may omit implicits, of which `SomeMonad s m` is one.

joelberkeley:
I think this is the same as this error
```idris
f : ({x : ()} -> ()) -> ()
g : {x : ()} -> ()

h : ()
h = f g
```

```
[ build ] Error: Unsolved holes:
[ build ] Main.{x:826} introduced at: 
[ build ] Main:9:7--9:8
[ build ]  5 | f : ({x : ()} -> ()) -> ()
[ build ]  6 | g : {x : ()} -> ()
[ build ]  7 |
[ build ]  8 | h : ()
[ build ]  9 | h = f g
[ build ]            ^
```

Issue #14, opened 2020-05-20
Implicit lambda arguments


edwinb:
<a href="https://github.com/clayrat"><img src="https://avatars0.githubusercontent.com/u/321557?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [clayrat](https://github.com/clayrat)**
_Sunday Jul 28, 2019 at 00:28 GMT_
_Originally opened as https://github.com/edwinb/Idris2-boot/issues/56_

----

(This is more of a feature request, but it's been a long-standing stumbling block for me in Idris 1, and I hope it's fairly uncontroversial)

# Steps to Reproduce
```
module Temp

import Data.Vect 

All : (a : i -> Type) -> Type
All a {i} = {j : i} -> a j

Vect' : Type -> Nat -> Type
Vect' = flip Vect 

infixr 1 :->
(:->) : (a, b : i -> Type) -> (i -> Type)
(:->) a b i = a i -> b i

showAll : All (Vect' Integer :-> Vect' String)
showAll = map show

justShowAll : Maybe (All (Vect' Integer :-> Vect' String))
justShowAll = Just ?hole
```
# Expected Behavior

`justShowAll = Just $ \{j} => map show`

# Observed Behavior

There doesn't seem to be a way to introduce the implicit to finish filling the hole.

edwinb:
<a href="https://github.com/edwinb"><img src="https://avatars1.githubusercontent.com/u/255926?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [edwinb](https://github.com/edwinb)**
_Sunday Jul 28, 2019 at 12:17 GMT_

----

Internally, these exist, but there's no syntax yet. There is a possible complication - implicits might appear in a different order than the lambdas, so perhaps there needs to be a way to indicate which implicit argument you mean (in much the same way as you write {n = foo} in a function call) and so it might need to be able to handle lambdas appearing in a different order than in the type.

On the other hand, a simple approach would be to say that if you have an implicit lambda, it needs to be "the next one".

Anyway, all of that is why I've never implemented it. But it would be nice to have.

edwinb:
<a href="https://github.com/clayrat"><img src="https://avatars0.githubusercontent.com/u/321557?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [clayrat](https://github.com/clayrat)**
_Sunday Jul 28, 2019 at 13:31 GMT_

----

[...]

PR #3842, opened 2026-08-21
[ new ] Support implicit arguments on lambdas

dunhamsteve:
# Description

This PR adds implicit lambda arguments to the surface syntax, following the [purely positional arguments](https://github.com/idris-lang/Idris2/issues/14#issuecomment-1206788173) proposed in issue #14.

The current syntax for _explicit_ arguments is `\ x, y : Nat, z => ...`. The change here is to allow wrapping each of those in braces to get an implicit argument in that position. So making the first two arguments implicit would look like `\ {x}, {y : Nat}, z => ...`. No commas inside the braces and no named arguments.  I think the syntax is a natural extension of the current lambda syntax. It varies a little from function definitions, but so do the commas in the current syntax.

I'll leave this as a draft while we sort out if there is a consensus and if this is useful. At that point we can update the documentation and add tests.

## Possible issues using implicit lambdas

An issue that may interfere with using this is automatic insertion of implicits.  I think there are two flavors that may come up:

**Idris2 will insert arguments on the left hand side for implicits referenced in the _normalized_ function type.** This insertion includes implicits after the last explicit argument given by the user. An implicit lambda expression at the beginning of the RHS will sometimes not work. (It may be a bug that Idris picks up implicits after the last argument or that it picks up implicits that aren't mentioned in the surface syntax.)

An example of this problem is:
```idris
WeirdType : Type
WeirdType = {0 a : Type} -> Nat -> Nat

oops : WeirdType -> WeirdType
oops f = \{0 a : Type} => f {a}
```
Here, `{0 a : Type}` is in the type _after normalization_, so Idris inserts `{a = a@{_}}` at the beginning of the LHS, even though that argument comes after the `f`.

**Idris2 will insert lambdas in front of most terms when the expected type begins with an implicit pi.** This does not happen if the term is lambda-shaped. If an implicit lambda is assigned to a name and then later used as an argument where an implicit Pi is expected, it won't typecheck. This issue is captured as #3223. I think it may be difficult to address because the insertion is sometimes desirable. This issue does not happen if a lambda is used directly as an argument.

So assuming these two issues are not yet addressed, I'd like to now if this feature is still useful for common use cases. And a test case would be nice.

A contrived example is:
```idris
data F : Nat -> Nat -> Type where

takesImpl : ({i : Nat} -> {j : Nat} -> F i j) -> ()
f : (i : Nat) -> F i j
g : (j : Nat) -> F i j

callImpl0 : ()
callImpl0 = takesImpl $ \ {i} => f i

callImpl3 : ()
callImpl3 = takesImpl $ \ {x} => f x
```

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -80,6 +80,7 @@ should target this file (`CHANGELOG_NEXT`).
   `%foreign_impl` strings [#3790](https://github.com/idris-lang/Idris2/issues/3790)
   * Improve related error messages
 * Fix exponential time issue in totality checking with large data on the left hand side (#3696).
+* Allow implicit lambdas using the syntax `\ {x} => ...`.
 * Removed `Borrowing` as a language extension.  This was never implemented in
   Idris2, so the only change is that `%language Borrowing` will now error rather
   than be accepted but do nothing.
--- a/src/Idris/Parser.idr
+++ b/src/Idris/Parser.idr
@@ -653,15 +653,27 @@ mutual
           _ => fail "Invalid multiplicity (must be 0 or 1)"
 
   bindList : OriginDesc -> IndentInfo ->
-             Rule (List (RigCount, WithBounds PTerm, PTerm))
+             Rule (List (RigCount, WithBounds PTerm, PTerm, PiInfo PTerm))
   bindList fname indents
-      = forget <$> sepBy1 (decoratedSymbol fname ",")
-                          (do rig <- multiplicity fname
-                              pat <- bounds (simpleExpr fname indents)
-                              ty <- option
-                                       (PInfer (boundToFC fname pat))
-                                       (decoratedSymbol fname ":" *> opExpr pdef fname indents)
-                              pure (rig, pat, ty))
+      = forget <$> sepBy1 (decoratedSymbol fname ",") (explicitBind <|> implicitBind)
+    where
+      explicitBind : Rule (RigCount, WithBounds PTerm, PTerm, PiInfo PTerm)
+      explicitBind = do
+        rig <- multiplicity fname
+        pat <- bounds (simpleExpr fname indents)
+        ty <- option
+           (PInfer (boundToFC fname pat))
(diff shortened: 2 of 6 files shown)


ohad:
Here's a less contrived example.

# Example 1: defining records with fields taking implicit arguments

Consider the following record type that defines a default value for a
given family of type-codes:
```idris
record DefaultCode (code : a -> Type) where
  constructor MkDefault
  choose : {0 ty : a} -> code ty

Ex0, Ex1,Ex2 : DefaultCode (Fin . S)
```
In this simple example, the type-checker can infer the implicit argument
 because `Fin` is injective.
```idris
Ex0 = MkDefault FZ
```
But in more complicated examples, we will be stuck and need to do
the following:
```idris
Ex1 = MkDefault (\{n} => FZ)
Ex2 = MkDefault {choose = \{n} => FZ}
```
Here's an example where `Ex0` won't work:
```idris
Code : Nat -> Type
Code 0 = Fin 1
Code n = Fin n

aux : Nat -> Nat
aux 0 = 0
aux n@(S k) = k

Ex0' : DefaultCode Code
Ex0' = MkDefault $ \{n} =>
  replace {p = id} (case n of
    0 => Refl
    (S k) => Refl
  ) (Fin.FZ {k = aux n})
```

This kind of issue happens all the time when we deal with multi-sorted algebraic operations, e.g., matrix-like multiplication that tracks dimensionality, etc.

andrevidela:
this diff makes me wish I had kept going with refactoring tuples out of the compiler. maybe I'll take a crack at it once this is merged

gallais:
Can we have a test with multiple implicit arguments all bound in the same lambda?
Ideally one that would not typecheck if they were bound in the wrong order.
Would also be nice to have a test with shadowing bindings.

This should make the test useful in case we also start throwing in different heuristics
(e.g. the name-based resolution used during the elaboration of implicit applications).

After that, happy to merge.

dunhamsteve:
I added a few more cases with two implicits of different types. Two of them were "failing" clauses because the implicit lambda returns the wrong argument - one because of shadowing.

PR #3847, opened 2026-08-25
[fixes #3846][ library ] Re-namespace quantifiers for easier disambiguation.

### LINKED ISSUE #3846: Library design to make namespace disambiguation easier to type.
When working with dependent types there will be times where you will have code that looks the same but is working with different values at the type-level. Consider the following code that fails to elaborate.

```
import Data.List.AtIndex
import Data.List.Quantifiers
import Data.Vect
import Data.Vect.Quantifiers
import Data.Vect.AtIndex

namespace List
  export
  extract : AtIndex x xs i -> All p xs -> p x

namespace Vect
  export
  extract : AtIndex x xs i -> All p xs -> p x
  
```

This fails as Idris cannot tell which version of `AtIndex` is being used.

To solve this, one needs to disambiguate *something*.

There are three *valid* routes:

1. Spin out the functions to their own module; [SOLID](https://en.wikipedia.org/wiki/SOLID) but requires an extra module;

2. Implicitly bind a value, that is roll-out the implicitly bound `xs` to introduce the binding site for `xs`, marking it `0`; valid but requires lots of typing;

3. namespace disambiguate the `AtIndex` to tell idris which one; my preferred solution,

The problem this proposal addresses is with number three. Depending on the namespace hierarchies involved, these names can become long and unwieldy. Consider the resolved names:

```
namespace List
  export
  extract : List.AtIndex.AtIndex x xs i -> All p xs -> p x

namespace Vect
  export
  extract : Vect.AtIndex.AtIndex x xs i -> All p xs -> p x
```

Here the subject of the quantifier is at the head of the namespace, with the verb, representing the quantifier itself, at the tail. The verb is even replicated.

The proposal is simple. When defining predicates acting on data, wrap them in a namespace named after their subject. Thus, shortening the hierarchy required for disambiguation.

Thus, turning our original example to:

```
namespace List
  export
  extract : List.AtIndex x xs i -> All p xs -> p x

namespace Vect
  export
  extract : Vect.AtIndex x xs i -> All p xs -> p x
```

For deeply nested hierarchies, such as `List.Quantifiers.All.All` this options makes the disambiguation much much shorter.

This will require a simple PR with the changes, and updates to code where things have been disambiguated using namespaces. We should also remind ourselves to check PRs for this design pattern.

I often use this pattern with my own libraries and think it is something we should do in `libs`.

I will be making a PR shortly.

- [X] I have read [CONTRIBUTING.md](https://github.com/idris-lang/Idris2/blob/main/CONTRIBUTING.md).
- [X] I have checked that there is no existing PR/issue about my proposal.
- [X] I confirm that this contribution did not involve GenerativeAI nor Large Language Models.

jfdm:
When working with dependent types there will be times where you will have code that looks the same but is working with different values at the type-level. Leading to namespace ambiguities. Consider the following:

```
namespace List
  export
  extract : AtIndex x xs i -> All p xs -> p x

namespace Vect
  export
  extract : AtIndex x xs i -> All p xs -> p x
```

We need to disambiguate the names, a quick solution is to use namespace disambiguation. Depending on the namespace hierarchies involved, however, these names can become long and unwieldy. Consider the resolved names:

```
namespace List
  export
  extract : List.AtIndex.AtIndex x xs i -> All p xs -> p x

namespace Vect
  export
  extract : Vect.AtIndex.AtIndex x xs i -> All p xs -> p x
```

Here the subject of the quantifier is at the head of the namespace, with the verb, representing the quantifier itself, at the tail. The verb is even replicated.

When defining predicates acting on data, wrap them in a namespace named after their subject. Thus, shortening the hierarchy required for disambiguation.

```
namespace List
  export
  extract : List.AtIndex x xs i -> All p xs -> p x

namespace Vect
  export
  extract : Vect.AtIndex x xs i -> All p xs -> p x
```

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -87,7 +87,7 @@ should target this file (`CHANGELOG_NEXT`).
   external tools to insert links from documentation to source code.
 * Fixed `parseDouble` dropping "-" sign when whole part is "0"
   [#3804](https://github.com/idris-lang/Idris2/issues/3804),
-
++ Changed module level namespace of `Libraries.Data.List01.Quantifiers` from `All` to `List01`
 ### Building/Packaging changes
 
 * Fix parsing of capitalised package names containing hyphens.
@@ -129,3 +129,5 @@ should target this file (`CHANGELOG_NEXT`).
 * Removed `writeIORef1`, which unsafely allowed a linear value to become unrestricted.
 * Made `xs` and `ys` explicit arguments of `SplitRecPair` in `Data.Vect.Views`.
 * Implemented core interfaces for `Data.Singleton`
++ Renamed-spaced quantifiers for `List`, `Vect`, `SnocList`, `List01` and other predicates,
+  so that their most recent parent is named for the datatype and not the quantifier.
--- a/libs/base/Data/List/AtIndex.idr
+++ b/libs/base/Data/List/AtIndex.idr
@@ -14,13 +14,14 @@ import Decidable.Equality
 
 %default total
 
-||| @AtIndex witnesses the fact that a natural number encodes a membership proof.
-||| It is meant to be used as a runtime-irrelevant gadget to guarantee that the
-||| natural number is indeed a valid index.
-public export
-data AtIndex : a -> List a -> Nat -> Type where
-  Z : AtIndex a (a :: as) Z
-  S : AtIndex a as n -> AtIndex a (b :: as) (S n)
+namespace List
+  ||| @AtIndex witnesses the fact that a natural number encodes a membership proof.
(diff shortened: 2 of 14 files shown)


stefan-hoeck:
While I see how this saves some typing, we'll definitely have to check how much existing code this breaks before merging.

gallais:
This breaks bootstrap so I think we should just not do it.

Especially given that client code using a qualified import is AFAIU an already working solution.

jfdm:
I will close this, more because of the bootstrapping issue than anything else. I nonetheless think it is a design pattern worth considering with base and prelude.

Issue #3836, opened 2026-08-17
[ perf ] disambiguation is extremely slow


FFFluoride:
# Steps to Reproduce
This is the smallest example I was able to make. I typed in my terminal 'idris2 Wontbuilt.idr'
```idris
module Wontbuilt

%default total

data A = B | C | D
data E = F | G | H

data I : E -> A -> Type where
  J : I F B
  K : I G C
  L : I H D

bad : (s ** I s i) -> (r ** I r j) -> ()
bad (a ** b) (H ** c) = ()
bad a (H ** b) = ()
bad (H ** L) a = ()
```
# Expected Behavior
It terminates and the code is either ok or there is an error (I stripped it down from some code that looked ok to me, but I did remove alot of cases)

# Observed Behavior
`1/1: Building Wontbuilt (Wontbuilt.idr)` in the terminal and it doesn't terminate

FFFluoride:
Nevermind, it just takes like 5 mins to fail on my machine. I'll be more thorough next time

dunhamsteve:
Ideally this wouldn't take that long.  Something exponential is happening. 

My theory is that it's the resolution of the ambiguous `**` (which can mean either `DPair` or `MkDPair`).  During elaboration, resolving this sometimes involves a bit of retrying and deferring the solution until later. I like the ambiguity resolution in Idris, but sometimes it's trouble.

And indeed, if you spell things out explicitly with `DPair` and `MkDPair`, it checks quickly:

```idris
module Wontbuilt

%default total

data A = B | C | D
data E = F | G | H

data I : E -> A -> Type where
  J : I F B
  K : I G C
  L : I H D

bad : (DPair _ (\s => I s i)) -> (DPair _ $ \r => I r j) -> ()
bad (MkDPair s b) (MkDPair H c) = ()
bad a (MkDPair H b) = ()
bad (MkDPair H L) a = ()
```

FFFluoride:
Thank you, I was confused about this (mostly `(**)` to be honest). Now that I know this I will try to be more explicit about it

gallais:
I think it should stay open; as Steve says it should not take this long to simply do disambiguation.

dunhamsteve:
I was completely wrong here. I thought I'd grab a log to take a look, and it was spinning somewhere with no logs. I have an idris compiler that runs in node, so I ran it in a debugger and grabbed a stacktrace of where it was spinning. It's the pretty-printer, while printing the error. Near the bottom of the stack:
[...]

Issue #1900, opened 2021-09-04
Lazy evaluation in `let` block does not behave as expected


stefan-hoeck:
It should be possible to delay evaluation in a `let` expression, in case we are dealing with an expensive computation, the result of which will not necessarily be needed. Right now, this doesn't seem to work (at least not when using `delay`, but probably I'm doing something wrong here).

# Steps to Reproduce

Build and run the following program:

```idris
import Debug.Trace

calc : Nat -> Nat
calc n0 =
  let n1  = delay $ (trace "foo" $ n0 + n0)
      n2  = delay $ (trace "bar" $ 2 * n1)
      n3  = delay $ (trace "baz" $ 2 * n2)
   in if n0 > 10 then n2 else n3

main : IO ()
main = do printLn (calc 100)
```

# Expected Behavior

Program prints

```
foo
bar
400
```

# Observed Behavior

Program prints

```
foo
bar
baz
400
```

It can be seen in the generated Javascript, that the call to `delay` is ignored (but this is not an issue of the JS backend. The schemes behave the same):

```javascript
function Main_calc($0) {
 const $1 = Debug_Trace_trace('foo', Prelude_Types_x2b_Num_Nat($0, $0));
 const $8 = Debug_Trace_trace('bar', Prelude_Types_x2a_Num_Nat(Prelude_Types_fromInteger_Num_Nat(2n), $1));
 const $11 = Debug_Trace_trace('baz', Prelude_Types_x2a_Num_Nat(Prelude_Types_fromInteger_Num_Nat(2n), $8));
 switch(Prelude_Types_x3e_Ord_Nat($0, Prelude_Types_fromInteger_Num_Nat(10n))) {
  case 1: return $8;
  case 0: return $11;
 }
}
```

gallais:
Do you get the same behaviour if you explicitly assign a `Lazy ?` type to these
let-bound expressions? It could be that inference comes up with a type that means
the thunk is forced!

stefan-hoeck:
You are right. If I replace `delay` with `the (Lazy Nat)`, I get the expected behavior.

stefan-hoeck:
So, how to move on from here? Personally, I think the inferred type here is not correct, as the call to `delay` should make it clear that we are interested in a delayed computation. However, it might be difficult to convey this to Idris, given that at other occasions we are happy that Idris inserts `force` for us automatically. As an alternative, we could mention this use case somewhere in the documentation and accept it as one of the quirks of Idris2.

[...]

PR #3848, opened 2026-08-26
[ fix ] pretty printer is sometimes exponential

dunhamsteve:
# Description

In #3836 a file takes a long time to check because of an exponential time issue while printing the error message.

On line 668, the second branch of a `Union` is delayed before being passed to `selectNicer`, but it is immediately passed to `fits`, which takes a non-lazy argument. The argument is forced, but not used by the `fits` function passed in by `layoutPretty`. I made that argument lazy.

That change did not completely fix the issue, because the delayed value is `let` assigned to a variable, and Idris appears to infer the type as non-lazy, dropping the `delay`. So I added a type annotation to the `let`.

Prior to this change, it took about a minute to check the file in #3836. After the change, it takes a third of a second.

I made the same changes to the copy of the code in contrib.

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -85,6 +85,7 @@ should target this file (`CHANGELOG_NEXT`).
   than be accepted but do nothing.
 * HTML files generated using `--mkdoc` now contain attributes, that allow
   external tools to insert links from documentation to source code.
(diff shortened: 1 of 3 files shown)


dunhamsteve:
It looks like the `let` issue is recorded as #1900.

gallais (approved):

PR #3849, opened 2026-08-26
[ fix ] don't force terms with explicit delay

dunhamsteve:
# Description

This PR fixes issue #1900 by telling the elaborator not to insert `force` in front of an explicit `delay`.  This issue also came up in #3836 and was worked around there by adding an annotation to the `let`.

When a user types:
```idris
expensive : Nat -> Foo

test = let x = delay (expensive 10) in ?hole
```
The type of `x` is `Foo` rather than `Lazy Foo`, despite the user explicitly typing `delay`. Because the type is inferred and a `force` is being inserted during conversion checking. 

We sometimes want to force the value before assigning it, to avoid duplicate work, but I think we can safely assume that the user wants a `Lazy` value if they explicitly write `delay`. So the PR disables this mechanism only if the term being checked begins with a `TDelay`.

### Alternatives considered

A user could work around the issue by annotating the `let` with a type:
```idris
test = let x : Lazy Foo = delay (expensive 10) in ?hole
```

## Self-check

### DIFF
--- a/CHANGELOG_NEXT.md
+++ b/CHANGELOG_NEXT.md
@@ -87,6 +87,7 @@ should target this file (`CHANGELOG_NEXT`).
   external tools to insert links from documentation to source code.
 * Fixed `parseDouble` dropping "-" sign when whole part is "0"
   [#3804](https://github.com/idris-lang/Idris2/issues/3804),
+* Do not insert `force` in front of an explicit `delay`.
 
 ### Building/Packaging changes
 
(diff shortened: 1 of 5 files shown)


dunhamsteve:
This appeared to work last night, but I must have misread something. Closing for now.

dunhamsteve:
I do have this working in the https://github.com/dunhamsteve/Idris2/tree/issue1900 branch. It's not as clean because I needed to look for an `App` of `Builtin.delay` instead of `TDelay`.

PR #3857, opened 2026-09-14
Add building of installers

transcendentalduncan:
# Description

Adds building of Idris2 installers. The installed Idris2 always uses the `IDRIS2_PREFIX` environment variable by setting it to its own installation as a custom installation prefix. The default is unused, and the installed Idris2 will always check if the user has set the environment variable manually at the start so that the user can still use it as usual.

If there is a mismatch between the Chez version that the installer was built using, and the version on the target machine, then the Chez-based installer will automatically recompile using the version on the target machine. This makes installation slower and is avoided when using the Racket-based installer.

The Chez-based installer has a runtime dependency on Chez scheme on the target machine. The Racket-based installer uses `raco distribute` to bundle the required runtime and thus does not require Racket to be installed on the target machine.

To build the installers, use `make distributable` with the `SCHEME` environment variable set to the location of Chez (on the machine that's building the installers) and the `RACO` environment variable set to the location of raco.

The Racket-based installer must be built using a Racket installation that corresponds to the target that the installer will be used for (for example, if the installer was built using Racket Nix package, the installed Idris2 would depend on Nix store paths at runtime). This is because the Racket-based installer uses `raco distribute` to bundle the runtime as mentioned, and thus the Chez-based installer does not have this constraint.

## Self-check


buzden:
I think whatever new functionality is added, tests of it should be added too

